PR #134, opened 2015-01-02
don't call our intro sections "tutorials", because they're not

qwcode:
our intro sections aren't "tutorials" right now, so let's not set up a false expectation for the reader.

tutorials involve literal tasks that the reader is guided though with clear results and checks along the way.

whether they should be tutorials or not is another matter, but they're not currently.  I don't have any plans to make them true tutorials.   I can imagine having actual tutorials in the advanced section for certain topics possibly.

In the title, I simply remove the word "Tutorial", but in the text, I replaced "tutorial" with "section".

this does not break any external inbound links.

### DIFF
--- a/source/additional.rst
+++ b/source/additional.rst
@@ -5,8 +5,8 @@ Advanced Topics
 :Last Reviewed: 2014-12-24
 
 This section covers a variety of packaging concepts and topics that don't fit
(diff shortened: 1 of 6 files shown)


ncoghlan:
+1 for this as a simple improvement that makes the headings better match the content.

PR #140, opened 2015-03-07
Consistent style in code blocks

limedaring:
A few of the code blocks had extraneous spacing. 

Also, it might be worth adding trailing commas to the code snippets to match https://github.com/pypa/sampleproject/blob/master/setup.py — would make it easier to copy/paste. Let me know if you'd like for me to add that into this PR.

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -151,7 +151,7 @@ name
 
 ::
 
-  name = 'sample'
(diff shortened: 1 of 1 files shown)


qwcode:
ok, add the commas.

limedaring:
Added!

PR #141, opened 2015-03-09
Add summary of pynsist in OS packaging and installers section

tobias47n9e:
I just learned about [pynsist](https://github.com/takluyver/pynsist), and saw that it is still missing from this guide. Maybe wait for @takluyver to look over the draft before merging.

### DIFF
--- a/source/deployment.rst
+++ b/source/deployment.rst
@@ -45,10 +45,35 @@ OS Packaging & Installers
 
   - Building rpm/debs for projects
   - Building rpms/debs for whole virtualenvs
-  - Building Windows installers for Python projects
   - Building Mac OS X installers for Python projects
 
+Windows
+-------
 
+::
+
+  FIXME
+
+  - Building Windows installers for Python projects
+
+Pynsist
+^^^^^^^
+
+`Pynsist <https://pypi.python.org/pypi/pynsist>`__ is a tool that bundles Python
+programs together with the Python-interpreter into a single installer based on
+NSIS. In most cases, packaging only requires the user to choose a version of
+the Python-interpreter and declare the dependencies of the program. Downloading
+and bundling of the dependencies is done automatically.
+
+The installer installs or updates the Python-interpreter on the users system,
+which can be used independently of the packaged program. The program itself,
+can be started from a shortcut, that the installer places in the start-menu.
+Uninstalling the program leaves the Python installation of the user intact.
+
+A big advantage of pynsist is that the Windows packages can be build on Linux.
+There are several examples for different kinds of programs (console, GUI) in
+the `documentation <http://pynsist.readthedocs.org>`__. The tool is released
+under the MIT-licence.
 
 Application Bundles
 ===================
--- a/source/deployment.rst
+++ b/source/deployment.rst
@@ -62,15 +62,16 @@ Pynsist
 `Pynsist <https://pypi.python.org/pypi/pynsist>`__ is a tool that bundles Python
 programs together with the Python-interpreter into a single installer based on
 NSIS. In most cases, packaging only requires the user to choose a version of
-the Python-interpreter and declare the dependencies of the program. Downloading
-and bundling of the dependencies is done automatically.
+the Python-interpreter and declare the dependencies of the program. The tool
+downloads the specified Python-interpreter for Windows and packages it with all
+the dependencies in a single Windows-executable installer.
 
 The installer installs or updates the Python-interpreter on the users system,
 which can be used independently of the packaged program. The program itself,
 can be started from a shortcut, that the installer places in the start-menu.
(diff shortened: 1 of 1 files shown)


source/deployment.rst
```diff
@@ -45,10 +45,35 @@ OS Packaging & Installers
...
+  - Building Windows installers for Python projects
+
+Pynsist
+^^^^^^^
+
+`Pynsist <https://pypi.python.org/pypi/pynsist>`__ is a tool that bundles Python
+programs together with the Python-interpreter into a single installer based on
+NSIS. In most cases, packaging only requires the user to choose a version of
+the Python-interpreter and declare the dependencies of the program. Downloading
+and bundling of the dependencies is done automatically.
```
takluyver: It doesn't really download or build dependencies at present - it just copies them into a directory and packages them up.

takluyver: The exceptions being Python itself and, for Python 2, the standalone version of the `py` launcher. It downloads MSIs for these.

source/deployment.rst
```diff
@@ -45,10 +45,35 @@ OS Packaging & Installers
...
+
+The installer installs or updates the Python-interpreter on the users system,
+which can be used independently of the packaged program. The program itself,
+can be started from a shortcut, that the installer places in the start-menu.
+Uninstalling the program leaves the Python installation of the user intact.
 
+A big advantage of pynsist is that the Windows packages can be build on Linux.
```
takluyver: s/build/built/

takluyver:
Thanks for thinking of this, @tobias47n9e

source/deployment.rst
tobias47n9e: Ha, I was already wondering why all the packages were "downloaded" so quickly :)

tobias47n9e:
@takluyver: You're welcome. Fixed the sections you pointed out. Glad I can help.

qwcode:
@pfmoore merge?

pfmoore:
pynist looks neat. But it's only 2 weeks old, according to the PyPI entry, and from what @takluyver seems to be saying in the comments on the original version it doesn't (yet?) download dependencies as promised. So maybe it's a bit early to be recommending it?

Also, what about long-established tools like ```py2exe``` and ```cx_freeze``` (which do the same job, but differently) and ```pyzzer``` and PEP 441's ```zipapp``` (which bundle apps to be run under an existing Python interpreter).

To be honest, I'm not 100% sure what this section is *meant* to cover. "Installers for Python projects" could mean any of the above, or none.

I think the guide is intended as a recommendation, and good though pynist is (I look forward to trying it out!) I don't think it's what anyone would recommend at the moment if asked "how do I package up my Python application"? I think the answers right now would be "a wheel with entry points" if you wanted to install it into an existing Python interpreter, or "py2exe and bundle the result using one of the standard Windows installers" if you wanted a standalone app.

takluyver:
* Version 1.5 is 2 weeks old. Pynsist is ~ a year old: https://github.com/takluyver/pynsist/releases
* I never promised that it would download dependencies ;-)

> I don't think it's what anyone would recommend at the moment if asked "how do I package up my Python application"?

I'm one of the major contributors to cx_Freeze, and yes, I would recommend pynsist over a freeze tool. I'd try to give the impression that this is a work in progress and you should expect the road to be somewhat rough, but none of the tools for packaging an application really work smoothly.

pfmoore:
@takluyver Thanks for the clarification. My apologies, that was inexcusably dense of me, I really should have realised I was only looking at the current version. And my comments on downloading dependencies was a misinterpretation of the comments on this PR - I should have actually read the pynsist docs first.

From the POV of letting people know about a useful tool, I've no problem with merging this. Maybe someone should add a section on py2exe, for balance, but I guess that needs to wait till someone contributes one...

takluyver:
No problem :-)

It would certainly make sense to have a description of at least one of the freeze tools as well. I think both cx_Freeze and PyInstaller are more actively developed than py2exe - the latter just has the best name.

tobias47n9e:
@pfmoore: I would gladly volunteer to write about the other tools, but I didn't have time to really try them out. And I am avoiding tools that don't have Python 3.4 support. So as you said, the completion of the chapter will require more contributors. Is there still something I should change or move around in this pull request?

pfmoore:
@takluyver py2exe was stagnant for a long time, but recently a Python 3 version has appeared and the project now seems active again. I'm not going to make any sort of judgement between tools, though (and IMO the PUG shouldn't be doing so either, but it's down to contributors to add text for their preferred tool).

@tobias47n9e No, I think this is fine. I'll go ahead and merge it.

takluyver:
I'm not sure I'd call it active - the last commit was in October. But I agree with the approach of leaving it up to contributors who're interested enough to write up a particular tool.

PR #150, opened 2015-04-13
Update science.rst

kerrywatson1:
Updated #the-conda-cross-platform-package-manager section about conda written 
in 2013 that was outdated several versions ago.

### DIFF
--- a/source/science.rst
+++ b/source/science.rst
@@ -116,22 +116,21 @@ and ``virtualenv`` based toolchain.
 The conda cross-platform package manager
 ----------------------------------------
 
-`Anaconda <https://store.continuum.io/cshop/anaconda/>`__ (in this context)
-is a SciPy distribution published by Continuum Analytics.
+`Anaconda <https://store.continuum.io/cshop/anaconda/>`__ is a Python 
+distribution published by Continuum Analytics. It is a stable collection of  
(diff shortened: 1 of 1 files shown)


source/science.rst
```diff
@@ -116,22 +116,21 @@ and ``virtualenv`` based toolchain.
...
+``conda`` an open source (BSD licensed) package management system and 
+environment management system included in Anaconda that allows users to 
+install multiple versions of binary software packages and their dependencies, 
+and easily switch between them. It is a cross-platform tool working on Windows, 
+OSX, and Linux. Conda can be used to package up and distribute all kinds of 
+packages, it is not limited to just Python packages.   It has full support 
+for native virtual environments. Conda makes environments first-class citizens, 
+making it easy to create independent environments even for C libraries. It is 
+written in Python, but is Python-agnostic. Conda installs Python directly in 
+contrast to pip which does not.  Conda is available in Anaconda and Miniconda
```
qwcode: I''m happy to merge this, but can you elaborate more on "Conda installs Python directly in contrast to pip which does not"

asmeurer: Maybe better wording would be something like "For instance, conda manages python itself as a package, so that 'conda update python' is possible, in contrast to pip, which only manages Python packages".

PR #154, opened 2015-06-04
Recommend uploading PKG-INFO to register.

The-Compiler:
This should be a lot easier than filling the form by hand, which is what the text seems to suggest.


qwcode:
merging, although I'm going to follow up with a bit more explanation

PR #156, opened 2015-06-08
Use A.B.x labels for pre-installed versions

jayvdb:
The labels of the pre-installed versions are currently wrong for 2.7
and 3.4, as Appveyor has updated the version, and install.ps1 does not
override the pre-installed version.


jayvdb:
Same update merged at https://github.com/ogrisel/python-appveyor-demo/pull/12

PR #157, opened 2015-06-08
Add 2.6.6 to the Appveyor build matrix

jayvdb:
Packagers may be needing to do a Python 2.6 Win32 release.


jayvdb:
Note that this makes the prose slightly wrong, as it talks about the build matrix only including pre-installed releases.

jayvdb:
Note merged in https://github.com/ogrisel/python-appveyor-demo/pull/11

PR #158, opened 2015-07-01
Remove wheel reference in tool recommendations

prophile:
This was a recommendation for speeding up subsequent installations, but `pip` now automatically builds wheels in a cache since 7.0, so there's questionable utility in keeping it here.

### DIFF
--- a/source/current.rst
+++ b/source/current.rst
@@ -20,11 +20,6 @@ Installation Tool Recommendations
 * Use :ref:`virtualenv`, or `pyvenv`_ to isolate application specific
   dependencies from a shared Python installation. [3]_
 
-* Use `pip wheel
-  <http://pip.pypa.io/en/latest/reference/pip_wheel.html>`_ to create a
-  cache of :term:`wheel` distributions, for the purpose of speeding up
-  subsequent installations. [4]_
-
 * If you're looking for management of fully integrated cross-platform software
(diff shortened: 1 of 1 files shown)


rbtcollins:
Seems reasonable to me.

ncoghlan:
Rather than dropping the reference entirely, it may be worth recommending having bdist_wheel installed - pip will only automatically create the cache if setuptools has the ability to build wheel files.

qwcode:
>  pip will only automatically create the cache if setuptools has the ability

not sure what you're saying nick?  if setuptools doesn't have the "ability to build wheels", then neither wheel caching or bdist_wheel works.

I think the reason for still mention `pip wheel` is for maybe an "index building tool", where you quickly want to populate  a wheel index for your project needs, and not have to pluck it from your cache

qwcode:
> recommending having bdist_wheel installed

ok, I think you mean confirm having the `wheel` project installed.
btw, as of pip v7,  get-pip.py installs wheels for you

I'm gonna go ahead and merge this  (thanks @prophile ), and post a followup PR to add a few things

ncoghlan:
Currently, ensurepip doesn't bundle bdist_wheel, so folks won't have it by default - they need to run "pip install wheel" to get it. We'll likely update ensurepip to bundle the extra dependency eventually, but both @dstufft and I have plenty of other things on our plate at the moment, so it's likely to take a while.

PR #159, opened 2015-07-04
Add sections on Multi-homed Packages and On-demand Package Resolution.

jaraco:
After some more thoughts spurred by [this PR](https://bitbucket.org/pypa/setuptools/pull-request/135/remove-import-sys-from-easy-installpth/diff), I wanted to capture two aspects that have impeded my personal transition to use pip instead of easy_install.

### DIFF
--- a/source/installing.rst
+++ b/source/installing.rst
@@ -310,7 +310,7 @@ from the src tree.
  pip install -e <path>
 
 
-You can also normally from src
+You can also install normally from src
 
 ::
 


source/multi-homed packages.rst
```diff
@@ -0,0 +1,30 @@
...
+introduction of namespace packages, this expectation
+is no longer necessarily maintained. A package such
+as ``zc.component`` shares the ``zc`` namespace with
+``zc.interface``. Theoretically, those packages could
+be installed in separate locations or together in a
+single ``zc`` directory.
+
+One consequence of the "flat" package installation used
+by pip is that a given package is expected to be merged
+into a single, canonical location. Pip provides no
```
dstufft: Is this actually true? I think that namespace packages actually work fine in general in pip, the problem (I think?) is that setuptools itself has two different mechanisms for implementing namespace packages and they do not play nice with each other.

If I don't use the setuptools style namespace packages, I can have a filesystem like:

```
.
├── README.tmpenv
├── a
│   └── foo
│       ├── __init__.py
│       └── bar
│           └── __init__.py
└── b
    └── foo
        ├── __init__.py
        └── frob
            └──__init__.py
```

The files locates at ``foo/__init__.py`` simply have the contents:

```python
__path__ = __import__("pkgutil").extend_path(__path__, __name__)
```

This is the package foo.bar and foo.frob "installed" via flat installation to two different locations (``a/`` and ``b/``) as far as I can tell. If I then do:

```
$ PYTHONPATH=a:b python -c "import foo.bar,foo.frob; print(foo.bar.__file__); print(foo.frob.__file__)"
/Users/dstufft/.virtualenvs/tmp-685862d013d0c8b/a/foo/bar/__init__.py
/Users/dstufft/.virtualenvs/tmp-685862d013d0c8b/b/foo/frob/__init__.py
```

That is a flat installed package, in two different locations, that seems to be working just fine. Am I missing something, or is this really just a limitation of setuptools that its namespace packages don't correctly support flat installed packages?

ncoghlan: I believe Jason may be referring to the distro namespace package problem: if you're using legacy namespace packages (rather than native ones),  and installing them all into a common location, then the different packages will conflict on the "__init__.py" file for the namespace.

In 3.3+, the answer to that is "don't use legacy namespace packages, use native ones". For older releases you either need to use separate install directories & .pth files to extend sys.path, or take the __init__.py files out of the packages and have the installer generate them (ala wrapper scripts).

The latter would be a good direction for us to go anyway, as it enables the migration to native namespace packages in Py3 without breaking Py2 compatibility.

dstufft: I don't think so, because neither of the two issues he referenced pypa/pip#3 or [https://bitbucket.org/pypa/setuptools/issue/250/](pypa/setuptools#250) has anything to do with distro packages, but instead they have to do with the fact that setuptools has multiple mechanisms for supporting namespace packages and the one that is used when you have setuptools install a package via ``--single-version-externally-managed`` (aka "flat", or what pip does) that conflicts when you install another package that uses the same namespace in the typical setuptools fashion (or what happens when pip does the ``setup.py develop`` for  ``-e``).

ncoghlan: Yes, distros use "--single-version-externally-managed" to create system packages, and hence have to take care of deleting the conflicting init files and moving them to a common dependency that "owns" the namespace package.

Upstream, such packages currently end up *actually* conflicting if installed with pip (whether from source or via a wheel file)

dstufft: Ok, but this entire section is comparing pip to easy_install and everything referenced is talking about pip and easy_install. I can't see anything in this PR, the pages that were edited, the links that were references, or anyone talking about distros here except for you.

ncoghlan: Yes, I'm saying Jason's description of the problem space is misleading, and that's the source of your confusion over "but pip does support namespace packages". The issue isn't with pip specifically, it's a general issue with the combination of the way legacy namespace packages work, the way "--single-version-externally-managed" works, and installing multiple contributions to a single namespace package into the same directory, which is why distro system packages hit the same problem.

The currently available solutions for eliminating the filesystem level conflict are:
* use native namespace packages in Python 3.3+ (which eliminates the conflicting `__init__.py` files entirely, but doesn't help with Python 2 namespace packages, unless you also depend on the `importlib2` backport)
* declare an "owning package" for the namespace that provides the `__init__.py` file and every other package contributing to the namespace depends on that one (this is what Linux distros do, but `setuptools` doesn't generate packages that work this way by default - each contributor to the namespace ships its own copy of the `__init__.py` file)
* install packages into their own directories and add them to sys.path via `*.pth` files (this is what `easy_install` does, but making `sys.path` longer can result in slower startup times)

As far as I am aware, there *isn't* a current readily available option that allows these helper `__init__.py` files to be generated at install time in a manner akin to the way wrapper executables are created at install time, nor am I aware of an easy way to turn off the `__init__.py` file generation for namespace packages when invoking setuptools with `--single-version-externally-managed`.

jaraco: Thanks for the discussion. You're both right. I'm not trying to capture issues with distro packages, but rather the conflict that arises between `--single-version-externally-managed` and default easy_installed packages. As pip uses the former and setuptools uses the latter, the issue arises. I don't believe it's possible to invoke pip in a way to produce the package structure in Donald's example. Am I mistaken?

To be sure, I'm not trying to impune pip, but only to capture accurately the differences. I think you're probably right that the statement here may be somewhat inaccurate or dated.  How would you write it differently to capture succinctly the issue revealed in this discussion?

dstufft: @jaraco Wouldn't that essentially be what it would be if you installed foo.bar normally and foo.frob as develop? foo.bar would be in site-packages and foo.frob would be in whatever directory the git checkout would be (added via egg-link).

jaraco: @dstufft: I don't think so, if by "installed normally", you mean installed with pip (without --egg). I think pip removes the `__init__.py` from namespace packages, so the pkgutil call wouldn't be present for foo.bar. If you installed foo.bar using distutils only, then it probably would work. I haven't seen any examples where a pip install would work for namespace packages (and hence why pip 3 is still open). If I'm mistaken, and there is a technique that works, then we should publish that instead.

jaraco:
Following dstufft's answer, I tried installing packages using pip into different homes and using PYTHONPATH to reference those homes:

```
$ python2.7 -m pip install jaraco.util -t install-here
Collecting jaraco.util
  Using cached jaraco.util-11.0.zip
Collecting six>=1.4.1 (from jaraco.util)
  Using cached six-1.10.0-py2.py3-none-any.whl
Installing collected packages: six, jaraco.util
  Running setup.py install for jaraco.util ... done
Successfully installed jaraco.util six-1.9.0
$ ls install-here/jaraco
lang	util
$ python2.7 --version
Python 2.7.10
$ python2.7 -m pip --version
pip 8.1.1 from /Library/Python/2.7/site-packages (python 2.7)
$ python2.7 -m pip install jaraco.functools -t install-there
Collecting jaraco.functools
  Downloading jaraco.functools-1.11-py2.py3-none-any.whl
Collecting backports.functools-lru-cache>=1.0.3 (from jaraco.functools)
  Using cached backports.functools_lru_cache-1.2.1-py2.py3-none-any.whl
Installing collected packages: backports.functools-lru-cache, jaraco.functools
Successfully installed backports.functools-lru-cache jaraco.functools
$ PYTHONPATH=install-there:install-here python2.7 -c "import jaraco.functools; import jaraco.util"
  File "<string>", line 1, in <module>
ImportError: No module named jaraco.functools
```

Consider another common case where some packages from a namespace are stored in site-packages but others from the same namespace are stored in dist-packages. Does pip support installing packages in that situation? Or in site-packages and user-site-packages?

```
$ python2.7 -m pip freeze
...
zope.interface==4.1.1
$ python2.7 -m pip install --user zope.component
Collecting zope.component
  Downloading zope.component-4.2.2.tar.gz (546kB)
    100% |████████████████████████████████| 552kB 1.9MB/s 
Requirement already satisfied (use --upgrade to upgrade): setuptools in /Library/Python/2.7/site-packages/setuptools-20.3.1-py2.7.egg (from zope.component)
Requirement already satisfied (use --upgrade to upgrade): zope.interface>=4.1.0 in /System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python (from zope.component)
Collecting zope.event (from zope.component)
  Downloading zope.event-4.2.0-py2-none-any.whl
Installing collected packages: zope.event, zope.component
  Running setup.py install for zope.component ... done
Successfully installed zope.component zope.event
$ python2.7 -c "import zope.interface"
  File "<string>", line 1, in <module>
ImportError: No module named interface
$ python2.7 -c "import zope.component"                       
  File "<string>", line 1, in <module>
  File "/Users/jaraco/Library/Python/2.7/lib/python/site-packages/zope/component/__init__.py", line 16, in <module>
    from zope.interface import Interface
ImportError: No module named interface
$ 
```

This shortcoming doesn't exist in environments managed by easy_install, and it's something I'd like to see solved in pip in order to effectively obviate easy_install in favor of pip. If there's better wording I could use to describe the desired behavior that's not available in pip, or if there's some way this need can be achieved in pip, please suggest it.

jaraco:
I don't think I meant to close this. It looks like maybe it got closed when I deleted the branches? I didn't think Github did that. Well, dang.

jaraco:
I see what happened with the commits. I forced master to match the upstream master, and put the commits in another branch. I'll have to re-open the PR from the new branch as Github doesn't support changing the branch after a PR has been initiated.

PR #160, opened 2015-07-06
Clarifying the data_files behaviour with pip 7.x and up

pombredanne:
Adding references to pending issues and explaining some of the differences between pip 6 and pip 7

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -357,12 +357,13 @@ For more information see the distutils section on `Installing Additional Files
 .. note::
 
   :ref:`setuptools` allows absolute "data_files" paths, and pip honors them as
-  absolute, when installing from :term:`sdist <Source Distribution (or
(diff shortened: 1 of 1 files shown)


source/distributing.rst
```diff
@@ -357,12 +357,13 @@ For more information see the distutils section on `Installing Additional Files
 .. note::
 
   :ref:`setuptools` allows absolute "data_files" paths, and pip honors them as
```
jdemeyer: This is simply not true. Setuptools does not support `data_files` in any useful way (neither with relative nor with absolute paths), see #212.

jonparrott:
Closing in favor of #212.

source/distributing.rst
pombredanne: FWIW, my point here was about pip and the changes that occurred with 7.0

PR #169, opened 2015-09-19
Installing pip/setuptools/wheel with Linux Package Managers

qwcode:


source/install_requirements_linux.rst
```diff
@@ -0,0 +1,126 @@
...
+
+
+Fedora
+~~~~~~
+
+* < Fedora 22:
+
+  * ``sudo yum install python-pip`` (use "python3-pip" for Python 3)
+  * setuptools is already installed
+  * wheel is not available
```
rkuska: If you refer to wheel as a package it's available for Fedora 21+
See http://koji.fedoraproject.org/koji/packageinfo?packageID=17411

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,126 @@
...
+
+  pip and setuptools are already installed. wheel is not available.
+
+To get newer versions of pip (and also setuptools and wheel), you can enable the
+`PyPA Copr Repo <https://copr.fedoraproject.org/coprs/pypa/pypa/>`_ using the
+`Copr Repo instructions <https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__,
+and run the following:
+
+* < Fedora 23:
+
```
rkuska: This should be changed to < Fedora 22 as dnf is the default package manager from Fedora22+.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,126 @@
...
+  ::
+
+    sudo yum install python-pip python-setuptools python-wheel
+
+  Replace "python" with "python3" for Python 3.
+
+* >= Fedora 23: same as above, but use "dnf", instead of "yum"
```
rkuska: It's easier to install packages from copr with dnf.

Install plugins for dnf:
`sudo dnf install dnf-plugins-core`
Enable copr pypa/pypa repo:
`sudo dnf copr enable pypa/pypa`
Install packages:
`sudo dnf install python-pip python-setuptools python-wheel`

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,126 @@
...
+
+* < Fedora 22:
+
+  * ``sudo yum install python-pip`` (use "python3-pip" for Python 3)
+  * setuptools is already installed
+  * wheel is not available
+
+* >= Fedora 22:
+
+  pip and setuptools are already installed. wheel is not available.
```
rkuska: This is not true for Fedora23. Fedora23 will have only Python3 installed by default (along with python3-pip and python3-setuptools).

qwcode: The Fedora23 cloud beta image has python 2.7.10 as the default Python?

rkuska: Default python for Fedora cloud image is python3 also but I think that python2 will be shipped along as ansible needs it ccing @Rorosha - fedora cloud member.

qwcode:
thanks @rkuska 

I updated the Fedora instructions.   I decided to just strictly cover the supported versions for now:  21, and 22

please correct any other mistakes.

source/install_requirements_linux.rst
qwcode: I looked at the Fedota 23 AMI, "python" was 2.7.10.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+To get newer versions of pip, setuptools, and wheel, you can enable the
+`PyPA Copr Repo <https://copr.fedoraproject.org/coprs/pypa/pypa/>`_ using the
+`Copr Repo instructions <https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__.
+
+
+CentOS/RHEL
+~~~~~~~~~~~
+
+CentOS and RHEL don't offer :ref:`pip` or :ref:`wheel` in their core repositories,
+although :ref:`setuptools` is installed by default.
```
soltysh: That's not true anymore, at least for 3.4, see how we install python-3.4 through SCL for [openshift/python-34-centos7](https://github.com/openshift/sti-python/blob/master/3.4/Dockerfile#L20) the same applies to [RHEL image](https://github.com/openshift/sti-python/blob/master/3.4/Dockerfile.rhel7#L26).

soltysh: I'm also not sure about setuptools being installed by default, again see above links.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+
+* Fedora 21:
+
+  * Python 2: ``sudo yum install python-pip python-wheel``
+  * Python 3: ``sudo yum install python3 python3-wheel``
+
+* Fedora 22:
+
+  * Python 2: ``sudo dnf install python-pip python-wheel``
+  * Python 3: ``sudo dnf install python3 python3-wheel``
```
soltysh: I need to confirm that but again, I'm not sure about setuptools and pip being installed by default.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+`Copr Repo instructions <https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__.
+
+
+CentOS/RHEL
+~~~~~~~~~~~
+
+CentOS and RHEL don't offer :ref:`pip` or :ref:`wheel` in their core repositories,
+although :ref:`setuptools` is installed by default.
+
+There's three options available to fill in :ref:`pip` and :ref:`wheel`, and to
```
soltysh: There are three options...

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+upgrade :ref:`setuptools`:
+
+1. Use the "Sofware Collections" feature to enable a collection that includes
+   pip, setuptools, and wheel.
+
+   * For Redhat, see here:
+     http://developers.redhat.com/products/softwarecollections/overview/
+   * For CentOS, see here: https://www.softwarecollections.org/en/
+
+   Although this has promise to be the preferred method in the future, as it is,
```
soltysh: This is current preferred way of installing newer software than available by default with RHEL/CentOS. This is how eg. openshift team is using those heavily to provide users with latest packages.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+2. You can install pip from the `EPEL repository
+   <https://fedoraproject.org/wiki/EPEL>`_. Enable EPEL using `these
+   instructions
+   <https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F>`__,
+   and install like so::
+
+     sudo yum install python-pip
+
+   Although EPEL tends to maintain a pretty recent version of :ref:`pip`, it
+   does not maintain :ref:`setuptools` or :ref:`wheel`, so you'd have to settle
```
soltysh: This sentence is true for setup tools but not for wheel, see  https://dl.fedoraproject.org/pub/epel/7/x86_64/repoview/python-wheel.html, but again that's assuming using EPEL7. Previous versions of EPEL does not contain those.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+3. You can enable the `PyPA Copr Repo
+   <https://copr.fedoraproject.org/coprs/pypa/pypa/>`_ using `these instructions
+   <https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__ [1]_, and run::
+
+     sudo yum install python-pip python-setuptools python-wheel
+
+
+Also, note that if you're using the `IUS repository
+<https://iuscommunity.org/pages/Repos.html>`_ to install alternative Python
+versions, IUS also maintains corresponding versions versions of pip, setuptools,
```
soltysh: maintains corresponding versions ~~versions~~ of pip

source/install_requirements_linux.rst
soltysh: I'm taking those back, I didn't notice *core repositories* there :wink: sorry

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+
+1. Use the "Sofware Collections" feature to enable a collection that includes
+   pip, setuptools, and wheel.
+
+   * For Redhat, see here:
+     http://developers.redhat.com/products/softwarecollections/overview/
+   * For CentOS, see here: https://www.softwarecollections.org/en/
+
+   Although this has promise to be the preferred method in the future, as it is,
+   you can't count on finding a collection that has recent versions.
```
soltysh: If you don't mind I'd say most recent versions.

source/install_requirements_linux.rst
rkuska: Yes they are. 
For python(2) it is from Fedora22 and higher.
See:
```
> $ rpm -q python --requires | grep python                                     
/usr/bin/python2.7
libpython2.7.so.1.0()(64bit)
python-libs(x86-64) = 2.7.10-4.fc22
python-pip
python-setuptools
```

For python3 it is from Fedora21 and higher.

rkuska:
@qwcode Good idea to only cover supported versions of Fedora. I am :+1: for the Fedora part (see my latest comment in code).

source/install_requirements_linux.rst
qwcode: but the pip version I'm seeing is 1.5.6, so it's hard for me to call it preferred.

https://www.softwarecollections.org/repos/rhscl/rh-python34/epel-7-x86_64/rh-python34-python-pip-1.5.6-4.el7/
https://www.softwarecollections.org/repos/rhscl/python27/epel-7-x86_64/python27-python-pip-1.5.6-5.el7/

source/install_requirements_linux.rst
qwcode: >  in Fedora22, to `sudo dnf install python-wheel` as python-pip is installed by default

I'm not seeing that on the fc22 AMI cloud image

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+~~~~~~
+
+* Fedora 21:
+
+  * Python 2: ``sudo yum install python-pip python-wheel``
+  * Python 3: ``sudo yum install python3 python3-wheel``
+
+* Fedora 22:
+
+  * Python 2: ``sudo dnf install python-pip python-wheel``
```
qwcode: I wonder if I should recommend an explicit upgrade of python-setuptools as well.
e.g. using the cloud image, the upgrade would take you from v12 to v17

source/install_requirements_linux.rst
soltysh: SCLs will never be on the bleeding edge, they will always fall behind a little bit. Python 3.4 which is over a year old, was just released as part of scl 2.0, the same is true for pip.

source/install_requirements_linux.rst
soltysh: From my local f22:
```
$ dnf search python-wheel
Last metadata expiration check performed 5 days, 18:32:38 ago on Tue Sep 15 16:34:30 2015.
====================================================== N/S Matched: python-wheel =======================================================
python-wheel.noarch : A built-package format for Python
```

source/install_requirements_linux.rst
soltysh: :+1:

source/install_requirements_linux.rst
carlgeorge: It might be best to avoid the phase "preferred" entirely.  Each approach has advantages and disadvantages, so it is best left to the users to choose which one works best for them.  Many users are turned off by the SCL workflow, despite it being Red Hat's preferred solution.

source/install_requirements_linux.rst
```diff
@@ -0,0 +1,116 @@
...
+
+
+CentOS/RHEL
+~~~~~~~~~~~
+
+CentOS and RHEL don't offer :ref:`pip` or :ref:`wheel` in their core repositories,
+although :ref:`setuptools` is installed by default.
+
+There's three options available to fill in :ref:`pip` and :ref:`wheel`, and to
+upgrade :ref:`setuptools`:
```
carlgeorge: This should be clarified.  The PyPA Copr is the only option that upgrades stock setuptools.  SCL and IUS are both parallel installs with stock, and are independent of the stock install of setuptools (and python and pip and wheel).  It is noted below that EPEL does not upgrade setuptools, as that is against policy.

source/install_requirements_linux.rst
soltysh: :+1: totally agree

source/install_requirements_linux.rst
qwcode: @soltysh the part I wasn't seeing on the fc22 AMI cloud image was "python-pip is installed by default"

rkuska: You were right @qwcode I checked and fc22 cloud image was shipped with older version of Python which didn't have a pip as a dependency so :+1:

qwcode:
ok, I think I've attempted to address all the comments...  take a look and let me know.  thanks.

soltysh:
LGTM

qwcode:
typo fixed along with one other fix.

merging....

PR #170, opened 2015-09-24
Document self hosting repositories

dstufft:


source/self_hosted_repository.rst
```diff
@@ -0,0 +1,56 @@
...
+    ├── bar
+    │   └── bar-0.1.tar.gz
+    └── foo
+        ├── Foo-1.0.tar.gz
+        └── Foo-2.0.tar.gz
+
+Once you have this layout, simply configure your webserver to serve the root
+directory with autoindex enabled. For an example using the built in Web server
+in `Twisted`_, you would simply run ``twistd -n web --path .`` and then
+instruct users to add the URL to their installer's configuration.
```
qwcode: ok, in this case, to be clear, there is no "simple" segment?  that confused me a bit, since the url example above has "simple" in the path  (https://python.example.com/simple/)

I guess be clear that a "simple" segment is not critical, and if it's not, maybe don't use it in the examples.

source/self_hosted_repository.rst
```diff
@@ -0,0 +1,56 @@
...
+:Page Status: Complete
+:Last Reviewed: 2015-09-24
+
+
+If you wish to host your own simple repository [1]_, you can either use a
+software package like `devpi`_ or you can use simply create the proper
+directory structure and use any web server that can serve static files and
+generate an autoindex.
+
+In either case, since you'll be relying on a repository that is likely not in
```
qwcode: the phrase "you'll be relying" confused me a bit

maybe  "since you'll be hosting a repository"

source/self_hosted_repository.rst
dstufft: I removed it from the example, does that resolve this?

source/self_hosted_repository.rst
dstufft: Fixed.

qwcode:
lgtm, merging

source/self_hosted_repository.rst
pfmoore: Maybe use the stdlib ```python -m http.server``` as the example, rather than Twisted?

dstufft: I used twisted because it's something you could actually use in production (it supports HTTPS, won't be vulnerable to attack via slowloris, etc). I can switch it to ``python -m http.server``, I just thought using something production grade would be better.

ncoghlan: Aye, recommending a production grade web server makes sense to me, and if "pip install twisted" is a problem, then running your own simple index may not be the best idea in the world :)

pfmoore: Does twisted work on Python 3, nowadays? It's ages since I've tried it. Also, if you're using twisted because it allows https, shouldn't you show how to do https??? That's not facetious, I wouldn't know how to set up https with Twisted - or with anything for that matter - and yet I have seriously considered a private index as behind the work firewall we can't see the internet (which is why http seems OK to me...)

Maybe include both?

dstufft: Twisted partially supports Python 3, I think the next release the command I listed above will work in Python 3 as well.

I don't think we need to go into details on how to configure the server completely. In your case HTTP would be fine if you trust the network (though HTTPS is still "better", just slightly so). I worry that going into configuring HTTPS quickly devolves into a rat hole (it requires a certificate file, do we explain how to get that? How about how to generate one?, etc). In particular though, the answer to your question is available in both the Twisted documentation and the help for the command:

```console
$ twistd -n web --path . --help
Usage: twistd [options] web [web options]
Options:
      --personal          Instead of generating a webserver, generate a
                          ResourcePublisher which listens on  the port given by
                          --port, or ~/.twistd-web-pb if --port is not
                          specified.
  -n, --notracebacks      Do not display tracebacks in broken web pages.
                          Displaying tracebacks to users may be security risk!
  -p, --port=             strports description of the port to start the server
                          on.
  -l, --logfile=          Path to web CLF (Combined Log Format) log file.
      --https=            Port to listen on for Secure HTTP.
  -c, --certificate=      SSL certificate to use for HTTPS. [default:
                          server.pem]
  -k, --privkey=          SSL certificate to use for HTTPS. [default:
                          server.pem]
  -i, --index=            Add the name of a file used to check for directory
                          indexes.         [default: index, index.html]
      --wsgi=             The FQPN of a WSGI application object to serve as the
                          root resource of         the webserver.
      --help              Display this help and exit.
      --class=            Create a Resource subclass with a zero-argument
                          constructor.
      --version           Display Twisted version and exit.
      --ignore-ext=       Specify an extension to ignore.  These will be
                          processed in order.
  -u, --user              Makes a server with ~/public_html and ~/.twistd-web-pb
                          support for         users.
      --allow-ignore-ext  Specify whether or not a request for 'foo' should
                          return 'foo.ext'
      --path=             <path> is either a specific file or a directory to be
                          set as the root         of the web server. Use this if
                          you have a directory full of HTML, cgi,         epy,
                          or rpy files or any other files that you want to be
                          served up raw.
      --processor=        `ext=class' where `class' is added as a Processor for
                          files ending         with `ext'.
  -m, --mime-type=        Specify the default mime-type for static files.
      --resource-script=  An .rpy file to be used as the root resource of the
                          webserver.

This starts a webserver.  If you specify no arguments, it will be a demo
webserver that has the Test class from twisted.web.demo in it.
```

I don't have a strong opinion on including ``http.server`` or not, really anything that serves static files with autoindex will work fine, I just picked Twisted because it's production ready and the configuration doesn't require a config file so it makes for a simple example.

pfmoore: Yeah, I realise this gets over-engineered fast. Leave it as twisted, and maybe just add "if you don't need a production-quality webserver, the http.server package in the standard library can provide http-only"

ncoghlan: I'd suggest a phrasing along the lines of ("If neither Twisted, nor any other production grade web server like gunicorn, nginx or Apache is available, then the standard library includes a basic web server that can be used to handle light loads on a trusted network: ``python3 -m http.server [port]``)

PR #175, opened 2015-10-01
Add some instructions on setting up compilers on Windows

pfmoore:


Rosuav:
+1 for having this info.

Is it possible to include actual links to the downloads, rather than simply saying "which is available from MS's web site"? Are those links sufficiently stable?

tjguk:
MS links are notoriously unstable. However, better some currently-good link (perhaps with a page title for searchability in the event of linkrot) than nothing.

pfmoore:
Agreed. I've added the links.

PR #184, opened 2015-11-12
Fix undefined label build warnings

ddbeck:
During the build process, several warnings appear about undefined labels linking to PEPs (via the PyPA intersphinx mapping). For example:

```
/Users/ddbeck/projects/python-packaging-user-guide/source/wheel_egg.rst:37: WARNING: undefined label: pypa:pep425s (if the link has no caption the label must precede a section header)
```

This pull request replaces the broken PyPA labels and hardcoded PEP URLs with `:pep:` inline text roles. After these changes, the build completes without any warnings or errors (at least for me).

### DIFF
--- a/source/current.rst
+++ b/source/current.rst
@@ -50,11 +50,10 @@ Packaging Tool Recommendations
        (which pip doesn't support).  For a detailed breakdown, see :ref:`pip vs
        easy_install`.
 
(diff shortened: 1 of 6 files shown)


qwcode:
looks good.  I had recently removed the PEPs page from pypa.io these were pointing too, but didn't double back and fix these.

Issue #182, opened 2015-11-09
Is there no preferred approach to single-sourcing the project version?


nchammas:
There are [many approaches](https://github.com/pypa/python-packaging-user-guide/blob/0dcf950850644900af322343fc32025ad8994207/source/single_source_version.rst) offered in the current documentation, but no approach seems to be "recommended" or otherwise suggested as the preferred approach to take. For a new developer looking to package their project, this can be confusing.

Looking at a few of the projects I'm familiar with, it seems that approach 3 is common.

> Set the value to a `__version__` global variable in a dedicated module in your project (e.g. version.py), then have setup.py read and exec the value into a variable.

Would it make sense to present this approach as more than just one among many? Or is there really no preferred way to single-source the project version, even for new projects?

pfmoore:
Typically any attempt to recommend a preferred solution is met with debate. There really doesn't seem to be a consensus.

FWIW, if I recall the common objection to the approach you mention here is that it doesn't work properly if your project build directory isn't on sys.path (which it may not be in certain situations).

The objections to the other options seem to me to be no stronger (but no weaker either).

nchammas:
> FWIW, if I recall the common objection to the approach you mention here is that it doesn't work properly if your project build directory isn't on sys.path (which it may not be in certain situations).

Is that a common occurrence? I'm not trying to start a debate about this now; I just ask as a total newcomer to packaging. :)

I suppose if there really is no consensus, or even no approach that we can offer as "works for most people", then we can close this issue.

pfmoore:
I don't know, it's not my objection...

qwcode:
It think approach #3 is becoming more common  (I use it now btw), but it's hard for me to say what is globally "more common", w/o doing some kind of analysis of what's being posted to pypi these days.

let's keep this issue open, and maybe #3 can become preferred at some point.

pfmoore:
BTW, my explanation of the objection I've seen is wrong (if you exec the file,you don't need it on sys.path).

One real issue is getting encodings right. The ```execfile``` approach is Python 2 only. The ```exec``` version needs an ```open()``` but Python 2 doesn't support supplying an encoding to ```open()```, and the default encoding depends on the user's PC settings. The best approach is only using ASCII in your ```version.py``` and relying on the fact that nobody's going to have a default encoding that isn't ASCII-compatible. But all it takes is a non-ASCII character in a comment and the read gets a decoding error :-(

[...]

PR #190, opened 2015-11-26
Improve single_source_version.rst

Changaco:
Solutions based on version control systems.

### DIFF
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -5,11 +5,11 @@ Single-sourcing the Project Version
 ===================================
 
 :Page Status: Complete
-:Last Reviewed: 2015-09-08
+:Last Reviewed: 2015-12-03
 
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+There are many techniques to maintain a single source of truth for the version
+number of your project:
 
 #.  Read the file in ``setup.py`` and parse the version with a regex. Example (
     from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
@@ -115,3 +115,8 @@ There are a few techniques to store the version in your project code without dup
     ``sample/__init__.py`` imports packages from ``install_requires``
     dependencies, which will very likely not be installed yet when ``setup.py``
     is run.
+
+
+#.  Keep the version number in the tags of a version control system (Git, Mercurial, etc)
+    instead of in the code, and automatically extract it from there using
+    `setuptools_scm <https://pypi.python.org/pypi/setuptools_scm>`_.


source/single_source_version.rst
```diff
@@ -115,3 +119,13 @@ There are a few techniques to store the version in your project code without dup
     ``sample/__init__.py`` imports packages from ``install_requires``
     dependencies, which will very likely not be installed yet when ``setup.py``
     is run.
+
+Using version control
+=====================
+
+#.  `version.py <https://github.com/Changaco/version.py>`_ is a small script that
+    extracts version numbers from git tags.
```
alexwlchan: Nitpick: "Git" is a proper noun; shouldn't it be capitalised?

Changaco: I guess so. I've made the change.

nchammas:
As a Python end-user, I dislike this change because it offers yet more options for me to choose from, as opposed to whittling things down to one or two clear choices. 

Single-sourcing the project version is such a basic part of packaging that I would assume the general Python community would prefer to converge on a standard solution rather than see things go off in many directions.

Per @qwcode [over in #182](https://github.com/pypa/python-packaging-user-guide/issues/182#issuecomment-155196130), it seems like the community is slowly converging on option 3 in the current guide:

> Set the value to a `__version__` global variable in a dedicated module in your project (e.g. `version.py`), then have `setup.py` read and `exec` the value into a variable.

I would rather see the guide evolve slowly in that direction than evolve to add more options.

Changaco:
In my opinion a solution based on VCS tags should be one of the recommended approaches, so adding them to the document is a necessary precondition to building a consensus in #182.

pfmoore:
I agree that we want fewer options, not more. The ```pbr``` project can explain the benefits of using ```pbr``` in their own documentation, I don't see the PUG as the right place to promote a 3rd party project. And your "small script" is either going to grow into another project in its own right (to support Mercurial, and alternate tagging conventions) or it'll simply be example code suitable for use by people with enough knowledge that they don't really need the sort of advice the PUG is offering.

I'd prefer that we keep it simple, and *don't* offer recommendations for how to integrate version control into the package management process.

Changaco:
Simply ignoring version control systems doesn't seem like a good strategy to me. If it isn't mentioned at all in the document then it's likely that others like me will open issues/PRs asking to complete it.

pfmoore:
It's not so much ignoring them as assuming that people with a sufficiently advanced VCS workflow that they want to derive versions from tags, will be capable of rolling their own solution.

Maybe we could add a comment to the current option 2 ("Use an external build tool...") to say that some such tools (e.g. pbr) derive the version number from VCS tag data, and leave it at that.

Changaco:
It seems to me that option 2 is badly worded (they're release tools, not build tools) and somewhat hidden in the middle when it should be a separate section. How about merging option 2 with the new section this PR proposes to add?

Changaco:
Since this PR has become tangled with #182 I've modified it to try to solve that issue.

pfmoore:
I don't particularly like the rewrite (for no specific reason, sorry - it just doesn't feel helpful to me), but at a minimum, can we include the script you're proposing for extracting git tags in the text? (BTW, you lost the note that version 1 needs you to deal with regular expressions...)

Changaco:
> can we include the script you're proposing for extracting git tags in the text?

Why? Adding the script and the instructions that go with it would add 50+ lines, and we'd have to keep them updated.

pfmoore:
Largely because it would be in line with the other sections, which include code inline. The code here is basically sample code we'd expect people to copy into their project, so it should be short & understandable. The script in this case is probably too big for that, which makes me think that it would be better packaged (and described) as a 3rd party project like pbr, rather than as an example of "how to do this".

One of my problems with this whole proposal is that basing the version on VCS tags is a bit too complex to recommend that people do it by hand. The fact that the sample script is too big to be plausibly included in the page just illustrates that point.

Changaco:
To clarify, my `version.py` script can't be packaged, it needs to be copied into the project that uses it. As such it's not an external tool, and it only handles the version number instead of the whole release process, so that's why I put it in its own item.

Changaco:
I've restored the note about regexps.

Changaco:
As I was thinking about packaging my script I stumbled upon the [setuptools_scm](https://github.com/pypa/setuptools_scm/) package, which is close to what my `version.py` does, except that it doesn't support git archives (https://github.com/pypa/setuptools_scm/issues/65).

I've modified this PR to link to `setuptools_scm` instead of my script, since it's a more generic solution.

pfmoore:
But ```setuptools_scm``` is just another 3rd party package much like ```bumpversion``` or ```pbr```, so it doesn't need its own section - just add it to the existing list (or don't - nobody said that list was intended to be exhaustive).

Changaco:
Like my `version.py`, `setuptools_scm` is not a release tool, it simply extracts the version number, that's why I didn't put it with `bumpversion` and the others.

Changaco:
I've made a few more changes.

@nchammas You opened #182, can I have your opinion on whether [the document as it is on this branch](https://github.com/Changaco/python-packaging-user-guide/blob/patch-1/source/single_source_version.rst) is better than the current one?

nchammas:
It reduces the number of categories of options presented to the user, which is good.

I can't comment more specifically other than to repeat, as we've been discussing in #182, that I hope this eventually evolves into something much simpler than what even this change proposes. But as a random user, I think this direction is good.

Changaco:
Thanks @nchammas. The document could be simplified further by dropping option 1, but @pfmoore has voted against that in https://github.com/pypa/python-packaging-user-guide/issues/182#issuecomment-160167053.

The other remaining issue is whether `setuptools_scm` deserves its own item. @pfmoore do you still think it needs to be grouped with the release tools?

pfmoore:
In my view the whole section is far too complex for the audience it's intended for anyway. It's aimed at addressing people who say "but I don't like the fact that the basic recommended approach asks me to maintain my version number in more than one place, what should I do?". Fundamentally, the answer is "it depends, you need to understand the trade-offs and devise a solution that suits you".

This section is basically showing a number of options, but it's still aimed at people who *don't* yet have the knowledge to assess things properly - that's why it's so frustrating that there isn't a "best practice" we can recommend. (But choosing something that a significant number of people disagree with is even more confusing for the newcomer, so we stick with what we have). So I see little point in separating out nuances of build tool vs release tool. The message for all of them is "let a 3rd party package do the job for you". Once someone knows the implications of choosing a "release tool" over a "build tool" or whatever other distinction is implied, they probably don't need this page any more.

FWIW, *I* don't understand the distinction you're trying to make - that's probably because I only have small single-maintainer projects to deal with, and so it's irrelevant to me. I don't want an explanation - until I have the problem, the explanation isn't important.

I don't really have anything more to say, so I'm going to drop out of the discussion at this point.

Changaco:
Okay, so we need someone else with write access to step in. Ping @dstufft @qwcode @xavfernandez (owners who commented on #182).

xavfernandez:
No real opinion except to keep all older options documented :)

Why did you remove the solution based on pkg_resources ?

Changaco:
Because it's not a reliable solution, it returns the number of the installed version, not the number of the version you're actually running.

xavfernandez:
Well the only version that is mandatory for packaging is the one from the metadata, i.e. the one returned by pkg_resources.

Nothing mandates that you add a duplicate information inside your python code to allow things like ` from my_pkg import VERSION`.

Changaco:
If you don't need the information in more than one place then you don't need a guide on how to single-source it.

xavfernandez:
Well the guide supposedly presents `a few techniques to store the version in your project code without duplicating the value stored in setup.py`. And one solution is to only store it in setup.py and access it via `pkg_resources`.
And since it is IMHO the cleanest solution, it should definitely stay in the guide. The only drawback being that `pkg_resources` is not in the standard library and might not be installed.

Changaco:
The `pkg_resources` technique doesn't "store the version in your project code". It's not a solution for you as project maintainer, it's a solution for people who are using your package and want to get its version number.

pfmoore:
Please note - this section is intended to enumerate the options available. We don't try to provide judgments. There are projects who use the ```pkg_resources``` technique, so it's reasonable to include it here, regardless of your opinion on its correctness. If there was a community consensus, we'd state that, but there isn't, so it's just a list of options people can choose from.

I'd suggest scaling back on the scope of this PR to simply add the extra option of getting the version from a VCS. That's relatively non-controversial (as long as you keep the text objective and don't try to claim one option is better than another). Just add a new section with your VCS option, and stop trying to adjust the existing text.

Changaco:
*Sigh*. I did that, a week ago, you and @nchammas both said that you wanted fewer options.

What you just said contradicts what @nchammas said in https://github.com/pypa/python-packaging-user-guide/issues/182#issuecomment-159998411 (which you acquiesced):

> The PUG is a guide, not an encyclopedia. Its purpose is to present common advice to users facing common problems, not to document every possible permutation that people have found useful.

pfmoore:
Your original version split the list into two parts "using version control" and "storing it in the code". I'd rather a single undifferentiated list. If all you'd done was add a new item to the existing list, I don't think your proposal would have caused so much debate.

I would still prefer there were fewer options, but what *I* meant by that was that I didn't see any great value in adding the option of getting the version from the VCS (it's an advanced enough technique that I feel that anyone who wants to do that is probably competent to work out how without needing a guide like this). Specifically I view "not to document every possible permutation that people have found useful" as arguing that *not* adding your proposed approach is entirely acceptable - that's what I was agreeing with.

My preference is to simply not merge this PR at all, but if it were just an extra option added to the list with no change to the existing text, I'd probably be willing to merge it, just because you feel strongly enough that it's worth mentioning to have put all the effort in.

I do appreciate the effort you put into trying to reword things to cram more options into a shorter format, but IMO the results didn't end up being an improvement.

Anyway, as I said, I'm not planning on doing anything further on the PR in its current form, so I'll leave it at that (once again...)

Changaco:
I've scaled back this PR, please review.

pfmoore:
I see no problem with this version, so I'll merge. Thanks for all the effort dealing with the feedback, I appreciate how exhausting it can get.

PR #193, opened 2015-12-03
Improve code snippets in single_source_version.rst

Changaco:
These are all the code improvements that were in #190 before it got scaled back, remade in small commits.

### DIFF
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -47,7 +47,12 @@ number of your project:
         version = {}
         with open(os.path.join(base_dir, "module_name", "version.py"), 'rb') as fp:
             exec(fp.read(), version)
-        # later on we use: version['__version__']
+
+        setup(
+           ...
+           version=version['__version__']
+           ...
+        )
 
     Example using this technique: `warehouse <https://github.com/pypa/warehouse/blob/master/warehouse/__about__.py>`_.
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -59,7 +59,7 @@ number of your project:
         base_dir = os.path.dirname(__file__)
         about = {}
         with open(os.path.join(base_dir, "module_name", "__about__.py"), 'rb') as fp:
-            exec(fp.read(), about)
+            exec(fp.read(), {}, about)
 
         setup(
            ...
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -43,8 +43,9 @@ number of your project:
     your project (e.g. ``version.py``), then have ``setup.py`` read and ``exec`` the
     value into a variable::
 
+        base_dir = os.path.dirname(__file__)
         version = {}
-        with open("...sample/version.py", 'rb') as fp:
+        with open(os.path.join(base_dir, "module_name", "version.py"), 'rb') as fp:
             exec(fp.read(), version)
         # later on we use: version['__version__']
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -14,6 +14,9 @@ number of your project:
 #.  Read the file in ``setup.py`` and parse the version with a regex. Example (
     from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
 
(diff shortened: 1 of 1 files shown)


source/single_source_version.rst
```diff
@@ -14,19 +14,18 @@ number of your project:
...
-            version_file = read(*file_paths)
-            version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]",
-                                      version_file, re.M)
-            if version_match:
-                return version_match.group(1)
+            # We read __version__ instead of importing it because that would fail if
+            # the module it's in tries to import things from dependencies that aren't
+            # installed yet when setup.py is being run.
+            # Moreover, imports can fail if sys.path isn't properly set.
+            with open(os.path.join(os.path.dirname(__file__), *file_paths), 'rb') as f:
```
qwcode: can we add some explanation of the use of 'rb' and 'br' to convince readers (and me) that this will always work.

pfmoore: I don't see this change as an advantage. It's technically correct, I guess, but the files in question are actually text files, and reading them as such seems more logical (even if it does mean you need to think about encodings).

Binary reading a text file just to avoid thinking about encodings feels like a hack (and the sort of thinking that the string changes in Python 3 were explicitly designed to discourage).

Changaco: What we're trying to extract, i.e. the version number, only uses a subset of ascii characters, so we don't need to deal with unicode. 'rb' tells python to read the file as bytes without trying to decode them, and since we're reading bytes we need the regexp pattern to be bytes (otherwise python 3 will raise `TypeError: cannot use a string pattern on a bytes-like object`), thus 'br'.

Changaco: Would this be satisfactory?

    # Since version numbers only use a subset of ascii, we can read the file
    # as bytes, thus avoiding unicode decode errors.

Changaco: Comment added.

Changaco: Note: I did search the stdlib for a function that properly decodes a python file (i.e. by looking for the `# coding:` header), but I couldn't find one, so reading the file as bytes is easier.

ncoghlan: These exist in Python 3 (as tokenize.open() and tokenize.detect_encoding()), but the 2/3 compatible code, assuming ASCII compatibility and reading the bytes is easier. We assume ASCII compatibility in other aspects of the packaging ecosystem (most notably distribution package names), so it's a reasonable assumption to make here as well.

Changaco: Thanks @ncoghlan :-)

Changaco:
If there are no other concerns then this is ready for merge.

Changaco:
Ping.

source/single_source_version.rst
```diff
@@ -48,25 +49,23 @@ number of your project:
...
-
-    Using ``exec``:
-
-    ::
+        # We exec the module instead of importing it because imports can fail if
+        # sys.path isn't properly set.
+        base_dir = os.path.dirname(__file__)
+        about = {}
+        with open(os.path.join(base_dir, "module_name", "__about__.py"), 'rb') as fp:
+            exec(fp.read(), about)
```
Ivoz: Why are you removing the Python 2 compatible example?

Changaco: > storing in locals might be better? `exec(fp.read(), {}, about)`

Why? By separating globals and locals it's no longer obvious that the variables will end up in the right dictionary.

pfmoore: Because globals gets cluttered with all the various builtins. Locals will just contain what ```__about__.py``` defines. It may not matter much for this specific use case, but if we're recommending code we should follow good practice (and  using locals is good practice here, IMO).

Changaco: Okay, that makes sense, I've made the change.

Changaco:
Anything else?

Changaco:
It looks like we all forgot about this PR. Is someone willing to merge it?

Changaco:
Okay, I give up.

Issue #194, opened 2015-12-03
[tracking] new tutorials and other revisions


ddbeck:
This issue is to track work on writing new, introductory tutorials for the guide, and other activities that are part of an effort to expand and revise the guide. [See this thread from the distutils-sig](https://mail.python.org/pipermail/distutils-sig/2015-November/027776.html) for more details about this project. Work on these tutorials will (at least initially) take place on the `develop` branch.

I am inviting the group of volunteers to use this issue to see what other issues are available and to use this as a place for general questions and help (i.e., questions that are not issue-specific), to minimize noise on other issues and the creation of new issues.

If you'd like to contribute to these sections, please see the (to be) linked issues for details. Additionally, please review the [developing guide to contributing](https://packaging.python.org/contribute/). If you start work on an issue, please consider saying so on the relevant issue, so others don't duplicate your work. Thanks!
## Outline

Broadly, the new intro to packaging tutorials have the following outline:
1. [Who is this for?](https://github.com/pypa/python-packaging-user-guide/issues/195)
2. [Prerequisites](https://github.com/pypa/python-packaging-user-guide/issues/196)
3. [What's a package?](https://github.com/pypa/python-packaging-user-guide/issues/199)
4. [Where do packages come from?](https://github.com/pypa/python-packaging-user-guide/issues/200)
5. [Install your first package](https://github.com/pypa/python-packaging-user-guide/issues/201)
6. [Use a virtual environment](https://github.com/pypa/python-packaging-user-guide/issues/202)
7. [Make your first package](https://github.com/pypa/python-packaging-user-guide/issues/203)
8. [Publish your first package](https://github.com/pypa/python-packaging-user-guide/issues/204)

Issues will soon be created to address each section specifically; the outline above will link to each section's issue.
## Bite size tasks
- #207: Add build instructions to the README
## Extra style notes

The intro to packaging tutorials have some extra style considerations on top of that of the guide generally:
- In contrast to more task-focused writing, the tutorials are meant to build fundamental packaging skills, rather than resolve a specific problem. With that in mind, try to invite readers to learn to do something without regard for its specific applications. So instead of something like "Read this section if you want to share a package…" try "After you complete this section, you'll be able to share a package…"

souravsingh:
I would like to work on the first two parts of the tutorial.

ddbeck:
[...]

PR #198, opened 2015-12-06
[WIP] Intro to Packaging tutorials

souravsingh:
Created the introduction page to Packaging tutorials.If any additions are needed,please feel free to tell me.

### DIFF
--- a/source/intro-tutorial.rst
+++ b/source/intro-tutorial.rst
@@ -8,7 +8,9 @@ Who is this for?
 ~~~~~~~~~~~~~~~~~
 This tutorial is for Python programmers who are at an "advanced beginner" level. There are broadly two classes of Python programmers who stand to benefit from these tutorials.
 
-You are a Python programmer who is looking to use a third party Python package. You should be reasonably comfortable making use of basic language features like classes, functions, conditionals and loops. In addition, you should know how to import classes and functions from standard library module and the what is meant by third party packages.
+The Python users who are looking to use a third party Python package. You should be reasonably comfortable making use of basic language features like classes, functions, conditionals and loops. In addition, it is expected that the user knows how to import classes and functions from standard library modules.
+
+The second category of Python programmers these tutorials aim to address are those who have built something useful in Python and are looking to share their work with the wider Python community by building and distributing their program as a package.
 
 
 What will be covered in this tutorial?
--- a/source/intro-tutorial.rst
+++ b/source/intro-tutorial.rst
@@ -6,12 +6,10 @@ This section aims to introduce people to Packaging of Python Scripts.
 
 Who is this for?
 ~~~~~~~~~~~~~~~~~
+This tutorial is for Python programmers who are at an "advanced beginner" level. There are broadly two classes of Python programmers who stand to benefit from these tutorials.
 
-This tutorial is for Python users who are at an "advanced beginner" level.This means that the users of Python can import a standard Python library and can make use of basic language features like ``if``, ``for`` and other such loops and can also differentiate between them.
+You are a Python programmer who is looking to use a third party Python package. You should be reasonably comfortable making use of basic language features like classes, functions, conditionals and loops. In addition, you should know how to import classes and functions from standard library module and the what is meant by third party packages.
 
-
-This tutorial is also aimed at people with an intermediate level experience in Python and want to make and distribute 
-a package for the first time,for Python users who need a refersher and advanced Python users who want to take on specialized packaging tasks.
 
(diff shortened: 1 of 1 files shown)


souravsingh:
Thanks a lot @amitsaha. I have implemented a few of your suggestions.

ddbeck:
@souravsingh Thanks so much for diving into this, Sourav! This is a great start. Having something concrete to think about has really clarified some things about how to write this introductory material. But that also means I came up with a bunch of ideas for ways to change the section (and ways to update the style guide and so on). I don't mean to dump a to-do list on you here; rather, I'm going to write down some ideas and you can choose to work on what you like, or we can merge the changes and I (or others) can carry on from there.

(And now that I'm writing this stuff down, I see that you've made some changes based on @amitsaha's comments already. Thanks to you both!)

Content idea: Newer Python users may lack context or confidence in their abilities (e.g., being able to use `if … elif … else` is different from knowing to call it a conditional). In addition to _describing_ the skills they need, I'd like to _show_ them too, as a way for readers to test-in to the tutorial, rather than opt-out. For example, we might show a short script or interactive session (in which something is `import`ed and used), and then tell the reader that if they understood most of it then they're ready to continue (and maybe a link to a language-learning resource if they don't).

Style ideas:

* I'd like to address the reader more directly. So rather than the current frame of "This tutorial is for Python programmers…" I'd prefer to see something more along the lines of "This tutorial is for you if you'd like to learn to…"
* On a related note, I'd like to be careful about talking abstractly about "programmers" (because lots of people [don't usually identify as programmers](https://packaging.python.org/en/develop/contribute/#audience)—myself included). Switching to the second person makes it easier to avoid, too.
* Mirroring something I've added to the style notes on #194:  We should invite the users to build skills (e.g., "After you complete this tutorial, you'll be able to…" ) rather than expect them to already have a specific need (e.g., "those who […] are looking to share their work").

Thanks again!

souravsingh:
About the Python Learning resource,I have a doubt about it
I am thinking of adding Hitchhiker's Guide to Python, is it good for novice resource?

ddbeck:
Yeah, I'm unsure too. I like the Hitchhiker's Guide, but it's definitely aimed at people who already know quite a bit about programming, just not a lot about _Python_.  The Python.org home page links to [this page](https://wiki.python.org/moin/BeginnersGuide), which I don't love, but it could be a good placeholder until someone comes along with a better suggestion.

ncoghlan:
For a Python Learning resource, I'd agree with just linking to the Beginner's wiki page for now, as interested folks can tweak that by asking for edit permissions as described in https://wiki.python.org/moin/FrontPage#use

Longer term, I suggested at the language summit a couple of years ago that it would make sense to split the main Python tutorial and the HOWTO guides out from the rest of the Python documentation, as the core developers of the reference interpreter aren't really the best people in the world to be maintaining a beginner's guide, and for both the core tutorial and the HOWTO guides, it's likely better to cover variations between versions inline within the text, rather than requiring people to choose the correct version up front.

There's a current discussion on replacing CPython's current Mercurial hosting setup with GitHub or GitLab (or a combination of the two), and once that's resolved, it becomes more feasible to talk about creating a separate docs repository for version independent guides. That would then create an opportunity to repeat what's being done here with the packaging tutorials: use the opportunity to get technical writers directly involved in the maintenance process.

amitsaha:
@ddbeck I like the idea of having some example code and making the reader comfortable. I am trying to think of a good example - simple but useful. Is there any specific package in your mind that we will possibly ask the reader to install as they progress through the documentation? Ideally, if we can give an example now - using a module from the standard library to do something and then using a third party package to do it better/or an advanced version of it, that would good I think.

ncoghlan:
One possible example would be the https://pypi.python.org/pypi/parse library lets you use str.format() style format strings to parse input, instead of producing output.

For an example that involves importing in the standard library version without introducing new concepts like regular expressions, then the standard library's pprint module could be compared with David Wolever's pprintpp.

For a more complex example that shows the benefits of wheel files, then the standard library re module could be compared with Matthew Barnett's regex module. However, that shouldn't be the main example, since it introduces both regular expressions and binary dependencies.

amitsaha:
Great suggestions @ncoghlan !

souravsingh:
Thanks for the suggestions @ncoglhan. I will start working on it.

jonparrott:
@souravsingh hi! This PR is now quite old and hasn't been touched in a while. I'm going to close, but if you still want to get this merged please let me know and I'll be happy to review and merge. :)

Issue #200, opened 2015-12-07
New tutorials: write section to describe sources of packages


ddbeck:
(For more information about this issue, please see #194).

We need a short section to follow the "What's a package?" section (see #199) (i.e., as its own file). It should satisfy the following goals (in no particular order):

* explain that packages can be distributed through many channels (e.g., downloaded from PyPI with a tool like `pip`, an archive downloaded from a website, a VCS repository)
* explain that PyPI is an extremely popular way of getting packages but not the only way
* explain that PyPI is an index of packages, but not the only possible index
* illustrate the point with an example (maybe mention a popular package, then describe how it can be found on PyPI, in a VCS repository, and downloaded from the project's website)

In short, this section should answer the question, _Where do packages come from?_

When opening pull requests for this issue, please submit to the `develop` branch. If you have questions that concern contributing more generally, please use issue #194. Otherwise, feel free to comment with questions or feedback. Thanks!

duims:
I'd be down to write this!

ddbeck:
@duims Great, thanks! Feel to get started and please comment if you run into any roadblocks.

PR #206, opened 2015-12-10
Created Package Sources section

duims:
Happy to hear comments- I'm not sure if the section on downloading the archive is too specific to Django or whether it's too detailed. 

Refer to issue #200 for goals.


ddbeck:
@duims This is a great start, thank you so much! I think you got the tone I was hoping for exactly. :100:

That said, I think you may be right that it's a little too detailed. I think because we're going to have more detailed instructions to download and install packages in subsequent sections (i.e., the reader will be explicitly asked to install them themselves), that we don't need to go into detail about the mechanics of downloading and installing in this section.

Would you like to update your fork to trim it down to just calling attention to the existence of the PyPI page, Git repo, and project website (and linking to each in turn)? I think that would be a good way to go, if you have the time available.

(Once we get the overall outline worked out, we can focus on some more detailed things like @amitsaha pointed out, but we don't need to worry about them yet.)

jonparrott:
@duims this PR is now very old and hasn't been updated in some time. If you still want to make this change, please feel free to open a new PR and cc me and I'll review and merge. :)

PR #208, opened 2015-12-18
Update distributing.rst "Development mode" section

ninz:
Installing project and dependencies in editable mode using local directories, not vcs, using current instructions does not work. The local directory is found, but then the dependency is overridden and fulfilled by PyPI. Changing the order of the editable mode dependencies (i.e. putting editable mode dependencies which reside in local directories at the top of the requirements file) fixes this problem.


xavfernandez:
This looks like a bug in pip (the different behavior between vcs/files editables), is there a related issue that could be linked ?

jonparrott:
Thanks, @ninz!

Issue #202, opened 2015-12-08
New tutorials: using a virtual environment


ddbeck:
(For more information about this issue, please see #194).

We need a short section to follow the install/uninstall section (see #201) (i.e., as its own file). It should satisfy the following goals:
- show the reader to how to install virtualenv (if applicable, Pythons 3.3+ should come with pyenv, I think)
- show the reader how to create a virtual environment
- show the reader how to install something inside that environment (and verify that installing worked, as in the previous section)
- show the reader how to destroy the environment

In short, we want to equip the reader with the skills to install and use packages isolated from the system's site packages.

When opening pull requests for this issue, please submit to the `develop` branch. If you have questions that concern contributing more generally, please use issue #194. Otherwise, feel free to comment with questions or feedback. Thanks!

amitsaha:
@ddbeck Can you please assign this one to me?

ddbeck:
@amitsaha feel to free to start on this (for some reason GitHub won't let me literally assign the issue to you, but don't let my confusion with the UI slow you down)

amitsaha:
@ddbeck  Great, as I commented in https://github.com/pypa/python-packaging-user-guide/pull/198 , can we decide on one third party package that we will use through out the docs? This can also be used in https://github.com/pypa/python-packaging-user-guide/pull/206 and elsewhere.

ddbeck:
@amitsaha I liked @ncoghlan's suggestion of using the `pprintpp` package. But feel free to choose something that you're comfortable working with now—a draft is easier to edit than an idea. We don't have to commit to having the same sample throughout or we can work on achieving consistency through the editing process, so don't worry much about matching up with the other sections just yet.

amitsaha:
@ddbeck  Just a note that I am slowly working on this at https://github.com/amitsaha/python-packaging-user-guide/blob/amitsaha_issue202/source/using-a-virtualenv.rst 

I have decided to use a separate package, because I felt this is something simpler.

ddbeck:
Great, keep up the good work and be sure to open a pull request when you're ready to entertain feedback.

amitsaha:
Thanks, i have opened a PR for preliminary feedback: https://github.com/pypa/python-packaging-user-guide/pull/211

ddbeck:
The PR has been merged to develop, plus some style changes: https://packaging.python.org/en/develop/using-a-virtualenv/

[...]

PR #211, opened 2016-01-05
Virtualenv tutorial (issue #202)

amitsaha:


source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
+================
+Using virtualenv
```
ddbeck: So this is going to be something that comes up in a couple places: in the general case, we should refer to "virtual environments" and only refer to "virtualenv" for Python 2 users and "venv" (and `pyvenv`) for Python 3 users. They both create virtual environments, but the tools have different names.

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+================
+
+By default, when we use ``pip`` to install packages, it will install
+it in a default location where all other packages are installed. This
```
ddbeck: A minor nitpick: "install them in a" for plural agreement

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+default behaviour may however not be desirable. The solution is to use
+a tool called `virtualenv
```
ddbeck: So here's an example of what I mean. It would probably help to wait to mention the tools until after this paragraph, so the Python 2/3 agnostic concept of creating an virtual environment is explained (and then maybe retitle the next subsection).

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+<https://virtualenv.readthedocs.org/en/latest/>`__. It allows us to
+create isolated **virtual environments** in which we can install our
+packages. You can think of isolated environments as a sandbox. By
+having a separate virtual environment for each of your projects, you
+will keep the packages installed in each separate from one another.
+
+Installing virtualenv
+---------------------
+
+If you are using Python 3, ``virtualenv`` should already be installed
```
ddbeck: "venv" instead of "virtualenv" for Python 3

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+and accessible via the ``pyenv`` command. If you are using Python 2,
```
ddbeck: "pyvenv" instead of "pyenv"

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+       1            2
+       5            2
+       7            2
+       8            2
+      10            2
+       2            1
+       4            1
+
+
+Feel free to explore the ``tabulate`` package more to see the
```
ddbeck: I'd make this a link to Tabulate's docs, as in your first mention.

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Once we are done working in our virtual environment, the
+``deactivate`` command will deactivate the virtual environment for us:
+
+.. code::
+
+   (tabulate) $ deactivate
+
+The same command will work on Windows as well.
+
```
ddbeck: One more thing I was hoping to cover in this section was deleting a virtual environment when it's no longer needed. If you want to add that, it would be great, but if you want to leave it to someone else, that's fine too. :smile:

ddbeck:
Thanks, @amitsaha! This is a great start. I'm very happy with you've done. I've made a few comments (a few minor things, one or two substantive things). Would you look over my comments and let me know if you have any questions? Thanks!

source/using-a-virtualenv.rst
dstufft: FWIW virtualenv still works on Python 3 and it's likely folks might still want to use that over ``pyvenv``. Reasoning:

* ``pyvenv`` only exists in 3.3+, so if you want to use 3.2 you're stuck.
* ``pyvenv`` only installs pip in 3.4+, so if you want to use 3.3 you have to get your own pip.
* ``pyvenv`` is tied to a specific version of Python, so you'll get different versions of pip (and thus different behavior from pip) depending on what version of Python you're using.
* ``pyvenv`` is tied to a specific version of Python, so you'll get different options, flags, etc available based on the Python version.


Longer term, virtualenv will be rewritten to take advantage of the isolation mechanism provided by ``pyvenv`` but still provide it's own pip and such to make it easier to have something creating a virtual environment with an up to date copy of the bundled tooling *and* that works the same across all versions of Python.

ddbeck: @dstufft Oh, those are some important distinctions (I didn't notice pyvenv wasn't in Python until 3.3)! In light of that, I'm tempted to say we should suggest virtualenv for all readers (Python 2 and 3 users alike), in the interest of having it Just Work. Is there a good reason _not_ to use virtualenv on Python 3.3+?

dstufft: Currently virtualenv doesn't take advantage of the better isolation mechanism introduced by pyvenv, it still uses the total black magic hack which can break sometimes (particularly on brand new versions of Python). Also it doesn't come with Python.

The first of those will hopefully be solved at some point in the future, the second of those will always be true.

source/using-a-virtualenv.rst
pfmoore: On Windows, ```pyvenv``` won't be in ```PATH``` by default. It would be better to invoke the Python 3 ```venv``` module as ```python -m venv``` rather than using the ```pyvenv``` wrapper.

source/using-a-virtualenv.rst
pfmoore: At the moment, "use ```virtualenv```" is probably the simplest "one solution suits everyone" advice. But I'm uncomfortable about not letting people know that Python 3 has a better native virtual environment facility. I suspect my discomfort merely reflects that I'm not a good person to write beginner-focused documentation, though :-)

ddbeck: OK, after sleeping on this one, I think we should go with "use virtualenv" for Python 2.7-3.2 users and "use pyvenv" for 3.3+. As much I'd like to go with a one-size fits all solution, I think there's something to be said for acknowledging that Python is changing thing. And if test readers say that doesn't work for them, then we can always change it.

amitsaha: Cool, thanks  all for the comments, I will address them in my next update.

amitsaha:
@ddbeck Please take a look at the updated doc now. Thanks for all the tips.

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,184 @@
...
+   $ source ~/work/virtualenvs/tabulate/bin/activate
+
+On Windows, the command to activate the virtual environment is:
+
+.. code::
+
+   > \work\virtualenvs\tabulate\Scripts\activate
+
+Now, we are in the virtual environment. To see the packages currently
+installed in our virtual enviroment, we can use the ``pip3 list``
```
Ivoz: We want `pip` instead of `pip3`.

ddbeck:
@amitsaha I'm very happy with this new draft! Thank you for making all those revisions! There are a few style guide-related things I'd like to revise, but they're nitpicks. I'll figure out how to merge your changes to `develop`, then make those style changes and, while I'm at it, fix the issue that @Ivoz noted.

Thanks again!

source/using-a-virtualenv.rst
```diff
@@ -0,0 +1,184 @@
...
+------------
+
+If you are using Python 2 or Python 3 (upto and including Python 3.2),
+install virtualenv using ``pip install virtualenv`` or ``pip3 install
+virtualenv`` respectively. You will then be able to create and work
+with virtual environments via the ``virtualenv`` command.
+
+If you are using Python 3.3 or a more recent version, support for
+creating virtual environments should already be installed and
+accessible via the ``pyvenv`` command.
```
pfmoore: This isn't true everywhere - on Windows at least, ```pyvenv``` is not on the path by default. I would suggest ```python -m venv``` as the canonical way of  running ```pyvenv```.

amitsaha:
Thank you all for the comments. @ddbeck do you mean you are happy with this for now?

ddbeck:
@amitsaha Yes, definitely! I had planned to set up a table of contents and merge your work this week but things sort of got away from me. I hope to catch up on this soon. Thanks for your patience. 😄

amitsaha:
@ddbeck No problems :+1: Looking forward to the docs.

ddbeck:
@amitsaha I have merged your changes (2e1d5aa0fb9ff5dac8f0c6639251cb308fc78e9a, 9682448de7ead14b5d20cf8cc849555c12910ff7) to `develop`! :tada: Thanks again for your patience!

Since I have merged the changes manually, I am going to close this PR. If you want to make any changes to this or other sections, feel free to start a new PR. 😄

amitsaha:
Thanks @ddbeck - looks great.

PR #212, opened 2016-01-13
Fix section on data_files

jdemeyer:
1) the fact that `data_files` might be installed in `sys.exec_prefix` is simply false, see https://bugs.python.org/issue25592

2) `data_files` is only fully supported by distutils. Setuptools doesn't support it at all and wheel supports only relative directories (i.e. relative to `sys.prefix`).

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -332,9 +332,7 @@ cases you may need to place data files *outside* of your :term:`packages
 
 Each (directory, files) pair in the sequence specifies the installation
 directory and the files to install there. If directory is a relative path, it is
-interpreted relative to the installation prefix (Python’s sys.prefix for
-pure-Python :term:`distributions <Distribution Package>`, sys.exec_prefix for
-distributions that contain extension modules). Each file name in files is
+interpreted relative to Python’s ``sys.prefix``. Each file name in files is
 interpreted relative to the ``setup.py`` script at the top of the project source
 distribution.
 
@@ -343,11 +341,13 @@ For more information see the distutils section on `Installing Additional Files
 
 .. note::
 
-  :ref:`setuptools` allows absolute "data_files" paths, and pip honors them as
-  absolute, when installing from :term:`sdist <Source Distribution (or
-  "sdist")>`.  This is not true when installing from :term:`wheel`
-  distributions. Wheels don't support absolute paths, and they end up being
-  installed relative to "site-packages".  For discussion see `wheel Issue #92
+  :ref:`setuptools` does not support ``data_files``, see
+  `setuptools #130
+  <https://bitbucket.org/pypa/setuptools/issues/130/install_data-doesnt-respect-prefix>`_.
+  Use ``distutils`` instead.
+  :term:`wheel` only supports relative paths: files with absolute paths
+  end up being installed relative to "site-packages".
+  For discussion see `wheel Issue #92
   <https://bitbucket.org/pypa/wheel/issue/92>`_.
 
 


ionelmc:
Another pr in the same area: https://github.com/pypa/python-packaging-user-guide/pull/160

jonparrott:
@jaraco @dstufft can you verify the assertion made by this change? Additionally, do we every want to suggest `distutils` for any purpose?

dstufft:
Uh, well setuptools *does* support them, it just does something different than distutils and wheels do :/

jdemeyer:
Honestly, I don't care anymore about `distutils` and `setuptools` since `pip` does The Right Thing even if stock `setuptools` doesn't.

jdemeyer:
> Uh, well setuptools does support them,

It puts `data_files` in the egg, making them effectively inaccessible by non-Python programs.

jdemeyer:
> Additionally, do we ever want to suggest distutils for any purpose?

Before pip, I certainly avoided setuptools and used distutils precisely because of the `data_files` issue. With pip, it doesn't matter so much anymore.

pombredanne:
a related question is whether we want to promote using any kind of `data_files` at all and instead lead readers towards using package data?

pombredanne:
@jdemeyer you wrote:
> ... since pip does The Right Thing ...

What do you mean by this? Do you care to elaborate?

jdemeyer:
> a related question is whether we want to promote using any kind of data_files at all and instead lead readers towards using package data?

`data_files` and `package_data` solve different problems. Depending on the situation, `data_files` or `package_data` is the right thing to use.

jdemeyer:
> What do you mean by this?

I mean that pip installs `data_files` in the expected location (at least the same location that `distutils` and `wheel` would use) instead of putting them inside an `.egg`.

pombredanne:
@jdemeyer MO the expected location has never bet clearly defined, but not big deal. 

My experience is that devs not being sure of anything in that area have relied on one of the distutils, setuptools or pip behaviors. And therefore when installing with pip the behavior is everything  but well defined for things already available on Pypi.

jdemeyer:
The distutils behaviour of `data_files` is clearly documented:
https://docs.python.org/2/distutils/setupscript.html#installing-additional-files

So it's perfectly fine for package authors to rely on that. And wheel and pip respect the distutils behaviour, setuptools (more precisely `easy_install` I guess) does not. So yes, I argue that setuptools is broken in this respect.

pfmoore:
Agreed, this is specifically something setuptools chose to interpret differently to support egg installs (AIUI). Unless we're talking about eggs, then the behaviour should be as documented in the reference quoted above (and if setuptools doesn't follow that, then that's a bug in setuptools). And we shouldn't be talking about eggs in this document as they are not a recommended approach.

Personally, I would say that using ```data_files``` with relative paths is ill-advised, and projects should strongly prefer ```package_data```. I gather that there is a use case for ```data_files``` with absolute path names, for allowing packages to install data files in system locations like ```/etc```, but it's not a usage that I'm personally convinced is justified. So I would be in favour of treating ```data_files``` as a specialist tool, that shouldn't be used by authors unless they are fully aware of the implications, and understand what they are doing (and in particular, the cross-platform implications of their choice).

jdemeyer:
> Personally, I would say that using data_files with relative paths is ill-advised

I would actually argue the opposite and would say that `data_files` with **absolute paths** is ill-advised. Installing files in arbitrary locations on the system (like `/etc`) totally goes against `virtualenv` for example. Also, wheel only supports relative paths, not absolute paths for `data_files`.

pfmoore:
I should clarify. I think absolute paths are a really, really bad idea. But as a Windows person, I can't really comment except in theory on what Unix people should do - and I have heard people claim that for a project that's only ever intended to work on Unix and needs a config file adding to ```/etc```, ```data_files``` with an absolute path is the right thing to do. I don't honestly agree with them, but it's not my call. If someone with Unix experience wants to document that absolute paths are a bad idea, though, I'm fine with that.

On the other hand, I don't think there's any good argument for using ```data_files``` with a relative path rather than ```package_data```. And given that there are specific Python APIs for referencing package data that for example are designed to be zipfile-safe, whereas you have to write your own code (which probably won't work in a zipfile) for ```data_files```, I see no reason for us to recommend anything but ```package_data``` for the relative pathname case.

jonparrott:
It seems that this guide has historically erred on the side of being *minimal* and *opinionated* instead of exhaustive. Because of that, I'm inclined to think that a feature that has a small applicative use case and lots of caveats doesn't belong in this particular doc. Perhaps this warrants instead a doc under "additional topics"?

pfmoore:
@jonparrott So you're saying delete the whole ```data_files``` section, and maybe add a note under "Additional topics" to the effect that ```data_files``` is available if you need to place files in a fixed location outside of your package structure, but you should know what you're doing if you want to use this feature? That sounds reasonable to me.

jonparrott:
@pfmoore yep, that's exactly what I'm suggesting.

dstufft:
That seems like a good idea to me, I think this guide *should* be minimal and opinionated and anything that requires a bunch of caveats should probably be left out unless the benefit of the thing is really really high.

jonparrott:
@jdemeyer do you mind updating your PR with the suggestion of moving this under additional topics and incorporating @pfmoore's feedback about this mostly being applicable to unixes?

pombredanne:
@jonparrott this is a great idea indeed. Moving `data_files`-related doc out of the main path would be great and makes sense since they are rarely needed and an advanced topic.

jdemeyer:
> On the other hand, I don't think there's any good argument for using data_files with a relative path rather than package_data. And given that there are specific Python APIs for referencing package data that for example are designed to be zipfile-safe

There are good reasons to use `data_files`. The *Python* APIs that you mention are Python specific. The main use case for `data_files` is to instal files not to be used by that package, but by other programs. Or maybe by other Python programs which don't use that specific API.

If a given Python package needs files for its own use, then `package_data` is probably the best choice. But for files meant for other programs, `package_data` is mostly useless.

pombredanne:
@jdemeyer you are entirely right
> If a given Python package needs files for its own use, then package_data is probably the best choice. But for files meant for other programs, package_data is mostly useless.

... which reinforces even further the need to keep the topic of data files for advanced usage, e.g. the advanced and less common use case of installing files with a Python package that can be used by other tools outside of the Python code of this package.

ncoghlan:
+1 for covering data files as an "Additional Topic". The key trade-off is between:

- installing bundled data files using relative paths, which are portable cross-platform and work well with Python virtual environments and other path-based isolation mechanisms for parallel installation support, but don't interoperate with other non-Python components
- installing system data files using absolute paths, which are platform specific and require the use of system level isolation mechanisms like chroot and containers for parallel installation support, but offer interoperability with other applications on the same system

When using absolute paths, we should recommend that folks also set the appropriate Trove classifiers for the operating system they're assuming: https://pypi.python.org/pypi?%3Aaction=list_classifiers (the `Operating System` section)

theacodes:
I'm going to go ahead and close this as it's well out of date. If you still want to contribute this - awesome! Please open a new pull request and I promise to review.

jdemeyer:
> I'm going to go ahead and close this as it's well out of date.

What does this mean? What has happened since the opening of the PR that makes it out of date?

jdemeyer:
OK, there is a merge conflict, but the underlying issue hasn't been fixed yet.

theacodes:
@jdemeyer I was clearing out PRs with no activity in the last 6 months that had conflicts. If you stillw ant to do this, I would love to have it!

PR #213, opened 2016-01-22
Add specifications section

ncoghlan:
Add a dedicated specifications section, allowing the following:

* noting all currently active PEPs in one *.python.org location
* recording caveats on PEPs that have been partially, but
  not wholly, superseded by other PEPs
* noting cases where current real world practice differs from
  the details of older PEPs from the era when distutils was
  still the reference implementation for packaging

### DIFF
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -57,7 +57,7 @@ Version Specifiers
 ==================
 
 Version numbering requirements and the semantics for specifying comparisons
-between versions are defined :pep:`440`.
+between versions are defined in :pep:`440`.
 
 The version specifiers section in this PEP supersedes the version specifiers
 section in :pep:`345`.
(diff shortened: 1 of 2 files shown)


source/specifications.rst
```diff
@@ -0,0 +1,114 @@
...
+
+It is legal to specify ``Provides-Extra:`` without referencing it in any
+``Requires-Dist:``.
+
+
+Version Specifiers
+==================
+
+Version numbering requirements and the semantics for specifying comparisons
+between versions are defined :pep:`440`.
```
xavfernandez: defined **in** :pep:`440`.

source/specifications.rst
```diff
@@ -0,0 +1,114 @@
...
+component is defined in :pep:`508`.
+
+The environment markers section in this PEP supersedes the environment markers
+section in :pep:`345`.
+
+Source Distribution Format
+==========================
+
+The source distribution format (``sdist``) is not currently formally defined.
+Instead, it's format is implicitly defined by the behaviour of the
```
xavfernandez: its

xavfernandez:
I like it.
I was wondering if it could end by mentioning the current work/peps in progress with a link to https://www.pypa.io/en/latest/roadmap/#core-standards ?

ncoghlan:
Cross linking to the road map may be a good idea in the future, but for this initial iteration, I've just fixed the typos you pointed out.

PR #216, opened 2016-01-25
Fix broken URL

neptunerx:
In the installing.rst document the bullet point, "Securely Download get-pip.py" contains an out of date link. I've updated it to use https://bootstrap.pypa.io/get-pip.py rather then https://raw.github.com/pypa/pip/master/contrib/get-pip.py


dstufft:
Thanks!

PR #217, opened 2016-01-26
Mention 'twine register'

mgedmin:


source/distributing.rst
```diff
@@ -748,7 +748,10 @@ Next, you need to register your project.  There are two ways to do this:
    then run ``python setup.py egg_info`` to have it generated. Using the form is
    a secure option over using #2 below, which passes your credentials over
    plaintext.
-2. Run ``python setup.py register``.  If you don't have a user account already,
+2. Use ``twine register dist/*``.  This is better than using ``setup.py register``
```
ncoghlan: There's no particular reason to prefer the PyPI form over using twine, so it may be best to switch to marking option 3 as **(Not Recommended)**.

Then the description of option 3 can say why it isn't recommended, and mention that it's only included here since some older guides may mention it, and we want to specifically advise against it.

mgedmin: Sounds good to me!

Would you care to propose an alternative PR?  ;-)

ncoghlan:
I created a larger PR at https://github.com/pypa/python-packaging-user-guide/pull/218 and have asked @dstufft to take a look (since I haven't actually used twine to register a new project, only update existing ones)

Thanks for pointing out that it was missing as an option!

PR #218, opened 2016-01-27
Cover package registration via twine

ncoghlan:
This change presents twine as a CLI option for registration as well.

It also moves the note about manual .pypirc creation to the section
on creating an account, as that is needed for `twine register` to work.

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -732,25 +732,9 @@ account. There are two options:
 1. Create an account manually `using the form on the PyPI website
    <https://pypi.python.org/pypi?%3Aaction=register_form>`_.
 
-2. Have an account created as part of registering your first project (see option
-   #2 below).
-
-
-Register your project
----------------------
-
-Next, you need to register your project.  There are two ways to do this:
-
-1. **(Recommended):** Use `the form on the PyPI website
-   <https://pypi.python.org/pypi?%3Aaction=submit_form>`_, to upload your
-   ``PKG-INFO`` info located in your local project tree at
-   ``myproject.egg-info/PKG-INFO``.  If you don't have that file or directory,
-   then run ``python setup.py egg_info`` to have it generated. Using the form is
-   a secure option over using #2 below, which passes your credentials over
-   plaintext.
-2. Run ``python setup.py register``.  If you don't have a user account already,
-   a wizard will create one for you.
-
+2. **(Not recommended):** Have an account created as part of
+   registering your first project (not recommended due to the
+   related security concerns, see option #3 below).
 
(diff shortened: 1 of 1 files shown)


ncoghlan:
@dstufft @mgedmin noted `twine register` wasn't covered, but in reviewing the PR at https://github.com/pypa/python-packaging-user-guide/pull/217 I noticed several other problems with the relevant section.

This PR aims to address those as well.

source/distributing.rst
```diff
@@ -786,7 +791,7 @@ There are two options:
    The biggest reason to use twine is that ``python setup.py upload`` (option #2
    below) uploads files over plaintext. This means anytime you use it you expose
    your username and password to a MITM attack. Twine uses only verified TLS to
-   upload to PyPI protecting your credentials from theft.
+   upload to PyPI in order protect your credentials from theft.
```
pfmoore: "in order to protect"

source/distributing.rst
```diff
@@ -765,8 +749,29 @@ write a ``~/.pypirc`` file like so.
...
+   ``PKG-INFO`` info located in your local project tree at
+   ``myproject.egg-info/PKG-INFO``.  If you don't have that file or directory,
+   then run ``python setup.py egg_info`` to have it generated.
+2. Run ``twine register dist/*``, and :ref:`twine` will register your project
+   based on the package metadata in the specified files. Your ``~/.pypirc``
+   must already be appropriately configured for twine to work.
+3. **(Not recommended):** Run ``python setup.py register``.  If you don't have
+   a user account already, a wizard will create one for you. This approach is
+   covered here due to it being mentioned in other guides, but it is not
+   recommended as it uses a plaintext HTTP connection, allowing your username
```
dstufft: I'm not that great at documentation, but this is slightly wrong. It *may* use a plaintext HTTP connection, or it *may* use an unverified HTTPS connection, or it *may* use a verified HTTPS connection based on what version of Python you're using and/or what patches have been applied to your Python by whomever you're getting it from.

It may be over complicated to go into that much detail though and it may be preferable to be technically wrong, but simpler and safer in this documentation.

dstufft:
One comment to consider, otherwise looks good to me.

For the record, there's no need to ever use an explicit register step in Warehouse, even in legacy PyPI it's only needed once to register the project to your name, in Warehouse you don't even need it that one time.

ncoghlan:
Thanks folks - 2nd commit addressed @dstufft's and @pfmoore's comments, so I've merged this and any further changes can go in another PR.

Thanks again for the prompt to address it, @mgedmin.

mgedmin:
:cake:

PR #219, opened 2016-01-28
Update installing.rst

blairg23:
Add a note about Windows needing to add the path to pip.exe in their environment variables.

### DIFF
--- a/source/installing.rst
+++ b/source/installing.rst
@@ -51,6 +51,8 @@ Install pip, setuptools, and wheel
     python -m pip install -U pip setuptools
 
   You will not have :ref:`wheel`, so you'll need to run: ``pip install wheel``
+  
+  **Note:** For most Windows blends, you'll need to add the path to the pip.exe file into the environment variables before you can use pip on the CLI.
 
 * If you're using a Python install on Linux that's managed by the system package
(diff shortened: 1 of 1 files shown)


source/installing.rst
```diff
@@ -51,6 +51,8 @@ Install pip, setuptools, and wheel
     python -m pip install -U pip setuptools
 
   You will not have :ref:`wheel`, so you'll need to run: ``pip install wheel``
+  
+  **Note:** For most Windows blends, you'll need to add the path to the pip.exe file into the environment variables before you can use pip on the CLI.
```
berkerpeksag: I can't comment on the content since I don't use Windows, but you can use a `.. note::` directive instead of `**Note:**`. Also, please wrap the line at 80 columns.

pfmoore: Given that the suggested command is ```python -m pip``` which doesn't use ```pip.exe```, the note is probably not needed. (That's *why* we recommend ```python -m pip``` over plain ```pip```)

jonparrott:
Hi @blairg23, this is quite old and has conflicting files. I'm going to close for now, but if you still think this change is necessary please let us know!

PR #222, opened 2016-01-30
Update virtualenv and pyvenv information in installing.rst

berkerpeksag:
Clarify that pyvenv is available on Python 3.3+ and virtualenv supports Python 3.5.

Also, updated to use "through" instead of "thru".

### DIFF
--- a/source/installing.rst
+++ b/source/installing.rst
@@ -50,8 +50,6 @@ Install pip, setuptools, and wheel
 
     python -m pip install -U pip setuptools
 
-  You will not have :ref:`wheel`, so you'll need to run: ``pip install wheel``
-
 * If you're using a Python install on Linux that's managed by the system package
   manager (e.g "yum", "apt-get" etc...), and you want to use the system package
   manager to install or upgrade pip, then see :ref:`Installing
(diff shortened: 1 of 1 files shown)


source/installing.rst
```diff
@@ -123,10 +123,10 @@ installation directories and they don’t share libraries with other virtual
 environments.
 
 Currently, there are two viable tools for creating Python virtual environments:
-:ref:`virtualenv` and `pyvenv`_. `pyvenv`_ is only available in Python 3.3 &
-3.4, and only in Python 3.4, is :ref:`pip` & :ref:`setuptools` installed into
-environments by default, whereas :ref:`virtualenv` supports Python 2.6 thru
-Python 3.4 and :ref:`pip` & :ref:`setuptools` are installed by default in every
+:ref:`virtualenv` and `pyvenv`_. `pyvenv`_ is available in Python 3.3 and later,
+and in Python 3.4 and later, is :ref:`pip` & :ref:`setuptools` installed into
```
ncoghlan: losing the "only" before "in Python 3.4 and later" has mangled the grammar in the rest of this sentence. However, the original phrasing was pretty awkward, so rather than adding it back, a better fix might be to turn this into a bulleted list:

> ... virtual environments:
> * `pyvenv`_ is available by default in Python 3.3 and later, and installs :ref:`pip` & :ref:`setuptools` into created virtual environments in Python 3.4 and later
> * :ref:`virtualenv` needs to be installed separately, but support Python 2.6 through Python 3.5,  and :ref:`pip` & :ref:`setuptools` are always installed into created virtual environments by default (regardless of Python version).

xavfernandez: Not sure if needed in this doc but virtualenv also installs the wheel package by default.

berkerpeksag: @xavfernandez thanks! ["Install pip, setuptools, and wheel"](https://github.com/pypa/python-packaging-user-guide/blob/master/source/installing.rst#install-pip-setuptools-and-wheel) section says "You will not have :ref:`wheel`, so you'll need to run: ``pip install wheel``" Do you want to me remove that sentence?

berkerpeksag: @ncoghlan thanks Nick! Your version is so much better. I'll update the pull request today.

berkerpeksag: Done.

Ivoz:
I believe saying "Python 2.6+ and Python 3.3+" is perfectly understandable, but also won't need to be updated for new python minor releases.

berkerpeksag:
> I believe saying "Python 2.6+ and Python 3.3+" is perfectly understandable, but also won't need to be updated for new python minor releases.

@Ivoz are you suggesting changing "[...] but supports Python 2.6 through Python 3.5, [...]" to "[...] but supports Python 2.6+ and Python 3.3+, [...]" in the second bullet point? Thanks.

Ivoz:
Yes. For instance, it doesn't support [3.0, 3.1, 3.2] anymore.

berkerpeksag:
Done. Thanks for the suggestion.

berkerpeksag:
Is there anything I can do to move this forward?

berkerpeksag:
Ping.

PR #224, opened 2016-02-06
PEP 440 formatting for version suffixes

bintoro:


jonparrott:
@bintoro I like this change, but can you provide some more detail examples to avoid ambiguity, for example:

```
* zero or more alpha releases denoted with a "aN" suffix, for example: ``1.0.0a4``.
```

jonparrott:
I'm going to close this PR as it has conflicts, but feel free to open a new one.

PR #226, opened 2016-02-16
virtualenv guide: Fix link

amitsaha:


Ivoz:
Could you change the url domain to `virtualenv.pypa.io` as well?

amitsaha:
@Ivoz sure, done!

ddbeck:
Thanks for the fix @amitsaha! :hammer:

PR #227, opened 2016-02-18
bad link to pypa

phaustin:


Ivoz:
https://pypa.io should normally be working fine, something is up with RTD

ddbeck:
https://pypa.io is working as expected again. :tada:

PR #229, opened 2016-03-13
Fix typo in sample appveyor.yml

federicobond:


techtonik:
@ncoghlan this is an easy one.

PR #244, opened 2016-06-14
Explaining where dist/ comes from

annaraven:
Issue ##231 indicated it is unclear where dist/ directory comes from or where to find it. Clarifying this situation so folks can register and upload their files to PyPI.


ncoghlan:
Thanks!

PR #245, opened 2016-06-15
Typo a fix in the PyPI distributing documentation.

illfelder:


ncoghlan:
Thanks!

source/distributing.rst
```diff
@@ -648,7 +648,7 @@ Only use the ``--universal`` setting, if:
    require 2to3).
 2. Your project does not have any C extensions.
 
-Beware that ``bdist_wheel`` does not currently have any checks to warn you if
+Beware that ``bdist_wheel`` does not currently have any checks to warn if you
```
Ivoz: This should have been `warn you if you`

illfelder: I don't think the first you is necessary, but I don't feel strongly if you want to make a followup change to add it back in.

PR #247, opened 2016-06-24
use venv instead of pyvenv

### LINKED ISSUE #242: Update Installing Packages to reflect deprecation of pyvenv
Per https://hg.python.org/cpython/rev/6ab285820db2#l1.8 `pyvenv` will be deprecated in 3.6 in favor of `python3 -m venv`. Suggest that [installing.rst](https://github.com/pypa/python-packaging-user-guide/blob/master/source/installing.rst) be reviewed and updated accordingly.

Include citations or references to reasoning for the change, perhaps in footnotes.

stevepiercy:
- update links to docs to reduce redirects
- closes #242

### DIFF
--- a/source/current.rst
+++ b/source/current.rst
@@ -5,7 +5,7 @@ Tool Recommendations
 ====================
 
 :Page Status: Complete
-:Last Reviewed: 2015-08-31
+:Last Reviewed: 2016-06-24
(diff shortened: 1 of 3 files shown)


ncoghlan:
Thanks!

PR #248, opened 2016-07-01
Added Android packaging options to deployment doc

kamathln:


ncoghlan:
Thanks! I've also filed https://github.com/pypa/python-packaging-user-guide/issues/267 to better document what's currently missing from the application bundling aspects of this topic.

PR #251, opened 2016-07-07
Update specifications.rst

xaxiclouddev:


ncoghlan:
PEP 440 still defines how version specifiers work - the full dependency specifiers in PEP 508 are a use case for PEP 440, not a replacement.

PR #258, opened 2016-07-29
Add Description-Content-Type field

msabramo:
This allows specifying the content type of the distribution's description (e.g.: `text/plain`, `text/x-rst`, `text/markdown`).

See https://github.com/msabramo/python-packaging-user-guide/blob/description-content-type/source/specifications.rst#description-content-type

Cc: @nicktimko, @ncoghlan, @dstufft, @jaraco, @xavfernandez 

Related:
- https://mail.python.org/pipermail/distutils-sig/2016-May/028689.html
- https://github.com/pypa/readme_renderer/pull/3
- https://github.com/pypa/setuptools/pull/708

### DIFF
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -55,8 +55,8 @@ It is legal to specify ``Provides-Extra:`` without referencing it in any
 Description-Content-Type
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-A string containing the format of the distribution's description, so that
-tools can intelligently render the description.
+A string stating the markup syntax (if any) used in the distribution's
+description, so that tools can intelligently render the description.
 
 Historically, PyPI supported descriptions in plain text and `reStructuredText
 (reST) <http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html>`_,
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -133,7 +133,7 @@ If a ``Description-Content-Type`` is an unrecognized value, then the assumed
 content type is ``text/plain`` (Although PyPI will probably reject anything
 with an unrecognized value).
 
-If the ``Description-Content-Type` is ``text/markdown`` and ``variant`` is not
+If the ``Description-Content-Type`` is ``text/markdown`` and ``variant`` is not
 specified or is set to an unrecognized value, then the assumed ``variant`` is
 ``CommonMark``.
 
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -87,8 +87,10 @@ The ``type/subtype`` part has only a few legal values:
 - ``text/x-rst``
 - ``text/markdown``
 
-There is one required parameter called ``charset``, to specify whether the text
-is UTF-8, ASCII, etc.
+One parameter is called ``charset``; it can be used to specify whether the
+character set in use is UTF-8, ASCII, etc. If ``charset`` is not provided, then
+it is recommended that the implementation (e.g.: PyPI) treat the content as
+UTF-8.
 
 Other parameters might be specific to the chosen subtype. For example, for the
 ``markdown`` subtype, there is a ``variant`` parameter that allows specifying
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -94,10 +94,16 @@ UTF-8.
 
 Other parameters might be specific to the chosen subtype. For example, for the
 ``markdown`` subtype, there is a ``variant`` parameter that allows specifying
-the variant of Markdown in use, such as ``Original`` for `Gruber's original
-Markdown syntax <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_ or
-``GFM`` for `GitHub Flavored Markdown (GFM)
-<https://tools.ietf.org/html/rfc7764#section-3.2>`_.
+the variant of Markdown in use, such as:
+
+- ``CommonMark`` for `CommonMark`
+  <https://tools.ietf.org/html/rfc7764#section-3.5>`_
+
+- ``GFM`` for `GitHub Flavored Markdown (GFM)
+  <https://tools.ietf.org/html/rfc7764#section-3.2>`_
+
+- ``Original`` for `Gruber's original Markdown syntax
+  <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_
 
 Example::
 
@@ -109,11 +115,15 @@ Example::
 
 Example::
 
-    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=CommonMark
 
 Example::
 
     Description-Content-Type: text/markdown; charset=UTF-8; variant=GFM
+
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
 
 If a ``Description-Content-Type`` is not specified, then the assumed content type
 is ``text/x-rst; charset=UTF-8``.
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -105,9 +105,9 @@ the variant of Markdown in use, such as:
 - ``Original`` for `Gruber's original Markdown syntax
   <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_
 
-If the subtype is ``markdown`` and no ``variant`` is specified, then the
-implementation (e.g.: PyPI) should assume that the ``variant`` is
-``CommonMark``.
+If the subtype is ``markdown`` and no ``variant`` is specified or the specfied
+``variant`` is not recognized, then the implementation (e.g.: PyPI) should
+assume that the ``variant`` is ``CommonMark``.
 
 Example::
 
@@ -129,8 +129,15 @@ Example::
 
     Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
 
-If a ``Description-Content-Type`` is not specified, then the assumed content type
-is ``text/x-rst; charset=UTF-8``.
+If a ``Description-Content-Type`` is not specified or it's set to an
+unrecognized value, then the assumed content type is ``text/x-rst;
+charset=UTF-8``.
+
+If the ``charset`` is not specified or it's set to an unrecognized value, then
+the assumed ``charset`` is ``UTF-8``.
+
+If the subtype is ``markdown`` and ``variant`` is not specified or it's set to
+an unrecognized value, then the assumed ``variant`` is ``CommonMark``.
 
 
 Version Specifiers
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -21,7 +21,7 @@ The current core metadata file format, version 1.2, is specified in :pep:`345`.
 
 However, the version specifiers and environment markers sections of that PEP
 have been superceded as described below. In addition, metadata files are
-permitted to contain the following additional field:
+permitted to contain the following additional fields:
 
 Provides-Extra (multiple use)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -51,6 +51,70 @@ respectively.
 
 It is legal to specify ``Provides-Extra:`` without referencing it in any
 ``Requires-Dist:``.
+
+Description-Content-Type
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+A string containing the format of the distribution's description, so that tools
+can intelligently render the description. Historically, distribution
+descriptions in plain text and in `reStructuredText (reST)
+<http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html>`_ have
+been supported and PyPI knows how to render reST into HTML. It is very common
+for distribution authors to write their description in `Markdown
+<https://daringfireball.net/projects/markdown/>`_ (`RFC 7763
+<https://tools.ietf.org/html/rfc7763>`_). Distribution authors commonly use
+Markdown probably because that's the markup language that they are most
+familiar with, but PyPI historically didn't know the format of the description
+and thus could not know to render a description as Markdown. This results in
+PyPI having many packages where the description is rendered in a very ugly way,
+because the description was written in Markdown, but PyPI is rendering it as
+reST. This field allows the distribution author to specify the format of their
+description and thus opens up the possibility for PyPI and other tools to be
+able to render Markdown and other formats.
+
+The format of this field is same as the ``Content-Type`` header in HTTP (e.g.:
+`RFC 1341 <https://www.w3.org/Protocols/rfc1341/4_Content-Type.html>`_).
+Briefly, this means that it has a ``type/subtype`` part and then it can
+optionally have a number of parameters:
+
+Format::
+
+    Description-Content-Type: <type>/<subtype>; charset=<charset>[; <param_name>=<param value> ...]
+
+The ``type/subtype`` part has only a few legal values:
+
+- ``text/plain``
+- ``text/x-rst``
+- ``text/markdown``
+
+There is one required parameter called ``charset``, to specify whether the text
+is UTF-8, ASCII, etc.
+
+Other parameters might be specific to the chosen subtype. For example, for the
+``markdown`` subtype, there is a ``variant`` parameter that allows specifying
+the variant of Markdown in use, such as ``Original`` for `Gruber's original
+Markdown syntax <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_ or
+``GFM`` for `GitHub Flavored Markdown (GFM)
+<https://tools.ietf.org/html/rfc7764#section-3.2>`_.
+
+Example::
+
+    Description-Content-Type: text/plain; charset=UTF-8
+
+Example::
+
+    Description-Content-Type: text/x-rst; charset=UTF-8
+
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=GFM
+
+If a ``Description-Content-Type`` is not specified, then the assumed content type
+is ``text/x-rst; charset=UTF-8``.
 
 
 Version Specifiers
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -96,7 +96,7 @@ Other parameters might be specific to the chosen subtype. For example, for the
 ``markdown`` subtype, there is a ``variant`` parameter that allows specifying
 the variant of Markdown in use, such as:
 
-- ``CommonMark`` for `CommonMark`
+- ``CommonMark`` for `CommonMark
   <https://tools.ietf.org/html/rfc7764#section-3.5>`_
 
 - ``GFM`` for `GitHub Flavored Markdown (GFM)
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -105,6 +105,10 @@ the variant of Markdown in use, such as:
 - ``Original`` for `Gruber's original Markdown syntax
   <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_
 
+If the subtype is ``markdown`` and no ``variant`` is specified, then the
+implementation (e.g.: PyPI) should assume that the ``variant`` is
+``CommonMark``.
+
 Example::
 
     Description-Content-Type: text/plain; charset=UTF-8
--- a/source/specifications.rst
+++ b/source/specifications.rst
@@ -125,12 +125,17 @@ Example::
 
     Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
 
-If a ``Description-Content-Type`` is not specified or it's set to an
-unrecognized value, then the assumed content type is ``text/x-rst;
-charset=UTF-8``.
+If a ``Description-Content-Type`` is not specified, then applications should
+attempt to render it as ``text/x-rst; charset=UTF-8`` and fall back to
+``text/plain`` if it is not valid rst.
 
-If the subtype is ``markdown`` and ``variant`` is not specified or is set to
-an unrecognized value, then the assumed ``variant`` is ``CommonMark``.
+If a ``Description-Content-Type`` is an unrecognized value, then the assumed
+content type is ``text/plain`` (Although PyPI will probably reject anything
+with an unrecognized value).
+
+If the ``Description-Content-Type` is ``text/markdown`` and ``variant`` is not
+specified or is set to an unrecognized value, then the assumed ``variant`` is
+``CommonMark``.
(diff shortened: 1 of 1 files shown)


nicktimko:
Bikeshedding: CommonMark? Can that be the default?

msabramo:
Maybe. I've never used it. Have a Python implementation that you like? https://pypi.python.org/pypi/CommonMark perhaps?

msabramo:
How is CommonMark different than the original Markdown? Why do you prefer it?

nicktimko:
CommonMark (http://commonmark.org/) is rigidly defined unlike Markdown which has bits of [ambiguity](http://johnmacfarlane.net/babelmark2/) (true "original" Markdown.pl is properly broken in some nesting cases). It improves a bunch of quirks, like spaces around lists, and standardizes some of the more convenient parts of GFM like fenced code blocks.

msabramo:
http://commonmark.org/ sounds interesting and has some big names behind it. I'm intrigued. I think I'll take a closer look and consider making CommonMark the default.

msabramo:
BTW, the only reason that I made original Markdown the default is because it's listed as the initial variant in RFC 7763 at https://tools.ietf.org/html/rfc7763#section-6.1.4, but then CommonMark is referenced in RFC 7764 at https://tools.ietf.org/html/rfc7764#section-4.5 (which I hadn't noticed before), so I guess nothing prevents it from being the default.

Part of me thinks that because there are so many competing Markdown variants and so much confusion around Markdown, that perhaps there should be **no default** and we should **require the variant to be specified**. Thoughts on that?

ncoghlan:
The overall default on the server side needs to be UTF-8 ReST (as in the current PR) for compatibility with existing projects that already correctly use ReST formatting.

However, when the media type is given as `text/markdown`, I'd actually suggest using either GFM (GitHub Flavored Markdown) or CommonMark (where GitHub are one of the backing organisations) as the default, as one of the more common reasons for using Markdown in README files these days is as a nice user-friendly front page for a GitHub repo.

So using CommonMark as the default (including falling back to it if asked to process an unknown variant) makes sense to me.

source/specifications.rst
```diff
@@ -52,6 +52,70 @@ respectively.
...
+    Description-Content-Type: <type>/<subtype>; charset=<charset>[; <param_name>=<param value> ...]
+
+The ``type/subtype`` part has only a few legal values:
+
+- ``text/plain``
+- ``text/x-rst``
+- ``text/markdown``
+
+There is one required parameter called ``charset``, to specify whether the text
+is UTF-8, ASCII, etc.
```
ncoghlan: I'd make `charset` optional, defaulting to `UTF-8` - that way `text/markdown` is sufficient to get Markdown rendering.

ncoghlan:
+1 from me for the general idea, but what you have so far is missing guidance on expected behaviour for tools reading the field to render the description when:

- they don't recognise the content type
- they don't recognise the charset
- they don't recognise the specific variant

Those additions would be SHOULD/MAY type guidelines rather than MUST requirements, but I think we still want to give some thought to how PyPI should handle those cases, and how those considerations can be turned into general recommendations.

source/specifications.rst
msabramo: My first instinct was to default to `charset` to `UTF-8`. 

But then I saw this:

  charset: Per Section 4.2.1 of [RFC6838], charset is REQUIRED.
         There is no default value because neither [MDSYNTAX] nor many
         popular implementations at the time of this registration do
         either.

That said, we can disagree with that RFC and do what we want, but I thought deferring to previous work would be a good start, since they probably had reasons and discussions around this.

msabramo: OK, I went ahead and made `charset` optional and defaulted it to UTF-8.

d454938

msabramo:
Added mentions of [`CommonMark`](https://tools.ietf.org/html/rfc7764#section-4.5) in 774d091.

msabramo:
Made CommonMark the default variant for Markdown in 33dc8c5.

source/specifications.rst
ncoghlan: In the general case, I agree that there's no sensible universal default, but Python project descriptions aren't the general case - they're a specific use case. That lets us make assumptions in the name of ease-of-use that wouldn't be valid given a broader context.

ncoghlan:
Latest draft looks good to me, but I'll defer to @dstufft for actual sign-off.

While I don't expect this idea to be controversial, we should also give folks a chance to discuss the idea on distutils-sig before committing to it.

msabramo:
@ncoghlan: I'm still planning to add some info on expected behavior when things are omitted or invalid, as you suggested in https://github.com/pypa/python-packaging-user-guide/pull/258#issuecomment-236351140.

But certainly, I'd love to have @dstufft and anyone else's comments.

source/specifications.rst
```diff
@@ -52,6 +52,86 @@ respectively.
...
+The ``type/subtype`` part has only a few legal values:
+
+- ``text/plain``
+- ``text/x-rst``
+- ``text/markdown``
+
+One parameter is called ``charset``; it can be used to specify whether the
+character set in use is UTF-8, ASCII, etc. If ``charset`` is not provided, then
+it is recommended that the implementation (e.g.: PyPI) treat the content as
+UTF-8.
```
pfmoore: Maybe move the sentence "If charset is not provided..." down below? (See next comment)

source/specifications.rst
```diff
@@ -52,6 +52,86 @@ respectively.
...
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=GFM
+
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+
+If a ``Description-Content-Type`` is not specified, then the assumed content type
+is ``text/x-rst; charset=UTF-8``.
+
```
pfmoore: Maybe include the case of a missing charset here?
```
If a Description-Content-Type is present, but has no charset specified, then the character set is assumed to be UTF-8
```

I'm not sure I see much benefit in distinguishing between "charset isn't given but the implementation should treat it as UTF-8" and "when charset isn't given, it defaults to UTF-8".

pfmoore:
Overall, LGTM (one minor comment in the diff).

msabramo:
e4b39db: Clarified how to treat missing/invalid values for `Description-Content-Type` (and `charset` and `variant`) as suggested by @ncoghlan at
https://github.com/pypa/python-packaging-user-guide/pull/258#issuecomment-236351140

source/specifications.rst
msabramo: @pfmoore Does e4b39dbb253ab64ab1022b3172e271377cf2bce7 address your concern?

source/specifications.rst
msabramo: Does e4b39dbb253ab64ab1022b3172e271377cf2bce7 help?

source/specifications.rst
pfmoore: Yep, looks fine to me, thanks.

source/specifications.rst
```diff
@@ -52,6 +52,93 @@ respectively.
...
+One parameter is called ``charset``; it can be used to specify whether the
+character set in use is UTF-8, ASCII, etc. If ``charset`` is not provided, then
+it is recommended that the implementation (e.g.: PyPI) treat the content as
+UTF-8.
+
+Other parameters might be specific to the chosen subtype. For example, for the
+``markdown`` subtype, there is a ``variant`` parameter that allows specifying
+the variant of Markdown in use, such as:
+
+- ``CommonMark`` for `CommonMark`
```
ekohl: The last backtick is causing incorrect rendering

source/specifications.rst
```diff
@@ -52,6 +52,93 @@ respectively.
...
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+
+If a ``Description-Content-Type`` is not specified or it's set to an
+unrecognized value, then the assumed content type is ``text/x-rst;
+charset=UTF-8``.
+
+If the ``charset`` is not specified or it's set to an unrecognized value, then
+the assumed ``charset`` is ``UTF-8``.
+
+If the subtype is ``markdown`` and ``variant`` is not specified or it's set to
```
ekohl: This line duplicates what's said above.

hadim:
If I am not wrong this PR is blocking for a lot of others PR (in `setuptools`, `readme_renderer` and `warehouse).

Beside that all the pieces seems to be in place to allow different markup languages support in the `long_description` field in `setup.py`.

Who can decide to merge this PR ? Should we/you start a kind of vote ?

I am happy to help but as I said all the PRs are here and they only need to be merged.

---

Here is a quick overview of the PRs (classified by possible merged order) :  

1. Add a new `Description-Content-Type` field in the `setup.py` specification : [`python-packaging-user-guide/#258`](https://github.com/pypa/python-packaging-user-guide/pull/258) : 

2. Implement the new specification : [`setuptools/#708`](https://github.com/pypa/setuptools/pull/708)

3. Add Markdown rendering support : [`readme_renderer/#3`](https://github.com/pypa/readme_renderer/pull/3)

4. Implement Markdown support in warehouse. No PR exists at this time but it should be super easy to make since all the hard work is done in `readme_renderer` : [`warehouse/869`](https://github.com/pypa/warehouse/issues/869)

source/specifications.rst
```diff
@@ -52,6 +52,93 @@ respectively.
...
+Markdown probably because that's the markup language that they are most
+familiar with, but PyPI historically didn't know the format of the description
+and thus could not know to render a description as Markdown. This results in
+PyPI having many packages where the description is rendered in a very ugly way,
+because the description was written in Markdown, but PyPI is rendering it as
+reST. This field allows the distribution author to specify the format of their
+description and thus opens up the possibility for PyPI and other tools to be
+able to render Markdown and other formats.
+
+The format of this field is same as the ``Content-Type`` header in HTTP (e.g.:
```
ncoghlan: Typo fix: "is same as" -> "is the same as"

source/specifications.rst
```diff
@@ -52,6 +52,93 @@ respectively.
...
+
+- ``GFM`` for `GitHub Flavored Markdown (GFM)
+  <https://tools.ietf.org/html/rfc7764#section-3.2>`_
+
+- ``Original`` for `Gruber's original Markdown syntax
+  <https://tools.ietf.org/html/rfc7763#section-6.1.4>`_
+
+If the subtype is ``markdown`` and no ``variant`` is specified or the specfied
+``variant`` is not recognized, then the implementation (e.g.: PyPI) should
+assume that the ``variant`` is ``CommonMark``.
```
ncoghlan: This sentence is duplicated down below - I suggest dropping this one, and keeping the one further down.

source/specifications.rst
ncoghlan: I agree with @pfmoore that this can be dropped in favour of the more concise specification below.

ncoghlan:
Aside from a few minor readability cleanups, I think the main piece we're missing is confirmation from @dstufft that he'd be happy to implement it in PyPI (at least for Warehouse, potentially the legacy one as well), and a final discussion thread on distutils-sig.

Once we have both of those addressed, then either @dstufft or I can approve the change by accepting the PR.

nicktimko:
Far from this being another thing to throw on to @dstufft's pile, I think this was just the first step that you (Coghlan) recommended for the GitHub plebs to implement 😛  https://mail.python.org/pipermail/distutils-sig/2016-June/029078.html (not that reviewing is a negligible task)

source/specifications.rst
```diff
@@ -55,24 +55,25 @@ It is legal to specify ``Provides-Extra:`` without referencing it in any
...
+Markdown READMEs, and authors would reuse the file for the description. PyPI
+didn't recognize the format and so could not render the description correctly.
+This resulted in many packages on PyPI with poorly-rendered descriptions when
+Markdown is left as plain text, or worse, was attempted to be rendered as reST.
+This field allows the distribution author to specify the format of their
+description, opening up the possibility for PyPI and other tools to be able to
+render Markdown and other formats.
+
+The format of this field is the same as the ``Content-Type`` header in HTTP
+(e.g.:
```
ncoghlan: This is an "i.e." ("for explanation") rather than an "e.g." ("for example"), since RFC 1341 is the relevant normative reference here.

source/specifications.rst
```diff
@@ -52,6 +52,90 @@ respectively.
...
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+
+If a ``Description-Content-Type`` is not specified or it's set to an
+unrecognized value, then the assumed content type is ``text/x-rst;
+charset=UTF-8``.
+
+If the ``charset`` is not specified or it's set to an unrecognized value, then
+the assumed ``charset`` is ``UTF-8``.
```
JelleZijlstra: Lines 92–94 above merely recommend that implementations assume the charset is UTF-8, but this line seems to require that they do so.

You should resolve the contradiction one way or the other. I'd prefer if UTF-8 is required to be the default for simplicity.

source/specifications.rst
```diff
@@ -52,6 +52,90 @@ respectively.
...
+
+    Description-Content-Type: <type>/<subtype>; charset=<charset>[; <param_name>=<param value> ...]
+
+The ``type/subtype`` part has only a few legal values:
+
+- ``text/plain``
+- ``text/x-rst``
+- ``text/markdown``
+
+The ``charset`` parameter can be used to specify whether the character set in
```
merwok: charset is a character encoding.  See for example https://www.w3.org/International/articles/http-charset/index

source/specifications.rst
nicktimko: I'd be all for UTF-8 only. What do others think?

pfmoore: I'm +1 on saying UTF-8 if otherwise unspecified, but I think it's potentially useful to allow an alternative user-specified encoding. I doubt it'll be used a lot, but it's consistent with similar MIME-type standards.

jonparrott:
@msabramo hi! This PR has some outstanding review comments and hasn't been touched in a while. Can you let us know if you're still working on it?

msabramo:
I will take a look at the comments soon and update this.

msabramo:
I just merged https://github.com/msabramo/python-packaging-user-guide/pull/2 which fixes a few f the comments above. I'm going to look through comments and see which ones are resolved and which ones are outstanding.

source/specifications.rst
msabramo: Thanks! Fixed in 4b17b0b3e85a8372c73c86110a4a1a3415f0e555.

source/specifications.rst
msabramo: Fixed in a3c07c54. Thanks!

source/specifications.rst
msabramo: Fixed in a3c07c54. Thanks!

source/specifications.rst
msabramo: Fixed in a3c07c54. Thanks!

source/specifications.rst
msabramo: Fixed in a3c07c54. Thanks!

source/specifications.rst
msabramo: @merwok: The doc has changed a bit since you reviewed in February,  mostly as a result of me merging https://github.com/msabramo/python-packaging-user-guide/pull/2. Does the explanation look good in the new version or do you still think it needs to be tweaked?

source/specifications.rst
```diff
@@ -52,6 +52,90 @@ respectively.
...
+use is UTF-8, ASCII, etc. If ``charset`` is not provided, then it is
+recommended that the implementation (e.g.: PyPI) treat the content as
+UTF-8.
+
+Other parameters might be specific to the chosen subtype. For example, for the
+``markdown`` subtype, there is a ``variant`` parameter that allows specifying
```
msabramo: The intention here (possibly needs to be clarified?) is that CommonMark is what PyPI would initially support, as it seems like the best standard right now, but I was trying to leave room for other variants so we can adapt to change.

I do wonder if I should remove the references to GFM and original Markdown, so that folks don't think they can use them and get their exact semantics (because it will likely be unrecognized by PyPI and thus treated as CommonMark). Folks are likely to see GFM and get excited and specify and then expect it to work exactly like GitHub and then complain when it doesn't.

Should I remove GFM and original Markdown and say CommonMark is the only legal variant right now?

source/specifications.rst
```diff
@@ -52,6 +52,90 @@ respectively.
...
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=GFM
+
+Example::
+
+    Description-Content-Type: text/markdown; charset=UTF-8; variant=Original
+
+If a ``Description-Content-Type`` is not specified or it's set to an
+unrecognized value, then the assumed content type is ``text/x-rst;
```
msabramo: Thanks! I used your wording in a8c3b77. Let me know if you have other comments.

msabramo:
First of all, sorry for the delay on this one. I let this languish for a while. I intend to give this the attention it needs to get it across the finish line.

I looked through all the comments and I think they're all addressed in one way or another (a lot of them by @nicktimko in https://github.com/msabramo/python-packaging-user-guide/pull/2; thanks again, Nick!) except that I have [this question](https://github.com/pypa/python-packaging-user-guide/pull/258#discussion_r115507243) (and you can see which way I'm leaning...

> Should I remove GFM and original Markdown and say CommonMark is the only legal variant right now?

nicktimko:
Because GFM is (will be) a ["strict superset of CommonMark"](https://github.github.com/gfm/#what-is-github-flavored-markdown-), I'd say CommonMark is sufficient for now, and if someone really wants tables/strikethrough, they could iterate on this later.

source/specifications.rst
nicktimko: As per the [GitHub Flavored Markdown spec](https://github.github.com/gfm/#what-is-github-flavored-markdown-), it's a superset of CommonMark with the following additional features ("extensions"):

* [Tables](https://github.github.com/gfm/#tables-extension-)
* Strikethrough using `~` like ~this~. Used more often for comedic effect.
* Task Lists
    - [ ] Probably
    - [ ] Irrelevant
    - [ ] For PyPI
* [More Autolinks](https://github.github.com/gfm/#autolinks). CM wants `<>` around links to flag them as being links. This is probably just a minor usability thing to make "quick" comments more useful. A negligible barrier for any document that's scrutinized for a couple seconds
    * Only on GFM:
        * www.google.com
        * www.commonmark.org/help
        * http://nick.solutions
    * Neither:
        * github.com
        * python.org
* [More HTML disallowed](https://github.github.com/gfm/#disallowed-raw-html-extension-)...but presumably there's going to be some (high) level of HTML sanitization going on anyways before/after conversion, so that might render it moot anyways. If the filters aren't turned up too high, users could do tables manually.

dstufft: The rendering mechanism used by both PyPI and Warehouse does not rely on the renderer to emit safe HTML and we do our own strict HTML sanitization after it has been rendered. Theoretically we could have an option for raw HTML as a long_description and it shouldn't provide any avenue for attack. The list of whitelisting tags/attributes can be found at [readme_renderer/clean.py#L24-48](https://github.com/pypa/readme_renderer/blob/master/readme_renderer/clean.py#L24-L48).

nicktimko: Looks like neither `<del>` or `<s>` is allowed anyway, so the ~strikethrough~ feature is borderline moot in GFM as there's 0 users using it on PyPI ;).

dstufft: We can of course add things to the list where needed as well! The current list is basically compiled by adding things whenever we found something that didn't render correctly on PyPI that we felt was a reasonable thing to want to have rendered. I didn't exhaustively look through every HTML tag and make a determination on it.

msabramo:
OK, see https://github.com/msabramo/python-packaging-user-guide/pull/3

msabramo:
Merged https://github.com/msabramo/python-packaging-user-guide/pull/3

This is ready to be reviewed again by interested parties.

source/specifications.rst
```diff
@@ -52,6 +52,86 @@ respectively.
 It is legal to specify ``Provides-Extra:`` without referencing it in any
 ``Requires-Dist:``.
 
+Description-Content-Type
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+A string containing the format of the distribution's description, so that
```
ncoghlan: "stating the markup syntax (if any) used in" might be clearer than "containing the format of"

msabramo: I like it. Changed to that in b371449.

msabramo:
@dstufft: How does it look to you now?

msabramo:
@dstufft: Any more comments on this?

jonparrott:
I imagine @dstufft is likely at pycon and will be a little slow to respond.
:)

dstufft:
Sadly I'm not at PyCon, I just haven't gotten a chance to circle back around to this yet :(

msabramo:
Yay, I see @dstufft merged this. Great news!

So what do we need to do next?

dstufft:
Probably implement support in setuptools, then wheel, then twine, then PyPI/Warehouse.

nicktimko:
Would it be better to have the server (PyPI/Warehouse) accepting it before the clients? Then newer clients wouldn't get errors if they support what the server doesn't

ncoghlan:
I think there are a few aspects with a required order, then some that can happen in arbitrary order:

1. Check that the main clients (pip, easy_install) will ignore the new field at install time. If not, update them, and document the minimum required versions of the client tools implied by using the new field.
2. Update PyPI/Warehouse to allow the new field (if they don't already)
3. Update setuptools and the other build systems to allow setting the new field, and PyPI/Warehouse to actually pay attention to it

(We should probably document this somewhere on pypi.io as a standard practice, since it applies to any interoperability metadata changes we make)

wimglenn:
@ncoghlan  or anyone else who knows about this, could you please explain how to get it working properly?  I added in `setup.cfg`:

```
[metadata]
long_description_content_type = long_description_content_type = text/x-rst; charset=UTF-8
```

This is syntax documented as supported in setuptools ([here](http://setuptools.readthedocs.io/en/latest/setuptools.html#metadata)).  But it doesn't seem to be respected or used for everything - the generated metadata in the distribution always just shows "UNKNOWN" here.  Is the correct value getting dropped in the installation at some stage?

I use the version of setuptools v38.2.4, and I use "python setup.py bdist_wheel" to build the dist.

di:
Hi @wimglenn, there's an open PR to setuptools to fix this: https://github.com/pypa/setuptools/pull/1207

wimglenn:
@di ah, yes, that’s exactly the spot I ended up at in pdb when I was looking at it. Thanks!

PR #260, opened 2016-08-04
Fix order of setup.py arguments for upload command

skirpichev:
If sdist was invoked before bdist_wheel - upload first tries sdist upload
and some metadata on pypi will be missing (notable, Requires
Distributions section).

c.f.

```
$ python setup.py sdist bdist_wheel upload -r https://pypi.python.org/pypi
[...]
running upload
Submitting dist/Diofant-0.8.0a2.tar.gz to https://pypi.python.org/pypi
HTTP Error 400: This filename has previously been used, you should use a different version.
error: HTTP Error 400: This filename has previously been used, you should use a different version.
```

vs
```
$ python setup.py bdist_wheel sdist upload -r https://pypi.python.org/pypi
[...]
running upload
Submitting /home/sk/src/diofant/dist/Diofant-0.8.0a2-py3-none-any.whl to https://pypi.python.org/pypi
HTTP Error 400: This filename has previously been used, you should use a different version.
error: HTTP Error 400: This filename has previously been used, you should use a different version.
```

Not sure if it's not a bug for setuptools, but it works in this way for recent setuptools too (checked with 25.1.1).

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -816,7 +816,7 @@ There are two options:
 
    ::
 
-    python setup.py sdist bdist_wheel upload
+    python setup.py bdist_wheel sdist upload
 
    This approach is covered here due to it being mentioned in other guides, but it
    is not recommended as it may use a plaintext HTTP or unverified HTTPS connection


jonparrott:
I'm okay with merging this as the change itself is trivial, however, I'm wondering if it's worthwhile to send a follow-up PR to document *why* this is necessary.

@jaraco (or anyone else) can you confirm the reason why metadata differs depending on the distribution type? Does it make a difference to `pip` for installation purposes?

jaraco:
I use [sdist then bdist_wheel in all of my projects](https://github.com/jaraco/skeleton/blob/84b53d9/setup.cfg#L3). I haven't had any complaints about metadata being missing. Moreover, when I look at [a release I've made recently](https://pypi.python.org/pypi/jaraco.mongodb/6.3.1), it has the Requires-Distributions metadata, so this report seems to be invalid to me. Of course, things may have changed in the interim, but it's not apparent from the OP that there's an actual issue here.

jonparrott:
If someone can present a reproducible case where this occurs we can document why it's necessary.

Thanks for the confirmation, @jaraco.

PR #264, opened 2016-08-15
Note that download_url should not be used

odinho:
I used a lot of time barking up that tree, only to find out it's not meant to be set.
Primarily because the python setuptools docs contain it.

Some background info in https://github.com/pypa/setuptools_scm/issues/95

### DIFF
--- a/source/distributing.rst
+++ b/source/distributing.rst
@@ -197,6 +197,12 @@ url
 
 Give a homepage url for your project.
 
+.. note::
+
+  :ref:`setuptools` also documents a field `download_url`, however this is
+  not generally used.  Your packages should rather be uploaded directly to
+  :term:`PyPI <Python Package Index (PyPI)>`.
+
 
 author
 ~~~~~~


jonparrott:
I like the idea of this, but I'm curious on whether we should just start a new section on discouraged setup.py arguments. I doubt this is the only one.

jwodder:
I believe the `requires`, `provides`, and `obsoletes` arguments are similarly discouraged.  `requires` dates from a pre-PyPI & setuptools time and has now been replaced by `install_requires`.  `provides` and `obsoletes` don't seem to have ever been used by anything, and there are apparently security/trust issues that prevent even trying to do anything with them.

I'm not sure if `platforms` is discouraged, but its functionality has definitely never been fully worked out.

`dependency_links` itself isn't deprecated, but pip's support for it definitely is.

jonparrott:
@ncoghlan what are you thoughts on a separate section/topic on discouraged/deprecated setup args? I'm leaning towards closing this and tracking that in an issue so we can get consensus around everything we want to discourage.

ncoghlan:
@jonparrott That sounds sensible to me, especially as such an issue could be handled across multiple PRs, starting with the ones that aren't controversial, and then moving on to the less clearcut ones.

jonparrott:
Closing in favor of #293 to more comprehensively document this stuff.

Issue #210, opened 2016-01-04
Guidlines for making a PyPI-friendly readme


ddbeck:
A [warehouse issue](https://github.com/pypa/warehouse/issues/876) brought up the idea of documenting how to format code in READMEs for PyPI. Rather than covering code formatting alone, I thought it might be useful to have a section that covers setting the contents of a project's PyPI home page:
- explaining where the content of a project's page on PyPI comes from (i.e., `long_description`)
- setting the contents of `long_description` to the content of a project's README file (like the [pypa sample project does](https://github.com/pypa/sampleproject/blob/master/setup.py#L16))
- using ReST to format the document and specifically calling attention to the PyPI-specific case of pygments lexers

Before I started writing, I thought I would open this issue for feedback or additional ideas.

dstufft:
Note: you can test a long_description using the exact same code that PyPI (and Warehouse) uses (modulo version numbers) using:

```console
$ pip install readme
$ python setup.py check -r -s
```

brainwane:
@njsmith suggested it might also be good to link to https://pypi.python.org/pypi/pyroma .

brainwane:
I now see that https://packaging.python.org/tutorials/distributing-packages/?highlight=https#uploading-your-project-to-pypi does mention [readme_renderer](https://github.com/pypa/readme_renderer/).

I do think a bit more explanation, mapping bits of the README to bits of a PyPI project detail page, would be useful.

brainwane:
The packaging and distribution tutorial currently says:

> All projects should contain a readme file that covers the goal of the project. The most common format is reStructuredText with an “rst” extension, although this is not a requirement.

I think that right now we should just say it is a requirement, to simplify the experience for users, till the multiple steps laid out in https://github.com/pypa/warehouse/issues/869#issuecomment-340928703 are completed. This would also reduce conflict with [the sample project README](https://github.com/pypa/sampleproject/blob/master/README.rst) which says that it should be written in reStructuredText.

This would reduce confusion. I was advising a person today who was trying out the new PyPI who asked,

> 2. The docs say that rst isn't the only format for readmes, but don't say what other formats are accepted. Can it handle markdown? 

Until we can confidently list what other formats will render well, let's just advise people to use `.rst`.

di:
I think the docs are implying that plaintext is also an option, if users don't want to use reStructuredText.

ncoghlan:
[...]

PR #266, opened 2016-08-30
Add documentation on documenting a package.

jaraco:
This new section attempts to capture my best understanding of the best practices for documenting a project. I welcome feedback and am happy to iterate it or for others to do so.

### DIFF
--- a/source/documentation.rst
+++ b/source/documentation.rst
@@ -76,6 +76,14 @@ If the docs rely on code in the project itself or its
 dependencies, you will also want to configure the
 "Install Project" option in Advanced Settings.
 
+Configure the Webhook
+---------------------
+
+In order for the documentation to be rebuilt when changes
+are published to the source repository, a webhook must
+be enabled. Follow `these directions to configure webhooks
+<http://docs.readthedocs.io/en/latest/webhooks.html>`_.
+
 Porting from Distutils-built Documentation
 ------------------------------------------
 
--- a/source/documentation.rst
+++ b/source/documentation.rst
@@ -7,8 +7,15 @@ Documenting the Package
 
 A well-rounded project should also provide documentation to its users
 to provide a narrative description of the purpose and use of the
-package. This section describes the best practices for maintaining
-and publishing documentation for a typical package.
+package.
+
+While a simple README file that provides both end user and
+developer documentation is a good starting point for a new project,
+as a project matures, it frequently becomes desirable to provide
+more structured user-facing documentation.
+
+This section describes the best practices for maintaining
+and publishing documentation for a such a package.
 
 .. contents:: Contents
    :local:
--- a/source/documentation.rst
+++ b/source/documentation.rst
@@ -157,3 +157,8 @@ This technique had some advantages that the RTD process does not::
  - Authentication and Authorization of maintainers in PyPI was re-used
    for authorization to publish documentation.
  - Uniform API that required no additional manual steps.
+ - A link to the docs would be provided automatically in the project page.
+ - Documentation builds could be readily tested locally, whereas with
+   RTD, one must iterate on the RTD servers.
+ - RTD is a complex service with many more opportunities for failure
+   and a higher bar for resolving issues when they arise.
--- a/source/documentation.rst
+++ b/source/documentation.rst
@@ -1,0 +1,144 @@
+=======================
+Documenting the Package
+=======================
+
+:Page Status: Draft
+:Last Reviewed: 2016-08
+
+A well-rounded project should also provide documentation to its users
+to provide a narrative description of the purpose and use of the
+package. This section describes the best practices for maintaining
+and publishing documentation for a typical package.
+
+.. contents:: Contents
+   :local:
+
+
+Layout
+======
+
+Author your documentation as ReStructuredText in ./docs.
+
+
+Publishing
+==========
+
+The PyPA recommends that the project maintainers publish through
+`Read The Docs <https://readthedocs.org>`_ (RTD). RTD requires
+that your project be hosted in a publically-accessible SCM repository,
+although `work is underway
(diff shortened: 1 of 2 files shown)


ncoghlan:
I like @jaraco's proposed addition, but when it comes to the suggestion to expand the example project to match, I'll note that when I tried to use a cookiecutter template that included both docs and CI skeletons to start a side project, I found it overwhelming and ended up abandoning the repo (I don't even remember what that project was now, just the feeling of encountering the intimidating number of reminders of "Hey, you're making a big future time commitment here" in the skeleton and the subsequent decision to abandon the idea). I would have been much better off starting with the more minimalist PyPA sample project, focusing on the idea I actually wanted to play with, and only adding other things as they became relevant (if they ever did).

Accordingly, I'd strongly encourage keeping the number of things we tell *new* software publishers they *need* to provide to the bare minimum - a README that serves as both basic docs and developer guide, and the necessary machinery to actually build the software, publish it to PyPI and check that it works as expected prior to publication.

Given that perspective, I'd suggest framing the introduction to the new section as something more along the lines of "While a simple README file that provides both end user and developer documentation is a good starting point for a new project, as a project matures, it frequently becomes desirable to provide more structured user facing documentation."

dstufft:
I am +1 on documenting the recommendation to use RTD when your needs grow beyond a simple ``README``. Though perhaps it'd be better to structure this a bit differently to provide more emphasis on the fact that a simple ``README`` is more than enough (and perhaps even document the fact the ``README`` can be a ``rst`` file that you pass into ``long_description`` so it shows up on PyPI and such. In other words, spell out the choice between simple README and RTD better.

I'm not sure the step-by-step guide for publishing to pythonhosted.org is needed, we're trying to move folks away from that so telling new people how to do it seems counterproductive and existing projects already know how to do it.

jaraco:
Part of the reason I wanted to document pythonhosted.org was because I wanted to capture the simplicity and elegance of that approach (mainly that everything could be contained in the repo with 0 additional commands or steps required). I would expect this section to be removed once the recommended approach reached a similar level of simplicity and functionality. As it stands, I'm not aware of any place I can reference that indicates what the old process was. Additionally, having this section helps provide context for the "porting" section.

ddbeck:
On the README file matter, I opened an issue about dealing with README files, #210. I think having this Sphinx/RtD coverage is a good idea. It'll serve nicely as a follow on to the README stuff, when I get around to wrapping up the draft I've started (I got a little sidetracked—I'll move it up my to do list, know that there's more interest in that sort of thing).

My only comment on the text here is that I'm a little uncomfortable with the word "simple" (as in "simple README" and "simple invocation"). It reads as minimizing the effort required to maintain a package, especially for someone's first time out. I'd hope that the reader would start to feel that it was simple after reading this, but this section might be better served by a show-don't-tell approach.

dholth:
docception

source/documentation.rst
```diff
@@ -0,0 +1,151 @@
...
+supplies an ``upload_docs`` command.
+
+Together, these commands enable the building and publishing of
+documentation as part of the package release process. One can
+create an alias, such as ``release`` below, which will upload the
+package assets and publish the documentation in one command::
+
+    # setup.cfg
+    [aliases]
+    release = sdist sphinx_build upload upload-docs
```
anarcat: `build_sphinx` is the actual command name, is this a typo?

source/documentation.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Next, ensure that sphinx is available during the build process by
+including it in the ``setup_requires`` of the setup script::
+
+    # setup.py
+    needs_sphinx = {'release', 'sphinx_build'}.intersection(sys.argv)
```
anarcat: same here

jaraco: You're right. Thanks for correcting this - it's fixed in a subsequent commit.

anarcat:
by the way, i am of the opinion that the current documentatin is great! it's fine to have two approaches provided as well, i don't see why we need to absolutely enforce RTD over pythonshoted.org. recommending it is fine, but we should provide usage for people that have needs that are not fulfulled by RTD or that dislike it for some reason.

source/documentation.rst
```diff
@@ -0,0 +1,144 @@
...
+and publishing documentation for a typical package.
+
+.. contents:: Contents
+   :local:
+
+
+Layout
+======
+
+Author your documentation as ReStructuredText in ./docs.
```
bignose-debian: The Docutils project asks that the word be spelled “reStructuredText”.

source/documentation.rst
```diff
@@ -0,0 +1,144 @@
...
+
+
+Publishing
+==========
+
+The PyPA recommends that the project maintainers publish through
+`Read The Docs <https://readthedocs.org>`_ (RTD). RTD requires
```
bignose-debian: RTD also requires that a new account, over and above the necessary PyPI account, is created and maintained with a different organisation.

This document should present tangible options for those who don't want a Read the Docs account, but *do* want documentation hosted for their PyPI package.

source/documentation.rst
```diff
@@ -0,0 +1,144 @@
...
+and not just a release or doc build.
+
+With that configuration, a simple invocation of::
+
+    python setup.py release
+
+will upload the source dist and publish the documentation for the
+project.
+
+This technique had some advantages that the RTD process does not::
```
bignose-debian: Advantages not yet listed here include:

* Advertising was not displayed.
* The same PyPI account was used to upload the distribution's program and its documentation.

source/documentation.rst
dstufft: I think we want to avoid adding a bunch of different options for people, RTD is the right answer for most people and if you're someone with specialized desires you're probably technically sufficient enough to know how to host some static files somewhere.

PR #273, opened 2016-12-14
Mention requirement of 'wheel' package for Universal Wheels

fboender:
Building "Universal Wheels" requires the "wheel" package to be installed. The manual doesn't mention this. This pull request fixes that.


jonparrott:
@fboender thanks for this, I'd prefer if you moved the `install the wheel library` part up into the preceding section title `Wheels` because it's equally application if you're building universal or platform wheel.s

fboender:
Will do!

jonparrott:
Closing in favor of #299

PR #277, opened 2017-01-16
Clarify confusing sentence

mgedmin:
We had a non-English speaker in #pypa asking about why the documentation seems to imply that `pip install -e` doesn't install dependencies.  Rewrite this to make it clear that dependencies will be installed, but --editable will have no effect on them.


ncoghlan:
Thanks for this - the new wording is much clearer!

PR #278, opened 2017-01-18
Update warehouse link to pypi.org

Flimm:
warehouse.python.org redirects to pypi.org, so let's update the links too.


ncoghlan:
Thanks!

Issue #231, opened 2016-03-21
Docs: "Upload your distributions"


guettli:
Most users don't read from top to bottom.

If you look at this part, then there is one information missing:

{{{
Upload your distributions

Finally, you can upload your distributions to PyPI.

There are two options:

    Use twine

    twine upload dist/*

}}}
From: http://python-packaging-user-guide.readthedocs.org/en/latest/distributing/#upload-your-distributions

I am missing: Where do the files in "dist/*" come from?

I search for "dist/" on the page (ctrl-f). The matches I found did not explain where these files in this directory come from.

I don't want much, just a sentence like this:

> You can now upload the files created in Link-To-Matching-Step-In-Docs with ...

annaraven:
So you want a phrase like "You can now upload the files created by packaging up your project as described in #packaging your project"? Given it's the very previous section, easily found by scrolling up one window(even on my tiny screen), I'm not convinced this would be helpful. 

However, I could see adding something like: "The first time you run the sdist or bdist command will create a new directory in your project, called dist/, and put the new .whl or .tar.gz there." might be helpful. 

Think that would be useful? You could easily find it then with ctrl-f even if you missed it in scrolling.

guettli:
@annaraven Yes, your sentence would make more sense, then mine:

> The first time you run the sdist or bdist command will create a new directory in your project, called dist/, and put the new .whl or .tar.gz there.

BTW, I don't like "or" in docs much. Is there a common term which matches for both file extensions?

annaraven:
Okay, how about "The first time you run the command to package up your project, a new directory will be created called dist/, and the new distribution file will be put there."

ncoghlan:
Closing this one based on @annaraven's PR.

guettli:
Thank you for fixing this small issue.

ncoghlan:
I mentioned the wrong issue in the commit message for https://github.com/pypa/python-packaging-user-guide/commit/094aad139bb2f3dcec4cee4d925970eacb1b62e5 - the correct reference is to #238

Issue #238, opened 2016-05-29
Update doc to include latest PEP 513 and many linux uploads to Pypi 


pombredanne:
The current doc is outdated and states only that Linux wheels are not supported on Pypi.

brettcannon:
This is specifically mentioned in https://packaging.python.org/distributing/#platform-wheels in a note.

edmorley:
It's not clear to me whether that note is now out of date, or whether PyPI doesn't yet support manylinux (and that it was just that support was added to pip so far).

ncoghlan:
The note is out of date, we just missed fixing it when support was enabled on PyPI. See numpy for example: https://pypi.python.org/pypi/numpy

edmorley:
Many thanks :-)

edmorley:
Whilst the docs now mention manylinux1 after #283, it would be great if it was easier to find example projects that have a working build/upload setup eg using Travis. I found this example project, but I wonder if there are any other real-world examples too?
https://github.com/pypa/python-manylinux-demo

PR #283, opened 2017-03-14
Issue #231: Cover manylinux1 in "Platform Wheels"

ncoghlan:


johnyf:
For posterity, see #238 for discussion.

PR #290, opened 2017-04-06
Add topic on namespace packages

### LINKED ISSUE #265: there's no documentation of how to package namespace packages
People seem to really want to do this, and they resort to black magic in order to make it happen.  See for example https://github.com/google/protobuf/issues/1296.  There should probably be clear documentation of how to achieve namespace package packaging in a reliable way.

I think https://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=namespace#namespace-package-support is authoritative in this case; should it be referenced?

jonparrott:
Resolves #265.

### DIFF
--- a/source/namespace_packages.rst
+++ b/source/namespace_packages.rst
@@ -1,0 +1,226 @@
+============================
+Packaging namespace packages
+============================
+
+:Page Status: Draft
+:Last Reviewed: 2017-05-04
+
+Namespace packages allow you to split a single Python package across multiple
+distributions. For instance, if you have the following package structure:
+
+.. code-block:: text
+    
+    mynamespace/
+        __init__.py
+        subpackage_a/
+            __init__.py
+            ...
+        subpackage_b/
+            __init__.py
+            ...
+    setup.py
+
+And you use this package in your code like so::
+
+    from mynamespace import subpackage_a
+    from mynamespace import subpackage_b
+
+Then you can break these sub-packages into two separate distributions:
+
+.. code-block:: text
+    
+    mynamespace-subpackage-a/
+        setup.py
+        mynamespace/
+            subpackage_a/
+                __init__.py
+
+    mynamespace-subpackage-b/
+        setup.py
+        mynamespace/
+            subpackage_b/
+                __init__.py
+
+Each sub-package can now be separately installed, used, and versioned.
+
+Namespace packages can be useful for a large collection of loosely-related
+packages (such as a large corpus of client libraries for multiple products from
+a single company). However, namespace packages come with several caveats and
+are not appropriate in all cases. A simple alternative is to use a prefix on
+all of your distributions such as ``import mynamespace_subpackage_a``.
+
+
+Creating a namespace package
+============================
+
+There are currently three different approaches to creating namespace packages:
+
+#. Use `PEP 420-style namespace packages`_. This is recommended if packages in
+   your namespace only ever need to support Python 3 and installation via
+   ``pip``.
+#. Use `pkgutil-style namespace packages`_. This is recommended for new
+   packages that need to support Python 2 and 3 and installation via both
+   ``pip`` and ``python setup.py install``.
+#. Use `pkg_resources-style namespace packages`_. This method is recommended if
+   you need compatibility with packages already using this method or if your
+   package needs to be zip-safe.
+
+The three methods of creating namespace packages are largely not
+cross-compatible. It's inadvisable to use different methods in distributions of
+the same namespace package.
+
+PEP 420-style namespace packages
+--------------------------------
+
+`PEP 420`_ outlined **implicit** namespace packages. All that is required to
+create an implicit namespace package is that you just omit ``__init__.py``
+from the namespace package directory. An example file structure:
+
+.. code-block:: text
+
+    setup.py
+    mynamespace/
+        # No __init__.py here.
+        subpackage_a/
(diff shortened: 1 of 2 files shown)


jonparrott:
> since the import system is covered in the language reference these days, I think we should adopt "native namespace packages" as the preferred term over "PEP-420-style namespace packages"

SGTM, I'll incorporate that.

>  at the import system level, the compatibility between native namespace packages and pkgutil-style namespace packages should be solid, and I think we should commit to also maintaining that compatibility across the PyPA tooling so folks can mix and match them freely (leaving pkg_resources-style namespace packages as the only "all-or-nothing" case)

For sure, the problem is that according to [my tests](https://github.com/jonparrott/namespace-pkg-tests/blob/master/table.md) there are a *few* issues with that. Notably, if you install the first package with setuptools and the second with pip it'll fail. I'm hesitate to advertise compatibility until that's resolved. @jaraco Any ideas what's going on in that case?

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
+============================
+Packaging namespace packages
+============================
+
+:Page Status: Draft
+:Last Reviewed: 2017-05-04
+
+Namespace packages allow you to split a single Python package across multiple
+distributions. For instance, if you have the following package structure:
```
jonparrott: Good point. I updated it to reference terms an disambiguate `distributions`. Let me know what you think. I suspect no matter what we do this sentence will be difficult to grok, but the example below is hopefully illustrative enough to convey understanding.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+            subpackage_b/
+                __init__.py
+
+Each sub-package can now be separately installed, used, and versioned.
+
+Namespace packages can be useful for a large collection of loosely-related
+packages (such as a large corpus of client libraries for multiple products from
+a single company). However, namespace packages come with several caveats and
+are not appropriate in all cases. A simple alternative is to use a prefix on
+all of your distributions such as ``import mynamespace_subpackage_a``.
```
jonparrott: Perhaps, but most style guides discourage aliasing. I've added it as an `or`. :)

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+
+
+Creating a namespace package
+============================
+
+There are currently three different approaches to creating namespace packages:
+
+#. Use `PEP 420-style namespace packages`_. This is recommended if packages in
```
jonparrott: Good point. Done.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+   your namespace only ever need to support Python 3 and installation via
+   ``pip``.
+#. Use `pkgutil-style namespace packages`_. This is recommended for new
+   packages that need to support Python 2 and 3 and installation via both
+   ``pip`` and ``python setup.py install``.
+#. Use `pkg_resources-style namespace packages`_. This method is recommended if
+   you need compatibility with packages already using this method or if your
+   package needs to be zip-safe.
+
+The three methods of creating namespace packages are largely not
```
jonparrott: Done.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+cross-compatible. It's inadvisable to use different methods in distributions of
+the same namespace package.
```
jonparrott: Good point. Done.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+
+.. code-block:: text
+
+    setup.py
+    mynamespace/
+        # No __init__.py here.
+        subpackage_a/
+            # Sub-packages have __init__.py.
+            __init__.py
+            module.py
```
jonparrott: Done in the very first section and examples and explicitly called out that sub-packages and modules can participate in namespace packages.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+
+It is extremely important that every distribution that uses the namespace
+package omits the ``__init__.py``. If any distribution does not, it will cause
+the namespace logic to fail and the other sub-packages will not be importable.
```
jonparrott: Done.

source/namespace_packages.rst
```diff
@@ -0,0 +1,226 @@
...
+
+The ``__init__.py`` file for the namespace package needs to contain **only**
+the following:
+
+.. code-block:: python
+
+    __path__ = __import__('pkgutil').extend_path(__path__, __name__)
+
+**Every** distribution that uses the namespace package must include an
+identical ``__init__.py``. If any distribution does not, it will cause the
```
jonparrott: Done.

jonparrott: Done via a `tip` in the native section.

jonparrott:
@ncoghlan Thanks for the review, it's ready for another round. :)

ncoghlan:
Very nice, thank you! In reviewing this, it made me realise that now that namespace packages are documented, it might be feasible to also cover the two main options to writing plugin systems for Python applications: https://github.com/pypa/python-packaging-user-guide/issues/291

jonparrott:
@ncoghlan my pleasure! One small note - the examples for namespace packages currently live on my personal github, it might make more sense for those to live under the pypa org. I'm happy to populate that repo if someone can create it.

PR #294, opened 2017-04-11
Add topic on plugin discovery

### LINKED ISSUE #291: Additional topic for plugin management?
Reviewing @jonparrott's PR for an Additional Topic on Namespace Packages (https://github.com/pypa/python-packaging-user-guide/pull/290) that reviewed the available options made me wonder whether or not it might be worth doing a similar comparison between entry points based plugins (http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins), and namespace based plugins using https://docs.python.org/3/library/pkgutil.html#pkgutil.iter_modules (or higher level helpers around that like https://straightplugin.readthedocs.io/en/latest/)

The key tradeoffs are that:

- pkgutil based plugin models work for any importable modules in the given namespace, regardless of whether or not they have install-time metadata available. However, there's no way to reliably get the list of available modules without actually looking for them through the Python import system.
- pkg_resources based plugin models require that installation metadata be available (so you'll need to use `pip -e .` or `python setup.py develop` to generate that during development), but offer the benefit that the plugins that will be installed can be determined based on that static metadata without actually installing the component first

So if you want to manage some kind of "plugin catalog", you either need to use the entry point based approach, or come up with your own scheme for publishing the plugin metadata.

If you don't need a programmatically generated catalog though (e.g. when the plugins are just a way of separating out backends for different components and services), then the pkgutil based approach can be simpler to work with.

jonparrott:
Resolves #291

### DIFF
--- a/source/plugin_discovery.rst
+++ b/source/plugin_discovery.rst
@@ -76,10 +76,17 @@ under that namespace:
 
     import myapp.plugins
 
+    def iter_namespace(ns_pkg):
+        # Specifying the second argument (prefix) to iter_modules makes the
+        # returned name an absolute name instead of a relative one. This allows
+        # import_module to work without having to do additional modification to
+        # the name.
+        return pkgutil.iter_modules(ns_pkg.__path__, ns_pkg.__name__ + ".")
+
     myapp_plugins = {
-        name: importlib.import_module(name, package='myapp.plugins')
+        name: importlib.import_module(name)
         for finder, name, ispkg
-        in pkgutil.iter_modules(myapp.plugins.__path__)
+        in iter_namespace(myapp.plugins)
     }
 
 Specifying ``myapp.plugins.__path__`` to :func:`~pkgutil.iter_modules` causes
--- a/source/plugin_discovery.rst
+++ b/source/plugin_discovery.rst
@@ -1,0 +1,141 @@
+=============================
+Plugin creation and discovery
+=============================
+
+:Page Status: Complete
+:Last Reviewed: 2017-04-10
+
+Often when creating a Python application or library you'll want the ability to
+provide customizations or extra features via **plugins**. Because Python
+packages can be separately distributed, your application or library may want to
+automatically **discover** all of the plugins available.
+
+There are three major approaches to doing automatic plugin discovery:
+
+#. `Using naming convention`_.
+#. `Using namespace packages`_.
+#. `Using package metadata`_.
+
+
+Using naming convention
+=======================
+
+If all of the plugins for your application follow the same naming convention,
+you can use :func:`pkgutil.iter_modules` to discover all of the top-level
+modules that match the naming convention. For example, `Flask`_ uses the
+naming convention ``flask_{plugin_name}``. If you wanted to automatically
+discover all of the Flask plugins installed:
+
+.. code-block:: python
+
+    import importlib
+    import pkgutil
+
+    flask_plugins = {
+        name: importlib.import_module(name)
+        for finder, name, ispkg
+        in pkgutil.iter_modules()
+        if name.startswith('flask_')
+    }
+
+If you had both the `Flask-SQLAlchemy`_ and `Flask-Talisman`_ plugins installed
+then ``flask_plugins`` would be:
+
+.. code-block:: python
+
+    {
+        'flask_sqlachemy': <module: 'flask_sqlalchemy'>,
+        'flask_talisman': <module: 'flask_talisman'>,
+    }
+
+.. _flask: https://flask.pocoo.org
+.. _Flask-SQLAlchemy: https://flask-sqlalchemy.pocoo.org/
+.. _Flask-Talisman: https://pypi.python.org/pypi/flask-talisman
+
+Using namespace packages
+========================
+
+:doc:`Namespace packages <namespace_packages>` can be used to provide a
+convention for where to place plugins and provides a way to perform discovery.
+For example, if you make the package ``myapp.plugins`` a namespace package
+other :term:`distributions <Distribution Package>` can provide modules and
+packages to that namespace. Once installed, you can use
+:func:`pkgutil.iter_modules` to discover all modules and packages installed
+under that namespace:
+
+.. code-block:: python
+
+    import pkgutil
+
+    import myapp.plugins
+
+    myapp_plugins = [
+        name for finder, name, ispkg
+        in pkgutil.iter_modules(myapp.plugins.__path__)]
+
+.. TODO:: Is there a simple way to import both of these at the same time as discovery?
+
+Specifying ``myapp.plugins.__path__`` to :func:`~pkgutil.iter_modules` causes
+it to only look for the modules directly under that namespace. For example,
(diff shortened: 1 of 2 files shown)


jonparrott:
@ncoghlan first draft, definitely needs a sanity check.

source/plugin_discovery.rst
```diff
@@ -0,0 +1,141 @@
...
+#. `Using naming convention`_.
+#. `Using namespace packages`_.
+#. `Using package metadata`_.
+
+
+Using naming convention
+=======================
+
+If all of the plugins for your application follow the same naming convention,
+you can use :func:`pkgutil.iter_modules` to discover all of the top-level
```
ncoghlan: Huh, I just learned something, as I previously thought `iter_modules` had the same problem as `walk_packages`, where the latter can cause problems by eagerly importing everything that it thinks looks like a Python package. That may be dozens or even hundreds of unwanted packages when running in the system Python on Linux.

However, it turns out iter_modules is safe from that problem - it's strictly a consequence of the recursive walk behaviour in walk_packages.

So the naming convention approach is even better than I thought - I previously thought you needed to combine it with something like http://pythonhosted.org/distlib/tutorial.html#querying-a-path-for-distributions to get the listing of what was installed locally.

source/plugin_discovery.rst
```diff
@@ -0,0 +1,141 @@
...
+.. code-block:: python
+
+    {
+        'flask_sqlachemy': <module: 'flask_sqlalchemy'>,
+        'flask_talisman': <module: 'flask_talisman'>,
+    }
+
+.. _flask: https://flask.pocoo.org
+.. _Flask-SQLAlchemy: https://flask-sqlalchemy.pocoo.org/
+.. _Flask-Talisman: https://pypi.python.org/pypi/flask-talisman
```
ncoghlan: Potentially worth mentioning here is that of the three approaches, the naming convention approach is the only one that can be reliably queried through the "simple" API for index servers: https://www.python.org/dev/peps/pep-0503/#specification

This is due to the fact that the simple API only exposes project names and artifact file names - it doesn't publish any of the other metadata that's available through the XML-RPC and JSON APIs.

And even if you do access those other APIs, they don't currently provide import module or entry point metadata anyway.

source/plugin_discovery.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Using namespace packages
+========================
+
+:doc:`Namespace packages <namespace_packages>` can be used to provide a
+convention for where to place plugins and provides a way to perform discovery.
+For example, if you make the package ``myapp.plugins`` a namespace package
```
ncoghlan: Missing comma at the end of the line.

source/plugin_discovery.rst
```diff
@@ -0,0 +1,141 @@
...
+
+    import pkgutil
+
+    import myapp.plugins
+
+    myapp_plugins = [
+        name for finder, name, ispkg
+        in pkgutil.iter_modules(myapp.plugins.__path__)]
+
+.. TODO:: Is there a simple way to import both of these at the same time as discovery?
```
simonvanderveldt: @jonparrott What kind of info are you looking for here?
If you just want to import the modules you find it's enough to simply add:
```
importlib.import_module(name)
```

If you only want to import packages and not modules you can check that based on the `ispkg` value

ncoghlan: Switching the comprehension to be `{name: importlib.import_module(name) for ...}` should make it line up with your example output below.

simonvanderveldt:
@jonparrott one question I recently ran into is wether or not to use a subpackage as a parent for the plugins. Your examples all use a subpackage `myapp.plugins`, was this a deliberate choice? If so, it might be of value to mention why?

source/plugin_discovery.rst
jonparrott: Yeah, I thought so too, happy change. :)

source/plugin_discovery.rst
jonparrott: Done.

source/plugin_discovery.rst
jonparrott: Intentional, add a `then` to make it clearer. :)

source/plugin_discovery.rst
jonparrott: So here's the rub, for a namespace packages it actually returns ['a', 'b'] not ['myapp.plugins.a', 'myapp.plugins.b']. I took a shot at something, but let me know if ya'll have a better way.

jonparrott:
> one question I recently ran into is wether or not to use a subpackage as a parent for the plugins. Your examples all use a subpackage myapp.plugins, was that a deliberate choice? If so, it might be of value to mention why?

@simonvanderveldt good call, I added a little note about that.

> (hope it's OK if I as a random bystander ask some questions/provide some feedback)

Absolutely!

@ncoghlan all comments addressed, please take another look? :)

source/plugin_discovery.rst
ncoghlan: Ah, right - because we feed in the path directly, `iter_modules` doesn't know what the package prefix should be unless we provide it explicitly.

Rather than passing `package` to `import_module` (that's for resolving relative references that start with a leading '.'), you'll instead want to set the package prefix appropriately:

    def iter_namespace(ns_pkg):
        return pkgutil.iter_modules(ns_pkg.__path__, ns_pkg.__name__ + ".")

jonparrott: @ncoghlan what do you think about the new approach I used of using the second argument to `importlib.import_module`? I think that's a bit cleaner than the string concatenation.

If you're good with that, let's merge. :)

ncoghlan: @jonparrott `importlib.import_module` ignores its second argument when given module names that look absolute (i.e. no leading `.`). That means to get it to work, you'd have to:

1. Specify `'.'` as the prefix when calling `iter_modules`
2. Specify `ns_pkg.__name__` as the package when calling `import_module`

Hence the suggestion of instead encapsulating those details in a helper function that just iterates over the given namespace package, producing absolute module names that can be passed directly to `importlib.import_module` or `importlib.util.find_spec`.

ncoghlan: Note that I also think `iter_submodules` (since it's not actually specific to namespace packages) would make a good addition to `pkgutil` in 3.7+, accepting an already imported package as its sole argument, rather than a list of path entries and an optional prefix. That's orthogonal to the recommendation for how folks can implement their own `iter_submodules` today, though.

ncoghlan:
All looks good to me now, with the one exception that I'm fairly sure the current example namespace iteration code won't work unless your current directory is the `myapp.plugins` directory (hence making its contents importable under their unprefixed names)

source/plugin_discovery.rst
jonparrott: Gotcha. Done. :)

jonparrott:
Fixed up the namespace sample and verified that it works. Merging. Thanks, @ncoghlan!

PR #302, opened 2017-04-19
Update links to namespace sample projects

### LINKED ISSUE #292: Transfer example namespace package repo, update links
As part of #290, @jonparrott created a sample namespace package repo at https://github.com/jonparrott/namespace-pkg-tests

We have a couple of options for what to do with this repo:

1. Transfer the repo to one of the PyPA org admins/owners and from there into the PyPA org namespace: https://help.github.com/articles/transferring-a-repository-owned-by-your-personal-account/

2. Create a new repo under PyPA, and move the example package repo's contents into there

Either way, the links in the guide will need to be updated, but in the first case they'll also redirect to the new location in the meantime.

@dstufft, @qwcode Do you have any particular preference as to which path we take?

(Note: I'm not currently a PyPA owner/admin, so for either option, either someone else would need to do the transfer/create the repo, or else my privileges would need to be bumped up at the org level)

jonparrott:
Resolves #292

### DIFF
--- a/source/namespace_packages.rst
+++ b/source/namespace_packages.rst
@@ -120,7 +120,7 @@ A complete working example of two native namespace packages can be found in
 the `native namespace package example project`_.
 
 .. _native namespace package example project:
-    https://github.com/jonparrott/namespace-pkg-tests
+    https://github.com/pypa/sample-namespace-packages/tree/master/native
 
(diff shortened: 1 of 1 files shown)


jonparrott:
link-only change, merging. If there's any concerns let me know and I'll send a follow-up PR.

Issue #301, opened 2017-04-19
Determine what to do with the 'develop' branch


theacodes:
@ddbeck the `develop` branch seems to be pretty stale now. Do we want to port anything over to master? Can we delete it to clean up?

jonparrott:
Looking at the [comparison](https://github.com/pypa/python-packaging-user-guide/compare/develop?expand=1). There's some really useful stuff here. I've independently sent a PR for travis in #300, but the contributing and style guide can be really useful.

ddbeck:
@jonparrott I do think there's some things worth keeping from `develop`, but the overall decision to have a separate branch was a mistake. It was difficult to manage practically and it was kind of demotivating to work on that in isolation. Let's salvage the good stuff and then retire the `develop` branch.

I'm willing to do the work to make this happen and I can commit to spend time on this repo after Saturday and at least through to the following Sunday. But I'm not quite sure what the proper workflow would be. Sync up `develop` with `master`, then cherry pick some new PRs? A giant merge PR? Some of it is fairly complete and isolated from `master` (like the contributing and style docs), but the tutorial work is a bit iffy (maybe we could "hide" that in the ToC until it's more complete). If you have ideas on the best way to get us to something that we could get consensus on merging, I'll hop on it.

jonparrott:
Personally for me it would be better to a new, separate PR based on master for the contributing and style guide (you can use cherry-pick for this). Then let's figure out what to do with the new tutorial content.

I'm thinking I'm gonna propose a new information architecture here -something along the lines of:

* tutorials
  * setting up packaging tools
  * installing a package
  * creating a package
  * distributing a package
  * ...
* reference
  * glossary
  * namespace packages
  * plugins
  * ...

ddbeck:
OK, that sounds like a plan to me. I promise at least one PR on this no later than Sunday, April 30 (though maybe sooner—I'm traveling this week, so it's a little hard for me to predict).

Regarding IA, a long while back I wrote [this proposed outline](https://docs.google.com/document/d/1wQ6tso247SJTQnrK7NcMgEj5On_zwtN1qsynoPWCPAI/edit?usp=sharing). It's old (so probably has some gaps) but sounds consistent with you've got in mind. It may serve as a good starting point.

jonparrott:
> OK, that sounds like a plan to me. I promise at least one PR on this no later than Sunday, April 30 (though maybe sooner—I'm traveling this week, so it's a little hard for me to predict).

[...]

PR #307, opened 2017-04-30
Add initial contributor and style guide

ddbeck:
This PR adds a contributing and style guide started on the `develop` branch (with a few edits). Based on the discussion with @jonparrott from #301, this is the first step in retiring the `develop` branch. Some of this doc is based on conventions that I've seen elsewhere in the guide, other bits are aspirational (e.g., headings use different styles in different places, so I tried to aim for easier to explain and maintain). I'm happy to make any changes needed to get this into mergeable state.

### DIFF
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -90,7 +90,8 @@ Here's a quick check: try reading aloud to get a sense for your writing's voice
 and tone. Does it sound like something you would say or does it sound like
 you're acting out a part or giving a speech? Feel free to use contractions and
 don't worry about sticking to fussy grammar rules. You are hereby granted
-permission to end a sentence in a preposition.
+permission to end a sentence in a preposition, if that's what you want to end it
+with.
 
 When writing the guide, adjust your tone for the seriousness and difficulty of
 the topic. If you're writing an introductory tutorial, it's OK to make a joke,
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -44,8 +44,8 @@ Purpose
 
 The purpose of the |PyPUG| is
 
-    to be the authoritative resource on how to package, publish and install
-    Python distributions using current tools
+    to be the authoritative resource on how to package, publish, and install
+    Python projects using current tools.
 
 
 Scope
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -132,8 +132,8 @@ Conventions and mechanics
 
 **Use a gender-neutral style**
   Often, you'll address the reader directly with *you*, *your* and *yours*.
-  Otherwise, user gender-neutral pronouns *they*, *their*, and *theirs* or
-  avoid pronouns entirely.
+  Otherwise, use gender-neutral pronouns *they*, *their*, and *theirs* or avoid
+  pronouns entirely.
 
   | Wrong: A maintainer uploads the file. Then he…
   | Right: A maintainer uploads the file. Then they…
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -1,0 +1,156 @@
+.. |PyPUG| replace:: Python Packaging User Guide
+
+************************
+Contribute to this guide
+************************
+
+The |PyPUG| welcomes contributors! There are lots of ways to help out,
+including:
+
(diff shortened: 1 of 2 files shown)


source/contribute.rst
```diff
@@ -0,0 +1,156 @@
...
+contributions to be accepted into the project.
+
+
+Purpose
+-------
+
+The purpose of the |PyPUG| is
+
+    to be the authoritative resource on how to package, publish and install
+    Python distributions using current tools
```
ncoghlan: We conceded the "distributions" terminology debate a while back, so I'd suggest rephrasing this to:

> to be the authoritative resource on how to define, build, publish, and install Python distribution packages using current tools

That rephrasing also avoids the ambiguity with Python distributions like ActivePython, EPD/Canopy, Anaconda/miniconda, pyenv, Linux packages, Mac homebrew, etc

I also considered reordering the subactivities in order of the number of people that engage in them, from most common (installing packages published by others) to least common (publishing your own packages), but that phrasing was much more awkward than using the chronological order of the publishing process. Instead, I just split "package" into two subtasks ("define" and "build") so it wasn't be used as both a noun and a verb in the same sentence.

source/contribute.rst
```diff
@@ -0,0 +1,156 @@
...
+Imagine you're working on a Python project with someone you know to be smart and
+skilled. You like working with them and they like working with you. That person
+has asked you a question and you know the answer. How do you respond? *That* is
+how you should write this guide.
+
+Here's a quick check: try reading aloud to get a sense for your writing's voice
+and tone. Does it sound like something you would say or does it sound like
+you're acting out a part or giving a speech? Feel free to use contractions and
+don't worry about sticking to fussy grammar rules. You are hereby granted
+permission to end a sentence in a preposition.
```
ncoghlan: I tried to figure out how to rephrase this sentence so it ended in a preposition, but no such luck :)

source/contribute.rst
```diff
@@ -0,0 +1,156 @@
...
+
+  | Wrong: Pip uses…
+  | Right: pip uses…
+  |
+  | Wrong: …hosted on github.
+  | Right: …hosted on GitHub.
+
+**Use a gender-neutral style**
+  Often, you'll address the reader directly with *you*, *your* and *yours*.
+  Otherwise, user gender-neutral pronouns *they*, *their*, and *theirs* or
```
ncoghlan: s/user/use/

source/contribute.rst
pfmoore: "You are hereby granted permission to end a sentence in a preposition, if that's what you want to end it with." :-) Sorry, couldn't resist.

source/contribute.rst
pfmoore: In this instance, I'd be inclined to use the term "Python projects" in place of "Python distribution packages". At this point, the reader can be assumed to have a limited feel for the subtleties of technical terms like "distribution package", and "project", while ambiguous, probably corresponds most closely with the reader's intuitive understanding.

source/contribute.rst
ddbeck: YES! Seriously, I spent entirely too much time trying to come up with something like that.

ncoghlan: Excellent!

source/contribute.rst
ncoghlan: +1 for "Python projects", which would give:

> to be the authoritative resource on how to package, publish, and install Python projects using current tools

(since "package" should be in there somewhere, and with the noun form removed the sentence, we're free to use it as a verb again)

ddbeck: I was quoting `source/index.rst` here. Should I update this text in both places?

jonparrott: I'd say go ahead. :)

ncoghlan: Yep, update it in both - the current phrasing presumably dates from when we were still trying to disambiguate distribution packages from import packages by introducing a different term rather than just qualifying "package" as needed.

ncoghlan:
And done! Thanks @ddbeck :)

PR #311, opened 2017-05-19
Replace copy/paste artifact with intended phrasing

jaraco:


jonparrott:
Good catch!

jonparrott:
Build is failing for unrelated reasons, I'll investigate as soon as I can (at pycon ATM).

PR #313, opened 2017-05-20
Pin sphinx to 1.5.6 to unbreak CI

jonparrott:
See: https://github.com/sphinx-doc/sphinx/issues/3754

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,5 +2,5 @@ language: python
 sudo: false
 python:
 - 2.7
-install: pip install sphinx
+install: pip install sphinx==1.5.6
 script: make SPHINXOPTS="-W" html


jonparrott:
@ncoghlan @dstufft This isn't broken on readthedocs yet (it [seems](https://readthedocs.org/projects/python-packaging-user-guide/builds/5444385/) to still be using Sphinx 1.5.3), but I imagine it will break at some point. I'm not an admin on there so I can't fix it, but I'm happy to do so if one of you can add me as a maintainer over there (I can also make sure things still work when we can upgrade to 1.6.2).

jonparrott:
Merging as this is a CI-only change.

Issue #317, opened 2017-05-26
Master documentation plan


jonparrott:
This is the master plan for restructuring this guide. This issue's description *will be edited* as discussions occur. It's the plan of record for what the end state is and what steps we will take to get to the end state. This is proposed as a *minimal* set of changes to improve these docs *and encourage contributions*, once we have this foundation we can build on this to improve even further. Specifically, the work started by @ddbeck for [new tutorials](https://github.com/pypa/python-packaging-user-guide/issues/194) will be significantly easier after this structure is in place.

# Documentation Types

This guide will consist of 3 distinct documentation types:

* **tutorials** are focused on teaching the reader new concepts by accomplishing a goal. They are opinionated step-by-step guides. They do not include extraneous warnings or information. [example](https://docs.djangoproject.com/en/1.11/intro/)
* **guides** are focused on accomplishing a specific task and can assume some level of pre-requisite knowledge.. These are similar to tutorials, but have a narrow and clear focus and can provide lots of caveats and additional information as needed. They may also discuss multiple approaches to accomplishing the task. [example](https://packaging.python.org/namespace_packages/)
* **discussions** are focused on understanding and information. These explore a specific topic without a specific goal in mind. [example](https://packaging.python.org/requirements/)

We may revisit and add additional documentation types as needed.

# Documentation plan

This is the set of docs that we want to provide, organized in the final toctree.

1. Landing page
2. Tutorials
   1. Installing packages
   2. Creating & publishing packages
3. Guides (items and titles in this section are most likely to change)
   1. Installing packaging tools with Linux package managers
   1. Installing scientific packages
   1. Multi-version installs
   1. Single-sourcing the package version
   1. Supporting multiple python versions
   1. Packaging binary extensions
   1. Building wheels for Windows using AppVeyor
   1. Packaging namespace packages
   1. Creating and discovering plugins
   1. Index mirrors & caches
   1. Hosting your own index
   1. Deploying applications
4. Discussions
    1. install-requires vs requirements.txt
    1. pip vs easy_install
    1. wheel vs egg
    1. setup() arguments
5. Specifications
    1. Package metadata
    1. Package index interfaces
5. Glossary
6. How to get help
7. How to get involved

These pages will be moved to other projects:
1. Key projects will be moved into pypa.io

# Steps to get there

- [x] Get agreement on this plan.
- [x] Migrate to new theme. (#305)
- [x] Add a script to track broken deep links. (#315)
[...]

PR #318, opened 2017-05-31
Re-organize existing documentation into new structure.

jonparrott:
Towards #317.

Content is unchanged in renamed files, except for a few page titles that were renamed for consistency.

After this PR is merged, a new PR to re-writing the landing page to better surface the docs will be created, so don't sweat the landing page so much right now.

I will create all redirects on RTD once this is merged (ugh I hate that this is manual).

This is staged here: http://temp.theadora.io/pypug-reorg/index.html


source/discussions/deploying-python-applications.rst
```diff
@@ -1,7 +1,7 @@
 
-======================
-Application Deployment
-======================
+=============================
+Deploying Python applications
```
ddbeck: How about "Deploy Python applications" or "Deploy a Python application" instead? I was a bit aspirational when I wrote the contributing guide, but one of the guidelines I suggested was to move toward an imperative form (or, really, a truncated question form), since it seems to be a bit more searcher friendly.

I'll go ahead and make similar notes elsewhere, though I don't know if I'd count them as blockers.

source/guides/creating-and-discovering-plugins.rst
```diff
@@ -1,6 +1,6 @@
-=============================
-Plugin creation and discovery
-=============================
+================================
+Creating and discovering plugins
```
ddbeck: Like before, maybe "Create and discover plugins"?

source/guides/hosting-your-own-index.rst
```diff
@@ -1,7 +1,7 @@
 .. _`Hosting your Own Simple Repository`:
 
 ==================================
-Hosting your Own Simple Repository
+Hosting your own simple repository
```
ddbeck: Like before, maybe "Host your own simple repository"?

source/guides/packaging-binary-extensions.rst
```diff
@@ -1,8 +1,8 @@
 .. _`Binary Extensions`:
 
-=================
-Binary Extensions
-=================
+===========================
+Packaging binary extensions
```
ddbeck: Like before, maybe "Package binary extensions"?

source/guides/single-sourcing-package-version.rst
```diff
@@ -1,7 +1,7 @@
 .. _`Single sourcing the version`:
 
 ===================================
-Single-sourcing the Project Version
+Single-sourcing the package version
```
ddbeck: Like before, maybe "Single source the package version"?

source/discussions/deploying-python-applications.rst
jonparrott: I general, I really really try to avoid imperative voice for titles in technical documentation, especially because imperative voice is used to tell the reader what to do to accomplish a task within a section. Gerund construction seems to sit better with me for informing what a section will cover. [Django](https://docs.djangoproject.com/en/1.11/topics/db/queries/) notably uses gerund construction. Try imaging that document with imperative voice - it's a bit harder to process.

(I'm gonna hold off on merging this until you have a chance to respond)

ddbeck: I think there's a trend away from the gerund in headings in technical writing, at least for task-oriented documents. The argument for the imperative form is that it's easier for readers to skim for the content that addresses their problem. By using the imperative form, a writer can align more closely with the words that the reader might use to describe their problem, rather than using the words that the writer would use to describe the problem that someone else is having. It's a little bit of trick really; while it's technically the imperative, it's being used a bit more like an infinitive (with the "to" omitted for brevity). There are some other arguments too (like differentiation in ToCs), but this is the big one.

So let's take the deployment heading as an example. Imagine you want to learn about deploying a Python application. You might ask questions like "How do I deploy my Python application?" or "What are the steps to deploy my Python application?" You can pretty readily skim for those formations. The gerund form isn't impossible, but it tends to be a bit less direct (e.g., "How do I go about deploying my Python application?") or doesn't address a task (e.g., "What is deploying a Python application?").

I think it's a good idea to write headings this way (incrementally—we don't need to go and retrofit everything at once) and since we you had changed some headings anyway, I figured now was a good time to bring it up. If you're not convinced, then it doesn't need to block the merge—but then we should probably revisit the headings guidance in the contributing guide.

(Oh about that Django doc: I don't love it in either case. There's a lot going on with the ORM abstraction and what it even means to _make queries_ when the abstraction is obscuring the fact that you are making a query. It's got problems with or without gerund headings.)

jonparrott: > I think there's a trend away from the gerund in headings in technical writing, at least for task-oriented documents.

Can you provide some samples? I want to see this trend and how it works in practice. Please don't take this as me not believing you, I'm open to changing my stance here (which is why I haven't merged this yet), it's just that there's so many places I see gerund construction over imperative voice, and that's generally been fairly constant in my career. This could be due to bias, as my company's internal style guide explicitly calls for gerund construction. I'm okay with going against the grain, but I'm going to need slightly more convincing.

>  think it's a good idea to write headings this way (incrementally—we don't need to go and retrofit everything at once) and since you had changed some headings anyway, I figured now was a good time to bring it up. If you're not convinced, then it doesn't need to block the merge—but then we should probably revisit the headings guidance in the contributing guide.

We should get this right now, as changing the heading breaks deep links (which is why I wrote the linkchecker script).

I definitely want you on board with whatever we do, as I don't want to alienate a contributor on a project that frankly needs help. :)

Another suggestion is that we can do a poll on distutils-sig if we can't reach agreement.

ddbeck: > Can you provide some samples?

Sure! I did some contract tech writing at Google recently and saw that their help centers (like [Inbox](https://support.google.com/inbox/?hl=en#topic=6067574) or [AdSense](https://support.google.com/adsense/?hl=en#topic=1250102)) use it often. You see it less often in their developer sites, though there are examples; I particularly like [this cool tutorial listing](https://cloud.google.com/getting-started/#quick-starts). I've also spotted it in some major sections in [the Docker documentation](https://docs.docker.com/engine/swarm/) (see the left ToC) recently. I can find more examples, if you like.

> it's just that there's so many places I see gerund construction over imperative voice, and that's generally been fairly constant in my career

I'll admit that the imperative form is a little bit trendy and the gerund is way more common. And, honestly, the gerund is not bad. I just think the imperative is a little better.

> I definitely want you on board with whatever we do, as I don't want to alienate a contributor on a project that frankly needs help. :)

You're giving the idea a good faith hearing. If you're not convinced, then that's fine. Sincere skepticism or disagreement isn't going to alienate me. Though I'll admit I like getting my way. 😃 

> Another suggestion is that we can do a poll on distutils-sig if we can't reach agreement.

Nah, that seems a bit much. Just let me know what you think of the examples. If you're not convinced, we can update the guidance and move on.

jonparrott: Thanks for all of this. I can definitely see how it could work, but I'm still leaning towards gerund construction.

I'm gonna leave this PR open until Friday and if @ncoghlan or @dstufft want to chime in and let me know their preference I can be swayed, otherwise I'll merge as-is :)

dstufft: I have no opinion and I am pretty sure I am not qualified to have an opinion.

ncoghlan: I think @ddbeck is write that it's a good way to structure a how to guide, but PyPUG is kind of a hybrid of a how to guide and a reference guide at the moment, so I think it makes more sense to defer such a style change until *after* the restructure.

And especially for the discussions, I think the gerund form more accurately conveys the kind of content to expect - they're more "here are the trade-offs to consider when deciding how to approach this task" rather than the imperative "here is how you should approach this task" (vendors and specific projects can be opinionated on that, while PyPUG avoids choosing sides)

jonparrott: Cool _ I'm gonna go ahead and merge this then. We can revisit as things evolve - I'm almost convince of a bit of a hybrid - for the *tutorials* I think I may move to @ddbeck's suggestion.

PR #324, opened 2017-06-15
Remove sidebar search to match Python docs

jonparrott:
Context: https://github.com/pypa/pypa-docs-theme/issues/3


jonparrott:
Merging as it's a trivial change, but as always holler if there's any concerns. :)

PR #329, opened 2017-06-18
Expand the section on the "name" argument

jwodder:
Rather than referring readers to a PEP via a link that doesn't even go to the right part of the page, this patch incorporates the salient points on project names from PEPs 426 and 503 into the section on the `name` argument to the `setup()` function.

### DIFF
--- a/source/tutorials/distributing-packages.rst
+++ b/source/tutorials/distributing-packages.rst
@@ -143,7 +143,7 @@ name
   name='sample',
 
 This is the name of your project, determining how your project is listed on
-:term:`PyPI <Python Package Index (PyPI)>`.  Per :pep:`426`, valid project
+:term:`PyPI <Python Package Index (PyPI)>`.  Per :pep:`508`, valid project
 names must:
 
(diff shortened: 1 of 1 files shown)


source/tutorials/distributing-packages.rst
```diff
@@ -142,10 +142,23 @@ name
 
   name='sample',
 
-This is the name of your project, and will determine how your project is listed
-on :term:`PyPI <Python Package Index (PyPI)>`. For details on permitted
-characters, see the :pep:`name <426#name>`
-section from :pep:`426`.
+This is the name of your project, determining how your project is listed on
+:term:`PyPI <Python Package Index (PyPI)>`.  Per :pep:`426`, valid project
```
ncoghlan: This can be updated to refer to the accepted PEP 508 now: https://www.python.org/dev/peps/pep-0508/#names

jwodder: Fixed.

ncoghlan:
Thanks!

PR #331, opened 2017-06-23
Add a license section to the distributing guide

### LINKED ISSUE #275: The used license should be distributed or not with the package?
Legal question:
Example of task: Creating Python pypa project with MIT license.
In your user guide the MIT license is only setup in setup.py is this good enough?
Or (MIT) LICENSE(.txt) file should be included in the package as well?
http://choosealicense.com/licenses/mit/

jonparrott:
Resolves #275


source/tutorials/distributing-packages.rst
```diff
@@ -106,6 +106,19 @@ For details on writing a ``MANIFEST.in`` file, see the `The MANIFEST.in template
...
+LICENSE.txt
+~~~~~~~~~~~
+
+Every package should include a license file detailing the terms of
+distribution. In many jurisdictions, packages without an explicit license can
+not be legally used or distributed by anyone other than the copyright holder.
+If you're unsure which license to choose, you can use resources such as
+`GitHub's Choose a License <https://choosealicense.com/>`_ or consult a lawyer.
+
+For an example, see the `LICENSE.txt
```
ddbeck: Less a change suggestion and more of an observation: I would write "see `file_name` from…" or "see the `file_name` file from…" This is a personal preference I had no idea I even had until this moment, so thank you for this learning experience. 😆

jonparrott: I'm trying to be consistent here. We'll eventually replace this whole page with new tutorials, so I'm just trying to get a band-aid in place for the time being. :)

PR #332, opened 2017-06-23
Note that MANIFEST.in does not affect wheels

### LINKED ISSUE #306: MANIFEST.in description should not mention bdist_wheel
The current description for MANIFEST.in gives the impression that MANIFEST.in can be used together with `python setup.py bdist_wheel`:
> A “MANIFEST.in” is needed in certain cases where you need to package additional files that python setup.py sdist (or bdist_wheel) don’t automatically include.

However, `python setup.py bdist_wheel` does not include in the resulting wheel the file "LICENSE.txt", which is included in MANIFEST.in.

EDIT: Tested with https://github.com/pypa/sampleproject.

jonparrott:
Resolves #306

### DIFF
--- a/source/tutorials/distributing-packages.rst
+++ b/source/tutorials/distributing-packages.rst
@@ -90,9 +90,8 @@ MANIFEST.in
 ~~~~~~~~~~~
 
 A "MANIFEST.in" is needed in certain cases where you need to package additional
-files that ``python setup.py sdist (or bdist_wheel)`` don't automatically
-include. To see a list of what's included by default, see the `Specifying the
-files to distribute
+files that are not automatically included in a source distribution. To see a
+list of what's included by default, see the `Specifying the files to distribute
 <https://docs.python.org/3.4/distutils/sourcedist.html#specifying-the-files-to-distribute>`_
 section from the :ref:`distutils` documentation.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/distributing-packages.rst
```diff
@@ -90,21 +90,21 @@ MANIFEST.in
 ~~~~~~~~~~~
 
 A "MANIFEST.in" is needed in certain cases where you need to package additional
```
ddbeck: Since we're here anyway, can you put the file name in backticks (or use Sphinx's `:file:` role) to be consistent with the note below?

source/tutorials/distributing-packages.rst
```diff
@@ -90,21 +90,21 @@ MANIFEST.in
...
 <https://github.com/pypa/sampleproject/blob/master/MANIFEST.in>`_ from the `PyPA
 sample project <https://github.com/pypa/sampleproject>`_
 
-
 For details on writing a ``MANIFEST.in`` file, see the `The MANIFEST.in template
 <https://docs.python.org/2/distutils/sourcedist.html#the-manifest-in-template>`_
 section from the :ref:`distutils` documentation.
 
+.. note:: ``MANIFEST.in`` does not affect binary distributions such as 
+  ``wheels``.
```
ddbeck: I don't think "wheels" needs to be code formatted (bonus: it fixes the weird source formatting).

jonparrott: yeah, good catch, leftover from when it was `bdist_wheel`.

source/tutorials/distributing-packages.rst
jonparrott: Sure.

PR #334, opened 2017-06-24
Cover the PyPI migration in Tool Recommendations

ncoghlan:
During the migration from the legacy PyPI implementation
to Warehouse, `pypi.python.org` and `pypi.org` are
essentially alternative tools for the tasks of publishing,
browsing, downloading and managing packages.

This adds them to the Tool Recommendations page, so
folks have a single place to check to be informed of the
current recommendations, rather than having to try to
keep up with any announcements made at the time the
recommendations change.

### DIFF
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -43,6 +43,71 @@ Packaging Tool Recommendations
   to :term:`PyPI <Python Package Index (PyPI)>`.
 
 
+Publishing Platform Recommendations
+===================================
+
+The original Python Package Index implementation (hosted at
+`https://pypi.python.org`_) is being phased out in favour of an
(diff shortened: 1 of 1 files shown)


ncoghlan:
Closed in favour of #335 (the deeplinks change means I need to do some local edits, and that's annoying for a PR I started entirely through the web UI)

jonparrott:
> the deeplinks change means I need to do some local edits, and that's annoying for a PR I started entirely through the web UI

Sorry! I can totally turn it off now that I'm down really moving stuff around, but I kind of wanted to keep it active until we write the new tutorials.

ncoghlan:
@jonparrott Sorry, the "annoying" there related to shifting a PR from web editing to local development, and that's entirely to do with GitHub not letting me create the branch in my own fork when I have commit access on the repo.

I like having the link stability checker :)

PR #339, opened 2017-06-28
Add a document on migrating uploads to PyPI.org

dstufft:
Feel free to edit this to be worded better. This is to support https://mail.python.org/pipermail/distutils-sig/2017-June/030766.html and I hope to link to this to instruct people how to migrate. If we're able to get this merged today that would be awesome.

### DIFF
--- a/source/guides/migrating-to-pypi-org.rst
+++ b/source/guides/migrating-to-pypi-org.rst
@@ -1,0 +1,52 @@
+.. _`Migrating to PyPI.org`:
+
+Migrating to PyPI.org
+=====================
+
+PyPI.org is a new, rewritten version of PyPI that is replacing the legacy code
+base located at pypi.python.org. As it becomes the default, and eventually only,
+version of PyPI people are expected to interact with, there will be a transition
+period where tooling and processes are expected to need to update to deal with
+the new location.
+
+This section covers how to migrate to the new PyPI.org for different tasks.
+
+
+Uploading
+---------
+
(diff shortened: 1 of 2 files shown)


ncoghlan:
@jonparrott Hmm, maybe we *should* turn the linkchecker off for now, at least in pre-merge CI - it does add a somewhat awkward extra step when it comes to adding new sections like this one.

source/guides/migrating-to-pypi-org.rst
```diff
@@ -0,0 +1,52 @@
...
+    password:yourpassword
+
+
+Then simply delete the line starting with ``repository`` and you will utilize
+the default version of your upload tool.
+
+If for some reason you're unable to upgrade the version of your tool to a
+version that defaults to using PyPI.org, then you may edit ``~/.pypirc`` and
+include the ``repository:`` line, but use the value
+``https://upload.pypi.org/legacy/`` instead.
```
ncoghlan: It may be worth including a parenthetical comment here that explains that the `legacy` in the URL refers to "Implementation of the legacy upload API", but I don't think that's a blocker for merging this.

ncoghlan:
I went ahead and merged this so you can start linking folks to it. Pacifying the link checker and iterating on the wording details can come later :)

jonparrott:
> @jonparrott Hmm, maybe we should turn the linkchecker off for now, at least in pre-merge CI - it does add a somewhat awkward extra step when it comes to adding new sections like this one.

I can turn it into a warning. #341.

PR #343, opened 2017-07-08
Changed a reference to the Mac operating system from "OS X" to "macOS" 

fmaida:


ddbeck:
Thanks for this PR, @fmaida. The change looks fine to me, though a quick check shows there's lots more instances of OS X or OSX in other pages, too. Is this something you'd be willing to find and replace elsewhere? I think it would help to be consistent throughout (or at least plan to get there at some point).

fmaida:
I can try to find as many as possible.

ddbeck:
Closing this in favor of @fmaida's follow-up PR, #345.

PR #345, opened 2017-07-09
Changing the reference from "OS X" to "macOS"

fmaida:


ddbeck:
@fmaida Thank you! Once I figure out why I can't get the linkchecker to pass, we'll be all set to merge this.

@jonparrott can you take a look at this linkchecking situation? Or maybe put some guidance in the README or contributor docs about the proper workflow for changing/adding links? I've fought with this for about a half hour and I'm making no progress.

For example, I can't get `nox.py` to run locally at all. This is as far as I can get:

```
$ python nox.py -s updatelinks
  File "nox.py", line 8, in <module>
    import nox
  File "/Users/ddbeck/Desktop/Managed Projects/Edit Python Packaging User Guide/python-packaging-user-guide/nox.py", line 11, in <module>
    @nox.session
AttributeError: 'module' object has no attribute 'session'
```

jwodder:
@ddbeck You're not supposed to run `nox.py` directly.  Instead, you need to install the `nox-automation` package and then run `nox -s updatelinks` from the directory containing `nox.py`.

ddbeck:
@jwodder thanks, that broke the logjam.

I'll open a separate issue for the linkcheck hassles and merge this PR.

PR #357, opened 2017-07-28
Add enscons to the list of key projects

### LINKED ISSUE #337: Add `enscons` to the Key Projects list as a non-PyPA project
@dholth's `enscons` is the main current example of using PEP 518 to bootstrap a full pip-installable C/C++ build system that also has the ability to emit Python packages, rather than working with the Python-specific distutils/setuptools model.

At the moment we don't mention it anywhere in the packaging user guide though, and I think it should at least be on the Key Projects list, since the two main questions we currently ask about new build system abstractions are:

* Does this work for flit?; and
* Does this work for enscons?

jonparrott:
Resolves #337.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -251,6 +251,23 @@ a complimentary command line tool to drive packaging, testing and release
 activities with Python.
 
 
+enscons
+=======
+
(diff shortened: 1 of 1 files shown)


ncoghlan:
Looks good to me! If @dholth would like any amendments we can put them in a new PR :)

PR #358, opened 2017-07-28
Add flit to the key projects list

### LINKED ISSUE #336: Add flit to the Key Projects list
`flit` 0.11 added the ability to upload sdists, addressing the one concern I had that kept me from personally recommending it as the default choice for packaging pure Python projects. (@takluyver pointed out on distutils-sig that 0.11 is also the version where `flit` switched to uploading to `pypi.org` by default)

`flit` should also have an entry in the Key Projects list.

@takluyver Would you prefer to add flit to the list as a non-PyPA project, or would you be interested in transferring the repo to be under the PyPA org? Being nominally part of PyPA doesn't actually make much difference in terms of the day-to-day operations of the project, it just means you're volunteering to be part of the PyPA collective that governs the default Python packaging experience (and you're effectively doing that already through your work on flit and the build system abstraction API).

jonparrott:
Resolves #336

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -251,6 +251,20 @@ a complimentary command line tool to drive packaging, testing and release
 activities with Python.
 
 
+flit
+====
+
+`Docs <https://flit.readthedocs.io/en/latest/>`__ |
+`Issues <https://github.com/takluyver/flit/issues>`__ |
+`PyPI <https://pypi.python.org/pypi/flit>`__
+
+Flit is a simple way to put Python packages and modules on PyPI. Flit packages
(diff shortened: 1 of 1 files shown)


ncoghlan:
@jonparrott Hmm, my attempt to resolve the merge conflict through the web UI appears to have confused GitHub, even though Travis is happy with the result :P

ncoghlan:
@jonparrott Ah, never mind, just needed to switch the merge strategy to Squash & Merge.

So, similar to the enscons entry, I figure if @takluyver would like any amendments, we can tackle those in a separate PR.

takluyver:
Thanks both! This looks good to me. :-)

PR #365, opened 2017-08-11
Add pypi.org term

### LINKED ISSUE #363: Glossary doesn't describe pypi.org
The glossary doesn't mention pypi.org (described in a note at the top of https://packaging.python.org/).  Based on that note and the page it links to, it seems like now that pypi.python.org is in read-only mode (post July 27, 2017), it should be described in the glossary.

https://packaging.python.org/glossary/#term-python-package-index-pypi

jonparrott:
Resolves #363

/cc @mpeckWhisker

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -137,9 +137,16 @@ Glossary
 
     Python Package Index (PyPI)
 
-        `PyPI <https://pypi.python.org/pypi>`_ is the default :term:`Package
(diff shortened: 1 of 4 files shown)


mpeckWhisker:
Thanks, this will help people discovering this in the future.

jonparrott:
Thanks for the quick review, @dstufft. I have another one coming for you in just a minute. :)

PR #369, opened 2017-08-29
Add a new tutorial on installing packages

jonparrott:
Towards #301 

Note this is written in the new tutorial format - it is intended to be short, concise, minimal, and opinionated. More guides and tutorials will be added to cover the topics presently covered by the existing installing tutorials.

Also note: this is in the new tutorials directory, so it isn't intended to be immediately consumed. Once the supplemental materials (guides, etc.) are available we can replace the current installing tutorial with this one.


source/new-tutorials/installing/installing-packaging-tools.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Afterwards, you should have pip:
+
+.. code-block:: bash
+
+    pip --version
+    pip 9.0.1 from /usr/local/lib/python2.7/dist-packages (python 2.7)
+
+
+Installing virtualenv
```
jonparrott: @ncoghlan @dstufft I'm curious if I even need this section - should we just tell them to use `venv` since this tutorial is Python 3 focused (we can add a *note* that for Python 2 they need to install and use `virtualenv` instead). WDYT?

source/new-tutorials/installing/installing-packaging-tools.rst
```diff
@@ -0,0 +1,104 @@
...
+Since you now have pip, you can install virtualenv by running:
+
+.. code-block:: bash
+
+    pip install --upgrade virtualenv
+
+.. Note:: If you're on Linux or macOS, you may need to run this with ``sudo``.
+    In general, you should avoid invoking pip with ``sudo`` as this installs
+    packages globally. However, virtualenv and pip are core utilities that
+    make sense to be installed globally.
```
ncoghlan: Nope, never ever recommend `sudo pip`, not even for `virtualenv` and `pip`. What folks *can* do is `pip install --upgrade --user virtualenv`, since their personally upgraded copy will shadow the system one *for them*, without being at risk of being overwritten by reinstallation of the system level package.

source/new-tutorials/installing/installing-packaging-tools.rst
ncoghlan: I like the idea of moving Python 2 content to a separate "Setting up Python 2 environments" guide, and then referencing that from relevant locations.

However, for this tutorial, I think this is a place where https://github.com/pypa/python-packaging-user-guide/issues/118 and https://github.com/pypa/python-packaging-user-guide/issues/367 apply, so we probably want to start recommending `pip install --user pipenv`, rather than recommending that folks manage their virtual environments by hand.

ncoghlan: (I initially suggested `pipsi install pew`, but that introduces a bootstrapping problem, where you first need to do `pip install --user pipsi` in order to get access to the `pipsi` command. We may still want to suggest that at some point, but it's orthogonal to the question of recommending an environment manager, and currently still has enough bootstrapping problems that it led to @kennethreitz describing the following optional `pipsi`-based bootstrapping approach for `pipenv`: http://docs.pipenv.org/en/latest/advanced.html#fancy-installation-of-pipenv)

source/new-tutorials/installing/index.rst
```diff
@@ -0,0 +1,19 @@
...
+.. toctree::
+   :maxdepth: 2
+   :hidden:
+
+   installing-packaging-tools
+   installing-and-using-packages
+
+This tutorial walks you through installing and using Python packages. It will
+show you how to install and use the necessary tools and make strong
+recommendations on best practices. Keep in mind that the Python packging
```
ncoghlan: Typo: `packging`

source/new-tutorials/installing/index.rst
```diff
@@ -0,0 +1,19 @@
...
+ecosystem is a varied one so this isn't the only way to install and use
```
ncoghlan: Slight wording tweak: `...a broad and varied one...`

source/new-tutorials/installing/index.rst
```diff
@@ -0,0 +1,19 @@
...
+Python packages, however, it does cover the most common use cases.
```
ncoghlan: Tweak: `most common tools and use cases`

source/new-tutorials/installing/installing-and-using-packages.rst
```diff
@@ -0,0 +1,141 @@
+Installing and using packages
+=============================
+
+This tutorial will show you how to install a Python package and use it. Before
+you read this tutorial make sure you have all the packaging tools by reading
+:doc:`installing-packaging-tools`.
+
+Creating a virtualenv
```
ncoghlan: As per https://github.com/pypa/python-packaging-user-guide/issues/367, let's lead with `pipenv` for the new tutorials. So this section would introduce `pipenv --three` (initialisation), `pipenv shell` (activation) and `pipenv install` (installing components)

source/new-tutorials/installing/installing-packaging-tools.rst
kennethreitz: Everyone I know uses virtualenv on Python 3, not venv.

kennethreitz:
Thanks for using Requests as the example! ✨🍰✨

source/new-tutorials/installing/installing-packaging-tools.rst
ncoghlan: @kennethreitz Same here, but I'm not sure I actually know anyone that didn't either learn Python 2 first, or else works professionally on maintaining Python 2 software :)

Either way, if we recommend `pipenv` as the default dependency & environment manager, then whether it's using `venv` or `virtualenv` underneath becomes a hidden implementation detail.

kennethreitz: @ncoghlan yeah, and pipsi requires users to modify their path, which is just asking too much for a general guide, i think.

jonparrott:
Thanks for reviewing @kennethreitz and @ncoghlan. I agree on pipenv - let me throw together a draft on that (hopefully tomorrow) and we'll go from there. That'll likely reduce this tutorial to just one page. :)

Two questions:

1. Does `--user` require additional setup? Doesn't the user need to put `site.USER_BASE` in their `PATH`? (According to https://github.com/pypa/pip/issues/3813 it looks like that's true).
2. Is pipenv well tested on Windows? I didn't see anything in the documentation about Windows (e.g. how does pipenv shell work?).

We can add separate guides for using pip/virtualenv for those who want to know.

kennethreitz:
@jonparrott yep, we have a full windows test suite (down at the moment, but it's historically been running and everything is kept up to date) and put an exhaustive amount of work to support windows. Pew provides windows support, thankfully.

jonparrott:
Good to hear. How much of a pain is it to bootstrap pipenv?

kennethreitz:
I have a [get-pipenv](https://github.com/kennethreitz/pipenv/blob/master/get-pipenv.py) executable (lifted from get-pip) that will install pip and then pipenv for you.

jonparrott:
Awesome. Thank you.

ncoghlan:
Good point re `PATH` - I'm used to `pip install --user` Just Working, since Fedora's defaults are set up that way :)

jonparrott:
Alright - re-wrote this to use Pipenv. :) Please take another look.

kennethreitz:
\o/

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+---------------------------------
+
+Before you go any further, make sure you have Python and that it's avalable
+from your command line. You can check this by simply running:
+
+.. code-block:: bash
+
+    python --version
+
+You should get some output like ``3.6.1``. If you do not have Python, please
```
kennethreitz: Why not 3.6.2?

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+with Node.js' `npm`_ or Ruby's `bundler`_, it is similar in spirit to those
+tools. While :ref:`pip` can install Python packages, Pipenv is recommended as
+it's a higher-level tool that simplifies dependency management for common use
+cases.
+
+Use ``pip`` to install Pipenv:
+
+.. code-block:: python
+
+    pip install --user pipenv
```
kennethreitz:
```
.. code-block:: shell

    $ pip install --user pipenv
```

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+
+
+.. Note:: This does a `user installation`_ to prevent breaking any system-wide
+    packages. If ``pipenv`` isn't available in your shell after installation,
+    you'll need to add the `user base`_'s ``bin`` directory to your ``PATH``.
+    You can find the user base by running ``import site; print(site.USER_BASE)``
+    in Python. For example, on Linux this will return ``~/.local`` so you'll
+    need to add ``~/.local/bin`` to your ``PATH``.
```
kennethreitz: Maybe add a note that this is typically done in `~/.profile`?

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+Installing packages for your project
+------------------------------------
+
+Pipenv manages dependencies on a per-project basis. To install packages,
+change into your project's directory (or just an empty directory for this
+tutorial) and run:
+
+.. code-block:: bash
+
+    cd myproject
```
kennethreitz: I really like having `$` preceding shell commands

kennethreitz: But I can understand why you'd want to not have that, if you don't want Windows users to feel alienated. I don't think it harms them, though.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+Pipenv will install the excellent `Requests`_ library and create a ``Pipfile``
+for you in your project's directory. The :ref:`Pipfile` is used to track which
+dependencies your project needs in case you need to re-install them, such as
+when you share your project with others. You should get output similar to this:
+
+.. code-block:: text
+
+    Creating a Pipfile for this project...
+    Creating a virtualenv for this project...
+    Using base prefix '/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6'
```
kennethreitz: I def think 3.6.2 should be used for the docs, as best practice :)

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+Now that Requests is installed you can create a simple ``main.py`` file to
+use it:
+
+.. code-block:: python
+
+    import requests
+
+    response = requests.get('https://httpbin.org/ip')
+
+    print('Your IP is {}'.format(response.json['origin']))
```
kennethreitz: needs to be `{0}` to work w/ Python 2.6. Not sure if that's a target or not.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: It's just example output (my mac has 3.6.1)

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Based on user research from cloud.google.com we determined that users indeed do get tripped up on this (as they tend to copy/paste). There's a css trick to add the $ in the margin so that it's not picked up on copy, but that would need to be done separately from this PR. The blocks are all annotated as bash, so if we do decide to do it at a later date it'll be easy.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Again, just example output. It'll be difficult to keep this up-to-date with the latest patch version, but we can likely try to bump for every minor version.

kennethreitz: I'd just patch it by hand :)

source/new-tutorials/installing-and-using-packages.rst
kennethreitz: Gotcha!

source/new-tutorials/installing-and-using-packages.rst
kennethreitz: I'd just document it as 3.6.2, personally, even if I had 3.6.1 installed.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+
+Using ``pipenv run`` ensures that your installed packages are available to
+your script. It's also possible to spawn a new shell that ensures all commands
+have access to your installed packages with ``pipenv shell``.
+
+
+Next steps
+----------
+
+Congratulations, you now know how to install and use Python packages! ✨ 🍰 ✨
```
kennethreitz: ;)

source/key_projects.rst
```diff
@@ -76,6 +76,16 @@ Dev irc:#pypa-dev
...
 A tool for installing Python packages.
 
+.. _Pipfile:
+
+Pipfile
+=======
+
+`Source <https://github.com/pypa/pipfile>`__
+
+``Pipfile`` and its sister ``Pipfile.lock`` are a replacement for the existing
```
ncoghlan: Rather than "... a replacement for the existing standard :ref:`pip`'s ...", I'd word this as "... a higher level application-centric alternative to :ref:`pip`'s lower level ...".

I think that's important, as Pipfile deliberately *isn't* a complete replacement for requirements.txt, and if you're doing things that don't quite fit the "develop and publish an integrated Python application" usage model, you're quite likely to end up wanting to drop back down to the less opinionated lower level interface.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
+Installing and using packages
+=============================
+
+This tutorial walks you through installing and using Python packages. It will
+show you how to install and use the necessary tools and make strong
+recommendations on best practices. Keep in mind that the Python packaging
+ecosystem is a broad and varied one so this isn't the only way to install and
+use Python packages, however, it does cover the most common tools and use
+cases.
```
ncoghlan: I think we may be able to provide better guidance on applicability here. Perhaps replace the second sentence with something like:

> Keep in mind that Python is used for a great many different purposes, and precisely how you want to manage your dependencies may change based on how you decide to publish your software. The guidance presented here is most directly applicable to the development and deployment of network services (including web applications), but is also very well suited to managing development and testing environments for any kind of project.

And then once we have a guide or discussion to link to, we can add a pointer to that with text like "If you find this approach isn't working well for you or your use case, you may want to explore these \[other options\](\<link\>)"

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+
+.. Note:: This guide is written for Python 3, however, these instructions
+    should work fine on Python 2.
```
ncoghlan: I'd qualify this as Python 2.7 (so we can ignore 2.6 compatibility)

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+---------------------------------
+
+Before you go any further, make sure you have Python and that it's avalable
+from your command line. You can check this by simply running:
+
+.. code-block:: bash
+
+    python --version
+
+You should get some output like ``3.6.2``. If you do not have Python, please
```
ncoghlan: Given your discussion with @kennethreitz , perhaps we could abuse the `|release|` notation to make it easy to update the Python version that we use for examples in one central place?

Alternatively, we could set `rst_prolog` to define a custom substitution field: https://stackoverflow.com/questions/36320992/sphinx-documentation-variables/36331678#36331678

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,166 @@
...
+
+.. code-block:: bash
+
+    cd myproject
+    pipenv install requests
+
+Pipenv will install the excellent `Requests`_ library and create a ``Pipfile``
+for you in your project's directory. The :ref:`Pipfile` is used to track which
+dependencies your project needs in case you need to re-install them, such as
+when you share your project with others. You should get output similar to this:
```
ncoghlan: I'd add in parentheses `(although the exact paths shown will vary based on your operating system and how you installed Python)`

source/new-tutorials/installing-and-using-packages.rst
ncoghlan: `python -m site` is a straightforward command to run that prints the value of `USER_BASE` (amongst other things)

source/new-tutorials/installing-and-using-packages.rst
ncoghlan: This is probably a good place to mention the following error with a caveat like "If you're new to programming and the command line in general, you may see an error like this":

    >>> python
      File "<stdin>", line 1, in <module>
    NameError: name 'python' is not defined

and explain that all of the code blocks in the tutorial are intended to be run at an *operating system* prompt, rather than at the interactive Python prompt.

https://opentechschool.github.io/python-beginners/en/getting_started.html#what-is-python-exactly was the best tutorial I could find on the exact details of how to open one.

jonparrott:
Thanks for the thorough review, @ncoghlan. I'll try to address all of these
tomorrow before I leave for vacation. :)

source/new-tutorials/installing-and-using-packages.rst
jonparrott: I tried the substitution but it doesn't work in code blocks or inline code markup, so it doesn't help here. :(

Great suggestion about the error. Added. :)

source/new-tutorials/installing-and-using-packages.rst
jonparrott: TIL! I switched to using that and added links to documentation on how to update your path.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -26,6 +28,20 @@ You should get some output like ``3.6.2``. If you do not have Python, please
...
+
+        >>> python
+        Traceback (most recent call last):
+          File "<stdin>", line 1, in <module>
+        NameError: name 'python' is not defined
+
+    It's because this command is intended to be run in a *shell* (also called
+    a *terminal* or *console*). See the Python for Beginners
+    `getting started tutorial`_ for an introduction to using your operating
+    system's shell and interacting with Python.
```
kennethreitz: love this

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -62,15 +79,19 @@ Use ``pip`` to install Pipenv:
...
     packages. If ``pipenv`` isn't available in your shell after installation,
     you'll need to add the `user base`_'s ``bin`` directory to your ``PATH``.
-    You can find the user base by running ``import site; print(site.USER_BASE)``
-    in Python. For example, on Linux this will return ``~/.local`` so you'll
-    need to add ``~/.local/bin`` to your ``PATH``.
+    You can find the user base by running ``python -m site`` which will print
+    site information including the user base. For example, on Linux this will
+    return ``USER_BASE: '~/.local'`` so you'll need to add ``~/.local/bin`` to
+    your ``PATH``. On Linux and macOS you can set your ``PATH`` permanently
+    by `modifying ~/.profile`_. On Windows you can set the user
```
kennethreitz: ```` around `~/.profile` ?

jonparrott: rst doesn't like it. :)

kennethreitz: Ah, the annoyances or ReST. It's also impossible to have both bold and italics at the same time!

jonparrott: It's the worst option except for all of the other options.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,193 @@
...
+This tutorial walks you through installing and using Python packages. It will
+show you how to install and use the necessary tools and make strong
+recommendations on best practices. Keep in mind that Python is used for a great
+many different purposes, and precisely how you want to manage your dependencies
+may change based on how you decide to publish your software. The guidance
+presented here is most directly applicable to the development and deployment of
+network services (including web applications), but is also very well suited to
+managing development and testing environments for any kind of project.
+
+.. Note:: This guide is written for Python 3, however, these instructions
```
duggelz: This renders as two lines of unformatted text in the preview, is that a quirk of the preview or a bug (or intentional)?  It looks like this:

```
Note

This guide is written for Python 3, however, these instructions should work fine on Python 2.7.
```

jonparrott: yeah github's preview is garbage

screen shot 2017-08-30 at 11 17 20 am

kennethreitz:
Really excited about this! If I can help accelerate this landing, let me know what needs done.

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,193 @@
...
+
+.. code-block:: bash
+
+    python --version
+
+You should get some output like ``3.6.2``. If you do not have Python, please
+install the latest 3.x version from `python.org`_ or refer to the
+`Installing Python`_ section of the Hitchhiker's Guide to Python.
+
+.. Note:: If you're newcomer and you get an error like this:
```
ncoghlan: Typo: Missing "a" before newcomer

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,193 @@
...
+    
+    .. code-block:: python
+
+        >>> python
+        Traceback (most recent call last):
+          File "<stdin>", line 1, in <module>
+        NameError: name 'python' is not defined
+
+    It's because this command is intended to be run in a *shell* (also called
```
ncoghlan: Might it be worth emphasising that this note also applies to all the other commands in the tutorial? For example:

> It's because this command, and the other suggested commands in this tutorial, are intended ...

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,193 @@
...
+
+Next steps
+----------
+
+Congratulations, you now know how to install and use Python packages! ✨ 🍰 ✨
+
+There's more resources you can look at to learn about installing and using
+Python packages:
+
+.. TODO Link to additional guides and resources.
```
ncoghlan: Maybe make these TODOs visible in the rendered version so there aren't just blank spaces while the link targets are being worked on?

source/new-tutorials/installing-and-using-packages.rst
```diff
@@ -0,0 +1,193 @@
...
+
+If you find this approach isn't working well for you or your use case, you may
+want to explore these other approaches:
+
+.. TODO Link to alternatives
```
ncoghlan: Suggested initial set of links:

* [briefcase](https://briefcase.readthedocs.io/en/latest/): briefcase is a tool that focuses on defining applications as Python packages, and then using that package definition to generate native application installers for popular client operating systems
* [pex](https://pex.readthedocs.io/en/stable/whatispex.html#whatispex): PEX is a single-file distribution format for executable Python virtual environments as zip archives containing a `__main__.py` file. Their main runtime requirement is that a suitable Python interpreter be available on the system running the application.
* [pipsi](https://github.com/mitsuhiko/pipsi#pipsi): Similar to `pex`, `pipsi` aims to run applications in self-contained virtual environments. Unlike `pex`, it generates these virtual environments dynamically inside the user's home directory for applications available as Python packages.
* [pew](https://github.com/berdario/pew#pew---python-env-wrapper): the underlying virtual environment manager used by `pipenv` (this is a lower level tool that offers more flexibility in where, when, and how virtual environments are created and updated)
* [pip-tools](https://github.com/jazzband/pip-tools#pip-tools--pip-compile--pip-sync): automated requirements.txt file management that distinguishes between abstract dependency combinations that are expected to work, and the specific pinned versions that will be tested and/or deployed (this is a lower level tool that offers more flexibility in when and how dependencies are installed and updated)
* [conda](https://conda.io/docs/): conda is a "cross-platform platform" that manages the Python interpreter and external binary dependencies in addition to managing Python modules. It is most useful in situations where it is easier to ensure that the conda package manager is preinstalled than it is to ensure that a suitable Python version is available.

And then make a note that the above list only covers cross-platform tools, and there are also other platform specific options available if users are only interested in a limited set of platforms.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Good catch, done.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: Done.

source/new-tutorials/installing-and-using-packages.rst
jonparrott: I don't want to put all of this in this doc, but it does make sense for a general overview guide. Moved these to https://github.com/pypa/python-packaging-user-guide/issues/367#issuecomment-326165340 for tracking. :)

kennethreitz:
💫✨🍰✨💫

kennethreitz:
Does this get auto-deployed?

jonparrott:
Yes - the Fastly config makes it take a second but it's now published here:

https://packaging.python.org/new-tutorials/installing-and-using-packages/

Feel free to share this link (we need feedback!), but keep in mind the URL is temporary. We'll eventually move this into `/tutorials/installing-and-using-packages` once I can ensure no content in the existing tutorial will be lost. :)

Thanks for reviewing @ncoghlan and @kennethreitz. :D

kennethreitz:
Replicated on Python-Guide.org and Pipenv.org:

- http://docs.python-guide.org/en/latest/dev/virtualenvs/
- http://docs.pipenv.org/en/latest/basics.html

Will link to this guide as the canonical source of truth, once the final URL has landed.

ncoghlan:
@jonparrott If you're currently busy, perhaps it would be easier if I put a PR together, so you only needed to review it, rather than write it?

jonparrott:
Go for it. :) I'm always happy to review.

ncoghlan:
New PR: https://github.com/pypa/python-packaging-user-guide/pull/402

Issue #367, opened 2017-08-24
Recommend dependency version management tooling based on use case


ncoghlan:
Similar to the problem in https://github.com/pypa/python-packaging-user-guide/issues/118 with regards to virtual environment management, our current tooling recommendations don't provide a pointer to higher level dependency version management tools like `pipenv` or `pip-tools`, they only refer to `pip` itself.

While we don't want to overwhelm people with choice, I think we're currently doing folks a disservice by letting them think `pip` is all there is, rather than guiding them towards more opinionated higher level tooling:

* `pipenv`: web application development (and potentially other forms of application development)
* `pip-tools`: general application development (useful when `pipenv` doesn't align with your needs)
* `conda`: language-independent cross-platform package (useful when you have extensive external dependencies)

ncoghlan:
`briefcase` would also be worth mentioning for cross-platform native app deployments (including web & mobile): https://pypi.python.org/pypi/briefcase

kennethreitz:
I'd love to see pipenv referenced! I think that'd do the community a great service (and pip-tools as well, of course). 

I don't do a good job of this myself on python-guide — I just point people to using pip. However, I just updated all my personal projects to recommending installation via pipenv last night. So, slowly encouraging adoption in that area. 

--------------------

My thoughts are that this is a good idea, if the information is presented simply, much in the way that Python-Guide presents the list of web frameworks:

http://docs.python-guide.org/en/latest/scenarios/web/

Note that it doesn't include every possible option, and it effectively instructs the user which framework to use.

ncoghlan:
@kennethreitz Heh, your advice on Django vs Flask is almost exactly how I was planning to describe the distinction between `pipenv` and `pip-tools` (i.e. if you don't have a strong opinion yourself, then you want to start with `pipenv`, since that's more comprehensive, whereas `pip-tools` is a better option if you need or want more flexibility over the precise dependency management behaviour).

kennethreitz:
✨🍰✨

jonparrott:
From @ncoghlan in https://github.com/pypa/python-packaging-user-guide/pull/369/files/b102d60fe4d8ac3cf6fce53fd5bf5f159d2f7fc8#r136228176

Suggested initial set of links:

* [briefcase](https://briefcase.readthedocs.io/en/latest/): briefcase is a tool that focuses on defining applications as Python packages, and then using that package definition to generate native application installers for popular client operating systems
[...]

PR #370, opened 2017-08-31
Fill out pipenv alternatives section

ncoghlan:
In the absence of a dependency management guide or discussion topic,
this just adds a list of links directly to the end of the pipenv tutorial page.

Longer term, I think it would be better to have a separate
"Dependency Management" discussion page that is linked from here,
but that would need to include more than just this list of links.

### DIFF
--- a/source/new-tutorials/installing-and-using-packages.rst
+++ b/source/new-tutorials/installing-and-using-packages.rst
@@ -186,7 +186,36 @@ Python packages:
 If you find this approach isn't working well for you or your use case, you may
 want to explore these other approaches:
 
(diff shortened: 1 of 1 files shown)


jonparrott:
Since this tutorial isn't yet out of experimental, I have a bit of a preference on not adding content that will need to be changed/removed later. I'd prefer to just fix the stuff #367 before we promote this- which I can do during my next pass after I get back from vacation.

But if you feel it's useful for now then lgtm. :)

ncoghlan:
Nah, I think you're right - this isn't urgent, so we can wait and use the content from this PR to help resolve #367 properly.

PR #371, opened 2017-08-31
Remove .git suffix from mercurial repository

pokoli:
Mercurial repositories does not have the .git suffix, so we should not use it on the example mercurial url.


jonparrott:
Thanks!

PR #385, opened 2017-09-29
Move existing installing packages tutorial into a guide

jonparrott:
This moves the current ["Installing Packages" tutorial](https://packaging.python.org/tutorials/installing-packages/) into a guide and updates some of the wording and recommendations.

Once this is merged, I will delete the current tutorial and replace it with the `pipenv` tutorial presently in `new-tutorials`.

### DIFF
--- a/source/guides/installing-using-pip-and-virtualenv.rst
+++ b/source/guides/installing-using-pip-and-virtualenv.rst
@@ -1,0 +1,377 @@
+Installing packages using pip and virtualenv
+============================================
+
+This guide discusses how to install packages using :ref:`pip` and
+:ref:`virtualenv`. These tools are recommended if :ref:`pipenv` does not suit
+your project's needs.
+
+.. note:: This doc uses the term **package** to refer to a
+    :term:`Distribution Package`  which is different from a :term:`Import
+    Package` that which is used to import modules in your Python source code.
+
+
+Installing pip
+--------------
+
+:ref:`pip` is the the Python package manager. It's used to install and update
+packages. You'll need to make sure you have the latest version of pip
+installed.
+
+
+Windows
++++++++
+
+The Python installers for Windows include pip. If you setup Python to be in
+your ``PATH`` you should have access to pip already:
+
+.. code-block:: bash
+
+    pip version
+    pip 9.0.1 from  c:\python36\lib\site-packages (Python 3.6.1)
+
+You should make sure that pip is up-to-date by running:
+
+.. code-block:: bash
+
+    pip install --upgrade pip
+
+
+Linux and macOS
+++++++++++++++++
+
+Debian and most other distributions include a `python-pip`_ package, if you
+want to use the Linux distribution-provided versions of pip see
+:doc:`/guides/installing-using-linux-tools`. 
+
+You can also install pip yourself to ensure you have the latest version:
+
+.. code-block:: bash
+
+    wget https://bootstrap.pypa.io/get-pip.py
+    sudo python get-pip.py
+
+Afterwards, you should have pip:
+
+.. code-block:: bash
+
+    pip --version
+    pip 9.0.1 from /usr/local/lib/python3.6/dist-packages (python 3.6)
+
+.. _python-pip: https://packages.debian.org/stable/python-pip
+
+Installing virtualenv
+---------------------
+
+:ref:`virtualenv` is used to manage Python packages for different projects.
+Using virtualenv allows you to avoid installing Python packages globally
+which could break system tools or other projects.
+
+You can install virtualenv by using pip:
+
+.. code-block:: bash
+
+    pip install --user virtualenv
+
+
+.. Note:: This does a `user installation`_ to prevent breaking any system-wide
+    packages. If ``virtualenv`` isn't available in your shell after
+    installation, you'll need to add the `user base`_'s ``bin`` directory to
+    your ``PATH``. See :ref:`the note in the installing packages tutorial
+    <pipenv-user-base>` for details on how to set your ``PATH`` correctly.
+
+.. Note:: If you are using Python 3.3 or newer the :mod:`venv` module is
+    included in the Python standard library. This can also create and manage
+    virtual environments, however, it only supports Python 3.
+
+.. _user base: https://docs.python.org/3/library/site.html#site.USER_BASE
+.. _user installation: https://pip.pypa.io/en/stable/user_guide/#user-installs
+
+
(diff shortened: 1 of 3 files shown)


jonparrott:
@pfmoore if you have time I would appreciate your review on this.

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -0,0 +1,377 @@
...
+Debian and most other distributions include a `python-pip`_ package, if you
+want to use the Linux distribution-provided versions of pip see
+:doc:`/guides/installing-using-linux-tools`. 
+
+You can also install pip yourself to ensure you have the latest version:
+
+.. code-block:: bash
+
+    wget https://bootstrap.pypa.io/get-pip.py
+    sudo python get-pip.py
```
ncoghlan: `sudo python` will typically run Python 2.7, so `sudo python3 get-pip.py` would be more reliable here.

However, we really shouldn't recommend this approach at all, since it's a good way for people to break either their system (if pip is provided natively), or else 3rd party repos like EPEL (if it isn't provided natively, but there are third party pacakges available). Instead, the better way to go is to install the system pip, and then use *that* to bootstrap a user level install of a more recent version (if the system one isn't already sufficiently recent).

Doing that looks like this:

    $ python3 -m pip --version
    pip 9.0.1 from /usr/lib/python3.6/site-packages (python 3.6)
    $ python3 -m pip install --user --upgrade pip
    Requirement already up-to-date: pip in ./.local/lib/python3.6/site-packages

Once the next version of `pip` is released, then the second command would actually install the newer version as a user install.

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -0,0 +1,377 @@
+Installing packages using pip and virtualenv
+============================================
+
+This guide discusses how to install packages using :ref:`pip` and
+:ref:`virtualenv`. These tools are recommended if :ref:`pipenv` does not suit
+your project's needs.
+
```
ncoghlan: This isn't quite right, as I'd actually recommend folks look at `pip-tools` and `pew` before dropping all the way down to the lowest level building blocks.

Essentially, the layering we have is:

* Comprehensive, but opinionated (the Django and Plone equivalents for Python package management):
    * `pipenv`: integrated Python-specific environment management & dependency management (default recommendation)
    * `hatch`: integrates version release management tooling as well (even more opinionated than `pipenv`)

* More narrowly focused packages with independent management of concerns (the Flask, SQL Alchemy, and Jinja2 equivalents for Python package management):
  * `pip-tools`: dependency management
  * `pew`: virtual environment management
  * `flit`: build and release management

* The lowest level building blocks (the `wsgiref` equivalents for Python package management)

  * `pip`: package installation management
  * `virtualenv`/`venv`: virtual environment creation

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -0,0 +1,377 @@
...
+:ref:`pip` is the the Python package manager. It's used to install and update
+packages. You'll need to make sure you have the latest version of pip
+installed.
+
+
+Windows
++++++++
+
+The Python installers for Windows include pip. If you setup Python to be in
+your ``PATH`` you should have access to pip already:
```
pfmoore: Do we want to presume users have Python on `PATH`? It's not the default as the `py` launcher is the recommended approach for accessing Python, and I'd prefer we stick with that as it's "what works out of the box" for all users.

On a default install, pip is available via `py -m pip`, and the upgrade command below would be `py -m pip install --upgrade pip`.

jonparrott:
Thanks for reviewing @pfmoore and @ncoghlan - I'm at a team offsite this week but will try to address all of these comments on Friday. :)

source/guides/installing-using-pip-and-virtualenv.rst
jonparrott: Gotcha, do you mind if I defer this for #367 and use the phrase "These are the lowest-level tools for managing Python packages and are recommended if higher-level tools do not suit your needs."?

Doing so will unblock us for now and give a good link point once #367 is ready.

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -0,0 +1,377 @@
...
+
+.. note:: This doc uses the term **package** to refer to a
+    :term:`Distribution Package`  which is different from a :term:`Import
+    Package` that which is used to import modules in your Python source code.
+
+
+Installing pip
+--------------
+
+:ref:`pip` is the the Python package manager. It's used to install and update
```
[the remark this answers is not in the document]
jonparrott: Done and done.

source/guides/installing-using-pip-and-virtualenv.rst
jonparrott: Great point. Done.

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -0,0 +1,377 @@
...
+.. code-block:: bash
+
+    pip version
+    pip 9.0.1 from  c:\python36\lib\site-packages (Python 3.6.1)
+
+You should make sure that pip is up-to-date by running:
+
+.. code-block:: bash
+
+    pip install --upgrade pip
```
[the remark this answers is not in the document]
jonparrott: Done.

source/guides/installing-using-pip-and-virtualenv.rst
jonparrott: Done, I've also switched to the `python -m` form for virtualenv commands throughout to avoid PATH issues. I left the post-virtualenv `pip` commands alone because activating a virtualenv will put everything on path.

jonparrott:
I think I addressed all comments, please take another look @pfmoore and @ncoghlan? (Thanks!)

pfmoore:
Looks good to me. I have some reservations about the whole section on activating a virtualenv (I've seen a lot of people get confused about what "activation" means and give it a greater significance than is needed) but I don't see a good way of addressing that in a document like this. Longer term, if we move towards recommending higher level tools like pew, this should solve itself, so I don't think it's a problem for now.

jonparrott:
@pfmoore agreed, I tried to make it a bit more clear by saying that it puts the environment-specific stuff in PATH even though that isn't the whole story. 

Thank you for reviewing, I'm gonna give @ncoghlan another chance to take a look before merging.

PR #386, opened 2017-09-29
Split specifications page into multiple pages

### LINKED ISSUE #319: Specifications page is hard to read
Looking at https://packaging.python.org/specifications/ the page is kind of hard to read the current grouping doesn't seem to make much sense. It currently groups everything under either "Package distribution metadata" or "Package index interfaces", but the latter only has one item in it and the former has a bunch, some of which are only tangentially related to each other.

Perhaps it'd be best to split this up more like:

* Core Metadata
  * PEP 345
  * Provides-Extra (multiple use)
  * Description-Content-Type
* Version specifiers
* Dependency Specifiers
* Declaring Build System Dependencies
* Distribution Formats
  * Source Distribution Format
  * Binary Distribution Format
* Platform Compatibility Tags
* Recording Installed Distributions
* Simple repository API

In addition, the text for the different things kind of run together, this isn't a page you're likely to read top to bottom, but one that you're going to be looking for the specific specification you're looking for. Maybe splitting each of the top level list items into their own sub page is a better idea here? Maybe the theming discussion in https://github.com/pypa/python-packaging-user-guide/issues/304 will make this easier to process. Perhaps using something different than different heading levels to lay out the documentation will make it easier. Maybe even this is just the best we can do!

I dunno, but the page as it stands is kind of hard to read for me.

di:
This should fix #319:

* Each section on `/specifications` is now it's own page;
* Included the contents of PEP 345 in `/specifications/core-metadata`;
* Added relevant `versionadded` directives to metadata field specs.


pradyunsg:
> Included the contents of PEP 345 in /specifications/core-metadata

Would this result in redundant information?

di:
@pradyunsg Yes, this will "duplicate" fields in PEP 345 for a short time until we can write a new PEP which supersedes it.

You might have missed that this was specifically requested in https://github.com/pypa/python-packaging-user-guide/issues/319#issuecomment-332405583.

pradyunsg:
> Yes, this will "duplicate" fields in PEP 345 for a short time until we can write a new PEP which supersedes it.

Oh, cool then. :)

> You might have missed that this was specifically requested

Indeed.

source/specifications/distribution-formats.rst
```diff
@@ -0,0 +1,21 @@
...
+====================
+
+
+Source Distribution Format
+==========================
+
+The source distribution format (``sdist``) is not currently formally defined.
+Instead, its format is implicitly defined by the behaviour of the
+standard library's ``distutils`` module when executing the ``setup.py sdist``
+command.
```
ncoghlan: Given PEP 517, this section isn't quite accurate any more, but fixing that is out of scope for this PR. I've filed https://github.com/pypa/python-packaging-user-guide/issues/388 to keep track of it.

source/specifications/core-metadata.rst
```diff
@@ -0,0 +1,538 @@
...
+
+============================
+Core Metadata Specifications
+============================
+
+The current core metadata file format, version 1.2, is specified in :pep:`345`.
+
+However, the version specifiers and environment markers sections of that PEP
+have been superceded as described below. In addition, metadata files are
+permitted to contain the following additional fields:
```
ncoghlan: Given the change to include all the existing metadata fields as well, this intro paragraph need to be updated accordingly.

I'd also include an explicit note saying that:

1. All fields added to later versions are considered optional in earlier versions (so metadata 1.3 fields can be listed even when the nominal metadata version is only 1.2)
2. Metadata version 1.3 will be defined in a forthcoming PEP that also makes this URL the canonical source for the latest metadata description, so the fields listed as being added in 1.3 aren't typos, they're already-defined fields that will be included without further changes in the next version of the metadata specification

source/specifications/core-metadata.rst
```diff
@@ -0,0 +1,538 @@
...
+.. contents:: Contents
+   :local:
+
+
+Metadata-Version
+================
+
+.. versionadded:: 1.0
+
+Version of the file format; "1.2" is the only legal value.
```
ncoghlan: Now that this is a multi-version format description, this should be reworded more in line with the description in PEP 416: https://www.python.org/dev/peps/pep-0426/#metadata-version

Currently permitted field values are then "1.0", "1.1", and "1.2"

di:
@ncoghlan Thanks for your feedback! I think I've addressed everything.

PR #389, opened 2017-10-03
update link from distributing-packages.rst to using-testpypi.rst

soxofaan:
distributing-packages.rst has an outdated link to https://wiki.python.org/moin/TestPyPI which links back to https://packaging.python.org/guides/using-testpypi/

better make that a real internal link

### DIFF
--- a/source/guides/using-testpypi.rst
+++ b/source/guides/using-testpypi.rst
@@ -1,3 +1,5 @@
+.. _using-test-pypi:
+
 ==============
 Using TestPyPI
 ==============
--- a/source/tutorials/distributing-packages.rst
+++ b/source/tutorials/distributing-packages.rst
@@ -799,9 +799,8 @@ distribution file(s) to upload.
 
 .. note:: Before releasing on main PyPI repo, you might prefer training with
   the `PyPI test site <https://testpypi.python.org/pypi>`_
-  which is cleaned on a semi regular basis. See
-  `these instructions <https://wiki.python.org/moin/TestPyPI>`_ on how
-  to setup your configuration in order to use it.
+  which is cleaned on a semi regular basis. See :ref:`using-test-pypi`
+  on how to setup your configuration in order to use it.
 
 .. warning:: In other resources you may encounter references to using
   ``python setup.py register`` and ``python setup.py upload``. These methods


README.rst
```diff
@@ -49,7 +49,7 @@ To build the guide, run the following bash command in the source folder::
 
 After the process has completed you can find the HTML output in the
 ``./build/html`` directory. You can open the ``index.html`` file to view the
-guide in web broswer, but it's recommended to serve the guide using an HTTP
+guide in web browser, but it's recommended to serve the guide using an HTTP
```
ncoghlan: I fixed this directly in the README rather than reviewing it as part of the PR: https://github.com/pypa/python-packaging-user-guide/commit/9b78751dc1afd0498bdb4fefce3642a094b52b01

source/guides/using-testpypi.rst
```diff
@@ -1,3 +1,5 @@
+.. _ `Using TestPyPI`:
```
ncoghlan: We use lowercase-with-hyphens for internal cross reference labels:

    .. _using-test-pypi:

source/tutorials/distributing-packages.rst
```diff
@@ -799,9 +799,8 @@ distribution file(s) to upload.
 
 .. note:: Before releasing on main PyPI repo, you might prefer training with
   the `PyPI test site <https://testpypi.python.org/pypi>`_
-  which is cleaned on a semi regular basis. See
-  `these instructions <https://wiki.python.org/moin/TestPyPI>`_ on how
-  to setup your configuration in order to use it.
+  which is cleaned on a semi regular basis. See :ref:`Using TestPyPI`
```
ncoghlan: Given the above label change, this would become:

    :ref:`using-test-pypi`

soxofaan:
changed accordingly and rebased

soxofaan:
FYI: I based the ref name based on another ref I saw in the file:
``` :ref:`install_requires vs Requirements files` ```

ncoghlan:
It's entirely possible that the label-naming-convention is more consistently applied in the CPython docs than it currently is here :)

That said, I still prefer the CPython convention, as it reads better when you're using the form that overrides the text used for the link:

    :ref:`link text goes here <link-target-goes-here>`

PR #390, opened 2017-10-19
Add a specification of the entry_points.txt file

takluyver:
As discussed on the distutils-sig mailing list, I would like entry points to be documented as an interoperable standard. I have a particular interest in this as I've written implementations of both creating entry points (flit) and loading them (entrypoints).

This is my attempt to document the entry points mechanism as it already exists, based on reading code and the [pkg_resources API docs](http://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=pkg_resources#creating-and-parsing).

### DIFF
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -22,7 +22,7 @@ Data model
 
 Conceptually, an entry point is defined by three required properties:
 
-The **group** an entry point belongs to indicates what sort of object it
+The **group** that an entry point belongs to indicates what sort of object it
 provides. For instance, the group ``console_scripts`` is for entry points
 referring to functions which can be used as a command, while
 ``pygments.styles`` is the group for classes defining pygments styles.
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -43,7 +43,8 @@ Conceptually, an entry point is defined by three required properties:
   point names should be unique. If different distributions provide the same
   name, the consumer decides how to handle such conflicts. The name may contain
   any characters except ``=``, but it cannot start or end with any whitespace
-  character, or start with ``[``.
+  character, or start with ``[``. For new entry points, it is recommended to
+  use only letters, numbers, underscores, dashes and dots (regex ``[\w-.]+``).
 
 - The **object reference** points to a Python object. It is either in the form
   ``importable.module``, or ``importable.module:object.attr``. Each of the parts
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -14,8 +14,13 @@ example:
   installed packages. For more about this, see
   :doc:`/guides/creating-and-discovering-plugins`.
 
-Entry points were developed as part of setuptools. This document aims to
-describe the mechanism that is already a de-facto standard.
+The entry point file format was originally developed to allow packages built
+with setuptools to provide integration point metadata that would be read at
+runtime with ``pkg_resources``. It is now defined as a PyPA interoperability
+specification in order to allow build tools other than setuptools to publish
+``pkg_resources`` compatible entry point metadata, and runtime libraries other
+than ``pkg_resources`` to portably read published entry point metadata
+(potentially with different caching and conflict resolution strategies).
 
 Data model
 ==========
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -26,7 +26,9 @@ Conceptually, an entry point is defined by three required properties:
   provides. For instance, the group ``console_scripts`` is for entry points
   referring to functions which can be used as a command, while
   ``pygments.styles`` is the group for classes defining pygments styles.
-  The consumer typically defines the expected interface.
+  The consumer typically defines the expected interface. To avoid clashes,
+  consumers defining a new group should use names starting with a PyPI name
+  owned by the consumer project, followed by ``.``.
 
 - The **name** identifies this entry point within its group. The precise meaning
   of this is up to the consumer. For console scripts, the name of the entry point
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -33,13 +33,17 @@ Conceptually, an entry point is defined by three required properties:
   ``pygments.styles`` is the group for classes defining pygments styles.
   The consumer typically defines the expected interface. To avoid clashes,
   consumers defining a new group should use names starting with a PyPI name
-  owned by the consumer project, followed by ``.``.
+  owned by the consumer project, followed by ``.``. Group names must be one or
+  more groups of letters, numbers and underscores, separated by dots (regex
+  ``^\w+(\.\w+)*$``).
 
 - The **name** identifies this entry point within its group. The precise meaning
   of this is up to the consumer. For console scripts, the name of the entry point
   is the command that will be used to launch it. Within a distribution, entry
   point names should be unique. If different distributions provide the same
-  name, the consumer decides how to handle such conflicts.
+  name, the consumer decides how to handle such conflicts. The name may contain
+  any characters except ``=``, but it cannot start or end with any whitespace
+  character, or start with ``[``.
 
 - The **object reference** points to a Python object. It is either in the form
   ``importable.module``, or ``importable.module:object.attr``. Each of the parts
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -147,3 +147,7 @@ Install tools are expected to set up wrappers for both ``console_scripts`` and
 ``gui_scripts`` in the scripts directory of the install scheme. They are not
 responsible for putting this directory in the ``PATH`` environment variable
 which defines where command-line tools are found.
+
+As files are created from the names, and some filesystems are case-insensitive,
+packages should avoid using names in these groups which differ only in case.
+The behaviour of install tools when names differ only in case is undefined.
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -1,0 +1,95 @@
+==========================
+Entry points specification
+==========================
+
+*Entry points* are a mechanism for an installed distribution to advertise
+components it provides to be discovered and used by other code. For
+example:
+
+- Distributions can specify ``console_scripts`` entry points, each referring to
+  a function. When *pip* installs the distribution, it will create a
+  command-line wrapper for each entry point.
+- Applications can use entry points to load plugins; e.g. Pygments (a syntax
+  highlighting tool) can use additional lexers and styles from separately
+  installed packages. For more about this, see
+  :doc:`/guides/creating-and-discovering-plugins`.
+
+Entry points were developed as part of setuptools. This document aims to
+describe the mechanism that is already a de-facto standard.
+
+Data model
+==========
+
+Conceptually, an entry point is defined by three required properties:
+
+The **group** an entry point belongs to indicates what sort of object it
+provides. For instance, the group ``console_scripts`` is for entry points
+referring to functions which can be used as a command, while
+``pygments.styles`` is the group for classes defining pygments styles.
+The consumer typically defines the expected interface.
+
+The **name** identifies this entry point within its group. The precise meaning
+of this is up to the consumer. For console scripts, the name of the entry point
+is the command that will be used to launch it.
+
+The **object path** points to a Python object. It is either in the form
+``importable.module``, or ``importable.module:object.attr``. Each of the parts
+delimited by dots and the colon is a valid Python identifier.
+It is intended to be looked up like this::
+
+    import importlib
+    if ':' in object_path:
+        modname, attrs = object_path.split(':')
+        obj = importlib.import_module(object_path)
+        for attr in attrs.split('.'):
+            obj = getattr(obj, attr)
+    else:
+        obj = importlib.import_module(object_path)
+
+.. note::
+   Some tools call this kind of object path by itself an 'entry point', for want
+   of a better term, especially where it points to a function to launch a
+   program.
+
+There is also an optional piece property: the **extras** are a set of strings
+identifying optional features of the distribution providing the entry point.
+If these are specified, the entry point requires the dependencies of those
+'extras'. See the metadata field :ref:`metadata_provides_extra`.
+
+File format
+===========
+
+Entry points are defined in a file called ``entry_points.txt`` in the
+``*.dist-info`` directory of the distribution. This is the directory described
+in :pep:`376` for installed distributions, and in :pep:`427` for wheels.
+The file uses the UTF-8 character encoding.
+
+The file contents are in INI format, as read by Python's :mod:`configparser`
+module. However, configparser treats names as case-insensitive by default,
+whereas entry point names are case sensitive. A case-sensitive config parser
+can be made like this::
+
+    import configparser
+
+    class CaseSensitiveConfigParser(configparser.ConfigParser):
+        optionxform = staticmethod(str)
+
+The entry points file always uses ``=`` to delimit names from values (whereas
+configparser also allows using ``:``).
+
+The sections of the config file represent entry point groups, the names are
+names, and the values encode both the object path and the optional extras.
+If extras are used, they are a comma-separated list inside square brackets.
+There may be zero, one or multiple spaces between the object name and the
+opening square bracket.
+
+For example::
+  
+    [console_scripts]
+    foo = foomod:main
+    # One which depends on extras:
+    foobar = foomod:main_bar [bar,baz]
+    
+    # pytest plugins refer to a module, so there is no ':obj'
+    [pytest11]
+    nbval = nbval.plugin
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -107,3 +107,34 @@ For example::
     # pytest plugins refer to a module, so there is no ':obj'
     [pytest11]
     nbval = nbval.plugin
+
+Use for scripts
+===============
+
+Two groups of entry points have special significant in packaging:
+``console_scripts`` and ``gui_scripts``. In both groups, the name of the entry
+point should be usable as a command in a system shell after the package is
+installed. The object reference points to a function which will be called with
+no arguments when this command is run. The function may return an integer to be
+used as a process exit code, and returning ``None`` is equivalent to returning
+``0``.
+
+For instance, the entry point ``mycmd = mymod:main`` would create a command
+``mycmd`` launching a script like this::
+
+    import sys
+    from mymod import main
+    sys.exit(main())
+
+The difference between ``console_scripts`` and ``gui_scripts`` only affects
+Windows systems. ``console_scripts`` are wrapped in a console executable,
+so they are attached to a console and can use ``sys.stdin``, ``sys.stdout`` and
+``sys.stderr`` for input and output. ``gui_scripts`` are wrapped in a GUI
+executable, so they can be started without a console, but cannot use standard
+streams unless application code redirects them. Other platforms do not have the
+same distinction.
+
+Install tools are expected to set up wrappers for both ``console_scripts`` and
+``gui_scripts`` in the scripts directory of the install scheme. They are not
+responsible for putting this directory in the ``PATH`` environment variable
+which defines where command-line tools are found.
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -56,6 +56,12 @@ identifying optional features of the distribution providing the entry point.
 If these are specified, the entry point requires the dependencies of those
 'extras'. See the metadata field :ref:`metadata_provides_extra`.
 
+The precise functionality of entry points with extras is tied to setuptools'
+model of adding ``.egg`` directories to ``sys.path`` at runtime. Other
+installation mechanisms (such as pip and virtualenv) work differently. Packages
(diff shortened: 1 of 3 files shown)


source/specifications/entry-points.rst
```diff
@@ -0,0 +1,95 @@
...
+        optionxform = staticmethod(str)
+
+The entry points file always uses ``=`` to delimit names from values (whereas
+configparser also allows using ``:``).
+
+The sections of the config file represent entry point groups, the names are
+names, and the values encode both the object path and the optional extras.
+If extras are used, they are a comma-separated list inside square brackets.
+There may be zero, one or multiple spaces between the object name and the
+opening square bracket.
```
pfmoore: Question: Are spaces allowed within the extras list? I.e., is `[   a,  b , c]` allowed? And is it equivalent to `[a,b,c]`? If I were writing a reader, I'd probably return a list of extras, so the spec needs to be clear on this.

If we want to be more formal as a specification (I'm not sure to what extent that would be useful, rather than over-pedantic) we should probably write out a proper syntax for the values. Things that are vague in the above:

1. How should content after the closing `]` of the extras be interpreted?
2. Are spaces allowed in the object path part?
3. What about control characters (tab, or other non-printable characters)
4. Non-ASCII data? Elements of the object path follow the rules for Python names, so they are fine. But what are the rules for extras? What about things like the Unicode non-breaking space character?

In practice, I'd go with a pedantic view that extras should follow the Python name syntax for consistency, and whitespace (defined as anything the `.strip()` string method will remove) is allowed between elements of the extras, but not in the object path. No trailing text (except whitespace) is allowed after the value.

So we have `name(.name)*(:name(.name)*)?([name(,name)*])`. `name` follows the rules for Python identifiers, whitespace is only allowed around `[`, `]` and `,` in the extras section, and no trailing data (including comments as defined by configparser) are allowed.

However, I came up with this purely in the abstract. As this is an existing format, the rules setuptools defines take precedence (if they can be confirmed).

pradyunsg: I'd also just drop in to say that it might be helpful to be a bit restrictive about the whitespace, if setuptools turns out to be extremely lenient since this would be a good opportunity to fix that.

I'm speculating so take it with a grain of salt fwiw. :P

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,95 @@
...
+            obj = getattr(obj, attr)
+    else:
+        obj = importlib.import_module(object_path)
+
+.. note::
+   Some tools call this kind of object path by itself an 'entry point', for want
+   of a better term, especially where it points to a function to launch a
+   program.
+
+There is also an optional piece property: the **extras** are a set of strings
```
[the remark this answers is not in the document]
takluyver: Thanks, will fix.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,95 @@
...
+identifying optional features of the distribution providing the entry point.
+If these are specified, the entry point requires the dependencies of those
+'extras'. See the metadata field :ref:`metadata_provides_extra`.
```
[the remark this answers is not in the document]
takluyver: I've deliberately left this somewhat ambiguous. From the pkg_resources API docs, it looks like the extras were meant to control the thing setuptools did with `.egg` directories added to `sys.path` according to what dependencies an application specified. Presumably,  loading an entry point would configure sys.path with the dependencies for its extras.

Since the recommended tools are now pip and virtualenvs, that approach doesn't quite make sense. None of the entry points in packages I have installed actually specify extras.

Maybe we should put a note that the extras aren't widely used, or even that they're no longer recommended? I'm not clearly how true that is, though.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,95 @@
...
+whereas entry point names are case sensitive. A case-sensitive config parser
+can be made like this::
+
+    import configparser
+
+    class CaseSensitiveConfigParser(configparser.ConfigParser):
+        optionxform = staticmethod(str)
+
+The entry points file always uses ``=`` to delimit names from values (whereas
+configparser also allows using ``:``).
```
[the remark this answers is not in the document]
takluyver: Thanks.

source/specifications/entry-points.rst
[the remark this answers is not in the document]
pfmoore: Maybe simply say that extras are a setuptools extension? Standard writers shouldn't write them, and readers should ignore them.

source/specifications/entry-points.rst
pfmoore: Again, rather than try too hard to understand what setuptools does, I'd be inclined to just specify that writers must be strict and readers must be lenient (using proper "specification-y" terms like "writers must not put whitespace between..." and "readers must accept and ignore whitespace..."). That way again, we allow any variations setuptools introduces to simply be setuptools extensions to the standard.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,106 @@
...
+
+Entry points were developed as part of setuptools. This document aims to
+describe the mechanism that is already a de-facto standard.
+
+Data model
+==========
+
+Conceptually, an entry point is defined by three required properties:
+
+The **group** an entry point belongs to indicates what sort of object it
```
pradyunsg: 📝 Maybe "group _that_ an entry point belongs to"?

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,106 @@
...
+delimiting them, and after the right square bracket.
+For tools writing the file, it is recommended only to insert a space between the
+object path and the left square bracket.
+
+For example::
+  
+    [console_scripts]
+    foo = foomod:main
+    # One which depends on extras:
+    foobar = foomod:main_bar [bar,baz]
```
pradyunsg: :art: Maybe `foobar = foomod:main_bar[bar,baz]`?

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,106 @@
...
+   Some tools call this kind of object path by itself an 'entry point', for want
+   of a better term, especially where it points to a function to launch a
+   program.
+
+There is also an optional property: the **extras** are a set of strings
+identifying optional features of the distribution providing the entry point.
+If these are specified, the entry point requires the dependencies of those
+'extras'. See the metadata field :ref:`metadata_provides_extra`.
+
+The precise functionality of entry points with extras is tied to setuptools'
```
pradyunsg: > I'd be inclined to just specify that writers must be strict and readers must be lenient (using proper "specification-y" terms like "writers must not put whitespace between..." and "readers must accept and ignore whitespace...").

+1 to what @pfmoore said.

(made a comment since this thread would only be marked outdated by GitHub if this paragraph is changed)

source/specifications/entry-points.rst
takluyver: Thanks, done.

source/specifications/entry-points.rst
takluyver: I added a note about the whitespace, in the file format section. Does that cover what you mean?

I'm reluctant to try to say much about what readers and writers should do with extras, because I just don't know. It looks to me like they only really make sense in a setuptools-managed egg world, but maybe other toolchains could have a way to use them. I think that the best I can do is describe what they *mean*, and warn that the behaviour is undefined (i.e. this paragraph).

source/specifications/entry-points.rst
takluyver: Examples I've seen leave a space there:

- http://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins
- http://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=pkg_resources#creating-and-parsing

And I marginally prefer it with a space for aesthetics. I'm happy to change it if most people prefer no space, though.

source/specifications/entry-points.rst
pfmoore: LGTM.

Agreed. The more I think about it, the more I'd say that from a "standard format" perspective, we can stick with treating extras as a setuptools-only proprietary extension that we acknowledge the existence of but that new tools should simply ignore.

pradyunsg: > Does that cover what you mean?

Indeed. :)

I'd missed that.

source/specifications/entry-points.rst
pradyunsg: No issues either way tbh.

dstufft:
As I've said on distutils-sig, I am -1 on this pull request unless it is specific to defining what we need for ``console_scripts`` and no more.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,106 @@
...
+Data model
+==========
+
+Conceptually, an entry point is defined by three required properties:
+
+The **group** an entry point belongs to indicates what sort of object it
+provides. For instance, the group ``console_scripts`` is for entry points
+referring to functions which can be used as a command, while
+``pygments.styles`` is the group for classes defining pygments styles.
+The consumer typically defines the expected interface.
```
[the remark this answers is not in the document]
takluyver: Done.

takluyver:
@ncoghlan 'group' may not be the clearest term, but it is already the term setuptools docs and API use for a group/category/interface of entry points, so I think it's worth using the same name here.

I couldn't see any clear existing name for what I chose to call an 'object path', so I've renamed it to 'object reference' as you suggested.

ncoghlan:
@takluyver +1 for maintaining consistency with established terminology where it exists.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,140 @@
...
+- Distributions can specify ``console_scripts`` entry points, each referring to
+  a function. When *pip* (or another console_scripts aware installer) installs
+  the distribution, it will create a command-line wrapper for each entry point.
+- Applications can use entry points to load plugins; e.g. Pygments (a syntax
+  highlighting tool) can use additional lexers and styles from separately
+  installed packages. For more about this, see
+  :doc:`/guides/creating-and-discovering-plugins`.
+
+Entry points were developed as part of setuptools. This document aims to
+describe the mechanism that is already a de-facto standard.
```
ncoghlan: Based on the distutils-sig discussion, I'd rephrase this paragraph:

> The entry point file format was originally developed to allow packages built with setuptools to provide integration point metadata that would be read at runtime with pkg_resources. It is now defined as a PyPA interoperability specification in order to allow build tools other than setuptools to publish pgk_resource compatible entry point metadata, and runtime libraries other than pkg_resources to portably read published entry point metadata (potentially with different caching and conflict resolution strategies).

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,140 @@
...
+  ``pygments.styles`` is the group for classes defining pygments styles.
+  The consumer typically defines the expected interface. To avoid clashes,
+  consumers defining a new group should use names starting with a PyPI name
+  owned by the consumer project, followed by ``.``.
+
+- The **name** identifies this entry point within its group. The precise meaning
+  of this is up to the consumer. For console scripts, the name of the entry point
+  is the command that will be used to launch it. Within a distribution, entry
+  point names should be unique. If different distributions provide the same
+  name, the consumer decides how to handle that.
```
ncoghlan: I'd suggest ending this with "... how to handle such conflicts." rather than "... how to handle that."

takluyver:
Thanks, I applied both of your suggestions.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+    if qualname_separator:
+        for attr in qualname.split('.'):
+            obj = getattr(obj, attr)
+
+.. note::
+   Some tools call this kind of object reference by itself an 'entry point', for
+   want of a better term, especially where it points to a function to launch a
+   program.
+
+There is also an optional property: the **extras** are a set of strings
```
dstufft: We should probably be stronger here and suggest that using extras as part of an entrypoint should not be used at all, but that parsers should be able to parse it due to historical reasons.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+declaring entry points should not assume that extras have any particular effect
+on the environment of the code using the entry points.
+
+File format
+===========
+
+Entry points are defined in a file called ``entry_points.txt`` in the
+``*.dist-info`` directory of the distribution. This is the directory described
+in :pep:`376` for installed distributions, and in :pep:`427` for wheels.
+The file uses the UTF-8 character encoding.
```
dstufft: I forget offhand what setuptools does here, but we should ensure that it's doing this.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+``pkg_resources`` compatible entry point metadata, and runtime libraries other
+than ``pkg_resources`` to portably read published entry point metadata
+(potentially with different caching and conflict resolution strategies).
+
+Data model
+==========
+
+Conceptually, an entry point is defined by three required properties:
+
+- The **group** that an entry point belongs to indicates what sort of object it
```
dstufft: I believe setuptools has restrictions on what a valid group name is, I'm not sure offhand. If so it should be encoded here.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+  provides. For instance, the group ``console_scripts`` is for entry points
+  referring to functions which can be used as a command, while
+  ``pygments.styles`` is the group for classes defining pygments styles.
+  The consumer typically defines the expected interface. To avoid clashes,
+  consumers defining a new group should use names starting with a PyPI name
+  owned by the consumer project, followed by ``.``.
+
+- The **name** identifies this entry point within its group. The precise meaning
```
dstufft: I believe setuptools has restrictions on what a valid name is, I'm not sure offhand. If should be encoded here. At the very least I know ``console_scripts`` does not allow anything with a ``/`` in the name.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+    
+    # pytest plugins refer to a module, so there is no ':obj'
+    [pytest11]
+    nbval = nbval.plugin
+
+Use for scripts
+===============
+
+Two groups of entry points have special significant in packaging:
+``console_scripts`` and ``gui_scripts``. In both groups, the name of the entry
```
dstufft: Off my memory, I believe that this is more complex then this. I know there are restrictions about having ``/`` in the name, and something makes me feel like we actually treat ``console_scripts`` as case insensitive. There is likely more.

source/specifications/entry-points.rst
takluyver: I tried to track it down through `pkg_resources`, and it looks like that is at least the intention.

https://github.com/pypa/setuptools/blob/403bfce4ab920823cc4ba0b5ca5ac0d1b213513d/pkg_resources/__init__.py#L1494

source/specifications/entry-points.rst
pfmoore: Because console scripts translate to filenames, they are case insensitive because (some of) the target filesystems are.

It obviously depends on the group what's allowed (and we can't do anything about that). However, the spec probably does need to say something about who enforces restrictions - is it down to the producer to reject invalid entry point names? Or to the consumer to deal with them? The former is basically impractical, but the consumer may well not be able to deal gracefully with the situation. We have this with pip - packages can specify two console scripts differing only in case, and pip then ends up having to write a broken installation on a case-insensitive system like Windows.

I don't think the spec can do anything other than say "any other restrictions are the responsibility of the publisher and subscriber to manage". That sucks, but it's in the nature of de facto standards that you have to live with such suckiness.

source/specifications/entry-points.rst
takluyver: Good call, it does:

https://github.com/pypa/setuptools/blob/403bfce4ab920823cc4ba0b5ca5ac0d1b213513d/pkg_resources/__init__.py#L2371

source/specifications/entry-points.rst
takluyver: Good call, it does:

https://github.com/pypa/setuptools/blob/403bfce4ab920823cc4ba0b5ca5ac0d1b213513d/pkg_resources/__init__.py#L2371

source/specifications/entry-points.rst
takluyver: The regex in `pkg_resources` appears to allow any character, though it cannot contain `=`, and can't end with a space. It would probably be appropriate to recommend a more restrictive character set for new entry names.

https://github.com/pypa/setuptools/blob/403bfce4ab920823cc4ba0b5ca5ac0d1b213513d/pkg_resources/__init__.py#L2449

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,145 @@
...
+  ``pygments.styles`` is the group for classes defining pygments styles.
+  The consumer typically defines the expected interface. To avoid clashes,
+  consumers defining a new group should use names starting with a PyPI name
+  owned by the consumer project, followed by ``.``.
+
+- The **name** identifies this entry point within its group. The precise meaning
+  of this is up to the consumer. For console scripts, the name of the entry point
+  is the command that will be used to launch it. Within a distribution, entry
+  point names should be unique. If different distributions provide the same
+  name, the consumer decides how to handle such conflicts.
```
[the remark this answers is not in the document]
takluyver: This can vary based on use case; if you're looking for plugins, you can ignore the names entirely and load all entry points, even if the names conflict. If you're looking for something with a specific name, there will be situations where it makes sense to error out on a conflict, to warn and pick one, or to take the first and stop looking.

`pkg_resources` leaves it up to the application what to do with conflicting entry points (its API is an iterator that yields all matching entry points). I don't think it makes sense to try to specify this.

dstufft:
Only one of the API presents that, the rest of them force a particular conflict resolution.

source/specifications/entry-points.rst
takluyver: I can't see any code in pip to block `/` in script names - does setuptools do something special?

I'm adding a note about case sensitivity.

takluyver:
Oh, my mistake, I missed that there are other APIs to load entry points. However, as far as I can see in the docs, the others only work on one distribution at a time. In any case, applications do get to choose how conflicts are handled.

source/specifications/entry-points.rst
```diff
@@ -147,3 +147,7 @@ Install tools are expected to set up wrappers for both ``console_scripts`` and
 ``gui_scripts`` in the scripts directory of the install scheme. They are not
 responsible for putting this directory in the ``PATH`` environment variable
 which defines where command-line tools are found.
+
+As files are created from the names, and some filesystems are case-insensitive,
+packages should avoid using names in these groups which differ only in case.
+The behaviour of install tools when names differ only in case is undefined.
```
pfmoore: IMO, this *is* starting to veer away from specifying the entry point file format into defining the specifics of how individual publishers and subscribers use them.

I think that the "entry points file format" spec should say nothing other than "the `console_scripts` and `gui_scripts` groups are reserved for use by packaging systems". We could have another spec that defines how wrapper scripts are defined using those entry points, but if we put that information in this spec, we do start to hit the issues @dstufft is concerned about, of conflating specifications that pip/setuptools/wheel/distlib care about with those that they don't.

takluyver: This final section is intended to document the *use* of entry points in packaging systems, after the earlier formats have described the file format and the general semantics. This is kind of combining two specifications into one, but they are clearly separated by the headings.

I can take this section out or move it to a separate document if people want. But it would be a pretty short document, so I'm saying 'practicality beats purity' unless there's a consensus to do that.

source/specifications/entry-points.rst
```diff
@@ -0,0 +1,154 @@
...
+    foobar = foomod:main_bar [bar,baz]
+    
+    # pytest plugins refer to a module, so there is no ':obj'
+    [pytest11]
+    nbval = nbval.plugin
+
+Use for scripts
+===============
+
+Two groups of entry points have special significant in packaging:
```
pfmoore: "significance"

jonparrott:
I'm okay with us documenting this, but considering this is under "specifications" it seems odd that this document does not reference a pep or a reference implementation - are we just documenting setuptools internals? If the answer is "yes" that's fine, I just want us to be clear about how "specific" the "specifications" are.

takluyver:
It's documenting an existing mechanism. I wouldn't call it setuptools 'internals' any more: multiple tools already have implementations to read and write it, so I consider it a de-facto interoperability standard. Donald disagrees with me.

On top of documenting the mechanism as it already exists, this makes a few recommendations about how tools should use it - e.g. namespacing entry point group names to avoid clashes.

ncoghlan:
The reference publishing tool is `setuptools` (as with PKG-INFO), the reference subscriber is `pkg_resources`.

source/specifications/entry-points.rst
ncoghlan: +1 for this approach- new publishers should feel free to not support this feature, and new consumers may want to disallow it as well (but will need to handle it if they want to support reading arbitrary existing entry point groups)

ncoghlan:
@jaraco's approval for elevation to an interoperability specification: https://github.com/pypa/setuptools/issues/1179#issuecomment-339330959

ncoghlan:
Thanks @takluyver - with @jaraco's +1, I've gone ahead and published this version.

If we find any further clarifications or corrections are needed, they can go in a new PR (as with any of the specifications).

takluyver:
Thanks Nick! :-)

PR #402, opened 2017-11-29
Introduce a new dependency management tutorial

ncoghlan:
- this promotes the draft pipenv tutorial to the main
  guide as a *new* dependency management tutorial
- several of the bootstrapping improvements have been
  incorporated into the initial package installation
  tutorial (that still needs more work, but it isn't
  the main focus of this update)
- the main index page has been rewrapped in editing
- the "new-tutorials" section has been removed
  completely for the time being (as I'm hoping that
  normal PR reviews will be sufficient for any
  further tutorial updates)


kennethreitz:
✨🍰✨

kennethreitz:
When can we expect this to land?

jonparrott:
> When can we expect this to land?

Right meow.

ncoghlan:
@pfmoore I tend to work on the theory that "me right now" and "me 6 months from now" aren't exactly the same person (since we won't remember exactly the same things), so pre-emptively collaborating with future me is still collaborating :)

pfmoore:
:-) 100% agreed with that. With my memory, "me 5 days from now" might as well be a different person. But nevertheless, to be serious for a moment, I doubt that's what people will think of when they see the word "collaborative". As I say, though, let's see how it works in practice before worrying.

ncoghlan:
After thinking about it for a while, I realised the key assumption that `Pipenv` makes that `pip` doesn't is that you're going to be checking `Pipfile` and `Pipfile.lock` into version control. That means the key differentiator isn't really collaboration, it's whether or not you're using version control. I'd just conflated the two, since you can often survive without version control on personal projects, while on collaborative projects it's pretty much essential.

I've tweaked the new bullet point on the main page accordingly: https://github.com/pypa/python-packaging-user-guide/commit/8564f7feee1aa46d81c06989b8643cb3381075cd

I haven't reviewed the tutorial intro on that basis, though.

PR #404, opened 2017-12-01
Add News page

### LINKED ISSUE #403: Add a manually updated "NEWS" page
Prompted by the discussion in https://github.com/python/psf-infra-meta/issues/1, I'm thinking it could be useful to have a manually updated page at "https://packaging.python.org/news" where we mention significant updates like the addition of the new "pipenv" tutorial, or updates to the pypi.org migration instructions.

jonparrott:
Resolves #403.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -358,7 +358,35 @@ intersphinx_mapping = {
 }
 
 
-# f this is True, todo and todolist produce output, else they produce nothing.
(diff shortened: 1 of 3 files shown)


ncoghlan:
Nice! I'll wait until https://github.com/python/psf-infra-meta/issues/1 progresses further before posting to distutils-sig about it (since that will help us define what we consider in scope for the PyPUG NEWS page)

PR #412, opened 2017-12-08
Metadata 2.1 updates

### LINKED ISSUE #410: Metadata spec: what's a valid name?
The requires-dist section says:

> The format of a requirement string is identical to that of a distutils project name (e.g., as found in the Name: field...

But the Name field provides no guidance on what is valid.

It looks like the correct answer is [in PEP 508](https://www.python.org/dev/peps/pep-0508/#names): names have to match this (case-insensitive) regex: `^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$` . This should be in the metadata spec document, so people don't have to go hunting through PEPs for it.

### LINKED ISSUE #411: Metadata spec: version spec without parentheses?
The requires-dist section again:

>  optionally followed by a version declaration within parentheses.

[PEP 508](https://www.python.org/dev/peps/pep-0508/#versions) appears to drop the need for the parentheses:

> The optional brackets around a version are present for compatibility with PEP-345 [2] but should not be generated, only accepted.

Does this supersede what's in the spec?

di:
This PR captures all changes we need to make now that [PEP 566](https://www.python.org/dev/peps/pep-0566/) is accepted.

Fixes #410, fixes #411, closes #449.

(cc @takluyver)

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -12,6 +12,13 @@ Fields defined in the following specification should be considered valid,
 complete and not subject to change. Fields should be considered "optional" for
 versions which predate their introduction.
 
+.. note:: *Interpreting old metadata:* In :pep:`566`, the version specifier
+   field format specification was relaxed to accept the syntax used by popular
+   publishing tools (namely to remove the requirement that version specifiers
+   must be surrounded by parentheses). Metadata consumers may want to use the
+   more relaxed formatting rules even for metadata files that are nominally
+   less than version 2.1.
+
 .. contents:: Contents
    :local:
 
@@ -118,6 +125,8 @@ Description (optional)
 ======================
 
 .. versionadded:: 1.0
+.. versionchanged:: 2.1
+   This field may be specified in the message body instead.
 
 A longer description of the distribution that can run to several
 paragraphs.  Software that deals with metadata should not assume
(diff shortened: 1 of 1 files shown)


source/specifications/core-metadata.rst
```diff
@@ -38,15 +35,21 @@ all of the needed fields.
...
 .. versionadded:: 1.0
+.. versionchanged:: 1.3
+   Added additional restrictions on format from :pep:`508`
+
+The name of the distribution. The name field is the primary identifier for a
+distribution. Distribution names are limited to those which match the following
+regex (run with ``re.IGNORECASE``)::
 
-The name of the distributions.
+    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$
```
takluyver: Is it worth adding a prose description of this as well? Here's some words we could use:

> I.e. A valid name consists only of ASCII letters and numbers, period, underscore and hyphen. It must start and end with a letter or number.

ncoghlan:
PEP 566 has been accepted: https://github.com/python/peps/commit/c83974875dcf14f8ef798a5893438a31b7f6cf4e

However, the PR requires a few updates to match the accepted version of the PEP:

- metadata version is now `2.1`
- needs versionadded notes for Description-Content-Type and Provides-Extra
- Description can now be the metadata body (and needs a versionchanged note for that)
- Requires-Dist et al should have a versionchanged note for 2.1 that says something like: "The field format specification was relaxed to accept the syntax used by popular publishing tools. Metadata consumers may want to use the more relaxed formatting rules even for metadata files that are nominally version 1.2" (Alternatively, it may make sense to have a separate section on interpreting older metadata, and reduce the note on each field to just the first sentence about relaxing the rules to match common practices).

di:
@ncoghlan I think I've addressed all your points (thank you for the checklist, BTW). Could you review/merge?

ncoghlan:
Huzzah! I'll go mark the PEP as Final as well :)

PR #415, opened 2017-12-11
Capitalize "Python" when describing the language.

gtback:
Per the [style
guide](https://devguide.python.org/documenting/#capitalization).
Instances involving URLs, commands, etc. were not modified.


jonparrott:
Thanks, @gtback!

Issue #409, opened 2017-12-08
Metadata spec does not mention environment markers


takluyver:
See [PEP 345](https://www.python.org/dev/peps/pep-0345/#environment-markers), which defines environment markers applying to several fields - most importantly, `Requires-Dist`.

The [core metadata specification](https://packaging.python.org/specifications/core-metadata/#requires-python) page does not say anything about these at present.

di:
I think that's what the [Dependency Specifiers](https://packaging.python.org/specifications/dependency-specifiers/) page is supposed to be. Maybe this should be renamed to "Environment Markers"?

takluyver:
Environment markers are part of that, but not all of it.

I think the core metadata page at least needs to make it clear which fields have environment markers, and where the canonical description of them is.

ncoghlan:
+1 (and enabling that kind of clarification without requiring a PEP update was a key part of the motivation for moving the specs into PyPUG)

PR #416, opened 2017-12-12
Note which fields can be used with environment markers

takluyver:
[PEP 345](https://www.python.org/dev/peps/pep-0345/#id34) lists six fields which can have environment markers. I've noted it for all six, but I don't know if I should. In particular, I've never seen a classifier listed with an environment marker, and it doesn't make much sense to me to allow that.

I've also updated the specification of `Requires-Dist` with a brief summary of PEP 508.

Closes gh-409

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -357,6 +357,8 @@ Classifier (multiple use)
 Each entry is a string giving a single classification value
 for the distribution.  Classifiers are described in PEP 301 [3]_.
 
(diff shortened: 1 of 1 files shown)


jonparrott:
Thanks @takluyver.

Re:

> In particular, I've never seen a classifier listed with an environment marker, and it doesn't make much sense to me to allow that.

Perhaps @ncoghlan or @dstufft has some insight here.

I'm going to go ahead and merge- we can address post-hoc comments in a follow-up PR.

PR #417, opened 2017-12-12
Add note about needing to rebuild changes

di:
Based on discussions with @lgh2. It may not be immediately clear to new users that the distribution command must be re-run if there are changes.

@lgh2, would you mind giving a review?


jonparrott:
This LGTM, but I'll let @lgh2 take a look before I merge.

lgh2:
It looks good, thank you for making the change!

PR #418, opened 2017-12-13
 Replace `~` with `$HOME` in guides, tutorials 

brainwane:
Be more explicit to help users who aren't as familiar with the tilde. Inspired by an experience @lgh2 had, figuring out where `.pypirc` files live:

`<lghampton>` I think it was the `~/.pypirc` syntax that confused me, it wasn't clear to me that it referred to `$HOME` or something relative to the project path

I left out instances where the tilde was explicitly explained in the paragraph where it was used (`source/tutorials/managing-dependencies.rst`, and where the tilde appeared as part of sample terminal output.

I might be wrong in proposing this PR. I'm open to being told, for instance, that we need to use `~` for platform compatibility reasons, or for uniformity with other documentation. In that case I'd like for us to explicitly mention what it means, contextually, as in `source/tutorials/managing-dependencies.rst`:

> this will typically print ``~/.local`` (with ``~`` expanded to the absolute path to your home directory)

cc @jonparrott

### DIFF
--- a/source/guides/installing-using-pip-and-virtualenv.rst
+++ b/source/guides/installing-using-pip-and-virtualenv.rst
@@ -55,7 +55,7 @@ Afterwards, you should have the newest pip installed in your user site:
 .. code-block:: bash
 
     python3 -m pip --version
-    pip 9.0.1 from ~/.local/lib/python3.6/site-packages (python 3.6)
+    pip 9.0.1 from $HOME/.local/lib/python3.6/site-packages (python 3.6)
 
 .. _python-pip: https://packages.debian.org/stable/python-pip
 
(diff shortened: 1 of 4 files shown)


jonparrott:
Thanks, @brainwane!

source/guides/installing-using-pip-and-virtualenv.rst
```diff
@@ -55,7 +55,7 @@ Afterwards, you should have the newest pip installed in your user site:
 .. code-block:: bash
 
     python3 -m pip --version
-    pip 9.0.1 from ~/.local/lib/python3.6/site-packages (python 3.6)
+    pip 9.0.1 from $HOME/.local/lib/python3.6/site-packages (python 3.6)
```
pradyunsg: Isn't this output?

jonparrott: Yeah but in both cases it's representative as pip actually outputs the whole path, e.g.:


```
pip 9.0.1 from /Users/jonwayne/.pyenv/versions/2.7.13/lib/python2.7/site-packages (python 2.7)
```

PR #421, opened 2018-01-01
Fix code block inside note.

danielroseman:
The double colon was causing the `..code-block` line to be interpreted
as part of the code itself and therefore displaying in the HTML.


ncoghlan:
Thanks!

PR #429, opened 2018-02-08
Core Metadata: Add example of multiple emails

di:
This is often asked for, has always been permitted by the spec, and is now accepted as valid by Warehouse (https://github.com/pypa/warehouse/pull/2904), so let's include an example here.


jonparrott:
Thanks, @di!

PR #433, opened 2018-02-17
Provide pointer to Google BigQuery download statistics

### LINKED ISSUE #428: Provide pointer to Google Big Query download statistics
(I thought there was already an open issue for this, but I can't find it, so filing this one)

We don't really have the Google Big Query download stats documented anywhere. https://mail.python.org/pipermail/distutils-sig/2016-May/028986.html is the original announcement, and https://langui.sh/2016/12/09/data-driven-decisions/ is a good blog post about them, but neither is particularly discoverable for package maintainers.

tswast:
This guide points to the PyPI download statistics dataset and describes
how to query it.

Closes #428.

### DIFF
--- a/source/guides/analyzing-pypi-package-downloads.rst
+++ b/source/guides/analyzing-pypi-package-downloads.rst
@@ -1,0 +1,224 @@
+================================
+Analyzing PyPI package downloads
+================================
+
+This section covers how to use the `PyPI package dataset`_ to learn more
+about downloads of a package (or packages) hosted on PyPI. For example, you can
+use it to discover the distribution of Python versions used to download a
+package.
+
+.. contents:: Contents
+   :local:
+
+
+Background
+==========
+
+PyPI does not display download statistics because they are difficult to
+collect and display accurately. Reasons for this are included in the
+`announcement email
+<https://mail.python.org/pipermail/distutils-sig/2013-May/020855.html>`__:
+
+    There are numerous reasons for [download counts] removal/deprecation some
+    of which are:
+
+        - Technically hard to make work with the new CDN
+            - The CDN is being donated to the PSF, and the donated tier does
+              not offer any form of log access
+            - The work around for not having log access would greatly reduce
+              the utility of the CDN
+        - Highly inaccurate
+            - A number of things prevent the download counts from being
+              inaccurate, some of which include:
+                - pip download cache
+                - Internal or unofficial mirrors
+                - Packages not hosted on PyPI (for comparisons sake)
+                - Mirrors or unofficial grab scripts causing inflated counts
+                  (Last I looked 25% of the downloads were from a known
+                  mirroring script).
+        - Not particularly useful
+            - Just because a project has been downloaded a lot doesn't mean
+              it's good
+            - Similarly just because a project hasn't been downloaded a lot
+              doesn't mean it's bad
+
+    In short because it's value is low for various reasons, and the tradeoffs
+    required to make it work are high It has been not an effective use of
+    resources.
+
+As an alternative, the `Linehaul project
+<https://github.com/pypa/linehaul>`__ streams download logs to `Google
+BigQuery`_ [#]_. Linehaul writes an entry in a
+``the-psf.pypi.downloadsYYYYMMDD`` table for each download. The table
+contains information about what file was downloaded and how it was
+downloaded. Some useful columns from the `table schema
+<https://bigquery.cloud.google.com/table/the-psf:pypi.downloads20161022?tab=schema>`__
+include:
+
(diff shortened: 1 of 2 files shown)


source/guides/analyzing-pypi-package-downloads.rst
```diff
@@ -0,0 +1,224 @@
...
+
+The following query counts the total number of downloads for the project
+"pytest".
+
+::
+
+    #standardSQL
+    SELECT COUNT(*) AS num_downloads
+    FROM `the-psf.pypi.downloads*`
+    WHERE file.project = 'pytest'
```
ncoghlan: Due to the 1 TB monthly data processing limit and the size of the PyPI data set, I think it would be good to use the "last 30 completed days" filter on all the example queries to make it easier for folks to copy and paste them.

That looks like this:

    FROM
        TABLE_DATE_RANGE( 
            [the-psf:pypi.downloads],
            DATE_ADD(CURRENT_TIMESTAMP(), -31, "day"),
            DATE_ADD(CURRENT_TIMESTAMP(), -1, "day") 
        )

source/guides/analyzing-pypi-package-downloads.rst
```diff
@@ -0,0 +1,224 @@
...
++---------------+
+| num_downloads |
++===============+
+| 31768554      |
++---------------+
+
+Package downloads over time
+---------------------------
+
+To group by monthly downloads, use the ``_TABLE_SUFFIX`` pseudo-column. Also
```
ncoghlan: I think the combination of `TABLE_DATE_RANGE` plus grouping by a suitably formatted `STRFTIME` column is a potentially clearer option here.

source/guides/analyzing-pypi-package-downloads.rst
```diff
@@ -0,0 +1,224 @@
...
+Useful queries
+==============
+
+Run queries in the `BigQuery web UI`_ by clicking the "Compose query" button.
+
+Note that the rows are stored in separate tables for each day, which helps
+limit costs if you are only interested in recent downloads. To analyze the
+full history, use `wildcard tables
+<https://cloud.google.com/bigquery/docs/querying-wildcard-tables>`__ to
+select all tables.
```
ncoghlan: If the examples are switched to use `TABLE_DATE_RANGE` by default, then this paragraph could be updated to explain that, and to point out that if you want to analyse *all* the available data, then you can use a wildcard query instead (but doing so will churn through a *lot* data quota as you iterate on other query details)

source/guides/analyzing-pypi-package-downloads.rst
```diff
@@ -0,0 +1,224 @@
...
+
+More queries
+------------
+
+- `Data driven decisions using PyPI download statistics
+  <https://langui.sh/2016/12/09/data-driven-decisions/>`__
+- `PyPI queries gist <https://gist.github.com/alex/4f100a9592b05e9b4d63>`__
+- `Python versions over time
+  <https://github.com/tswast/code-snippets/blob/master/2018/python-community-insights/Python%20Community%20Insights.ipynb>`__
+
```
ncoghlan: https://bigquery.cloud.google.com/savedquery/51422494423:ff1976af63614ad4a1258d8821dd7785 is an example of breaking down non-Windows downloads by platform

source/guides/analyzing-pypi-package-downloads.rst
tswast: Good idea. You're right that querying all of history would use up quota fast.

Done (sort-of).

source/guides/analyzing-pypi-package-downloads.rst
tswast: I use the table suffix pseudo-column instead of `TABLE_DATE_RANGE` because `TABLE_DATE_RANGE` is for legacy SQL not standard SQL. I and the rest of the BigQuery team would prefer that new examples use standard SQL, as that is where all the development of new features is going. (I do realize that the UI still defaults to legacy SQL, though.)

source/guides/analyzing-pypi-package-downloads.rst
tswast: Added a link to it, thanks!

source/guides/analyzing-pypi-package-downloads.rst
ncoghlan: Oh, nice - I'd missed that Big Query had gained standard SQL support. Given that some of the linked blog posts use the original big-query-specific syntax, it's probably worth calling this out explicitly, but I don't think that should block merging an initial version of the guide.

ofek:
@tswast Thanks for recommending `pypinfo`!

PR #435, opened 2018-02-19
Tidy up some build warnings

ncoghlan:


source/tutorials/installing-packages.rst
```diff
@@ -63,7 +63,7 @@ please install the latest 3.x version from `python.org`_ or refer to the
     can run system commands like those in this tutorial by prefacing them with
     a ``!`` character:
 
-    .. code-block:: python
+    ::
```
pradyunsg: This would need ipython's lexer to work. If we don't want to pick up this build-dependency, I think the line before should be changed to have :: at the end.

ncoghlan: I'm fine with adding a dependency on IPython's lexer, I just didn't want to do it as part of this PR (mainly because it wasn't obvious to me from a Google search where the current docs for that are: the top Google hit generates a warning for "You're looking at an old version of IPython's docs", and clicking the link in the warning takes you to the docs home page for the newer version).

Issue #401, opened 2017-11-26
Consolidate pypi.org migration notes


ncoghlan:
We currently have two sets of pypi.org migration notes:

* https://packaging.python.org/guides/tool-recommendations/#publishing-platform-migration
* https://packaging.python.org/guides/migrating-to-pypi-org/

Aside from duplicating information, the first set also kind of overwhelms that tool recommendation lists that are intended to be the main purpose of that particular guide.

It would be preferable to consolidate all the actual migration notes into the migration guide, and update the section on the tool recommendations page to just link to the migration guide.

ncoghlan:
I'm going to handle this now, as we need to add info about user registration being turned off in legacy PyPI as a result of https://status.python.org/incidents/mgjw1g5yjy5j

ncoghlan:
#401 handles the consolidation, and also adds the note about user registration migrating.

brainwane:
I missed the chance to review #436 before it was merged. Thanks for it, @ncoghlan.

As of earlier this week [we do have package and release management features in Warehouse](https://github.com/pypa/warehouse/milestone/8?closed=1) so I'll update the ["Managing published packages and releases"](https://github.com/pypa/python-packaging-user-guide/blob/master/source/guides/migrating-to-pypi-org.rst#managing-published-packages-and-releases) section with a new PR shortly.

PR #436, opened 2018-02-19
Issue #401: Consolidate PyPI migration info

ncoghlan:
- continue to describe the migration on the recommendations
  page, but link to the migration page for details
- merge the info from the recommendations page into the
  migration page
- add a new section about user registration migrating due
  to a spam attack against the legacy service


ncoghlan:
@brainwane FYI

PR #439, opened 2018-02-21
Update PyPI migration info

brainwane:
Issue #401: update PyPI migration info to advise users on package, release management features.


brainwane:
cc @ncoghlan @jonparrott @di

PR #446, opened 2018-02-25
Add external links to extension module guide

ApproximateIdentity:
I'm opening up this PR due to feedback I've received for my posts on the python core mentorship mailing list as recommended by @ncoghlan :

`relevant quotes from the mailing list:`
```
On 23 February 2018 at 20:09, Thomas Nyberg <tomuxiong@gmx.com> wrote:

    On 02/23/2018 03:08 AM, Nick Coghlan wrote:
    > I think even an update that added links to the articles you've already
    > written would be quite helpful for folks just getting into extension
    > module development and looking to learn more about the underlying
    > concepts involved :)
    >

    I've read through this and feel I can definitely add some things. It's a
    bit difficult to find a good middle ground between this and my writings.
    Obviously my writings are bit too long-winded for a doc like this one,
    but maybe I can extract some useful kernels of info. Also unfortunately
    I've only ever worked with debian so I'm a bit less familiar with the
    build process on other OSs. (Though I am that type of crazy person
    that's always been curious about this stuff so this might be a good time
    to learn...) Of course just adding the links is also a reasonable
    approach even if it feels a bit lazy on my part.


I'm responsible for most of the "we should really write this some day" notes
 in those docs, so the bar for "at least it's an improvement over the status 
quo" isn't especially high. One of the advantages of linking out to external 
docs is that those can be pitched at different audiences - yours are good 
for folks that are wanting to *understand* what the tools are actually doing 
internally, while resources like the milksnake and cffi docs will be better for 
folks that just want to get a C library wrapped, and don't really care about 
the technical details of how it works.

Cheers,
Nick.
```

I have built this version of the documentation and so I know the links work, but I don't know if my PR or my actual changes match what people expect in this repo. Please definitely let me know if I could change things. 

Cheers,
Thomas

### DIFF
--- a/source/guides/packaging-binary-extensions.rst
+++ b/source/guides/packaging-binary-extensions.rst
@@ -295,3 +295,13 @@ For interim guidance on this topic, see the discussion in
    mention the fact that Linux distros have a requirement to build from
    source in their own build systems, so binary-only releases are strongly
    discouraged
+
+External Links
+==============
+
+For a more in depth explanation of how extension modules are used by cpython on
+a debian system, see the following articles:
+
+* `What are (c)python extension modules? <https://thomasnyberg.com/what_are_extension_modules.html>`_
+* `Releasing the gil <https://thomasnyberg.com/releasing_the_gil.html>`_
+* `Writing cpython extension modules using C++ <https://thomasnyberg.com/cpp_extension_modules.html>`_


source/guides/packaging-binary-extensions.rst
```diff
@@ -295,3 +295,13 @@ For interim guidance on this topic, see the discussion in
    mention the fact that Linux distros have a requirement to build from
    source in their own build systems, so binary-only releases are strongly
    discouraged
+
+External Links
```
ncoghlan: I'll amend the title of this section to "External Resources", but other than that I think this is a fine way to structure the additions.

ApproximateIdentity:
Great! I didn't like how what I wrote the section introduction and feel like yours is much better so I'm alright with all of this as long as you are. Thanks a lot!

PR #456, opened 2018-03-12
Clarify long description classifier on pypi.org

willingc:
Related to https://github.com/pypa/warehouse/issues/3229


jonparrott:
Thanks @willingc and @brainwane!

PR #457, opened 2018-03-18
Added long_description_content_type

9999years:
With [PEP 566](https://www.python.org/dev/peps/pep-0566/#description-content-type-optional), the `long_description` can be Markdown with the `long_description_content_type` arg. This adds documentation for it

There were also a couple cases where filenames were enclosed with quotes rather than as code; I've changed those to be `code` as well for consistency.

See also:

 * [Relevant section in PEP 566](https://www.python.org/dev/peps/pep-0566/#description-content-type-optional)
 * [The CommonMark specification](http://spec.commonmark.org)
 * [The Description-Content-Type field documentation in the Python
   packaging documentation](https://packaging.python.org/specifications/core-metadata/#description-content-type-optional)

Note that this references the `README.md` added in pypa/sampleproject#66, which is not yet merged. Merging that will break the link here, and merging this will result in a broken link until that PR is merged. Not sure what the best route there is.

### DIFF
--- a/source/tutorials/distributing-packages.rst
+++ b/source/tutorials/distributing-packages.rst
@@ -49,12 +49,12 @@ Initial Files
 setup.py
 ~~~~~~~~
 
-The most important file is "setup.py" which exists at the root of your project
+The most important file is ``setup.py`` which exists at the root of your project
 directory. For an example, see the `setup.py
 <https://github.com/pypa/sampleproject/blob/master/setup.py>`_ in the `PyPA
 sample project <https://github.com/pypa/sampleproject>`_.
 
-"setup.py" serves two primary functions:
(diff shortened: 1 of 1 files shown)


jonparrott:
@brainwane the `:file:` role is great, but the inline `code` role (via ``) is fine as well.

jonparrott:
Thanks, @9999years!

9999years:
Great! I’ll open another for the `:file:` role (I took the 10 minutes to run through the repo and change it everywhere).

brainwane:
Thanks for this, @9999years! Any chance we'll see you at [one of our sprints](https://wiki.python.org/psf/PackagingSprints) this year?

PR #458, opened 2018-03-18
Change ``filename`` to :file:`filename` throughout

9999years:
Changes filenames written with code backticks to use the Sphinx [file directive](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-file); ultimately, it renders the same.

This PR changes no *text*, only formatting.


jonparrott:
Thanks again, @9999years!

Issue #450, opened 2018-03-04
Guide request: gracefully dropping support for older Python versions


ncoghlan:
Talking to @tonybaloney about https://twitter.com/anthonypjshaw/status/970088412177776640, I discovered that our support for gracefully dropping support for older Python versions is currently better than I realised, but it's also far from obvious how to actually do it and the potential errors to avoid.

The key moving parts are:

* the `Requires-Python` field in metadata 1.2+: https://packaging.python.org/specifications/core-metadata/#requires-python
* the `data-requires-python` link attribute in the index server release listing API: https://www.python.org/dev/peps/pep-0503/#specification
* a `data-requires-python` aware installer (e.g. `pip` 9+)
* a publishing client that allows setting the `Requires-Python` field in the project metadata (e.g. the `python_requires` setting in `setuptools` 24.2.0+)
* an index server that sets `data-requires-python` correctly (e.g. `pypi.python.org` since https://github.com/pypa/pypi-legacy/commit/3e1ce8fb2acdad76574352d4db50a06109d82e74)

The most conservative approach to implicitly pinning older clients to the last compatible release would be to release an alpha or development release that sets "Requires-Python: >= 3.6" (for example) in the metadata, and includes an import-time warning snippet like the one below, but *doesn't* actually require the new version yet:

    _MINIMUM_SUPPORTED_VERSION = (3, 6, 0) # f-strings! Async generators!
    if sys.version_info < _MINIMUM_SUPPORTED_VERSION:
        import warnings
        warnings.warn("This project no longer supports this version of Python, "
                      "but the release metadata indicating that has been ignored. "
                      "Please update to a newer installation client (e.g. pip 9+) and "
                      "ensure that any caching proxies in use support PEP 503's"
                      "`data-requires-python` link attribute.", FutureWarning)

(Using `FutureWarning` ensures that the message will be visible by default even on versions where `DeprecationWarning` is hidden by default)

This conservative approach has a few key benefits:

1. It allows for opt-in testing of the metadata setting by folks running `pip install --pre` while others on recent `pip` versions will ignore it even if `Requires-Python` didn't get set correctly (since it's a pre-release)
2. It means that folks with older deployment toolchains (even truly ancient ones that allow installation of pre-releases by default) will only get a (valid!) warning on stderr instead of a completely broken deployment.
[...]

PR #459, opened 2018-03-21
Tutorial for phasing out Python versions

tonybaloney:
See #450 for discussion


source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+in the package metadata. If they do not match, it will attempt to install the last package distribution that supported that Python runtime.
+
+This mechanism can be leveraged to drop support for older Python versions, by amending the "Requires-Python" attribute in the package metadata.
+
+Requirements
+------------
+
+This workflow requires that:
+
+1. The publisher be using the latest version of :ref:`setuptools`,
```
hugovk: -> "The publisher is using"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+2. The latest version of :ref:`twine` be used to upload the package,
```
hugovk: -> "is used"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+3. The user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
+
+Defining the Python version required
+------------------------------------
+
+1. Download the newest version of Setuptools
```
hugovk: The requirements also say the latest version of Pip and Twine are needed.

Shall we make this section cover all of those, and update the command to `pip install --upgrade pip setuptools twine` all at once?

Or, because Pip can be a bit of a [special case](https://github.com/pypa/python-packaging-user-guide/blob/master/source/guides/installing-using-pip-and-virtualenv.rst#installing-pip), at least for Setuptools and Twine.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ensure that before you generate source-distributions or binary-distributions, you update setuptools.
```
hugovk: -> "source distributions"
-> "binary distributions"
-> "Setuptools"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Steps::
+
+   pip install --upgrade setuptools
+
+Version should be above 24.0.0
```
hugovk: End with a fullstop.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+2. Specify the version ranges for supported Python distributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can specify version ranges and exclusion rules, such as at least Python 3. Or, Python 2.7 and 3.4 beyond.
+
+Examples::
+
+    Requires-Python: ">=3"
+    Requires-Python: ">2.7,!=3.0.*,!=3.1.*, !=3.2.*,!=3.3.*"
```
hugovk: Add spaces after commas for consistency and readability.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+Steps::
+
+   pip install --upgrade setuptools
+
+Version should be above 24.0.0
+
+2. Specify the version ranges for supported Python distributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can specify version ranges and exclusion rules, such as at least Python 3. Or, Python 2.7 and 3.4 beyond.
```
hugovk: -> "3.4 and beyond."

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Examples::
+
+    Requires-Python: ">=3"
+    Requires-Python: ">2.7,!=3.0.*,!=3.1.*, !=3.2.*,!=3.3.*"
+
+The way to set those values is within the call to setup within your setup.py script. This will insert the Requires-Python
```
hugovk: -> "call to `setup`"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+metadata values based on the argument you provide in `python_requires`.
+
+..code-block:: Python
+
+    from setuptools import setup
+
+
+    setup(
+        # Your setup arguments
+        python_requires='>=2.7', # Your supported Python ranges
```
hugovk: Two spaces before inline comments.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+3. Validating the Metadata before publishing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Within a Python source package (the zip or the tar-gz file you download) is a text file called PKG-INFO.
+
+This file is generated by distutils or :ref:`setuptools` when it generates the source package.
```
hugovk: -> "Distutils"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+Within a Python source package (the zip or the tar-gz file you download) is a text file called PKG-INFO.
+
+This file is generated by distutils or :ref:`setuptools` when it generates the source package. 
+The file contains a set of keys and values, the list of keys is part of the PyPa standard metadata format. 
+
+You can see the contents of the generated file like this:
+
+    tar xvfz dist/my-package-1.0.0.tar.gz -O | cat */PKG-INFO
+
+Validate that the following is in place, before publishing the package
```
hugovk: End with a colon

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+- If you have upgraded correctly, the Metadata-Version value should be 1.2 or higher. 
+- The "Requires-Python" field is set and matches your specification in setup.py
```
hugovk: Remove quotes for consistency, and end with a fullstop.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+4. Using Twine to publish
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Twine has a number of advantages, apart from being faster is now the supported method for publishing packages.
```
hugovk: Add "it" -> "apart from being faster it is now"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+Make sure you are using the newest version of Twine, at least 1.9.
+
+Dropping a Python release
+-------------------------
+
+Once you have published a package with the Requires-Python metadata, you can then make a further update removing that Python runtime from support.
+
+It must be done in this order for the automated fail-back to work.
+
+For example, if you published the Requires-Python: ">=2.7" as version 1.0.0 of your package.
```
hugovk: Remove "if"

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+If you were then to update the version string to ">=3.5", and publish a new version, 2.0.0 of your package, any users running pip 9.0+ from version 2.7 of Python will
```
hugovk: Remove comma:
-> "publish a new version 2.0.0 of your package"

hugovk: Caps:
-> "Pip 9.0+"

hugovk: "from version 2.7 of Python will"
->
"from Python 2.7 will"

tonybaloney:
@hugovk all comments taken into account in 003466d

tonybaloney:
and thank you for the thorough review! You are quite the wordsmith :-)

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+This mechanism can be used to drop support for older Python versions, by amending the "Requires-Python" attribute in the package metadata.
+
+Requirements
+------------
+
+This workflow requires that:
+
+1. The publisher is using the latest version of :ref:`setuptools`,
+2. The latest version of :ref:`twine` is used to upload the package,
+3. The user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
```
ncoghlan: There's potentially a 4th requirement here, which is that the package index server has to populate the `data-requires-python` link attribute described in PEP 503. Both pypi.org (the warehouse codebase) and pypi.python.org (the legacy PyPI codebase) do that, but we're not sure about the various caching proxies that are out there - those may confound the client's version requirement detection.

However, I think we can leave that out for now, and potentially add it if folks report problems that get traced back to that issue.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+Dropping Support for Older Python Versions
+==========================================
+
+Dropping support for older Python versions is supported by the standard :ref:`core-metadata` 1.2 specification via a "Requires-Python" attribute.
+
+Metadata 1.2+ clients, such as Pip 9.0+, will adhere to this specification by matching the current Python runtime and comparing it with the required version 
+in the package metadata. If they do not match, it will attempt to install the last package distribution that supported that Python runtime.
+
+This mechanism can be used to drop support for older Python versions, by amending the "Requires-Python" attribute in the package metadata.
+
```
ncoghlan: We should include a caveat here that this guide is specifically for users of `setuptools`, or users of `distutils` that will need to upgrade to `setuptools` in order to be able to include the `Requires-Python` field in their metadata.

Users of other packaging tools will need to consult the relevant project's documentation on how to set `Requires-Python`, but the general approach described here will otherwise hold.

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+    setup(
+        # Your setup arguments
+        python_requires='>=2.7',  # Your supported Python ranges
+    )
+
+3. Validating the Metadata before publishing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Within a Python source package (the zip or the tar-gz file you download) is a text file called PKG-INFO.
```
ncoghlan: While I think this section will be fine if we add an "Aimed at `distutils`/`setuptools` users" caveat at the top, it does suggest to me that a `twine upload --dry-run` option that printed out the headers that would be sent without actually making a POST request to the server could be very helpful. (I'll add a comment to https://github.com/pypa/twine/issues/207 pointing back to this PR as an example use case)

source/guides/dropping-older-python-versions.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Validate that the following is in place, before publishing the package:
+
+- If you have upgraded correctly, the Metadata-Version value should be 1.2 or higher. 
+- The Requires-Python field is set and matches your specification in setup.py.
+
+4. Using Twine to publish
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Twine has a number of advantages, apart from being faster it is now the supported method for publishing packages.
```
ncoghlan: In addition to mentioning support here, I'd also suggest noting that it's the most reliable when it comes to correctly passing the artifact metadata to the publication server (the server doesn't inspect the artifacts directly - it relies on the client to supply the relevant metadata).

source/guides/dropping-older-python-versions.rst
tonybaloney: I'll leave this out I think

source/guides/dropping-older-python-versions.rst
tonybaloney: added in top comment

source/guides/dropping-older-python-versions.rst
tonybaloney: 👍

tonybaloney:
@ncoghlan added a comment in a04ed9c

ncoghlan:
I spotted a couple of things in the rendered version that I'd missed on review: https://github.com/pypa/python-packaging-user-guide/commit/e88a088045fcefa908a5a51f1a9b32b7b0cb7609

PR #460, opened 2018-03-28
Fix 'installing packages' link in Packaging and Distributing Packages.

andportnoy:
The Packaging and Distributing Projects tutorial assumes familiarity
with the Installing Packages tutorial, but links instead to the pip
Installation tutorial. This commit replaces the pip link with the link
to the Installing Packages tutorial.


jonparrott:
Thanks, @andportnoy!

PR #461, opened 2018-03-31
Add README guide

ddbeck:
This adds some guidelines for writing and setting metadata for READMEs that will play nicely with PyPI. This should cover a lot of the things discussed in #210.

### DIFF
--- a/source/guides/making-a-pypi-friendly-readme.rst
+++ b/source/guides/making-a-pypi-friendly-readme.rst
@@ -1,0 +1,82 @@
+Making a PyPI-friendly README
+=============================
+
+README files can help your users understand your project and can be used to set your project's description on PyPI.
+This guide helps you:
+
+* create a README in a PyPI-friendly format
+* include your README in your package so it appears on PyPI
+* find resources for writing README content that will help your users
+
+
+Creating a README file
+----------------------
+
+README files are often named ``README`` or ``README.txt`` for plain-text READMEs,
+or :samp:`README.{extension}` where *extension* indicates a markup language
+(such as ``rst`` for reStructuredText or ``md`` for Markdown).
+
+For your README to display properly on PyPI, choose a markup language supported by PyPI.
+Formats supported by `PyPI's README renderer <https://github.com/pypa/readme_renderer>`_ are:
+
+* plain text
+* `reStructuredText <http://docutils.sourceforge.net/rst.html>`_
+* Markdown (`GitHub Flavored Markdown <https://github.github.com/gfm/>`_ by default,
+  or `CommonMark <http://commonmark.org/>`_)
+
+It's customary to save your README file in the root of your project, in the same directory as your ``setup.py`` file.
+
+
+Write your README file
+----------------------
+
+Your README file can contain whatever you want, but it's common to include information that helps your reader understand:
+
+* what your project is and how it can help them
+* how to install or use your project
+* where to go to contribute or get help
+
+For more information about writing READMEs, check out these resources:
+
+* `README checklist <https://github.com/ddbeck/readme-checklist>`_
+* `Awesome README <https://github.com/matiassingers/awesome-readme>`_
+
+
+Including your README in your package's metadata
+------------------------------------------------
+
+To include your README's contents as your package description,
+set your project's ``Description`` and ``Description-Content-Type`` metadata.
+
+For example, to set these values in your package's ``setup.py`` file,
+use ``setup()``'s ``long_description`` and ``long_description_content_type``.
+
+Set the value of ``long_description`` to the contents of the README file itself.
+Set the ``long_description_content_type`` to an accepted ``Content-Type``-style value for your README file's markup,
+such as ``text/plain``, ``text/x-rst`` (for reStructuredText), or ``text/markdown``.
+
+.. seealso::
+
+   * :any:`metadata_description`
+   * :any:`metadata_description_content_type`
+
+For example, see this example :file:`setup.py` file,
+which reads the contents of :file:`README.md` as ``long_description`` and
+identifies the markup as GitHub-flavored markdown:
+
+.. code-block:: python
+
+   from setuptools import setup
+
+   # read the contents of your README file
+   from os import path
+   this_directory = path.abspath(path.dirname(__file__))
+   with open(path.join(this_directory, 'README.md'), encoding='utf-8') as f:
+       long_description = f.read()
+
+   setup(
+       name='an_example_package',
+       # other arguments omitted
+       long_description=long_description,
+       long_description_content_type='text/markdown; variant=gfm'
+   )
--- a/source/guides/index.rst
+++ b/source/guides/index.rst
@@ -24,3 +24,4 @@ introduction to packaging, see :doc:`/tutorials/index`.
    hosting-your-own-index
    migrating-to-pypi-org
    using-testpypi
+   making-a-pypi-friendly-readme
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -121,6 +121,8 @@ Example::
     Summary: A module for collecting votes from beagles.
 
 
+.. _metadata_description:
+
 Description (optional)
 ======================
 
@@ -164,6 +166,8 @@ Alternatively, the distribution's description may instead be provided in the
 message body (i.e., after a completely blank line following the headers, with
 no indentation or other special formatting necessary).
 
+
+.. _metadata_description_content_type:
 
 Description-Content-Type (optional)
 ===================================


source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,82 @@
...
+
+For your README to display properly on PyPI, choose a markup language supported by PyPI.
+Formats supported by `PyPI's README renderer <https://github.com/pypa/readme_renderer>`_ are:
+
+* plain text
+* `reStructuredText <http://docutils.sourceforge.net/rst.html>`_
+* Markdown (`GitHub Flavored Markdown <https://github.github.com/gfm/>`_ by default,
+  or `CommonMark <http://commonmark.org/>`_)
+
+It's customary to save your README file in the root of your project, in the same directory as your ``setup.py`` file.
```
pradyunsg: I think saying "setup.py or pyproject.toml" here would be nice.

ddbeck: This is a good idea, @pradyunsg ! I'll rework this a little to mention `pyproject.toml`.

Even better though would be to include a first-class `pyproject.toml` example. Do you know of any projects I could look at that specify a README and markup in their `pyproject.toml`?

ncoghlan: I think we want to hold off on pushing `pyproject.toml` too heavily until there's a version of `pip` available with initial PEP 517 support for pluggable build backends.

Then we can add a sample `flit` project with only `pyproject.toml` (assuming flit supports getting its settings from there).

ddbeck: OK, well, I still reworked things to leave room for other ways of specifying—it shouldn't require too much rewriting to add `pyproject.toml` when it becomes a practical option.

source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,83 @@
...
+* create a README in a PyPI-friendly format
+* include your README in your package so it appears on PyPI
+* find resources for writing README content that will help your users
+
+
+Creating a README file
+----------------------
+
+README files are often named ``README`` or ``README.txt`` for plain-text READMEs,
+or :samp:`README.{extension}` where *extension* indicates a markup language
```
jonparrott: why not just enumerate all of them? We only support 3 different content types `README.txt`, `README.md`, and `README.rst`.

source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Write your README file
+----------------------
+
+Your README file can contain whatever you want, but it's common to include information that helps your reader understand:
+
+* what your project is and how it can help them
+* how to install or use your project
+* where to go to contribute or get help
+
```
jonparrott: I would love a short example here.

source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,83 @@
...
+
+Write your README file
+----------------------
+
+Your README file can contain whatever you want, but it's common to include information that helps your reader understand:
+
+* what your project is and how it can help them
+* how to install or use your project
+* where to go to contribute or get help
```
jonparrott: The readme should also contain licensing information.

source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. seealso::
+
+   * :any:`metadata_description`
+   * :any:`metadata_description_content_type`
+
+For example, to set these values in a package's :file:`setup.py` file,
+use ``setup()``'s ``long_description`` and ``long_description_content_type``.
+
+Set the value of ``long_description`` to the contents of the README file itself.
```
jonparrott: suggestion: make it clear **not** to use the readme file name:

```
Set the value of ``long_description`` to the contents of the README file itself, not to the file name of the README file.
```

source/guides/making-a-pypi-friendly-readme.rst
```diff
@@ -0,0 +1,83 @@
...
+   from os import path
+   this_directory = path.abspath(path.dirname(__file__))
+   with open(path.join(this_directory, 'README.md'), encoding='utf-8') as f:
+       long_description = f.read()
+
+   setup(
+       name='an_example_package',
+       # other arguments omitted
+       long_description=long_description,
+       long_description_content_type='text/markdown; variant=gfm'
```
jonparrott: drop the variant here, as GFM is the default.

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: I don't want to go overboard about details of README content and I'm not sure there's a brief way to describe describing a license in a README file.

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: Because other formats exist even if we don't support them yet?

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: I worry that even a short example would be long enough to dominate the page. I'd rather link to resources and examples elsewhere.

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: OK, I'll change this.

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: OK, I can fix this up.

source/guides/making-a-pypi-friendly-readme.rst
di: I agree, this is what the license classifiers and the `License` metadata field are for.

source/guides/making-a-pypi-friendly-readme.rst
jonparrott: This is written as if in the abstract. This guide should focus on the *actual*:

```
READMEs for Python projects are generally named ``README``, ``README.rst``, or ``README.md``.
```

source/guides/making-a-pypi-friendly-readme.rst
jonparrott: That's fine considering you link to a sample.

source/guides/making-a-pypi-friendly-readme.rst
jonparrott: There absolutely is a brief way of describing it:

```
* the name of the license your project is under.
```

See [here](https://github.com/GoogleCloudPlatform/google-auth-library-python#license) for example.

It's so incredibly critical that this information is in the project description as many packages can omit the license metadata.

source/guides/making-a-pypi-friendly-readme.rst
ddbeck: OK.

source/guides/making-a-pypi-friendly-readme.rst
ncoghlan: Aside from "It was the early 2000s and licensing discipline wasn't anywhere near as well understood back then as some of understand it now", PyPI also doesn't require explicit licence metadata because it's the Terms of Service that set the minimum licensing terms (i.e. the ability to make verbatim copies of the uploaded files, even if you're not supposed to do anything else with them).

That said, I agree with @ddbeck that it makes sense to remain silent on the topic of including licensing info in README files for now, especially since platforms like GitHub will extract and display the license from your repo anyway. (Clarity in open source licensing for uploaded projects is also an area that could use a whole lot of improvements in general, and a single bullet point here isn't going to resolve those issues)

jonparrott: I don't mean to be a jerk, and I'm happy to be outvoted here, but please here my last plea to include this here. I consider this critically important to OSS, so, please, hear me out just once more.

---

LIcensing is like security. It should be done *in depth*. There's a reason Apache and friends tell you to put a *license header* on every file. It's to reduce the amount of ambiguity around what license things are under. Naming the license in the README is just another avenue to reduce ambiguity (I *never* suggested it should be the only thing you do). Naming the license *is sufficient* for this task (it's akin to the license header in a source file). Keep in mind that in lots of situations it's entirely possible for the package metadata to be lost in one way or another (repackaging, etc.).

---

I also don't understand why there is so much push back on this item. I would argue that there are lots of projects that *will not* accept contributions, lots of projects that aren't meant to be installed broadly, and lots of projects that are made for just one person. This sort of argument can be used to just remove all of these bullet points. But critically, every package uploaded to the package index should indicate, somehow, even if it's just by name, what the terms of use are.

ddbeck: Thanks for your comment @jonparrott and your patience while I got back to you on this.

To go meta for a moment, you asked why there was push back here. I can't speak for anyone else, but I put a lot of care into this doc and I had good reasons for pretty much everything I wrote from the outset. And I have bit of a personal attachment to the topic generally, since "Daniel cares a lot about READMEs" seems to be my personal brand of late. What can I say? I care about this.

Now for the actual issue: I see where you're coming from. I agree with you that licensing is important. But I don't think that providing licensing information to users, in isolation, is a primary goal of README files. Nor do I think licensing represents a major problem for README authors: when I surveyed READMEs for a conference talk, I found that remembering to name your project and link to the repo or homepage is a more significant and pervasive failure on the part of README authors than licensing. 😬

But I'm also uncomfortable with implying that naming a license is significant and meaningful (by itself, outside a bigger discussion on software licensing). I believe that approach contributes to a ritualistic view of copyright (and law broadly) in which people believe [magic words will bind others' actions](https://www.pixsy.com/does-no-copyright-infringement-intended-have-any-legal-significance/). It feels like a sort of anti-civics lesson.

I have to say though, I really like this general idea of licensing in-depth. I would _love_ to see a doc on how to properly license, top to bottom, a Python package (and maybe even address some Python specific examples—like what's the relationship between `import somelib` and its license? I wish I knew!). I think my ideal outcome would be to link to that page from this one, rather than trying to cover a tiny slice of it here.

jonparrott: > I would love to see a doc on how to properly license, top to bottom, a Python package (and maybe even address some Python specific examples—like what's the relationship between import somelib and its license? I wish I knew!). I think my ideal outcome would be to link to that page from this one, rather than trying to cover a tiny slice of it here.

I would love that, want to file a separate issue?

---

Back to the topic at hand - Let me rephrase your argument *against* adding this as an argument against adding one of the items you already have in this list:

>  But I don't think that providing *contribution* information to users, in isolation, is a primary goal of README files. Nor do I think *accepting contributions* represents a major problem for README authors.

My argument is this: either this section should contain *just the bare minimum* which you identified through survey as just "the project name and homepage" or it we should amend this list to include the name of the license. If it's going to be an arbitrary list of common items, license should be there. If it's going to be the bare minimum, then none of these other things should be there.

jonparrott: Also want to offer a reasonable bridge here: We could go the bare minimum route here and dp something to this effect:

> Your README should at minimum contain your project's name and a link to your project's homepage where users can find out more information. For a information on writing a more comprehensive project README, we recommend using [readme-checklist](https://github.com/ddbeck/readme-checklist/blob/master/checklist.md).

Your README checklist project more than satisfies my desires in terms of documenting the license and I am more than happy to promote it here.

jonparrott: And of course, re-reading the changeset I see that you already link to it, which makes me more in favor of the minimalist approach.

ddbeck: @jonparrott please don't change my words to say something I would not have argued and then put it in a block quote. This feels a half step removed from the mocking SpongeBob meme.

Anyway, I'm chaffing a bit at the suggestion that my recommendations were arbitrary, so I'm pushing a change to drop the README content section entirely. Accept the PR or don't, but I'm done being needled over this.

jonparrott:
FYI: chatted with @ddbeck over email and I'm cool with these changes. Merging. :)

Thanks, @ddbeck!

PR #464, opened 2018-04-04
fix typo

wimglenn:
licencse is a good compromise between licence and license, but unfortunately not a real word  ;)


jonparrott:
Thanks, @wimglenn!

PR #466, opened 2018-04-09
Remove GPG references from publishing tutorial

ncoghlan:
Warehouse is moving away from publishing GPG signatures,
and even if it wasn't, it isn't a topic for the "Publishing your first package"
tutorial.


dstufft:
tenor

PR #469, opened 2018-04-15
Describe required metadata fields at the start of the spec

takluyver:
And remove "(optional)" from field headings. As described on distutils-sig, this was not applied accurately - its absence implied that some optional fields were required.

@ncoghlan suggested using explicit rst link targets to preserve old links to the sections, like `...#description-optional`, and I've done this.

Remaining question: is `Summary` required? It is not marked optional in any of the PEPs specifying metadata 1.0-1.2, but @dstufft says that PyPI does not require it, so it is possible to upload packages without it.

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -9,8 +9,13 @@ It defines the following specification as the canonical source for the core
 metadata file format.
 
 Fields defined in the following specification should be considered valid,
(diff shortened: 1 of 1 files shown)


ncoghlan:
Thanks! Let's run with the pragmatic approach of letting Summary be optional at the interoperability layer, and leave it up to publishing tool UX to encourage adding it.

PR #471, opened 2018-04-16
adds instructions for Warehouse

lgh2:
This updates the guide to add instructions for interacting with PyPI.  I have left the instructions on how to update .pypirc and the error message when registering project names, so if people have problems the solutions will hopefully show up when they search for them. 

When I was writing this I assumed, please correct me if I am wrong, that ``pypi.python.org`` had already been redirected and that ``pypi.org`` was the default for downloading packages.

### DIFF
--- a/source/guides/migrating-to-pypi-org.rst
+++ b/source/guides/migrating-to-pypi-org.rst
@@ -3,21 +3,17 @@
 Migrating to PyPI.org
 =====================
 
-:term:`PyPI.org` is a new, rewritten version of PyPI that is replacing the
-legacy code base located at `pypi.python.org`. As it becomes the default, and
-eventually only, version of PyPI people are expected to interact with, there
-will be a transition period where tooling and processes are expected to need to
-update to deal with the new location.
-
-This section covers how to migrate to the new PyPI.org for different tasks.
-
+:term:`PyPI.org` is the new, rewritten version of PyPI that has replaced the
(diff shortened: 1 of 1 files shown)


source/guides/migrating-to-pypi-org.rst
```diff
@@ -111,46 +107,24 @@ Registering new user accounts
...
-
-``pypi.org`` is fully functional for purposes of browsing available packages, and
-some users may choose to opt in to using it.
-
-``pypi.org`` is expected to become the default recommended interface for browsing
-once the limitations in the next section are addressed (at which point
-attempts to access ``pypi.python.org`` will automatically be redirected to
-``pypi.org``).
+While ``pypi.python.org`` is may still be used in links from other PyPA
+documentation, etc, the default interface for browsing packages is ``pypi.org``. During the transition, at some point``pypi.python.org`` will automatically be redirected to ``pypi.org``).
```
di: Maybe change:

> During the transition, at some point``pypi.python.org`` will automatically be redirected to ``pypi.org``).

to:

> The domain ``pypi.python.org`` now redirects to ``pypi.org``, and may be disabled sometime in the future.

as this will happen tomorrow.

jonparrott:
I will merge this once we're able to get a green build - it's currently failing because of some turbulence with the pypi.org cutover.

PR #475, opened 2018-04-16
Update instructions and status for PyPI launch

brainwane:


brainwane:
would love a review, @alanbato @jonparrott

jonparrott:
Thanks, @brainwane!

PR #477, opened 2018-04-24
Update pypi.python.org links to pypi.org links

### LINKED ISSUE #476: Update pypi.python.org links to pypi.org links
I think this would be a pretty easy task for a new first-timer.

1. Follow `README.rst` to clone and build the guide locally.

2. In the root of the repository, run:

`git grep "https://pypi.python.org"`

3. Go through those files and replace most of the old pypi.python.org URLs with new pypi.org URLs. Guides: https://packaging.python.org/guides/migrating-to-pypi-org/ and https://warehouse.readthedocs.io/api-reference/integration-guide/#migrating-to-the-new-pypi . If you want to, get fancy with regular expressions -- http://regexr.com/ might help. Be careful not to change URLs where we deliberately mean to mention the deprecated old address.

4. In the root of the repository, build the guide. Verify that all the links you changed work.

5. Submit PR. :)

pgadige:
Fixes #476 

Updated `pypi.python.org` links to `pypi.org` links.
Changed `https://pypi.python.org/pypi?%3Aaction=list_classifiers` to `https://pypi.org/classifiers/`

### DIFF
--- a/source/guides/migrating-to-pypi-org.rst
+++ b/source/guides/migrating-to-pypi-org.rst
@@ -40,7 +40,7 @@ configured in a file located at ``$HOME/.pypirc``. If you see a file like:
         pypi
 
     [pypi]
-    repository:https://pypi.python.org/pypi
+    repository:https://pypi.org/pypi
     username:yourusername
     password:yourpassword
 
--- a/source/discussions/deploying-python-applications.rst
+++ b/source/discussions/deploying-python-applications.rst
@@ -60,7 +60,7 @@ Windows
 Pynsist
 ^^^^^^^
 
-`Pynsist <https://pypi.python.org/pypi/pynsist>`__ is a tool that bundles Python
+`Pynsist <https://pypi.org/pypi/pynsist>`__ is a tool that bundles Python
 programs together with the Python-interpreter into a single installer based on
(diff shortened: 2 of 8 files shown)


source/guides/migrating-to-pypi-org.rst
```diff
@@ -40,7 +40,7 @@ configured in a file located at ``$HOME/.pypirc``. If you see a file like:
         pypi
 
     [pypi]
-    repository:https://pypi.python.org/pypi
+    repository:https://pypi.org/pypi
```
di: This should not change, as it represents the state of `.pypirc` _before_ migrating, so it may have old links.

pgadige: Thank you for the feedback. I've reverted to old link.

pgadige:
Thank you for the feedback, @di! 
I changed links of the form `https://pypi.org/pypi/something` to `https://pypi.org/project/something` to avoid redirects.
I also changed `https://pypi.python.org/pypi?%3Aaction=list_classifiers` to `https://pypi.org/classifiers/` in one more file, `source/specifications/core-metadata.rst`

jonparrott:
Thank you, @pgadige and @di!

PR #490, opened 2018-05-14
Fix section on data_files

jdemeyer:
This is a reboot of #212.

I no longer mention distutils at all and the fact that `data_files` only works with relative paths is stated as fact, not as possible bug.


theacodes:
Thanks, @jdemeyer! Reviewing now.

PR #491, opened 2018-05-14
"nox -s preview" was failing, since it tried to open "build/html".  I…

reuven:
… changed (thanks to Thea)

The command to open "build" rather than "build/html", and it now works.


theacodes:
Thanks, @reuven!

PR #498, opened 2018-05-17
Create new, simpler packaging tutorial

### LINKED ISSUE #203: New tutorials: make a package
(For more information about this issue, please see #194).

We need a short section to follow the section about making a distribution package (see #202) (i.e., as its own file) that will guide the reader through making a bare module or (import) package ready for installing with `pip` (and, in the following section, publishing on PyPI). The section should satisfy these goals:
- show how to make a `setup.py` file that uses `setup()`
- briefly acquaint the reader with friendly practices such as including a `README` file, a license, and using a PEP 440-compliant versioning scheme (link generously to resources that cover these things in detail)
- ~~show how to install the package in editable/develop mode~~
- show how to create an sdist and wheel
- show how to test that the package is, in fact, installable
- show the work in the context of a virtual environment (as this will have been covered in the previous section)

Please note that this section should not assume the reader already has modules to package (so provide some short sample code where applicable). Also, don't cover publishing the package to PyPI or elsewhere—that's going to be covered in the next tutorial section.

When opening pull requests for this issue, please submit to the `develop` branch. If you have questions that concern contributing more generally, please use issue #194. Otherwise, feel free to comment with questions or feedback. Thanks!

### LINKED ISSUE #298: Expand on using TestPyPI in the distributing guide
Original context: https://github.com/pypa/python-packaging-user-guide/pull/271#issuecomment-293357309

We already mention that TestPyPI is an option, however, we link off to separate documentation on how to use it. We should self-host this documentation and better surface it in our guide.

### LINKED ISSUE #494: Feedback from PyCon sprint on package creation tutorial
(@brainwane asked me to write this, despite not being a well-formed issue)

During the PyCon sprint for Python packaging, I worked through the [tutorial](https://packaging.python.org/tutorials/distributing-packages/#configuring-your-project) and wrote down pain points or places I got stuck.

* The material is good, but it is written more as a reference than as a tutorial.  I had to read through a [long section on optional arguments to `setup()`](https://packaging.python.org/tutorials/distributing-packages/#setup-args) before getting to how to build and upload a package.
* It would be more instructive to create a minimal example package, build an sdist, and upload it to <https://test.pypi.org>.  The reference material can come later, or on a separate page.
* Minor issue: it's not clear to me after doing the tutorial what the difference between an sdist and a universal wheel is, or when I would prefer one over the other.

Otherwise, I think the content is accurate and the right level of detail.

theacodes:
Move the existing tutorial into guides/distributing-packages-using-setuptools to be cleaned up later.

Staged: http://temp.theadora.io/pypug/tutorials/packaging-projects.html

Resolves #494 
Resolves #298 
Resolves #203

### DIFF
--- a/source/tutorials/packaging-libraries.rst
+++ b/source/tutorials/packaging-libraries.rst
@@ -1,0 +1,344 @@
+Packaging Python Libraries
+==========================
+
+This tutorial walks you through how to package a simple Python library. It will
+show you how to add the necessary files and structure to create the package, how
+to build the package, and how to upload it to the Python Package Index.
+
+
+A simple library
+----------------
+
+This tutorial uses a simple library named `example_pkg`. If you are unfamiliar
+with Python's modules and :term:`import packages <import package>`, take a few
+minutes to read over the `Python documentation for packages and modules`_.
+
+To create this library locally, create the following file structure:
+
+.. code-block:: text
+
+    /example_pkg
+      /example_pkg
+        __init__.py
+
+
+Once you create this structure, you'll want to run all of the commands in this
+tutorial within the top-level folder - so be sure to ``cd example-pkg``.
+
+You should also edit :file:`example_pkg/__init__.py` and put the following
+code in there:
+
+.. code-block:: python
+    
+    name = "example_pkg"
+
+This is just so that you can verify that it installed correctly later in this
+tutorial.
+
+.. _Python documentation for packages and modules:
+    https://docs.python.org/3/tutorial/modules.html#packages
+
+
+Creating the package files
+--------------------------
+
+You will now create a handful of files to package up this library and prepare it
+for distribution. Create the new files listed below - you will add content to
+them in the following steps.
+
+.. code-block:: text
+
+    /example_pkg
+      /example_pkg
+        __init__.py
+      setup.py
+      LICENSE
+      README.md
+
+
+Creating setup.py
+-----------------
+
+:file:`setup.py` is the build script for :ref:`setuptools`. It tells setuptools
+about your package (such as the name and version) as well as which code files
+to include.
+
+Open :file:`setup.py` and enter the following content, you can personalize
+the values if you want:
+
+.. code-block:: python
+
+    import setuptools
+
+    with open("README.rst", "r") as fh:
+        long_description = fh.read()
+
+    setuptools.setup(
+        name="example_pkg",
+        version="0.0.1",
+        author="Example Author",
+        author_email="author@example.com",
+        description="A small example package",
+        long_description=long_description,
+        long_description_content_type="text/markdown",
+        url="https://github.com/pypa/example-project",
+        packages=setuptools.find_packages(),
+        classifiers=(
+            "Programming Language :: Python :: 3",
+            "License :: OSI Approved :: MIT License",
+            "Operating System :: OS Independent",
+        ),
+    )
+
+
+:func:`setup` takes several arguments. This example package uses a relatively
+minimal set:
+
+- ``name`` is the name of your package. This can be any name as long as only
+  contains letters, numbers, ``_`` , and ``-``. It also must not already
+  taken on pypi.org.
+- ``version`` is the package version see PEPTODO for more details on
+  versions.
+- ``author`` and ``author_email`` are used to identify the author of the
+  package.
+- ``description`` is a short, one-sentence summary of the package.
+- ``long_description`` is a detailed description of the package. This is
+  shown on the package detail package on the Python Package Index. In
+  this case, the long description is loaded from :file:`README.md` which is
+  a common pattern.
+- ``long_description_content_type`` tells the index what type of markup is
+  used for the long description. In this case, it's Markdown.
+- ``url`` is the URL for the homepage of the project. For many projects, this
+  will just be a link to GitHub, GitLab, Bitbucket, or similar code hosting
+  service.
+- ``packages`` is a list of all Python :term:`import packages <Import
+  Package>` that should be included in the :term:`distribution package`. 
+  Instead of listing each package manually, we can use :func:`find_packages`
+  to automatically discover all packages and subpackages. In this case, the
+  list of packages will be `example_pkg` as that's the only package present.
+- ``classifiers`` tell the index and :ref:`pip` some additional metadata
+  about your pckage. In this case, the package is only compatible with Python
+  3, is licensed under the MIT license, and is OS-independent. You should
+  always include at least which version(s) of Python your package works on,
+  which license your package is available under, and which operating systems
+  your package will work on. For a complete list of classifiers, see
+  https://pypi.org/classifiers/.
+
+There are many more than the ones mentioned here. See
+:doc:`/guides/distributing-packages-using-setuptools` for more details.
+
+
+Creating README.md
+------------------
+
+Open :file:`README.md` and enter the following content. You can customize this
+if you'd like.
+
+.. code-block:: md
+
+    # Example Package
+
+    This is a simple example package. You can use
+    [Github-flavored Markdown](http://TODO) to write your content.
+
+
+Creating a LICENSE
+------------------
+
+It's important for every package uploaded to the Python Package Index to include
+a license. This tells users who install your package the terms under which they
+can use your package. For help picking a license, see
+https://choosealicense.com/. For now, open :file:`LICENSE` and enter the
+following content - which is the MIT license.
+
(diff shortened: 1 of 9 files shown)


source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
+Packaging Python Libraries
+==========================
+
+This tutorial walks you through how to package a simple Python library. It will
+show you how to add the necessary files and structure to create the package, how
+to build the package, and how to upload it to the Python Package Index.
```
di: Would it make sense to link to various terms in the glossary as they are introduced in this guide? In addition, "library" is not even in the glossary, so it might be confusing if this is the first page of the guide a user sees.

source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
...
+
+
+Creating a LICENSE
+------------------
+
+It's important for every package uploaded to the Python Package Index to include
+a license. This tells users who install your package the terms under which they
+can use your package. For help picking a license, see
+https://choosealicense.com/. For now, open :file:`LICENSE` and enter the
+following content - which is the MIT license.
```
di: This almost seems like we prefer the MIT license. It might be good to say something like:

> Once you have chosen a license, open `LICENSE` and enter the license text. For example, if you had chosen the MIT license:

source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
...
+.. note:: If you run into trouble here, please copy the output and file an issue
+  over on `packaging problems`_ and we'll do our best to help you!
+
+.. _packaging problems:
+  https://github.com/pypa/packaging-problems/issues/new?title=Trouble+following+packaging+libraries+tutorial
+
+
+The ``tar.gz`` file is a :term:`source archive` - it is intended mostly for
+downstream packagers (such as Linux distros) or users who have disabled
+installing built distributions. The ``.whl`` file is a :term:`built
```
di: I think we should be a little stronger about why users should upload source distributions.

The "it is intended mostly for downstream packagers" bit will probably read to a new user (who is unfamiliar with who or what downstream packagers are) that it's unnecessary.

One current headache on PyPI is that sometimes maintainers only upload wheels for a release, and users who have incompatible platform have no way to install that given release. In addition, source distributions are also a little easier to audit.

Something as strong as "You should always upload a source distribution for a release" would be ideal, IMO.

source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
...
+:doc:`/guides/using-testpypi`.
+
+Now that you are registered, you can use :ref:`twine` to upload the
+distribution packages. You'll need to install twine:
+
+.. code-block:: bash
+
+    python3 -m pip install --user --upgrade twine
+
+Once install, run twine to upload all of the archives under :file:`dist`:
```
di: "Once _installed_"

source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
...
+    >>> import example_pkg
+    >>> example_pkg.name
+    'example_pkg'
+
+Next steps
+----------
+
+**Congratulations, you've packaged and distributed a Python library!**
+✨ 🍰 ✨
+
```
di: This might be a good place to remind readers that TestPyPI is ephemeral, and that this package will go away someday, so if they truly want to make it available to users they will need to upload it to PyPI.

source/tutorials/packaging-libraries.rst
ncoghlan: We mainly stuck with the phrase "project" to avoid having to make the distinction, and saying "Python projects use `pyproject.toml`" (once this guide is able to switch to that) would make sense.

theacodes: Good call, I went ahead and replaced "library" with "project".

@di I tried to be really liberal with links to the glossary, but let me know if I missed any.

source/tutorials/packaging-libraries.rst
```diff
@@ -0,0 +1,344 @@
...
+Open :file:`README.md` and enter the following content. You can customize this
+if you'd like.
+
+.. code-block:: md
+
+    # Example Package
+
+    This is a simple example package. You can use
+    [Github-flavored Markdown](http://TODO) to write your content.
```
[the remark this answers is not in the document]
theacodes: Good catch. Fixed.

source/tutorials/packaging-libraries.rst
theacodes: Good call, done.

source/tutorials/packaging-libraries.rst
theacodes: Agreed, done.

source/tutorials/packaging-libraries.rst
theacodes: Good catch. Fixed.

source/tutorials/packaging-libraries.rst
theacodes: Good call, done.

theacodes:
Restaged over at: http://temp.theadora.io/pypug/tutorials/packaging-projects.html

source/tutorials/packaging-projects.rst
```diff
@@ -0,0 +1,345 @@
...
+
+This tutorial walks you through how to package a simple Python project. It will
+show you how to add the necessary files and structure to create the package, how
+to build the package, and how to upload it to the Python Package Index.
+
+
+A simple project
+----------------
+
+This tutorial uses a simple project named `example_pkg`. If you are unfamiliar
```
di: Should this be `example_pkg` and not _example_pkg_?

source/tutorials/packaging-projects.rst
```diff
@@ -0,0 +1,345 @@
...
+
+    Collecting example_pkg
+      Downloading https://test-files.pythonhosted.org/packages/.../example_pkg-0.0.1-py3-none-any.whl
+    Installing collected packages: example-pkg
+    Successfully installed example-pkg-0.0.1
+
+You can test that it was installed correctly by importing the module and
+referencing the ``name`` property you put in :file:`__init__.py` earlier.
+
+Run python (make sure you're still in your virtualenv):
```
di: Maybe "Run the Python interpreter?" Either way I think "Python" should be capitalized.

source/tutorials/packaging-projects.rst
```diff
@@ -0,0 +1,345 @@
...
+    Enter your password:
+    Uploading example_pkg-0.0.1-py3-none-any.whl
+    100%|█████████████████████| 4.65k/4.65k [00:01<00:00, 2.88kB/s]
+    Uploading example_pkg-0.0.1.tar.gz
+    100%|█████████████████████| 4.25k/4.25k [00:01<00:00, 3.05kB/s]
+
+.. note:: If you get an error that says ``The user '[your username]' isn't
+  allowed to upload to project 'example-pkg'``, you'll need to go and pick
+  a unique name for your package. A good choice is
+  ``example_pkg_your_username``. Update the ``name`` argument in
```
pradyunsg: minor nit -- one place used dash, suggested workaround uses underscore; both should be the same.

theacodes:
@ncoghlan *very* gentle ping

source/tutorials/packaging-projects.rst
```diff
@@ -313,23 +313,24 @@ And then import the module and print out the ``name`` property. This should be
...
-**Congratulations, you've packaged and distributed a Python library!**
+**Congratulations, you've packaged and distributed a Python project!**
 ✨ 🍰 ✨
 
 If you want to upload your package to the real Python Package Index you can do
 it by registering an account on https://pypi.org and following the same
 instructions above but omitting the ``--repository-url`` argument to ``twine``
-and the ``--index-url`` argument to ``pip``.
+and the ``--index-url`` argument to ``pip``. Note that Test PyPI is emphermal
+and it's not unusual for packages and accounts to be deleted occasionally.
```
ncoghlan: Typo for "ephemeral", but I'm also wondering if this might read better as a separate paragraph preceding this one. That way the paragraph about the real PyPI reads like an answer to the follow-on question of "OK, you've reminded me that this is ephemeral, so how do I make it persistent?"

theacodes: Great point. Done - just made it the leading sentence in this paragraph.

theacodes:
Thanks everyone!

PR #501, opened 2018-05-20
Remove outdated warning about Python version mixing with Pipenv

### LINKED ISSUE #452: Revise `pipenv` version mixing caveat
`pipenv` 11 uses the Python in the virtual environment to generate the lock file now, so version differences between the host Python and venv Python are no longer a problem. The caveat should be reworded to:

1. Make sure folks are using `pipenv` 11+
2. Note that `pipenv` folks on a single target version, and running `pipenv install` on a different version may implicitly install extra unpinned dependencies (if `requires_python` isn't explicitly set in `Pipfile`). (Alternatively, we could just recommend always setting `requires_python`, and note that while possible, dependency locking when targeting multiple versions of Python can be a bit tricky due to version dependent dependencies)

theacodes:
Resolves #452

### DIFF
--- a/source/tutorials/managing-dependencies.rst
+++ b/source/tutorials/managing-dependencies.rst
@@ -113,16 +113,7 @@ when you share your project with others. You should get output similar to this
     Adding requests to Pipfile's [packages]...
     P.S. You have excellent taste! ✨ 🍰 ✨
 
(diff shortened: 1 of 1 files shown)


Pareone (commented):
Hola

PR #516, opened 2018-05-30
Modify system command to account for Jupyter notebook

### LINKED ISSUE #407: Account for Jupyter shell/kernel interactions
The tutorials now mention `!python -m pip ...` as a way of running installation commands.

However, this isn't quite right due to the problems @jakevdp describes in https://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/index.html

At the very least, we should update those examples to use the variant of the command that ensures the right environment is being updated.

pgadige:
Fixes #407 
To accurately distinguish between Jupyter shell and kernel interactions, change existing python system command to its Jupyter variant.
Wrap the paragraph about Jupyter notebook into a separate note.

### DIFF
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -69,6 +69,11 @@ please install the latest 3.x version from `python.org`_ or refer to the
                 !{sys.executable} --version
         Python 3.6.3
 
+   It's recommended to write ``{sys.executable}`` rather than plain ``python`` in
+   order to ensure that commands are run in the Python installation matching
+   the currently running notebook (which may not be the same Python
+   installation that the ``python`` command refers to).
+
 .. Note:: Due to the way most Linux distributions are handling the Python 3
    migration, Linux users using the system Python without creating a virtual
    environment first should replace the ``python`` command in this tutorial
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -59,13 +59,14 @@ please install the latest 3.x version from `python.org`_ or refer to the
     an introduction to using your operating system's shell and interacting with
     Python.
 
-    If you're using an enhanced shell like IPython or the Jupyter notebook, you
-    can run system commands like those in this tutorial by prefacing them with
-    a ``!`` character:
+.. Note:: If you're using an enhanced shell like IPython or the Jupyter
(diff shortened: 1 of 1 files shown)


source/tutorials/installing-packages.rst
```diff
@@ -59,13 +59,14 @@ please install the latest 3.x version from `python.org`_ or refer to the
...
+.. Note:: If you're using an enhanced shell like IPython or the Jupyter
+   notebook, you can run system commands like those in this tutorial by
+   prefacing them with a ``!`` character:
 
     ::
 
-        In [1]: !python --version
+        In [1]: import sys
+                !{sys.executable} --version
         Python 3.6.3
```
ncoghlan: Nice! I'm wondering if it may be worth briefly stating why we're using `sys.executable` here, rather than a plain `python`. Something like this just after the example:

> It's recommended to write `{sys.executable}` rather than `python` in order to ensure that commands are run in the Python installation matching the currently running notebook (which may not be the same Python installation that the `python` command refers to).

@theacodes, your thoughts?

theacodes: Yeah I think that would be wise.

pgadige: Thank you for your valuable feedback @ncoghlan and @theacodes ! I made the desired changes right away. Could you please review and let me know if any changes are required?

source/tutorials/installing-packages.rst
```diff
@@ -59,15 +59,21 @@ please install the latest 3.x version from `python.org`_ or refer to the
...
     Python.
 
-    If you're using an enhanced shell like IPython or the Jupyter notebook, you
-    can run system commands like those in this tutorial by prefacing them with
-    a ``!`` character:
+.. Note:: If you're using an enhanced shell like IPython or the Jupyter
+   notebook, you can run system commands like those in this tutorial by
+   prefacing them with a ``!`` character:
 
     ::
```
pradyunsg: minor style nit: it's possible to get rid of this line and just have to `::` after `character` instead.

theacodes: Eh, let's leave this as-is for now. I really prefer the more verbose `..code-block:: syntax`.

theacodes:
Thank you @pgadige!

Issue #310, opened 2017-05-10
Possible reference article on "The Packaging Gradient"


ncoghlan:
@mahmoud just published a good article looking at the many possible layers at which developers can consider the question of application packaging and deployment: http://sedimental.org/the_packaging_gradient.html

I don't think we want to pull that directly into PyPUG, but it seems like it could be useful to reference somewhere as background material for some of the "Additional Topics" sections.

jonparrott:
I'm definitely for this, as it's a fantastic resource for *application* packaging options as opposed to *library* packaging.

mahmoud:
I'm glad you liked it! It's definitely a good introductory that feels good to get out there, and I couldn't have asked for a warmer reception. Even [HN](https://news.ycombinator.com/item?id=14309168) didn't totally hate it! If there's anything I can do to help adapt the content please let me know. I may need some background on the guide (and some time to recover from PyCon), but it sounds like a fun summer activity :)

pzwang:
Just stopping by to add an emphatic +1 on this.  In discussions on Python packaging, we can sometimes get lost on the "how" and the "what", and lose sight of the "why".  Mahmoud's presentation nicely distills down the user job stories about *why* people want to package something, and therefore what is even within the scope of definition of "binary artifact".

(IMO I think 90% of the misunderstandings about conda stem from this simple issue: different people mean vastly different things when they use the word "platform" or "runtime".)

ncoghlan:
Closed via #519 :)

PR #519, opened 2018-05-31
Add outline / rough draft of new Overview page

mahmoud:
Following up on @theacodes and my discussion at the PyCon 2018 packaging sprint, this PR brings in a new Overview document, meant to be linked from the [packaging.python.org](http://packaging.python.org/) landing page. Also per Thea's instruction, I'm submitting the PR much closer to an "outline" state than a finished document.

This PR also addresses #310, providing much more direct coverage of packaging applications written in Python. This initial PR took a bit longer than expected because there are some intentional changes in direction from [the original inspiration](http://sedimental.org/the_packaging_gradient.html):

1. More direct approach. Not assuming the audience has packaged Python before.
2. Coverage of web, mobile, PaaS, and other frameworks.
3. Brief explanation of *tools* -- (often smaller) applications with a developer audience
4. Shorter sections so as to function more like a portal

TODOs for this document before the draft is final:

- [x] Flesh out and edit down sections for an even reading experience.
- [x] Documentation links for each technology.
- [x] Open-source project example links (the idea for which gave me a great idea for a survey we could run...)
- [x] Possibly import illustrations, if folks are interested

Some features I was planning for this doc may best be broken out into dedicated articles for each approach, e.g., "Packaging Python Applications with Snappy", just to keep it maximally browsable. Specifically I think a "Packaging Python Applications with Freezers" would be a great place for a table comparing the myriad options.

Anyways, hope you like the new direction, looking forward to continuing work on this!


source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+
+A Python file, provided it only relies on the standard library, can be
+redistributed and reused. You will also need to ensure it's written
+for the right version of Python.
+
+Python source distributions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If your code consists of multiple Python files, it's usually organized
+into a *package*. A package is any directory containing Python files,
```
ncoghlan: We have both [import package](https://packaging.python.org/glossary/#term-import-package) and [distribution package](https://packaging.python.org/glossary/#term-distribution-package) defined in the glossary, so I'd suggest adjusting this section a bit to cover the ambiguity of the "package" term, and linking to those terms the first time they're used.

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+
+In all these cases, it makes sense to work backwards, from the
+framework's packaging and deployment story. Frameworks wrap the
+technologies outlined in the rest of the guide, and can make your
+production experience easier and more reliable.
+
+Service platforms
+*****************
+
+If you're developing a "Platform-as-a-Service" or "PaaS" like Heroku
```
ncoghlan: Missing "for" after "developing"

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+or Google App Engine, you are going to want to follow their respective
+packaging guides.
+
+* Heroku
+* Google App Engine
+* PythonAnywhere
+* OpenShift
```
ncoghlan: Zappa for AWS Lambda would also be worth mentioning here.

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+the coming sections.
+
+.. TODO: Another way of thinking about packaging solutions is by how
+   much they include. All solutions include your code, plus some
+   amount of your code's library and service dependencies. PEX
+   includes Python libraries. RPM includes a list of dependencies on
+   libraries and local services. Images can be built to include
+   everything.
+
+Depending on a framework
```
ncoghlan: "deployment system" might be a better phrase here (I initially thought you meant web framework)

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+
+In all these setups, the platform takes care of packaging and
+deployment, as long as you follow their patterns. Most software does
+not fit these templates, hence the existence of all the other options
+below.
+
+If you're developing software that will be deployed to machines you
+own, users' personal computers, or any other arrangement, read on.
+
+Web and mobile platforms
```
ncoghlan: While technically accurate, I think mentioning "Web and ..." in the heading here is more confusing than helpful (since the previous section also mostly targets the web, just for backend network services rather than frontend single-page applications).

Alternatively, expanding it to `Web browsers and mobile platforms` would likely also work, as that makes it clearer that it isn't referring to backend web services the way the previous section does.

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+************************
+
+Python's advances are leading it into new spaces. These days you can
+write a mobile app or web application frontend in Python. While the
```
ncoghlan: Along a similar vein to tinkering with the section title, I think sticking a "single page" qualifier before "web application" will help distinguish the scope of this from more traditional server-side template rendering in a web framework.

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+Depending on a pre-installed Python
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Depending on the host system to have Python installed. Common in
+controlled environments like data centers, and local environments of
+tech savvy people. Technically includes pretty much every major Linux
+and Mac OS version for many years now.
+
+* PEX
+* zipapp (doesn't include library dependencies, requires Python 3.5+)
```
ncoghlan: s/include/help manage/ (if you use `pip install --target` to add zip-compatible pure Python deps, `zipapp` will happily add them to the archive)

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+* shiv (requires Python 3)
+
+Depending on a new Python ecosystem
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Depending on the host system to have an alternative ecosystem
+installed, like Anaconda. Increasingly common in academic, analytical,
+and other data-oriented environments. Also used in production services.
+
+* conda/Anaconda
```
ncoghlan: Possible alternative structure here: rather than saying "new Python ecosystem", perhaps say "language independent packaging ecosystem", and mention:

* conda/Anaconda (cross-platform)
* homebrew (Mac OS X only)
* rpm (selected Linux distros only)
* deb (selected Linux distros only)

source/overview.rst
```diff
@@ -0,0 +1,248 @@
...
+* OpenStack
+
+Bringing your own hardware
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Depending on your host to have electricity.
+
+Embed your code on an Adafruit or a Micropython, or some other
+hardware, and just ship it to the datacenter, or your users' homes,
+and call it good.
```
ncoghlan: Hey, I resemble this remark :)

(DC fast chargers are a teensy bit bigger than a micro:bit, though)

ncoghlan:
If you have images you want to include, I think that could definitely be a good idea - I know a lot of folks found the layered diagrams useful for explain IaaS vs PaaS vs SaaS, and much of that applies here as well (at least on the application packaging side).

source/overview.rst
mahmoud: For sure! Definitely want lots of internal links, that's the reason for the season :)

source/overview.rst
mahmoud: Haha, good catch. Developing a new PaaS would at the very least be a separate document.

source/overview.rst
mahmoud: Yeah, web clients/browsers and mobile apps would definitely be the better explicit choice. I'll make that change.

source/overview.rst
mahmoud: Interesting, yeah, I'll try that out!

source/overview.rst
mahmoud: Yeah, the bullets in these sections are really anemic. I'll be sure to include more discussion of the pros and cons of each option. Even then, I might summarize by saying "minimal support" or something, because that is a really small subset of the libraries out there :)

source/overview.rst
mahmoud: Interesting. Yeah, I suppose there are Python applications out there that depend on RPM-packaged/managed libraries (python-*.i386.rpm, etc.). I'm not sure if I'd recommend it. I probably would want to give conda its own heading regardless, sort of like omnibus, below, because it does represent a pretty unique solution.

source/overview.rst
mahmoud: I need to find more projects like that to link to from here. I'm pretty confident we can find open-source exemplars of applications for all the solutions above, but do keep an eye out for good hardware applications.

mahmoud:
@ncoghlan Glad you like it, and thanks for the initial review! As far as pictures, I think you're right. A diagram is more helpful than the flavor illustrations of the original blog post. I'll probably make a version of this:

image

I'll ping back when I integrate these changes and make some progress on the TODOs.

source/overview.rst
mahmoud: I was kind of looking for a term that could encapsulate the conventions of Heroku/PaaS as well as stuff like Kivy/Beeware, where as long as you stay within the lines, stuff should just work.

mahmoud: I guess I could include a clarifications that not all frameworks include deployment systems, so evaluate carefully, especially if the option isn't on the list.

ncoghlan:
If you want more examples of "Anaconda-like" situations, you might consider situations like ESRi's ArcGIS Pro, Enthought Canopy, VFX Platform, etc.

VFX Platform in particular is an interesting example, since it's a platform spec that agrees on common versions of various dependencies rather than a particular implementation: http://www.vfxplatform.com/

mahmoud:
Just pushed a commit that addresses most of the points above. Still need to address my initial TODOs.

As far as Canopy and VFX Platform, I think the main thing I didn't find for those was an actual guide to making your own package for the platform. VFX is great, and kinda reminds me of manylinux, but without a reference/repo with the docker image for building and auditing.

Canopy is definitely like Anaconda in that it's a distribution, but all the docs I found were about using it as a consumer, not as a distributor of an original application.

Anaconda has [multiple tutorials](https://conda.io/docs/user-guide/tutorials/index.html) for building packages (applicable for both library and applications), which can, like OS package managers, depend on Postgres and other non-Python dependencies. Unlike OS package managers, it's OS agnostic and [has a tutorial](https://conda.io/docs/user-guide/tutorials/build-pkgs-skeleton.html) for smoothly transitioning from a Python-native ecosystem (PyPI), to the bigger pond of cross-language.

Like, I'm tempted to include it in a more comprehensive OS-native distribution section, but conda really is a pretty unique case, especially taking into account its particular mutual affinity with Python community.

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+* Is your software installed individually, or to many computers at once?
+
+Packaging is all about target environment and deployment
+experience. There are many answers to the questions above and each
+combination of circumstances has its own solutions.
+
+Packaging libraries and tools
+-----------------------------
+
+You may have heard about PyPI and ``setup.py`` and wheel files. These
```
pradyunsg: nit: two "and", first can become a comma.

source/overview.rst
```diff
@@ -0,0 +1,256 @@
+================
+Packaging Python
+================
+
+Python is a general-purpose programming lanaguage, meaning you can use
```
pradyunsg: nit: language

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+Python's steady advances are leading it into new spaces. These days
+you can write a mobile app or web application frontend in
+Python. While the language may be familiar, the packaging and
+deployment practices are brand new.
+
+If you're planning on releasing to these new frontiers, you'll want to
+check out the following frameworks, and refer to their packaging
+guides:
+
+* Kivy
```
pradyunsg: Should we link to these project's pages?

mahmoud: Definitely! I actually have quite a few projects I still haven't mentioned, but when I get a spare hour or two, I plan on going through and hyperlinking everything :)

source/overview.rst
```diff
@@ -0,0 +1,256 @@
+================
+Packaging Python
```
theacodes: Could you make the title something like "An overview of Python packaging"? It makes it better for linking and SEO.

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+================
+
+Python is a general-purpose programming language, meaning you can use
+it for many things. You can build robots or server software or a game
+for your friends to play.
+
+For this reason, the first step in every Python project must be to
```
theacodes: Nit: this paragraph should merge with the one above (just delete the newlines).

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+think about the project's audience and the corresponding target
+environment. Using this information, this overview will guide you to
+the packaging technologies best suited to your project.
+
+It might seem strange to think about packaging before writing code,
+but this process does wonders for avoiding headaches later on.
+
+* Who are your software's users? Are they other developers doing
```
theacodes: Introduce this list with something like:

"Some of the questions you may need to ask to determine your packaging strategy are:"

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+provided one of those files is named ``__init__.py``, comprises an
+:term:`import package`.
+
+Because packages consist of multiple files, they are harder to
+distribute. Most protocols support transferring only one file at a
+time (when was the last time you clicked a link and it downloaded
+multiple files?). It's easier to get incomplete transfers, and harder
+to guarantee code integrity at the destination.
+
+So long as your code contains nothing but pure Python code, and you
```
theacodes: Hmm, this is a bit weird. Even packages with native components have source distributions, and conversely, pure-Python packages can still have built distributions (wheels).

source/overview.rst
```diff
@@ -0,0 +1,256 @@
...
+libxml2 in the case of lxml, or BLAS libraries in the case of numpy),
+you will want to read on.
+
+.. TODO: "Did you know?" about distributions providing multiple
+   versions of the same package. Python packaging superpower!
+
+Python binary distribution
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Python's real power comes from its ability to integrate with the
```
theacodes: "Python's real power" can read as dismissive. I'd recommend "A great feature of Python is its ability to..."

mahmoud: Totally. This section is very underdeveloped, but I've adjusted this in my most recent commit..

pradyunsg:
This PR has been idle for a month. Is there anything I can do to help get this moving again; by contributing content or otherwise?

pradyunsg:
No issues @mahmoud. :)

Thanks for making the time to work on this again.

mahmoud:
I daresay this PR is finally done! I only imported diagrams, not illustrations, and there are a couple ideas for sections commented, but otherwise the overview reads and looks great on my local. Added a link to the home page, and I think this is ready to merge. Let me know if there are any requested changes. Otherwise, very much looking forward to seeing this go live!

theacodes:
Excellent! I will try to review this tomorrow. Thank you so much!

source/overview.rst
```diff
@@ -0,0 +1,419 @@
...
+technologies, and a good amount of effort.
+
+A selection of Python freezers:
+
+* `pyInstaller <http://www.pyinstaller.org/>`_ - Cross-platform
+* `cx_Freeze <https://anthony-tuininga.github.io/cx_Freeze/>`_ - Cross-platform
+* `constructor <https://github.com/conda/constructor>`_ - For command-line installers
+* `py2exe <http://www.py2exe.org/>`_ - Windows only
+* `py2app <https://py2app.readthedocs.io/en/latest/>`_ - Mac only
+* `bbFreeze <https://pypi.python.org/pypi/bbfreeze>`_ - Windows, Linux, Python 2 only
```
gtback: Trivial thing: should be https://pypi.org/project/bbfreeze/ (though this URL will redirect correctly)😁 

P.S. Very excited about having this guide!

mahmoud:
@gtback, dangit, I got all of them but one! Good catch.

Fixed, with some additional improvements as well. No TODOs left in the doc :)

source/overview.rst
```diff
@@ -89,11 +89,19 @@ If you rely on any non-Python code, or non-Python packages (such as
...
+.. note:: Python and PyPI support multiple distributions providing
+   different implementations of the same package. For instance the
+   unmaintained-but-seminal `PIL distribution
+   <https://pypi.org/project/PIL/>`_ provides the PIL package, and so
+   does `Pillow <https://pypi.org/project/Pillow/>`_, an
+   actively-maintained fork of PIL!
+
+   This Python packaging superpower makes it possible for Pillow to be
+   a drop-in replacement for PIL, just by changing your project's
+   `q`install_requires`` or ``requirements.txt``.
```
pradyunsg: There sees to be a typo here.

theacodes:
Preview at d366f79 staged here: http://temp.theadora.io/pypug/overview.html if anyone wants to look at it.

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+files. These are just a few of the tools Python's ecosystem provides
+for distributing Python code to developers, which you can read about in
+:doc:`guides/distributing-packages-using-setuptools`.
+
+The following approaches to packaging are meant for libraries and
+tools used by technical audience in a development setting. If you're
+looking for ways to package Python for a non-technical audience and/or
+a production setting, skip ahead to :ref:`packaging-applications`.
+
+Python modules
```
theacodes: This section reads a bit bare and doesn't quite mesh with the context. What are we trying to convey to the user here? What action do we expect them to take with this information?

mahmoud: So this section is under two pressures:

1. Python developers from their first `pip install` are steeped in `setup.py` and "for-other-developers" packaging. I vividly remember the times I really struggled to make the mental leap from "I should build an egg" to "I should build an RPM" for my production application. It wasn't until much later that I realized the  deployment model and even the audience were completely different. So this section is trying to place these technologies in the decision tree, as early and separate as possible from the applications. Without it, the immediate question becomes "well, what about wheels/sdists/pip/eggs/etc."?
2. The section is on the bare side (even shorter than [the sections in the initial post](http://sedimental.org/the_packaging_gradient.html#prelude-the-humble-script)) because while I want to acknowledge the technologies, the rest of the site is a much better guide. I wanted it to be more content-light and link-heavy.

If you think it could stand do be expanded a bit more without overshadowing more extensive writings on the rest of the guide, I'd be happy to do what I can. Alternatively, as someone who knows the guide much better than me, I would really welcome more links to things that need linking!

source/overview.rst
mahmoud: Fixed, thanks :)

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+* `Flexx <http://flexx.readthedocs.io/en/latest/>`_
+
+If you are *not* interested in using a framework or platform, or just
+wonder about some of the technologies and techniques utilized by the
+frameworks above, continue reading below.
+
+Depending on a pre-installed Python
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Pick an arbitrary computer, and depending the context, there's a very
```
ncoghlan: Typo: missing "on" after "depending"

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+
+For a long time many operating systems, including Mac and Windows,
+lacked built-in package management. Only recently did these OSes gain
+so-called "app stores", but even those focus on consumer applications
+and offer little for developers.
+
+Developers long sought remedies, and in this struggle, emerged with
+new their own package management solutions -- with some notable
+benefits for Python developers in particular. The most prominent, an
+alternative package ecosystem called Anaconda is built around Python
```
ncoghlan: For this doc, I think the name of the tool ("conda") would be a better fit than the name of the company (I know it's ambiguous, since "Anaconda" is the name of both the company and the base package set installed, so I won't push against it if you prefer to keep the full distribution name).

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+          this also makes for the smallest package, as small as
+          single-digit megabytes, or even kilobytes.
+
+          In general, decreasing the dependency on the target system
+          increases the size of our package, so the solutions here
+          are roughly arranged by increasing size of output.
+
+.. _depending-on-a-new-python:
+
+Depending on a new Python ecosystem
```
ncoghlan: It would be good to adjust the title here, as "new" is ambiguous: it could refer to "new-to-the-target-environment" or "new-to-the-world-at-large" (and neither is entirely accurate for the scope of the section anyway).

My suggestion would be "Depending on another software distribution ecosystem". That way if we later decide to add other options like Nix or homebrew in here, they can just be an extra paragraph at the end of this section, rather than needing their own section.

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+
+Bringing your own hardware
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The most all-encompassing way to ship your software would be to ship
+it already-installed on some hardware. This way, your software's user
+would require only electricity.
+
+Whereas the virtual machines described above are primarily reserved
+for the tech-savvy, hardware appliances used by the most advanced data
```
ncoghlan: Typo: missing "are" before "used"

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+What about...
+-------------
+
+The sections above can only summarize so much, and you might be wondering about some of the more conspicuous gaps.
+
+Operating system packages
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+As mentioned in :ref:`depending-on-a-new-python` above, some operating
+have package managers of their own. If you're very sure of the
```
ncoghlan: Typo: missing "systems" in "operating systems"

This cross-ref will also need to change if that section title changes.

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+operating system you're targeting, you can depend directly on a format
+like `deb <https://en.wikipedia.org/wiki/Deb_(file_format)>`_ (for
+Debian, Ubuntu, etc.) or `RPM
+<https://en.wikipedia.org/wiki/RPM_Package_Manager>`_ (for Red Hat,
+Fedora, etc.), and use that built-in package manager to take care of
+installation, and even deployment.
```
ncoghlan: It would be good to directly mention FPM here, since that can take a fair bit of pain out of converting Python projects into Linux system packages: https://fpm.readthedocs.io/en/latest/source/virtualenv.html

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+<https://www.google.com/search?channel=fs&q=static+vs+dynamic+linking>`_.
+
+Wrap up
+-------
+
+Packaging in Python has a bit of a reputation for being a bumpy
+ride. This impression is mostly a byproduct of Python's
+versatility. Once you understand the natural boundaries between each
+packaging solution, you begin to realize that the varied landscape is
+a small price Python programmers pay for using the most balanced,
```
ncoghlan: I'd tweak the praise of the language in this last sentence to be something unequivocally true. Something like:

> ... for using one of the most balanced, flexible, and broadly applicable languages available.

(The "one of" does the equivocation work there: "most" is debatable, "one of the most" is not)

source/overview.rst
mahmoud: good catch! wrt the crossref, I think I made an explicit heading anchor ref thing in this case.

source/overview.rst
mahmoud: (working from bottom to top has its disadvantages, I see what you mean from down below)

source/overview.rst
mahmoud: yeah, used to be conda was the tool, Anaconda was the ecosystem, and Continuum was the company. Now I still think it's correct to say Anaconda to refer to the ecosystem, but I'll check with Kale and see if this needs a touch up later.

source/overview.rst
mahmoud: Thanks!

source/overview.rst
mahmoud: fixed!

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+redistributed and reused. You will also need to ensure it's written
+for the right version of Python.
+
+Python source distributions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If your code consists of multiple Python files, it's usually organized
+into a directory structure. Any directory containing Python files,
+provided one of those files is named ``__init__.py``, comprises an
+:term:`import package`.
```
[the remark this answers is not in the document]
mahmoud: I've removed the qualifier, but it's kind of tricky. Maybe I'm getting philosophical here, but a directory without an expression of intent like `__init__.py`  doesn't necessarily become an import package until someone imports it. And that was more a choice of the importer than the publisher.

I think I'll always prefer to the `__init__.py` for explicit package creation to avoid this sort of quandary. :)

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+-------------------------
+
+Packages exist to be installed (or *deployed*), so before you package
+anything, you'll want to have some answers to the deployment questions
+below:
+
+* Who are your software's users? Will your software be installed by
+  other developers doing software development, operations people in a
+  datacenter, or a less software-savvy group?
+* Is your software intended to run on servers, desktops, or embedded
```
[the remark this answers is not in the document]
mahmoud: sure!

source/overview.rst
mahmoud: Absolutely! Added, with a link to bottle, and boltons, only because I specifically talk about this [in the docs](http://boltons.readthedocs.io/en/latest/architecture.html#architecture).

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+   unmaintained-but-seminal `PIL distribution
+   <https://pypi.org/project/PIL/>`_ provides the PIL package, and so
+   does `Pillow <https://pypi.org/project/Pillow/>`_, an
+   actively-maintained fork of PIL!
+
+   This Python packaging superpower makes it possible for Pillow to be
+   a drop-in replacement for PIL, just by changing your project's
+   ``install_requires`` or ``requirements.txt``.
+
+Python binary distributions
```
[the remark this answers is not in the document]
mahmoud: Yeah, I also want to recommend universal wheels for pure-Python packages. That said, in trying to "ramp" users based on their needs, I also don't want to overwhelm them. I've put a couple sentences in to tie the two together as best practice, but more technical discussions should probably be on the linked article (tutorials/packaging-projects.rst). Maybe an aside to explicitly recommend doing both could work?

source/overview.rst
mahmoud: Heh, I'd take that debate anyday! But sure, not trying to make any enemies of anyone who read to the bottom of the page :)

source/overview.rst
mahmoud: Added

mahmoud:
OK, made some changes which I think mostly address the concerns raised. 

@ncoghlan, regarding your big point, I'm OK expanding the packaging-libraries section somewhat, I just want to avoid having a big technical bump in the middle covering a topic that's really well-covered by links to the rest of the guide. Take a look at the sentences added in https://github.com/pypa/python-packaging-user-guide/pull/519/commits/8d3aae2433c5aa369b683cd8be65071ee8c62434, and if those aren't enough, would you mind writing a short aside (something I can put in a `.. note:`) as mentioned in the comment above?

source/overview.rst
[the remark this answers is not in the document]
ncoghlan: Yeah, I default to `__init__.py` as well, and there are still plenty of tools that only handle explicit import packages by default.

Perhaps we could say something like:

> Any directory containing Python files can optionally be accessed at runtime as an :term:`import package`, rather than as a collection of independent scripts (and if one of the files in the directory is named ``__init__.py``, it makes it explicit that the submodules are intended to be accessed via the package rather than directly).

source/overview.rst
```diff
@@ -0,0 +1,445 @@
...
+          this also makes for the smallest package, as small as
+          single-digit megabytes, or even kilobytes.
+
+          In general, decreasing the dependency on the target system
+          increases the size of our package, so the solutions here
+          are roughly arranged by increasing size of output.
+
+.. _depending-on-a-new-ecosystem:
+
+Depending on a new software distribution ecosystem
```
ncoghlan: "new" is still a bit ambiguous here ("new" in general? "new" to you? "new" to your users?).

Perhaps "a new" -> "another"?

mahmoud: Oops, meant to fix that one before, fixed now!

source/overview.rst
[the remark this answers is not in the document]
mahmoud: Hmm, I guess my main concern there is that right now the document is organized by "class of distributable artifact, and what it is capable of distributing". If I'm understanding correctly, this would add a section about wheels, which may be a confusing redundancy.

Another point is that `sdist + wheel` is absolutely a best practice on PyPI, but not necessarily so in other packaging workflows. I've used these formats "off PyPI" and always sdist only or wheel only. I'm not sure having both would have done much in those other packaging approaches. So in the interest of maintaining the introductory tone, I think it helps to introduce the formats and their intended purposes in this overview, then the best practices in the practical docs. (If someone manages to figure out how to push to PyPI from the overview alone, I'll be very surprised!)

source/overview.rst
```diff
@@ -0,0 +1,429 @@
...
+the coming sections.
+
+Depending on a framework
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some types of Python applications, like web sites and services, are
+common enough that they have frameworks to enable their development
+and packaging. Other types of applications, like web and mobile
+clients, are advanced enough that a framework becomes more than a
+convenience.
```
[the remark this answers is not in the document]
mahmoud: almost missed this one, too! just fixed. (hopefully it doesn't seem like I'm ignoring anything here, GitHub just loves collapsing things)

source/overview.rst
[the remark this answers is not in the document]
ncoghlan: I don't see "class of distributable artifact" as the main organising principle in the current overview - I see it as being primarily organised by target audience, and then working back from that to the appropriate artifacts to use.

And when the target audience is "other Python developers", the appropriate artifacts to publish are both an sdist and a wheel archive.

The only time it makes sense to choose one or the other exclusively is when the target audience is "next stage in a specific automated build pipeline", and I don't think that's a scenario the overview really needs to emphasise.

At most, that use case feels like a potential side note along the lines of "Default to publishing both sdist and wheel archives, *unless* you're creating artifacts for a very specific use case where you know the recipient only needs one or the other".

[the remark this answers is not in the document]
mahmoud: I liked that last line, incorporated it in one of the recent commits. :)

mahmoud:
Just finished adding a few more tweaks based on feedback, and I think in many ways this document is actually a lot more comprehensive and broadly-targeted than [the original post](https://sedimental.org/the_packaging_gradient.html), in all the best ways. I think it's safe to say at this point I think it is/I am done :) 

For the moment, I've got a (my) wedding to attend to, so I don't think I'll have much more time to work on it. I think even for any of its remaining, arguable flaws, having the overview available on the primary site would help a lot more than it would hurt, so I would encourage publishing it in its current state and tweaking it further in the form of future PRs. Thanks again for all the feedback, everyone!

theacodes:
LGTM, too! @ncoghlan if you have any further concerns please let me know (or file bugs) and I'll be happy to fix them in follow-up PRs.

@mahmoud this is monumental work, thank you so much!

ncoghlan:
https://github.com/pypa/python-packaging-user-guide/commit/8d3aae2433c5aa369b683cd8be65071ee8c62434 and https://github.com/pypa/python-packaging-user-guide/pull/519/commits/88a1bc98e709102a54f08e2842f8251822680d76 cover my wheel vs sdist concerns without requiring major changes to that section, so +1 for that approach :)

Thanks again @mahmoud for this incredibly useful addition!

PR #525, opened 2018-06-14
updating the information on the doc related to the correct url

waldirio:
Hi,

Actually following the docs I got error, after do a simple research I got a lot of cases related to the same issue and the fix is just related to the url, just changing according below

from
---
twine upload --repository-url https://test.pypi.org/legacy/ dist/*
---

to
---
twine upload --repository-url https://upload.pypi.org/legacy/ dist/*
---

I believe this change will upload the online docs and to the next *new users* which try to upload their packages, everything will be fine.

Thank you.

### DIFF
--- a/source/guides/using-testpypi.rst
+++ b/source/guides/using-testpypi.rst
@@ -28,7 +28,7 @@ in the ``--repository-url`` flag
 
 .. code::
 
-    $ twine upload --repository-url https://test.pypi.org/legacy/ dist/*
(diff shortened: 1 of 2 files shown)


theacodes:
Hi @waldirio! This is working as intended. The page you changed is for *Test PyPI*, so the url is correct. Merging this change would mean that users would mistakenly upload to the *real* PyPI.

PR #531, opened 2018-07-03
Improve binary extension docs

thewtex:


theacodes:
Thanks, @thewtex!

thewtex:
Thanks for the lightning quick review and integration @theacodes ! :zap:

dhermes:
https://packaging.python.org/guides/packaging-binary-extensions/ still says "Last Reviewed: 2013-12-08". Need to update?

theacodes:
@dhermes probably. We've been removing those things from pages, so it'd be nice if someone sent a PR to do that to this page.

dhermes:
I feel the nudge to send the PR :grinning: but I like those `:Page Status:` :Last Reviewed:` extras so it'd be sad if I gave them the ax. (Though I understand why they'd get stale.)

Issue #529, opened 2018-07-01
example setup.py references nonexisting github URL


gitqlt:
https://packaging.python.org/tutorials/packaging-projects/, Creating setup.py suggests the
```
url="https://github.com/pypa/example-project"
```
which does not exist.

aritro01:
Can I have this?

ncoghlan:
Done, thanks!

PR #533, opened 2018-07-10
Updating the nonexistent URL in example setup.py references 

aritro01:
Updated the nonexistent URL addressed in issue #529


ncoghlan:
Thanks!

PR #538, opened 2018-08-13
Update Python version supported by virtualenv

cdunklau:
Virtualenv 16.0.0 dropped support for Python 2.6
https://virtualenv.pypa.io/en/stable/changes/


theacodes:
Thank you!

PR #540, opened 2018-08-13
Move the Overview to a task based lead in along with the others

dstufft:
This moves the overview that was added in #519 from a dedicated call out at the start of the "Get Started" section, into the list of task based documents.

The overview document is not the globally correct choice for users hitting this page who want to get started-- often these users will be looking for a simple tutorial that just tells them what to do rather than a deep dive into all of the various options they *could* pick. The new location brings these different tasks ("I want an overview" vs "I want to package software" vs "I want to manage dependencies for an application" vs "I want to install software") back into alignment with each other, and generally fits the overall feel of python.packaging.org better.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -33,8 +33,6 @@ happily accept any :doc:`contributions and feedback <contribute>`. 😊
 Get started
 ===========
 
-Thinking about shipping a new Python library or application? Start with :doc:`overview`.
-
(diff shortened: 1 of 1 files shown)


dstufft:
Wording suggests on this appreciated. I'm not the greatest at wording things.

pfmoore:
LGTM

PR #541, opened 2018-08-15
Refine overview lead in

mahmoud:
While I appreciate the expeditious velocity in #540, I agree with @dstufft that the wording needs some work. The overview does not cover "Python's packaging options", but rather "options for packaging Python libraries and applications". In fact, most of the options do not originate from official Python efforts or groups.

TLDR: "an overview of Python's packaging options" -> "an overview of packaging options for Python libraries and applications"


mahmoud:
Sigh, I see commit squashing has taken its toll. I'll freshen this up.

PR #545, opened 2018-08-15
Fix tutorials/packaging-project's twine example

Great-Li-Xin:
In some operations systems such as mine (macOS Mojave 10.14 Beta (18A365a)), after installing twine the system can not find twine unless you call it explicatively.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -249,7 +249,7 @@ Once installed, run Twine to upload all of the archives under :file:`dist`:
 
 .. code-block:: bash
 
-    twine upload --repository-url https://test.pypi.org/legacy/ dist/*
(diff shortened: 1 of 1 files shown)


di:
@Great-Li-Xin That shouldn't be the case. Can you give us some more details about how you have Python installed on that platform or if you're using the system Python? Exactly how did you install twine?

Great-Li-Xin:
@di I installed twine through pip. And it was **NOT** the system python. Precisely, the python I was using is the python 3.7 downloaded from [the python website](https://www.python.org). 
### Twine installed through the following command below:
``` bash
pip3 install twine
```

theacodes:
I also think that using `python3 -m {x}` is a bit more explicit and is generally immune to issues where the user base isn't in `path`, so I'm going to merge this.

Thank you, @Great-Li-Xin!

Issue #548, opened 2018-09-03
https://packaging.python.org/tutorials/installing-packages/#


martinlarsalbert:
Maybe mention that windows cmd uses double quotes:" instead of: '

Linux:
`pip install 'SomeProject==1.4'`

Windows cmd:
`pip install "SomeProject==1.4"`

ncoghlan:
Double quotes work on Linux (etc) as well, so if single quotes don't work on Windows, it will likely make more sense to just switch to double quotes for every example that isn't using a wildcard in the string.

pradyunsg:
IOW, @ncoghlan is saying we'll be happy to accept a PR for this, if someone comes around to this before one of the editors can. :)

PR #550, opened 2018-09-06
Changed to double quotes in the pip install SomeProject==1.4

martinlarsalbert:
since it is the only thing working in Windows cmd. And @ncoghlan said that this will also work in Linux. Resolving #548


theacodes:
Thank you!

PR #553, opened 2018-09-15
Improve `packages` and `py_modules` wording

jwodder:
I felt the wording of the sections on `packages` and `py_modules` could be improved.  I also felt the two sections belonged right next to each other.


theacodes:
Thanks again, @jwodder!

PR #562, opened 2018-10-04
Mention not using Sphinx extensions in "Making a PyPI-friendly README"

wcooley:
I thought I was doing the right thing by including [Sphinx cross-references](http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects)
in my `README.rst` and through that in my package's `long_description`, but only after
several failed validations and searching did I learn that Sphinx extensions are not allowed.

The error message from `twine check` (or `rst-lint`) is not obvious about what is wrong: e.g.,
'Error: Unknown interpreted text role "py:func".' and to those just figuring out Sphinx and rST
the distinction is not necessarily obvious.

(I am submitting this PR against "master" because the other 2 open PRs do that; I read the contributing guide but it gave no advice about which branch to use.)

### DIFF
--- a/source/guides/making-a-pypi-friendly-readme.rst
+++ b/source/guides/making-a-pypi-friendly-readme.rst
@@ -14,7 +14,7 @@ For your README to display properly on PyPI, choose a markup language supported 
 Formats supported by `PyPI's README renderer <https://github.com/pypa/readme_renderer>`_ are:
 
 * plain text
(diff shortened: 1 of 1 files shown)


theacodes:
Thanks, @wcooley!

webknjaz:
@wcooley please also note that the opposite might be useful: you can include parts of your readme into sphinx docs and it might be useful to add some markers in README, which won't be rendered :)

PR #565, opened 2018-10-13
Update link to __about__ example

ssaamm:
This file has been removed, but we could still link to the historical version.


CMeza99:
FWIW, I agree with this change. ATM, the documentation has a broken link. This PR would resolved that and point to the correct data in Git history.

theacodes:
Thanks, @ssaamm!

PR #573, opened 2018-11-15
Add example for single-sourcing pgk ver using pkg_resources

CMeza99:
Add `setuptools` as an example of single-sourcing package versioning using `pkg_resources`.


theacodes:
Thanks!

PR #574, opened 2018-11-15
Readme as reST instead of MD

CMeza99:
README.md should be README.rst because:
* [The markup used for the Python documentation is reStructuredText](https://devguide.python.org/documenting/)
* README.rst is a file distutils will automatic recognize for sdist, not README.md. https://docs.python.org/3/distutils/sourcedist.html#specifying-the-files-to-distribute

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -53,7 +53,7 @@ them in the following steps.
         __init__.py
       setup.py
       LICENSE
-      README.md
+      README.rst
 
 
 Creating setup.py
@@ -70,7 +70,7 @@ the values if you want:
 
     import setuptools
 
-    with open("README.md", "r") as fh:
+    with open("README.rst", "r") as fh:
         long_description = fh.read()
 
     setuptools.setup(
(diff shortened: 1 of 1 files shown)


jwodder:
Your first link is about the documentation that makes up docs.python.org.  Documentation for third-party Python packages is free to be in any format that it wants, and PyPI has supported Markdown in project descriptions for several months now.  [Setuptools has also recognized `README.md` files for about the same amount of time](https://github.com/pypa/setuptools/pull/1133); quite frankly, who cares what distutils recognizes?

di:
I agree with @jwodder, support for Markdown is on-par with reStructuredText (with the exception of distutils which is frozen in time and shouldn't be used directly).

Furthermore, the fact that this guide suggests using Markdown [was not a mistake](https://github.com/pypa/python-packaging-user-guide/pull/498) -- it's much more beginner-friendly than reStructuredText. You are of course still free to use rST for your project, but since this guide is for folks unfamiliar with packaging, recommending Markdown is ideal in my opinion.

theacodes:
Yep, the tutorial is written as intended. Thank you for taking the time to write up a pull request for this, but I don't think it's the right thing to merge at this time.

PR #575, opened 2018-11-20
Remove landing page link to PyPI migration guide

pradyunsg:
The migration is complete. ;)

I've let the guide stay, as a historical record. I don't think it'd serve much use though now, since the migration is over, so I'll be happy to delete the file if that's what we want to do.


theacodes:
Thanks!

PR #576, opened 2018-11-25
python3: virtualenv module is now venv

divyekapoor:
Updating the docs to match Python 3.7.1 on Mac.
See below for proof of change.

Divyes-MacBook-Air:~ divyekapoor$ python3 --version
Python 3.7.1
Divyes-MacBook-Air:~ divyekapoor$ python3 -m virtualenv
/usr/local/opt/python/bin/python3.7: No module named virtualenv
Divyes-MacBook-Air:~ divyekapoor$ python3 -m venv
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear]
            [--upgrade] [--without-pip] [--prompt PROMPT]
            ENV_DIR [ENV_DIR ...]
...

### DIFF
--- a/source/guides/installing-using-pip-and-virtualenv.rst
+++ b/source/guides/installing-using-pip-and-virtualenv.rst
@@ -104,13 +104,15 @@ On macOS and Linux:
 
 .. code-block:: bash
 
-    python3 -m virtualenv env
+    python3 -m venv env
+    OR
+    python -m virtualenv env
 
 On Windows:
 
 .. code-block:: bash
 
-    py -m virtualenv env
+    py -m venv env
 
 The second argument is the location to create the virtualenv. Generally, you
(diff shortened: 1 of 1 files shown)


jwodder:
[virtualenv](https://virtualenv.pypa.io/) and [venv](https://docs.python.org/3/library/venv.html) are two separate modules, and the page you've edited is explicitly about using virtualenv.  virtualenv has not become venv; virtualenv, unlike venv, must be installed with pip before you can use it, [just as the guide says](https://packaging.python.org/guides/installing-using-pip-and-virtualenv/#installing-virtualenv).

divyekapoor:
John: Thanks for the context. Dropping the pull request.
Cheers!

pradyunsg:
Can someone close this PR?

PR #581, opened 2018-12-21
Reference twine from Python3

ChrisRackauckas:
pip doesn't necessarily add it to the path.


theacodes:
Thank you, @ChrisRackauckas!

PR #583, opened 2018-12-31
Add more clarity to the packaging tutorial dir structure

dogweather:
The tutorial begins by instructing the reader to create this confusing hierarchy without explanation:

```
example_pkg/
    example_pkg/
        __init__.py
```

The tutorial then continues on, referring to `example_pkg` frequently, but not clarifying _which_ one is meant. Confusing! And not necessary, in my opinion. So this PR renames the top-level dir, better describing its purpose:

```
packaging_tutorial/
    example_pkg/
        __init__.py
```

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -48,7 +48,7 @@ them in the following steps.
 
 .. code-block:: text
 
-    /example_pkg
+    /packaging_tutorial
       /example_pkg
         __init__.py
       setup.py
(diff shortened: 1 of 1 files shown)


ncoghlan:
@theacodes This change seems like a good idea to me - the current structure came from the fact that in a typical single-package project the VCS repo and the top-level Python import package share a name (so you end up with this doubly-nested layout), but for someone running through the tutorial in a throwaway directory, that reasoning doesn't apply.

theacodes:
Agreed. Thanks for doing this, @dogweather!

PR #595, opened 2019-02-01
Update packaging-projects.rst

iurisegtovich:
The tutorial tells the reader not to forget to have a LICENSE file but, in the end, the LICENSE file was not shipped.
IMHO, an extra line to the minimalist setup.py ensures consistency between what is said and what is done.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -87,6 +87,7 @@ Open :file:`setup.py` and enter the following content. You **should** update the
             "License :: OSI Approved :: MIT License",
             "Operating System :: OS Independent",
         ],
+        data_files = [("", ["LICENSE"])],
     )
 
 


jwodder:
I believe the *de facto* standard way to include a license file in a wheel is via the `license_file` `[metadata]` option in `setup.cfg`, not via `data_files`.

iurisegtovich:
thank you, I didn't know about that, I just started reading about
packaging. do you think it's worth to tell about this configuration files
in the minimalist tutorial?

theacodes:
I think keeping this as minimal as possible is certainly ideal. @ncoghlan WDYT?

ncoghlan:
The sample project actually uses the approach @jwodder describes: https://github.com/pypa/sampleproject/blob/master/setup.cfg

So updating the tutorial to include a `setup.cfg` file that specifies the license file to ship would probably provide a better on-ramp to the full guide at https://packaging.python.org/guides/distributing-packages-using-setuptools/

The obvious downside of doing that is that it means burdening the tutorial with details on why the information is split across the two files, but the upside of explaining it is that it's a question that's inevitably going to come up when using `setuptools` in a real project anyway.

PR #596, opened 2019-02-01
Improve sentence.

porridge:
Add "according to" for the sentence to make more sense.


ncoghlan:
Thanks! I'm guessing some words got lost in an edit somewhere along the way :)

PR #602, opened 2019-02-27
Elaborate on how the tutorial differs from the real packaging process

alchemyst:
This addresses [this issue](https://github.com/pypa/packaging-problems/issues/246). It is meant to help people who want to use the tutorial as a scaffold for getting their own package onto PyPI to understand what steps are required and which ones are just for the tutorial.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -95,7 +95,7 @@ minimal set:
 
 - ``name`` is the *distribution name* of your package. This can be any name as long as only
   contains letters, numbers, ``_`` , and ``-``. It also must not already
-  taken on pypi.org. **Be sure to update this with your username,** as this ensures you won't try to upload a package with the same name as one which already exists when you upload the package.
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -63,7 +63,7 @@ Creating setup.py
 about your package (such as the name and version) as well as which code files
 to include.
 
-Open :file:`setup.py` and enter the following content. You **should** update the package name to include your username (for example, ``example-pkg-theacodes``. You can personalize the other values if you'd like:
+Open :file:`setup.py` and enter the following content. To avoid lots of projects with the same name from people doing this tutorial, you **should** update the package name to include your username (for example, ``example-pkg-theacodes``). You can personalize the other values if you'd like:
```
theacodes: Maybe phrase this as `Update the package name to include your username (for example, ``example-pkg-theacodes``), this ensures that you have a unique package name and that your package doesn't conflict with packages uploaded by other people following this tutorial.`

alchemyst: ✅

theacodes: Hrm, I think you updated the wrong one. The one below should remain short (just a reminder of the instructions here).

alchemyst: Sorry, should be fixed now.

PR #605, opened 2019-03-28
SIngle-sourcing the package version: Remove a py2-only option; simplify a py3 option

paulmelnikow:
Thanks for this guide! I was using this and tried `execfile()` without realizing it didn't work in Python 3.

I thought about leaving the Python 2 version here with a disclaimer, though this list is already very long, and it seems more helpful to steer people toward a good solution than to exhaustively document an obsolete solution.

I simplified the Python 3 option to a version which seems equally safe without the context handler (it seems the file handle should be closed when it goes out of scope), which should be immediately upon return.

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -48,16 +48,6 @@ number of your project:
     your project (e.g. :file:`version.py`), then have :file:`setup.py` read and
     ``exec`` the value into a variable.
 
-    Using ``execfile``:
-
-    ::
-
-        execfile('...sample/version.py')
-        # now we have a `__version__` variable
-        # later on we use: __version__
-
-    Using ``exec``:
-
     ::
 
         version = {}
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -51,8 +51,7 @@ number of your project:
     ::
 
         version = {}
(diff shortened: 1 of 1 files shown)


source/guides/single-sourcing-package-version.rst
```diff
@@ -48,21 +48,10 @@ number of your project:
...
-        # later on we use: __version__
-
-    Using ``exec``:
-
     ::
 
         version = {}
-        with open("...sample/version.py") as fp:
-            exec(fp.read(), version)
+        exec(open("...sample/version.py").read(), version)
```
theacodes: Could you leave this example as-is?

paulmelnikow: Out of curiosity, is there an advantage to the longer context-handler form?

pfmoore: The `exec(open(...).read(), ...` version relies on the garbage collector to close the file, whereas the context manager closes it deterministically.

webknjaz: FTR There's also `Pathlib.Path.read_text()` but it requires python 3.5 or so

webknjaz:
```suggestion
        with open("...sample/version.py") as fp:
            exec(fp.read(), version)
```

theacodes:
Thanks, @paulmelnikow!

PR #606, opened 2019-04-20
reduce emphasis on virtualenv

cs01:
* Replace use of `virtualenv` with `virtual environment` when referring to a virtual environment 
  * Only use the term `virtualenv` when referring to the tool. 
* Recommend venv over virtualenv. Make it clear that venv is preferred for Python 3.

### DIFF
--- a/source/guides/index.rst
+++ b/source/guides/index.rst
@@ -9,7 +9,7 @@ introduction to packaging, see :doc:`/tutorials/index`.
    :maxdepth: 1
 
    tool-recommendations
-   installing-using-pip-and-virtualenv
+   installing-using-pip-and-virtual-environments
    installing-stand-alone-command-line-tools
    installing-using-linux-tools
    installing-scientific-packages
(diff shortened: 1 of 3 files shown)


ncoghlan:
@theacodes This looks good to me, but the file name change has the potential to break incoming links.

Do you still have something set up in nox to detect & add the redirects to deal with that? (I remember we had to turn it off in CI because it had trouble coping with new sections being added to existing pages)

theacodes:
Yes, this will definitely break links. 

I can either add redirects after this is merged or we can keep the old file name. I have no preference.

I turned off the link checker when we stopped having so much churn.

ncoghlan:
I think the canonical URL change makes sense, so if you're happy to add the redirects, let's go ahead!

ncoghlan:
It helps if I push the *right* button...

theacodes:
Cool, docs have been built and the redirect is in place. Thanks, everyone!

PR #607, opened 2019-04-24
[pipx] replace ensurepath with userpath

cs01:
pipx is now relying on the PyPI package [userpath](https://pypi.org/project/userpath/) and has deprecated its `ensurepath` command.

### DIFF
--- a/source/guides/installing-stand-alone-command-line-tools.rst
+++ b/source/guides/installing-stand-alone-command-line-tools.rst
@@ -24,8 +24,8 @@ packages, and allows you to safely run the program from anywhere.
 
 ::
 
-  $ pip install --user pipx
-  $ userpath ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
+  $ python3 -m pip install --user pipx
+  $ python3 -m userpath ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
 
 .. Note:: You may need to restart your terminal for the path updates to take effect.
 
--- a/source/guides/installing-stand-alone-command-line-tools.rst
+++ b/source/guides/installing-stand-alone-command-line-tools.rst
@@ -25,7 +25,7 @@ packages, and allows you to safely run the program from anywhere.
 ::
 
   $ pip install --user pipx
-  $ pipx ensurepath  # ensures the path of the CLI application directory is on your $PATH
+  $ userpath ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
 
 .. Note:: You may need to restart your terminal for the path updates to take effect.
 
--- a/source/guides/installing-stand-alone-command-line-tools.rst
+++ b/source/guides/installing-stand-alone-command-line-tools.rst
@@ -25,7 +25,7 @@ packages, and allows you to safely run the program from anywhere.
 ::
 
   $ python3 -m pip install --user pipx
-  $ python3 -m userpath ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
+  $ python3 -m userpath append ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
 
 .. Note:: You may need to restart your terminal for the path updates to take effect.
 


source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -25,7 +25,7 @@ packages, and allows you to safely run the program from anywhere.
 ::
 
   $ pip install --user pipx
-  $ pipx ensurepath  # ensures the path of the CLI application directory is on your $PATH
+  $ userpath ~/.local/bin  # ensures the path of the CLI application directory is on your $PATH
```
ncoghlan: Hmm, this makes me wonder if both of these commands should be using a `python3 -m ...` invocation.

cs01: What does?

ncoghlan: If you do `pip install --user pipx`, and `~/.local/bin` isn't already on your path, how is either `pipx` or `userpath` actually going to run when invoked directly rather than via `python -m ...`?

I've realised that the new command is problematic for a different reason, though: it isn't right for Windows. I don't remember off the top of my head what the per-user install path is there, I just remember it isn't `~/.local/bin`.

cs01: Are you saying packages that install with `pip install --user` are not guaranteed to have their entry points on the user's PATH?

pfmoore: That's certainly true on Windows, and I suspect it might be true on Unix as well (but @ncoghlan can probably confirm that better than I can).

ncoghlan:
Correct, at least Debian & Ubuntu don't have `~/.local/bin` on `PATH` by default.

I just forgot about that problem when reviewing the original version of this guide.

source/guides/installing-stand-alone-command-line-tools.rst
ncoghlan: Correct, at least Debian & Ubuntu don't have `~/.local/bin` on `PATH` by default.

I just forgot about that problem when reviewing the original version of this guide.

ncoghlan: Oh, both Paul and I answered the wrong question.

The issue is that neither `ensurepath` nor `userpath` are guaranteed to be executable after a per-user install, but  `python3 -m ensurepath` and `python3 -m userpath` are.

cs01:
Ah yes that all makes sense. That is assuming `userpath` and `ensurepath` are packages with a `__main__.py` file, which happens to be true for `userpath`. `ensurepath` is not a package though, it was a pipx command: `pipx ensurepath`. So the move to use `userpath` fixes that too :smile: . Either way, should be good to go now. Thank you for your help.

cs01:
Oh, I'm not a Windows user by I don't think `python3` works on Windows. It's supposed to be `python3.exe` or `py` I think.

theacodes:
@cs01 is this still active? Should I close it for now?

cs01:
We can close it for now.

Ivoz:
Not sure why this should be closed, the changes are still relevant.

`pip` on many linuxes will be trying to use python 2 to install pipx. Not entirely sure how that is/should be dealt with over the entire guide. 

`$> pipx` on many linuxes, after installing it with `--user`, will indeed possibly not be found.

`pipx` has indeed [deprecated ](https://github.com/pipxproject/pipx/blob/0938682dd390f76fa713b940b9de4b66994ff67e/pipx/commands.py#L651) `$> pipx ensurepath`

In clarification, on Windows & Archlinux then simply `python` would correctly invoke a python 3 interpreter

cs01:
Thanks, you are right @Ivoz there are still improvements to be made here. I am working on getting feedback for un-deprecating `pipx ensurepath` in https://github.com/pipxproject/pipx/pull/179, then I'll re-open this PR.

PR #608, opened 2019-04-25
Add a blurb for shiv in the key projects section.

lorencarvalho:
Hi 👋 

Just as the subject states, just adding a blurb for `shiv` to the key projects section. Thanks for looking!


theacodes:
Thanks, @lorencarvalho!

lorencarvalho:
thank you!

PR #620, opened 2019-06-03
Fix :file:``filename`` instances

treszkai:
Issue #458 introduced minor formatting errors in `entry-points.rst`, when it changed 

```
``foobar.txt``
```

to

```
:file:``foobar.txt``
```

The correct version is

```
:file:`foobar.txt`
```


theacodes:
Thank you!

PR #621, opened 2019-06-19
Code changes as per #266

PadmarajBhat:
- improving the clarity on file structure
- avoiding the possibility of user deviating from parent directory.


pradyunsg:
Thanks for filing this! It seems to me that there's already #611 for the same issue.

PR #624, opened 2019-07-05
guides/packaging-namespace-packages.rst: mention setuptools.find_name…

### LINKED ISSUE #580: Mention find_namespace_package function in "Packaging namespace packages" guide.
I was struggling with a namespace packge for a long time until I found out about `setuptools.find_namespace_packages()` in a github issue.

I think it would be nice if `find_namespace_packages()` was mentioned in this guide:

https://packaging.python.org/guides/packaging-namespace-packages/#packaging-namespace-packages

There is this section already:

> Because mynamespace doesn’t contain an __init__.py, setuptools.find_packages() won’t find the sub-package. You must explicitly list all packages in your setup.py. For example:

I think it would be nice if it was updated to something like this:

> Because mynamespace doesn’t contain an __init__.py, setuptools.find_packages() won’t find the sub-package. To get the sub-packages, you must either use setuptools.find_namespace_packages() or explicitly list all sub-packages in your setup.py. For example:

plemelin:
Adds a mention to setuptools.find_namespace_packages() in the packaging guide since setuptools.find_packages() will not work on PEP 420 packages

fixes #580

### DIFF
--- a/source/guides/packaging-namespace-packages.rst
+++ b/source/guides/packaging-namespace-packages.rst
@@ -100,7 +100,8 @@ package omits the :file:`__init__.py` or uses a pkgutil-style
 logic to fail and the other sub-packages will not be importable.
 
 Because ``mynamespace`` doesn't contain an :file:`__init__.py`,
-:func:`setuptools.find_packages` won't find the sub-package. You must
+:func:`setuptools.find_packages` won't find the sub-package. To get the
+sub-packages, you must either use :func:`setuptools.find_namespace_packages` or
 explicitly list all packages in your :file:`setup.py`. For example:
 
 .. code-block:: python


ncoghlan:
Build failure looked like an issue with loading the intersphinx inventories, so I restarted it. However, it failed again, so we may have a config issue somewhere.

On the PR itself, not that `setuptools` has a function for this, perhaps we should update the examples to use it? That could be a separate follow-up PR after this one, though - it doesn't need to be part of the initial change.

@theacodes, your thoughts?

theacodes:
Hrm I'm really not sure what's going on there. I'll have to take a look
once I have some more time.

plemelin:
I tried running the jobs the CI is running in a local venv, on the master branch, and I have the same failures.

theacodes:
Filed #625 to fix the build issues. Also closing this in favor of #622. Thanks, everyone!

PR #626, opened 2019-07-16
pypa.io -> www.pypa.io

Redoubts:
pypa.io doesn't appear to be a navigable link. Safari & Firefox complain they can't connect because the site is misconfigured, and they don't seem to redirect to www.pypa.io, which does work.

Screen Shot 2019-07-16 at 7 05 31 AM


webknjaz:
@ewdurbin ^

ewdurbin:
Thanks for highlight, looking now.

webknjaz:
Looks like certbot is broken. The certificate is valid until July 4.

ewdurbin:
Yeah, just confirmed that cert renewals were failing due to a cron typo. That should be resolved now.

Redoubts:
Fixed for me. Thanks for the quick work.

PR #630, opened 2019-07-30
clarify test property in example module

belm0:


source/tutorials/packaging-projects.rst
```diff
@@ -303,14 +303,14 @@ Run the Python interpreter (make sure you're still in your virtualenv):
 
     python
 
-And then import the module and print out the ``name`` property. This should be
+And then import the module and print out the ``__name__`` property. This should be
```
di: I don't think this should be changed. This is referring to an attribute that is set earlier in the tutorial:
https://github.com/pypa/packaging.python.org/blob/14eff73b7501ec1b54de3ba286314d817c09f369/source/tutorials/packaging-projects.rst#L33

I think the idea is that this is some data that will be consistent between installations of this package. I agree it's a bit confusing that this looks similar to `__name__` and that the actual value is the module name -- perhaps we should change it to something else entirely.

belm0: Oh-- yes, that's definitely confusing.  I was just following the instructions loosely with my own package.

I'll change this PR to make the example and its context more obvious.

belm0: @di would you review?

di: After taking a second look at this, I think the tutorial is trying to show that `name="example-pkg-your-username",` in `setup.py` and `example_pkg.__name__` can be different, i.e. that you can have a module with a different name than the package name.

This is kind of a funny way of showing it though. I think we could simplify a bit here by:
* just creating the `__init__.py` file
* just showing that after installing `example-pkg-your-username` you can import `example_pkg`

aixtools:
:grin: reading this makes me realize I need to follow this tutorial and figure out how to dig ans see, I think, the meta data as it develops. Why the grin? because this looks terribly scary aka confusing - how many (different) ways are there to specify "name"? Or should I be asking - how many attributes have "name" as part of their keyword, or is it just RST notation, or "scary"!
:grin: all meant in a productive way. (Don´t get much more noob than I when it comes to packaging internals)

di:
@belm0 I think we should simplify this even further:

* just creating the `__init__.py` file
* just showing that after installing `example-pkg-your-username` you can `import example_pkg`

I don't think we need all that `test_string` and `__name__` stuff.

belm0:
hopefully ready to merge

Issue #453, opened 2018-03-07
tutorials/installing-packages is problematic in the way it deals with multiples operating systems.


sffjunkie:
The installing packages page has the following note

*"Due to the way most Linux distributions are handling the Python 3 migration, Linux users using the system Python without creating a virtual environment first should replace the python command in this tutorial with python3"*

However most of the commands are already changed to `python3` in the rest of the text. Therefore on Windows when following the tutorial the command `python3` is not found because it does not exist.

ncoghlan:
The instructions that start with `python3` relating to creating virtual environments, which typically isn't needed on non-Linux systems.

It's definitely a problem with trying to handle Linux and Mac OS X in the same guide as we use to handle Windows, though - the baseline starting states for the 3 operating systems are *very* different, so we end up with cases like this where it isn't obvious that folks on Windows are likely better off not worrying about virtual environments when they're just getting started :(

(See also https://github.com/pypa/python-packaging-user-guide/issues/393 and https://github.com/pypa/python-packaging-user-guide/issues/396)

jeanas:
This should be fine now. The [guide](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/) in question has “Unix/macOS” and “Windows” tabs, with `python3` only in the former.

PR #631, opened 2019-08-01
use python3 -m when installing pipx

cs01:
As noted in https://github.com/pypa/packaging.python.org/pull/607, `pipx` may not be on the user's PATH when installed with the `--user` flag. This PR adds the `python3 -m` prefix when ensuring the PATH has paths pipx expects. Note that pipx now considers `ensurepath` best practice again (it's no longer deprecated).


cs01:
cc @ncoghlan @Ivoz

ncoghlan:
Yeah, there's an existing issue for `python3` being problematic: https://github.com/pypa/packaging.python.org/issues/453

PR #641, opened 2019-08-25
Properly exclude test/ and tests/ in find_packages() example

jwodder:
This pull request edits the code snippet [in the "packages" `setup()` keyword description](https://packaging.python.org/guides/distributing-packages-using-setuptools/#packages) so that `find_packages()` properly excludes `test` *and* `tests` *and* everything under them.

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -353,7 +353,7 @@ packages
 
 ::
 
-  packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
+  packages=find_packages(exclude=['contrib', 'docs', 'test', 'test.*', 'tests', 'tests.*']),
 
 Set ``packages`` to a list of all :term:`packages <Import Package>` in your
 project, including their subpackages, sub-subpackages, etc.  Although the


webknjaz:
@pganssle WDYT?

pganssle:
> @pganssle WDYT?

I think it's kinda overkill. Presumably the point of this is to explain how to exclude things with `find_packages` and to list a few different things that might be excluded. I think adding some sort of exhaustive list of everything that might match a test will just distract from that purpose.

I personally think that you should probably not be bothering with `exclude` and `include` except in rare circumstances, and the best thing to do is to have your package in an `src/` directory and everything that is *not* your package (including tests) in a directory alongside your `src/` directory. That is a much more bulletproof solution than trying to come up with a bunch of glob patterns that might match some test file you erroneously put into your package.

In `dateutil` I have the `dateutil.test` package in my package root for historical reasons, and I [simply exclude it directly](https://github.com/dateutil/dateutil/blob/master/setup.cfg#L39), which is more in line with what the original example says.

jwodder:
This PR was made under the assumption that a nontrivial number of developers are liable to just copy & paste the `packages=` line from the guide and use it in their own `setup.py`.  Under this assumption, I feel the guide should, at a minimum, exclude the entire `tests` hierarchy correctly (The guide currently uses `'tests*'`, which is wrong; it should be `'tests'` and `'tests.*'`).

I am aware that some people want the PyPUG to switch to all-`src/` layout, but until that happens (which shows no sign of being soon), PyPUG should at least present quality non-`src/` configuration.

pganssle:
> I am aware that some people want the PyPUG to switch to all-`src/` layout, but until that happens (which shows no sign of being soon), PyPUG should at least present quality non-`src/` configuration.

At the Packaging Summit we actually basically agreed to switch it over, but no one has actually done it yet, so hopefully that's a sign that it should be soon-ish.

As for the rest of it, I don't think it changes my mind, but I am not a maintainer of this repository. I don't think people should mindlessly copy this and assume things will work out.

How about instead of showing how to use `find_packages` with `exclude`, we switch it over to a whitelist with `include`. That a much better pattern than trying to exclude common things that aren't packages anyway. Note that in [the setuptools documentation](https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages), it explicitly mentions that the `exclude` directive was designed for simpler use cases than what you are proposing here.

jwodder:
New PR created using `include`: #642

PR #648, opened 2019-09-24
Modify `Requests` package link to correct URL

shayan-7:
The `https://python-requests.org` is currently unavailable; So modified to `https://2.python-requests.org` valid URL.

### DIFF
--- a/source/tutorials/managing-dependencies.rst
+++ b/source/tutorials/managing-dependencies.rst
@@ -95,7 +95,7 @@ when you share your project with others. You should get output similar to this
 
     Adding requests to Pipfile's [packages]...
 
-.. _Requests: https://python-requests.org
+.. _Requests: https://2.python-requests.org
 .. _pipenv-tox: https://docs.pipenv.org/advanced/#tox-automation-project
 
 


pfmoore:
Do you have a pointer to an explanation of *why* https://python-requests is unavailable, which confirms that the alternative you give is "official"? Generally, I'd assume something like this was a temporary outage and just report it to the website owner, not try to fix up the documentation (there's likely a *lot* of docs that point to the requests site).

pfmoore:
@webknjaz Ah, thanks for the clarification. I guess we could use the http form, if the project isn't going to fix the https redirect.

pfmoore:
Agreed. I'm not strongly against the https://2.python-requests.org URL, it's just not one I've seen documented anywhere else, and it looks to me more like a direct link to one node of a CDN or similar than a legitimate URL for "Requests 2", so I'm concerned it might cause confusion.

The best solution would be if requests fix their redirect.

shayan-7:
@pfmoore Do you suggest using http://python-requests.org? Or modifying the URL, in this case, is invalid?

pfmoore:
My preferences, in order:

1. Get the requests project to fix their https redirect. This seems like a bug at their end so better if they just get it fixed. Presumably the current URL used to work, so there's no reason for them to have deliberately blocked it all of a sudden.
2. Use the http URL because it's the "Homepage" as defined on the requests PyPI page.
3. Use the https://2.python-requests.org URL "because it works".

But my preferences aren't strong. I only made my initial comment because the proposed new URL looked odd to me, so I wanted to understand better why it was needed.

PR #650, opened 2019-10-05
Moved language around, adding how to activate a venv on Windows

### LINKED ISSUE #393: venv section in current tutorial excludes Windows
In https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments (and the section just before it), we show examples using the `source` command.

On Windows, the activate script is just run directly, since there's no `source` equivalent in an interactive shell.

ak9999:
I saw that issue #393 was still open. If it still is open, would this be helpful?

Fixes #393.


di:
Thanks!

ak9999:
Glad this helps!

PR #651, opened 2019-10-06
Binary extension use cases: provide examples

bsolomon1124:
Provide a simple CPython example of each of the three use cases mentioned in "Packaging binary extensions."

Additions highlighted:

Screen Shot 2019-10-06 at 6 22 07 PM

This commit also bolds the 3 terms themselves to break up the long text block.


di:
Thanks!

PR #656, opened 2019-10-11
Update manylinux tags section

di:
Updates the section on `manylinux` per https://github.com/pypa/manylinux/issues/179 and https://github.com/pypa/manylinux/issues/338.

### DIFF
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -9,16 +9,25 @@ Platform compatibility tags allow build tools to mark distributions as being
 compatible with specific platforms, and allows installers to understand which
 distributions are compatible with the system they are running on.
 
-General platform tags
-=====================
-
 The platform compatibility tagging model used for the ``wheel`` distribution
 format is defined in :pep:`425`.
 
-.. _manylinux:
+Platform tags for Windows
+-------------------------
 
-Platform tags for Linux platforms
-=================================
+The scheme defined in :pep:`425` covers public distribution of wheel files to
+systems running Windows.
+
+Platform tags for macOS (Mac OS X)
+----------------------------------
+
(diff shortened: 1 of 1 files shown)


source/specifications/platform-compatibility-tags.rst
```diff
@@ -5,24 +5,72 @@
...
 
 .. _manylinux:
 
-Manylinux tags
-==============
+Platform tags for Linux platforms
+=================================
+
+The scheme defined in :pep:`425` is insufficient for public distribution of
+wheel files (and \*nix wheel files in general) to Linux platforms, due to the
```
ncoghlan: The parenthetical here seems a bit out of place. It's not *wrong*, just not especially relevant in a section titled "Platform tags for Linux platforms".

What if we went for a top level structure that looked like:

    Platform compatibility tags
    ===========================
        ... existing intro paragraph

    The platform compatibility tagging model used for the ``wheel`` distribution format is defined in :pep:`425`.
    

    Platform tags for Windows
    -------------------------

    The scheme defined in :pep:`425` covers public distribution of wheel files to systems running Windows.

    Platform tags for macOS (Mac OS X)
    ----------------------------------

    The scheme defined in :pep:`425` covers public distribution of wheel files to systems running macOS (previously known as Mac OS X).

    Platform tags for common Linux distributions
    --------------------------------------------

    ... new section from the PR

    Platform tags for other \*nix platforms
    ---------------------------------------

    The scheme defined in :pep:`425` is not generally sufficient for public distribution of wheel files to other \*nix platforms. Efforts are currently (albeit intermittently) under way to define improved compatibility tagging schemes for AIX and for Alpine Linux.

di: I added these changes verbatim.

PR #663, opened 2019-10-24
Add field name for the Project-URL example

wimglenn:
Every other example in the guide has the field name included in the example, I think it was likely just an oversight that this one was missing it.


di:
Thanks @wimglenn!

wimglenn:
No prob, FYI I've also got https://github.com/python/peps/pull/1216 open to change the wording in the PEP (where this was presumably copy-pasted from).  Will ping here if it gets accepted.

PR #671, opened 2019-11-09
distributing - clarify keywords list is a string

wesinator:


jwodder:
This PR is not entirely accurate:

* The `keywords` string doesn't have to be space-separated; it can also be comma- or semicolon-separated; cf. the `format_tags()` function in [`Warehouse/warehouse/filters.py`](https://github.com/pypa/warehouse/blob/master/warehouse/filters.py).

* setuptools accepts setting `keywords` to a list, in which case it will join the elements of the list with commas.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -327,7 +327,7 @@ keywords
 
   keywords='sample setuptools development',
 
-List keywords that describe your project.
+Separated string, or list, of keywords that describe your project.
```
webknjaz, line 330: I'm not sure this is worded correctly. Before it was an imperative call to action and now it's just a statement that doesn't invite to do anything...

PR #672, opened 2019-11-10
Closes #667: advise against using distutils directly

### LINKED ISSUE #667: Formally drop support for distutils-only projects?
The guide already assumes that people will use setuptools when building projects, but it falls short (as far as I can see) of stating that projects should not use "pure" distutils without setuptools. Newer packaging standards (for example, PEP 517) no longer support the use of distutils without setuptools.

Should the packaging guide make a formal statement that use of distutils alone is no longer supported? Or is there somewhere else that such a statement could be made?

See https://github.com/giampaolo/psutil/pull/1565 for some background.

ncoghlan:


ncoghlan:
Arising from both the discussion in #667 (and the linked psutil issue), as well as https://github.com/pypa/packaging-problems/issues/288 (where `bdist_wheel` was failing for a user because their `setup.py` was still importing from `distutils.core`)

source/tutorials/packaging-projects.rst
```diff
@@ -120,6 +120,16 @@ minimal set:
...
 :doc:`/guides/distributing-packages-using-setuptools` for more details.
 
+.. note::
+
+   You may see some existing projects or other Python packaging tutorials that
+   import their ``setup`` function from ``distutils.core`` rather than
+   ``setuptools``. This is a `legacy approach`_ that installers [1]_ support
+   for backwards compatibility purposes, but using the legacy ``distutils`` API
+   directly in new projects is strongly discouraged, as it means that newer
+   build commands, like ``setup.py bdist_wheel``, won't work.
```
pfmoore: I'd like to see this worded a little more strongly, to address 2 points:

1. The "backward compatibility" support is limited, and may not be compatible with newer installer features (like build isolation). It also won't be maintained indefinitely, which leads to....
2. Existing projects are strongly advised to move away from using the legacy `distutils` API to ensure that they will continue working as installers evolve.

source/tutorials/packaging-projects.rst
```diff
@@ -339,3 +349,15 @@ some things you can do:
...
 .. _poetry: https://github.com/sdispater/poetry
+
+
+----
+
+.. [1] Some legacy Python environments may not have ``setuptools``
+       pre-installed, and the operators of those environments may still be
+       requiring users to install packages by running ``setup.py install``
+       commands, rather than providing an installer like ``pip`` that
+       automatically installes required build dependendencies. These
```
di:
```suggestion
       automatically installs required build dependendencies. These
```

webknjaz:
Hey @ncoghlan, could you solve the merge conflict here?

source/tutorials/packaging-projects.rst
webknjaz: Hey @ncoghlan, are you planning to address this?

webknjaz:
@ncoghlan also, there's a huge merge conflict that needs to be solved manually (the browser-based UI is not very helpful here).

astrojuanlu:
I'd be happy to take over this and bring it to a mergeable state.

webknjaz:
@astrojuanlu feel free to send a PR based on this.

astrojuanlu:
Done: #1028

PR #674, opened 2019-11-19
Change keywords field to be comma separated

takluyver:
distutils and setuptools have gone against the spec for years, possibly without anyone noticing. Changing the spec to record what's in widespread use is much easier than trying to fix the tools to match the spec.

Discussion on distutils-sig: https://mail.python.org/archives/list/distutils-sig@python.org/thread/LFTLGMIIB5PRG7NUUIGGGX4COGX53DMR/


SaltyCatFish:
Can this issue be closed?

PR #676, opened 2019-11-25
Update creating-and-discovering-plugins.rst

maggyero:
Fix typos and name the `plugins` variable consistently.

### DIFF
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -28,7 +28,7 @@ discover all of the Flask plugins installed:
     import importlib
     import pkgutil
 
-    flask_plugins = {
+    plugins = {
         name: importlib.import_module(name)
         for finder, name, ispkg
         in pkgutil.iter_modules()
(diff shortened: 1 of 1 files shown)


source/guides/creating-and-discovering-plugins.rst
```diff
@@ -80,16 +80,16 @@ under that namespace:
...
         for finder, name, ispkg
         in iter_namespace(myapp.plugins)
     }
 
 Specifying ``myapp.plugins.__path__`` to :func:`~pkgutil.iter_modules` causes
 it to only look for the modules directly under that namespace. For example,
-if you have installed distributions that provide the modules ``myapp.plugin.a``
-and ``myapp.plugin.b`` then ``myapp_plugins`` in this case would be:
+if you have installed distributions that provide the modules ``myapp.plugins.a``
+and ``myapp.plugins.b`` then ``plugins`` in this case would be:
```
di: I think this is getting a little confusing here. With this edit there are two different `plugins` variables:  one is a submodule of `myapp` and the other is the result of the comprehension in this file.

I think the name of the variable in all these three examples should be something like `discovered_plugins` to avoid this confusion.

maggyero: @di I agree, I have updated the variable names to `discovered_plugins`.

maggyero:
Thanks @di for reviewing this.

PR #678, opened 2019-11-28
Update contributing purpose formatting

jrbourbeau:
This PR contains a small update to the "Purpose" subsection of the contributing page, which is currently rendered as a block quote:

Screen Shot 2019-11-27 at 10 47 16 PM

Happy to close this PR if this formatting is intended


ncoghlan:
I think we actually were quoting something when we originally added this page, but I agree it just looks odd now.

Thanks for the update!

PR #682, opened 2019-12-12
Use utf-8 to read README

methane:
On Windows, default file encoding is not UTF-8 in most environments.


domdfcoding:
Is this PR ready to be merged?

webknjaz:
I think so. @di :point_up:

PR #690, opened 2020-01-25
Add reference to PEP 610

sbidoul:
As proposed by @ncoghlan in https://discuss.python.org/t/draft-pep-recording-the-origin-of-distributions-installed-from-direct-url-references/1535/41?u=sbidoul

### DIFF
--- a/source/specifications/direct-url.rst
+++ b/source/specifications/direct-url.rst
@@ -1,0 +1,303 @@
+
+.. _direct-url:
+
+==========================================================
+Recording the Direct URL Origin of installed distributions
+==========================================================
+
+This document specifies a :file:`direct_url.json` file in the
+:file:`*.dist-info` directory of an installed distribution, to record the
+Direct URL Origin of the distribution. The layout of this file was originally
+specified in :pep:`610` and is formally documented here.
+
+.. contents:: Contents
+   :local:
+
+Specification
+=============
+
+The :file:`direct_url.json` file MUST be created in the :file:`*.dist-info`
+directory by installers when installing a distribution from a requirement
+specifying a direct URL reference (including a VCS URL).
+
+This file MUST NOT be created when installing a distribution from an other
+type of requirement (i.e. name plus version specifier).
+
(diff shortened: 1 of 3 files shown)


sbidoul:
@ncoghlan ah ok, I was not sure since, since many other entries in the specifications page are merely pointers to the related PEP.

At which point should we copy the spec here? I assume the PEP should move out of Draft state first?

ncoghlan:
@sbidoul We used to use the PEPs as the primary reference, so we've just been opportunistically migrating them as they get updated for other reasons (e.g. PEP 566 was really about defining metadata 2.1, but it *also* migrated to using packaging.python.org as the main reference page for the core metadata spec).

But yeah, for the initial draft, it makes sense to get it to Accepted before duplicating all the details over here. The key thing to add to the PEP is the long term URL you plan to use, and I think your suggestion of https://packaging.python.org/specifications/direct-url is a clear and sensible choice.

sbidoul:
> The key thing to add to the PEP is the long term URL you plan to use, and I think your suggestion of https://packaging.python.org/specifications/direct-url is a clear and sensible choice

@ncoghlan done in https://github.com/python/peps/pull/1283

sbidoul:
I updated this PR with the full specification extracted from PEP 610.

cc/ @ncoghlan @pradyunsg

pradyunsg (approved):
I've skimmed this to check the major bits, and not gone line by line. OTOH, I'm pretty sure that @sbidoul isn't gonna sneakily change a detail from their own PEP. 🙃

sbidoul:
@pradyunsg indeed, I only tweaked rst formatting (title underlines and roles), and adjusted the first few lines so it reads better.

This reminds me, though, that I had [said](https://discuss.python.org/t/draft-pep-recording-the-origin-of-distributions-installed-from-direct-url-references/1535/59?u=sbidoul) that I'd leave out the part about `resolved_revision` and `resolved_revision_type` in the spec, because it is currently not implemented anywhere. So I've just added a commit removing that.

ncoghlan (approved):

PR #691, opened 2020-02-02
Update wording, links in key projects page

brainwane:
A few link fixes, clearer explanations regarding wheel, packaging, pip, and distlib, and a few wording/capitalization fixes.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -227,12 +227,13 @@ User irc:#pypa  |
 Dev irc:#pypa-dev
 
 virtualenv is a tool which uses the command-line path environment
-variable to (like venv) create isolated Python :term:`Virtual
-Environments <Virtual Environment>`. virtualenv goes beyond venv by
-supporting Python 2.7 and by providing convenient features for
-configuring, maintaining, duplicating, and troubleshooting the virtual
-environments. For more information, see the section on :ref:`Creating
-and using Virtual Environments`.
+variable to create isolated Python :term:`Virtual Environments
+<Virtual Environment>`, much as ``venv`` does. virtualenv provides
+additional functionality, compared to ``venv``, by supporting Python
+2.7 and by providing convenient features for configuring, maintaining,
+duplicating, and troubleshooting the virtual environments. For more
+information, see the section on :ref:`Creating and using Virtual
+Environments`.
 
 
 .. _warehouse:
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -69,7 +69,7 @@ Dev irc:#pypa-dev
 
 Core utilities for Python packaging used by :ref:`pip` and :ref:`setuptools`.
 
-The core utilities in the packaging bundle handle version handling,
+The core utilities in the packaging library handle version handling,
 specifiers, markers, requirements, tags, and similar attributes and
 tasks for Python packages. Most Python users rely on this bundle
 without needing to explicitly call it; developers of the other Python
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -22,7 +22,7 @@ bandersnatch
 `Issues <https://github.com/pypa/bandersnatch/issues>`__ |
 `GitHub <https://github.com/pypa/bandersnatch>`__ |
 `PyPI <https://pypi.org/project/bandersnatch>`__ |
-Dev irc:#bandersnatch
+Dev IRC:`#bandersnatch <https://webchat.freenode.net/?channels=%23bandersnatch>`__
 
 ``bandersnatch`` is a PyPI mirroring client designed to efficiently
 create a complete mirror of the contents of PyPI. Organizations thus
@@ -68,8 +68,8 @@ packaging
 `Issues <https://github.com/pypa/packaging/issues>`__ |
 `GitHub <https://github.com/pypa/packaging>`__ |
 `PyPI <https://pypi.org/project/packaging>`__ |
-User irc:#pypa |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 Core utilities for Python packaging used by :ref:`pip` and :ref:`setuptools`.
 
@@ -101,8 +101,8 @@ pip
 `Issues <https://github.com/pypa/pip/issues>`__ |
 `GitHub <https://github.com/pypa/pip>`__ |
 `PyPI <https://pypi.org/project/pip/>`__ |
-User irc:#pypa |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 The most popular tool for installing Python packages, and the one
 included with modern versions of Python.
@@ -153,8 +153,8 @@ Python Packaging User Guide
 `Mailing list <http://mail.python.org/mailman/listinfo/distutils-sig>`__ |
 `Issues <https://github.com/pypa/python-packaging-user-guide/issues>`__ |
 `GitHub <https://github.com/pypa/python-packaging-user-guide>`__ |
-User irc:#pypa |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 This guide!
 
@@ -185,8 +185,8 @@ setuptools
 `Issues <https://github.com/pypa/setuptools/issues>`__ |
 `GitHub <https://github.com/pypa/setuptools>`__ |
 `PyPI <https://pypi.org/project/setuptools>`__ |
-User irc:#pypa  |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 
 setuptools (which includes ``easy_install``) is a collection of
@@ -226,8 +226,8 @@ virtualenv
 `Issues <https://github.com/pypa/virtualenv/issues>`__ |
 `GitHub <https://github.com/pypa/virtualenv>`__ |
 `PyPI <https://pypi.org/project/virtualenv/>`__ |
-User irc:#pypa  |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 virtualenv is a tool which uses the command-line path environment
 variable to create isolated Python :term:`Virtual Environments
@@ -248,7 +248,8 @@ Warehouse
 `Mailing list <http://mail.python.org/mailman/listinfo/distutils-sig>`__ [2]_ |
 `Issues <https://github.com/pypa/warehouse/issues>`__ |
 `GitHub <https://github.com/pypa/warehouse>`__ |
-Dev irc:#pypa-dev
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
+
 
 
 The current codebase powering the :term:`Python Package Index
@@ -266,8 +267,8 @@ wheel
 `Issues <https://github.com/pypa/wheel/issues>`__ |
 `GitHub <https://github.com/pypa/wheel>`__ |
 `PyPI <https://pypi.org/project/wheel>`__ |
-User irc:#pypa  |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 Primarily, the wheel project offers the ``bdist_wheel`` :ref:`setuptools` extension for
 creating :term:`wheel distributions <Wheel>`.  Additionally, it offers its own
@@ -310,7 +311,7 @@ buildout
 `Issues <https://bugs.launchpad.net/zc.buildout>`__ |
 `PyPI <https://pypi.org/project/zc.buildout>`__ |
 `GitHub <https://github.com/buildout/buildout/>`__ |
-irc:#buildout
+IRC:`#buildout <https://webchat.freenode.net/?channels=%23buildout>`__
 
 Buildout is a Python-based build system for creating, assembling and deploying
 applications from multiple parts, some of which may be non-Python-based.  It
@@ -499,7 +500,7 @@ run on Raspbian use Piwheels as their primary Python package index.
     Note that Piwheels `does not yet fully support
     <https://github.com/piwheels/piwheels/issues/208>`__ :pep:`503` and
     thus some users have trouble installing certain wheels; this is in
-    progress.)
+    progress.
 
 .. _poetry:
 
@@ -628,8 +629,8 @@ distutils
 `Docs <https://docs.python.org/3/library/distutils.html>`__ |
 `User list <http://mail.python.org/mailman/listinfo/distutils-sig>`__ [2]_ |
 `Issues <http://bugs.python.org>`__ |
-User irc:#pypa  |
-Dev irc:#pypa-dev
+User IRC:`#pypa <https://webchat.freenode.net/?channels=%23pypa>`__ |
+Dev IRC:`#pypa-dev <https://webchat.freenode.net/?channels=%23pypa-dev>`__
 
 The original Python packaging system, added to the standard library in
 Python 2.0.
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -104,7 +104,7 @@ The most popular tool for installing Python packages, and the one
 included with modern versions of Python.
 
 It provides the essential core features for finding, downloading, and
-installing packages from pypi.org and other index servers, and can be
+installing packages from PyPI and other Python package indexes, and can be
 incorporated into a wide range of development workflows via its
 command-line interface (CLI).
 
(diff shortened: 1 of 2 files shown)


source/key_projects.rst
```diff
@@ -435,7 +454,7 @@ distutils
 User irc:#pypa  |
 Dev irc:#pypa-dev
 
-A package in the Python Standard Library that has support for creating and
```
webknjaz: Technically it's a package https://github.com/python/cpython/tree/master/Lib/distutils
(module == one py file, package == dir)

Also, this change is inconsistent with the paragraph about ensurepip above.

source/key_projects.rst
```diff
@@ -54,13 +59,20 @@ packaging
...
 
 Core utilities for Python packaging used by :ref:`pip` and :ref:`setuptools`.
 
+The core utilities in the packaging bundle handle version handling,
+specifiers, markers, requirements, tags, and similar attributes and
+tasks for Python packages. Most Python users rely on this bundle
+without needing to explicitly call it; developers of the other Python
+packaging, distribution, and installation tools listed here often use
+its functionality to parse, discover, and otherwise handle dependency
+attributes.
```
ncoghlan:
```suggestion
attributes.

This project specifically focuses on implementing the modern Python packaging
interoperability standards defined at :ref:`packaging-specifications`, and will report
errors for sufficiently old legacy packages that are incompatible with those
standards. The :ref:`distlib` project is a more permissive library that attempts to
provide a plausible reading of ambiguous metadata in cases where ``packaging``
will instead report on error.
```

It seems worth highlighting that the essential difference between "packaging" and "distlib" is that the latter includes extra behaviour that isn't in the approved PEPs, and one of the key aspects of that is making more guesses when confronted with ambiguous legacy metadata.

source/key_projects.rst
```diff
@@ -71,12 +83,13 @@ pip
...
 `Issues <https://github.com/pypa/pip/issues>`__ |
-`Github <https://github.com/pypa/pip>`__ |
+`GitHub <https://github.com/pypa/pip>`__ |
 `PyPI <https://pypi.org/project/pip/>`__ |
 User irc:#pypa |
 Dev irc:#pypa-dev
 
-A tool for installing Python packages.
+The most popular tool for installing Python packages, and the one
+included with modern versions of Python.
```
ncoghlan:
```suggestion
included with modern versions of Python.

It provides the essential core features for finding, downloading, and installing packages
from pypi.org and other index servers, and can be incorporated into a wide range of
development workflows via its CLI interface.
```

source/key_projects.rst
```diff
@@ -150,7 +163,7 @@ twine
 
 `Mailing list <http://mail.python.org/mailman/listinfo/distutils-sig>`__ [2]_ |
 `Issues <https://github.com/pypa/twine/issues>`__ |
-`Github <https://github.com/pypa/twine>`__ |
+`GitHub <https://github.com/pypa/twine>`__ |
 `PyPI <https://pypi.org/project/twine>`__
 
 Twine is a utility for interacting with PyPI, that offers a secure replacement for
```
ncoghlan: We should update this description, as ``setup.py upload`` is secure in modern versions of Python (since 3.5.3 or so).

``twine`` is still better - just for reasons other than this one.

source/key_projects.rst
```diff
@@ -435,7 +454,7 @@ distutils
 User irc:#pypa  |
 Dev irc:#pypa-dev
 
-A package in the Python Standard Library that has support for creating and
+A module in the Python Standard Library that has support for creating and
```
ncoghlan:
```suggestion
The original Python packaging system, added to the standard library in Python 2.0.
Due to the challenges of maintaining a packaging system where feature updates are tightly coupled to language runtime updates, direct usage of ``distutils`` is now actively discouraged, with :ref:`Setuptools` being the preferred replacement. :ref:`Setuptools` not only provides features that plain ``distutils`` doesn't offer (such as dependency declarations and entry point declarations), it also provides a consistent build interface and feature set across all supported Python versions.
```

source/key_projects.rst
brainwane: Thanks. Resolved.

source/key_projects.rst
```diff
@@ -39,12 +39,17 @@ distlib
...
-packaging and distribution of Python software.  It consists in part of the
-functions from the `distutils2 <https://pypi.org/project/Distutils2>`_
-project, which was intended to be released as ``packaging`` in the Python 3.3
-stdlib, but was removed shortly before Python 3.3 entered beta testing.
-
+Distlib is a library which implements low-level functions that relate
+to packaging and distribution of Python software.  It consists in part
+of the functions from the `distutils2
+<https://pypi.org/project/Distutils2>`_ project, which was intended to
+be released as ``packaging`` in the Python 3.3 stdlib, but was removed
```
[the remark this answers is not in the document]
brainwane: Thanks! Fixed, I believe..

source/key_projects.rst
```diff
@@ -39,12 +39,17 @@ distlib
...
+shortly before Python 3.3 entered beta testing. distlib implements
+several relevant PEPs (Python Enhancement Proposal standards) and is
+useful for developers of third-party packaging tools to make and
+upload binary and source distributions, achieve interoperability,
+resolve dependencies, manage package resources, and do other similar
+functions.
```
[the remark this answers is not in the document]
brainwane: Thanks - added.

source/key_projects.rst
brainwane: Thanks so much! Added.

source/key_projects.rst
brainwane: Thank you! Added.

source/key_projects.rst
brainwane: I updated it more thoroughly. Thanks.

source/key_projects.rst
brainwane: Thanks so much! Incorporated your addition.

brainwane:
> Several suggestions inline. Perhaps it would make sense to split the purely mechanical GitHub word mark fix out from the more subjective updates to the project descriptions?

I appreciate the note; if I'd been more disciplined I would have done that! If this PR drags on for several days then I'll probably do that. :-)

source/key_projects.rst
```diff
@@ -150,12 +180,14 @@ twine
...
 `Issues <https://github.com/pypa/twine/issues>`__ |
-`Github <https://github.com/pypa/twine>`__ |
+`GitHub <https://github.com/pypa/twine>`__ |
 `PyPI <https://pypi.org/project/twine>`__
 
-Twine is a utility for interacting with PyPI, that offers a secure replacement for
-``setup.py upload``.
-
+Twine is the primary tool developers use to upload packages to the
+Python Package Index or other Python package indexes. It is a
```
bhrutledge:
```suggestion
PyPI or other Python package indexes. It is a
```

source/key_projects.rst
```diff
@@ -71,13 +95,18 @@ pip
...
 `PyPI <https://pypi.org/project/pip/>`__ |
 User irc:#pypa |
 Dev irc:#pypa-dev
 
-A tool for installing Python packages.
+The most popular tool for installing Python packages, and the one
+included with modern versions of Python.
 
+It provides the essential core features for finding, downloading, and
+installing packages from pypi.org and other index servers, and can be
```
bhrutledge: This feels more in line with other references to these concepts:

```suggestion
installing packages from PyPI and other Python package indexes, and can be
```

source/key_projects.rst
bhrutledge: How about ths, to avoid the repetition of "package index"?

```
Twine is the primary tool developers use to upload packages to
PyPI or other Python package indexes. It is a
```

It looks like the acronym is more common in this document than the words.

source/key_projects.rst
```diff
@@ -348,19 +386,59 @@ pipx
...
 
+piwheels
+========
+
+`Website <https://piwheels.org/>`__ |
+`Docs <https://piwheels.readthedocs.io/>`__ |
+`GitHub <https://github.com/piwheels/piwheels/>`__
+
+Piwheels is a website, and software underpinning it, that fetches
+source code distribution packages from the Python Package Index and
```
bhrutledge:
```suggestion
source code distribution packages from PyPI and
```

Similar to my comment re: Twine

source/key_projects.rst
```diff
@@ -348,19 +386,59 @@ pipx
 ====
 
 `Docs <https://github.com/pipxproject/pipx>`__ |
```
bhrutledge:
```suggestion
`Docs <https://pipxproject.github.io/pipx/>`__ |
```

source/key_projects.rst
```diff
@@ -54,13 +62,29 @@ packaging
...
 `Issues <https://github.com/pypa/packaging/issues>`__ |
-`Github <https://github.com/pypa/packaging>`__ |
+`GitHub <https://github.com/pypa/packaging>`__ |
 `PyPI <https://pypi.org/project/packaging>`__ |
 User irc:#pypa |
 Dev irc:#pypa-dev
 
 Core utilities for Python packaging used by :ref:`pip` and :ref:`setuptools`.
 
+The core utilities in the packaging bundle handle version handling,
```
bhrutledge: The word "bundle" surprised me; I hadn't seen it used in the context of Python packaging. Wondering if "library" is more accurate?

```suggestion
The core utilities in the packaging library handle version handling,
```

source/key_projects.rst
```diff
@@ -54,13 +62,29 @@ packaging
...
+specifiers, markers, requirements, tags, and similar attributes and
+tasks for Python packages. Most Python users rely on this bundle
```
bhrutledge:
```suggestion
tasks for Python packages. Most Python users rely on this library
```

source/key_projects.rst
```diff
@@ -54,13 +62,29 @@ packaging
...
+its functionality to parse, discover, and otherwise handle dependency
+attributes.
+
+This project specifically focuses on implementing the modern Python
+packaging interoperability standards defined at
+:ref:`packaging-specifications`, and will report errors for
+sufficiently old legacy packages that are incompatible with those
+standards. In contrast, the :ref:`distlib` project is a more
+permissive library that attempts to provide a plausible reading of
+ambiguous metadata in cases where ``packaging`` will instead report on
```
bhrutledge: Not sure what the convention is (or should be) around monospacing package names. It seems like most package names are *not* monospaced, except here, once in "Distlib", and in "distutils".

bhrutledge:
Also, thanks for putting the time into this, and the attention to detail. I think it’s a nice improvement.

source/key_projects.rst
```diff
@@ -407,6 +485,19 @@ Spack is not in PyPI (yet), but it requires no installation and can be
...
 
 
+zest.releaser
+=============
+
+`Docs <https://zestreleaser.readthedocs.io/en/latest/>`__ |
+`GitHub <https://github.com/zestsoftware/zest.releaser/>`__ |
+`PyPI <https://pypi.org/project/zest.releaser/>`__
+
+zest.releaser is a Python package release tool providing an
```
webknjaz: How about this?
```suggestion
``zest.releaser`` is a Python package release tool providing an
```
It seems like there's no consistency for marking dist names with `<code>` on this page. You added it for distutils so maybe others deserve this too.

source/key_projects.rst
```diff
@@ -348,19 +386,59 @@ pipx
...
+`Website <https://piwheels.org/>`__ |
+`Docs <https://piwheels.readthedocs.io/>`__ |
+`GitHub <https://github.com/piwheels/piwheels/>`__
+
+Piwheels is a website, and software underpinning it, that fetches
+source code distribution packages from the Python Package Index and
+compiles them into binary wheels that are optimized for installation
+onto Raspberry Pi computers. Many Raspberry Pi users who use or
+develop Python tools to run on Raspbian use piwheels as their primary
+Python package index.
```
webknjaz: Recently, I've discovered that this index doesn't support PEP 503 which makes Pip explode for wheels that specify `Requires-Python` under interpreter versions that don't match the condition instead of selecting wheels properly on the download stage. (https://github.com/cherrypy/cherrypy/issues/1837)

I think this may be a good place to add a known issue warning about this.

source/key_projects.rst
webknjaz: This suggested change breaks indentation :point_up: (which is a known bug @ GH)
Here's a fixed suggestion:
```suggestion
 `Docs <https://pipxproject.github.io/pipx/>`__ |
```

source/key_projects.rst
brainwane: It is. I'm guessing a lot of people come to section anchor tags in this document and skim around, and I think it's okay to say the full name once in a while. I'm not going to figure out a consistent style guide and apply it in this particular PR, but if someone else does so in a later change and it changes my choices in this regard, then I will be fine with that; I will not cry out to the heavens and rend my clothes or anything,

source/key_projects.rst
brainwane: Yeah, I think there is no consistent convention. Future person who decides on a style guide, consider this question!

source/key_projects.rst
brainwane: I will do so; I'll link to https://github.com/piwheels/piwheels/issues/208 which is where they are fixing it. Thanks for the heads-up.

brainwane:
Added several projects and improved wording and detail for several existing projects.

source/key_projects.rst
```diff
@@ -347,20 +443,82 @@ make deployment of Python applications as simple as ``cp``.
...
+`Docs <https://piwheels.readthedocs.io/>`__ |
+`GitHub <https://github.com/piwheels/piwheels/>`__
+
+Piwheels is a website, and software underpinning it, that fetches
+source code distribution packages from PyPI and compiles them into
+binary wheels that are optimized for installation onto Raspberry Pi
+computers. Many Raspberry Pi users who use or develop Python tools to
+run on Raspbian use Piwheels as their primary Python package index.
+
+(Note that Piwheels `does not yet fully support
```
webknjaz: This should probably be a warning box marked in RST...

webknjaz:
```suggestion
  .. warning::

    Note that Piwheels `does not yet fully support
```

source/key_projects.rst
```diff
@@ -347,20 +443,82 @@ make deployment of Python applications as simple as ``cp``.
...
+<https://github.com/piwheels/piwheels/issues/208>`__ :pep:`503` and
+thus some wheels have trouble with inconsistent ``Requires-Python``
```
webknjaz: This sounds wrong. Wheels don't have troubles, they are correct, it's Pip that has troubles because it does not get enough info from the index and cannot properly identify what to start downloading.

webknjaz: I'd maybe go as far as pointing to pip-tools as a workaround mentioning that this will help end-users pin versions of transitive deps.

source/key_projects.rst
```diff
@@ -167,12 +221,18 @@ virtualenv
...
-`Github <https://github.com/pypa/virtualenv>`__ |
+`GitHub <https://github.com/pypa/virtualenv>`__ |
 `PyPI <https://pypi.org/project/virtualenv/>`__ |
 User irc:#pypa  |
 Dev irc:#pypa-dev
 
-A tool for creating isolated Python environments.
+virtualenv is a tool which uses the command-line path environment
+variable to (like venv) create isolated Python :term:`Virtual
+Environments <Virtual Environment>`. virtualenv goes beyond venv by
```
pradyunsg:
```
virtualenv is a tool to create isolated Python :term:`Virtual
Environments <Virtual Environment>` (like venv). virtualenv provides
additional functionality compared to venv by...
```

source/key_projects.rst
```diff
@@ -199,16 +259,22 @@ wheel
...
 `PyPI <https://pypi.org/project/wheel>`__ |
 User irc:#pypa  |
 Dev irc:#pypa-dev
 
-
 Primarily, the wheel project offers the ``bdist_wheel`` :ref:`setuptools` extension for
 creating :term:`wheel distributions <Wheel>`.  Additionally, it offers its own
 command line utility for creating and installing wheels.
 
+See also `auditwheel <https://github.com/pypa/auditwheel>`__, a tool
```
pradyunsg: Any specific reason to not make auditwheel a dedicated heading?

uranusjr:
I’m on my phone right now; there are several casing and format inconsistencies that could use some fixing. I see at least three different rendering in distlib alone: Distlib, distlib, and `distlib`.

source/key_projects.rst
brainwane: Two reasons.

1. `auditwheel` is, as I understand it, only useful in the context of `wheel`.
2. Alphabetically, they're super far from each other.

But I don't particularly care. If you would like to block this PR on that, please say so; otherwise, I figure if someone would like to do that, they can do so and I will not object. :-)

source/key_projects.rst
brainwane: Thanks. Fixed.

source/key_projects.rst
brainwane: At this point I'm going to write a very simplified line of explanation here and I suggest you make a separate PR or commit with the language that you prefer. :-)

brainwane:
I've fixed up a few formatting inconsistencies (addressing @uranusjr's comments) as best I could given the inconsistency of the formatting beforehand. Also I've now added web links to the IRC channels.

I believe this is ready to merge and would substantially improve the Key Projects page when merged, with better basic overviews of many important packaging projects. I respect that this PR is a chance to fix a bunch of formatting and wording stuff in a piece of documentation that hasn't gotten a ton of love lately. May I ask that we try to merge this is, and make follow-up PRs or issues about further improving casing/formatting consistency, advising users in more detail about gotchas, and so on?

source/key_projects.rst
webknjaz: This is questionable: technically, it's useful for making manylinux wheels but `wheel` dist isn't used directly. You could create a wheel using any tool you like and then use `auditwheel` to inspect it and/or make it manylinux compliant. Most of the guides just use `pip wheel` command to make that first step. So whether it's only useful in the context of `wheel` probably depends on the PoV.

ncoghlan: I was going to comment on this in my initial review, but came to the same conclusion as @brainwane: with the current layout of this page, giving auditwheel its own section would look quite odd, whereas it fits better here.

If we were to give it a top level section, I'd suggest doing it in a separate PR where we categorise this list by project type (e.g. installation, build & publishing, package hosting, utility library) rather than splitting it out here.

pradyunsg: > If you would like to block this PR on that, please say so

Nope. :)

PR #692, opened 2020-02-04
Edit piwheels section of key projects

bennuttall:
Thanks to @brainwane for adding piwheels to key projects in #691. Just wanted to propose a couple of small changes:

- un-capitalise Piwheels -> piwheels
- use canonical url "**www**.piwheels.org"
- change "some devs use piwheels as primary index" -> "Raspbian is pre-confiigured to use piwheels as an additional index"

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -485,19 +485,19 @@ create layers of constraints for the program to obey.
 piwheels
 ========
 
-`Website <https://piwheels.org/>`__ |
+`Website <https://www.piwheels.org/>`__ |
 `Docs <https://piwheels.readthedocs.io/>`__ |
 `GitHub <https://github.com/piwheels/piwheels/>`__
 
-Piwheels is a website, and software underpinning it, that fetches
+piwheels is a website, and software underpinning it, that fetches
 source code distribution packages from PyPI and compiles them into
 binary wheels that are optimized for installation onto Raspberry Pi
-computers. Many Raspberry Pi users who use or develop Python tools to
-run on Raspbian use Piwheels as their primary Python package index.
+computers. pip in Raspbian is pre-configured to use piwheels.org as
+an additional index to PyPI.
 
   .. warning::
 
-    Note that Piwheels `does not yet fully support
+    Note that piwheels `does not yet fully support
     <https://github.com/piwheels/piwheels/issues/208>`__ :pep:`503` and
     thus some users have trouble installing certain wheels; this is in
     progress.


source/key_projects.rst
```diff
@@ -485,19 +485,19 @@ create layers of constraints for the program to obey.
...
 `GitHub <https://github.com/piwheels/piwheels/>`__
 
-Piwheels is a website, and software underpinning it, that fetches
+piwheels is a website, and software underpinning it, that fetches
 source code distribution packages from PyPI and compiles them into
 binary wheels that are optimized for installation onto Raspberry Pi
-computers. Many Raspberry Pi users who use or develop Python tools to
-run on Raspbian use Piwheels as their primary Python package index.
+computers. pip in Raspbian is pre-configured to use piwheels.org as
+an additional index to PyPI.
```
webknjaz: Is it additional? I have a feeling that it is the main one and PyPI is additional. It probably goes first because Pip seems to find things there if they exist, which only depends on the loading order...

source/key_projects.rst
```diff
@@ -485,19 +485,19 @@ create layers of constraints for the program to obey.
...
 `Docs <https://piwheels.readthedocs.io/>`__ |
 `GitHub <https://github.com/piwheels/piwheels/>`__
 
-Piwheels is a website, and software underpinning it, that fetches
+piwheels is a website, and software underpinning it, that fetches
 source code distribution packages from PyPI and compiles them into
 binary wheels that are optimized for installation onto Raspberry Pi
-computers. Many Raspberry Pi users who use or develop Python tools to
-run on Raspbian use Piwheels as their primary Python package index.
+computers. pip in Raspbian is pre-configured to use piwheels.org as
```
webknjaz: If I'm not mistaken, Pip preferred spelling is capitalized:
```suggestion
computers. Pip in Raspbian is pre-configured to use piwheels.org as
```

webknjaz: ...or not. But it still seems weird that the first word in the sentence is lowercase

source/key_projects.rst
bennuttall: This is how it's configured:

```
[global]
extra-index-url=https://www.piwheels.org/simple
```

It's only generally "first" because pip prioritises bdist over sdist. If both provided bdist, it would go with PyPI.

source/key_projects.rst
bennuttall: "pip" is not capitalised anywhere else on that page. Yes it does seem weird at the start of a sentence. "conda" starts three sentences and two are capitalised...

uranusjr: The preferred version seems to be “pip” since all occurrences on the [documentation’s main page](https://pip.pypa.io/en/stable/) are lowercased.

source/key_projects.rst
webknjaz: From my experience, if both have exactly the same wheel, piwheels one is selected. I think, I asked @pradyunsg about multiple indexes and he told me that whatever is found first wins, meaning that if indexes are randomly ordered internally, there's no guarantee that one will be prioritized over the other. Hence different people can get slightly different behavior...

uranusjr: The behaviour is not random; `index-url` is always first, `extra-index-url` are selected based on the order of declaration. The above config would choose PyPI over piwheels (if the project publishes pi-compatible wheels).

webknjaz: Well, then there's a bug. Because I hit the behavior I described a few times with Raspbian...

source/key_projects.rst
pradyunsg: Yep -- the preferred capitalization for pip is indeed all lowercase. I want to note that explicitly somewhere in our documentation but haven't spent the time to figure out the right place to do so + do a push to fix incorrect occurrences in various visible places.

source/key_projects.rst
pradyunsg: If you could file a bug with a minimum repro in pip's issue tracker, that'd be great! ^>^

bennuttall: I think it's more likely that was for another reason. If you can reproduce can you share with me? An issue on piwheels would be fine.

webknjaz: @pradyunsg I think that the last time I faced it (https://github.com/cherrypy/cherrypy/issues/1837#issuecomment-579541994) is happening because of that scenario I sent you in Telegram. Not sure about the previous time which was about a year ago, I won't be able to reproduce the conditions now, I think. Maybe it was the same, maybe not.

@bennuttall looks like it's essentially because of  https://github.com/piwheels/piwheels/issues/208. So there's nothing else to file. Except maybe for an issue on the pip tracker with a proposal to improve warnings in this case.

brainwane:
If I get a little time in the next few days I will review this PR. Thanks for making it.

Since the site and this page are already inconsistent in some matters of capitalization, I suggest that reviewers avoid bikeshedding over capitalization in this PR. If PyPUG has a style guide saying, for instance, how to deal with a lower-cased proper noun at the start of a sentence, then let's follow that. If there is no such style guide, let's either file an issue to fix up all of PyPUG to follow the English Wikipedia Manual of Style, or let's file an issue to make a similar document for PyPUG.

PR #695, opened 2020-02-11
Fix invalid regex for entry point names

### LINKED ISSUE #694: Regex for entry point name is invalid
See under [Data model, end of second bullet](https://packaging.python.org/specifications/entry-points/#data-model).

The dash character should be at the beginning or ending of a regex range.

randomir:
Closes  #694.

### DIFF
--- a/source/specifications/entry-points.rst
+++ b/source/specifications/entry-points.rst
@@ -44,7 +44,7 @@ Conceptually, an entry point is defined by three required properties:
   name, the consumer decides how to handle such conflicts. The name may contain
   any characters except ``=``, but it cannot start or end with any whitespace
   character, or start with ``[``. For new entry points, it is recommended to
-  use only letters, numbers, underscores, dashes and dots (regex ``[\w-.]+``).
+  use only letters, numbers, underscores, dots and dashes (regex ``[\w.-]+``).
 
 - The **object reference** points to a Python object. It is either in the form
   ``importable.module``, or ``importable.module:object.attr``. Each of the parts


webknjaz:
Why?

randomir:
@webknjaz, see #694. The regex is not valid.

webknjaz:
I saw that. Still, it doesn't explain why you consider it invalid. Looks ok to me.

randomir:
Try compiling it.

randomir:
Check the [ERE specs, section on Bracket Expression](https://pubs.opengroup.org/onlinepubs/007908799/xbd/re.html#tag_007_003_005), near the end of the section:
> The hyphen character will be treated as itself if it occurs first (after an initial ^, if any) or last in the list, or as an ending range point in a range expression. As examples, the expressions [-ac] and [ac-] are equivalent and match any of the characters a, c or -; [^-ac] and [^ac-] are equivalent and match any characters except a, c or -; the expression [%--] matches any of the characters between % and - inclusive; the expression [--@] matches any of the characters between - and @ inclusive; and the expression [a--@] is invalid, because the letter a follows the symbol - in the POSIX locale. To use a hyphen as the starting range point, it must either come first in the bracket expression or be specified as a collating symbol, for example: [][.-.]-0], which matches either a right bracket or any character or collating element that collates between hyphen and 0, inclusive.

webknjaz:
Okay, I see now. For the future, it's a good idea to explain things in the PR description and/or issue because if it's obvious to you it doesn't mean that it's obvious to everyone else. You just caused me to ask for the explanation twice and only then hinted to run something in the interpreter which could've been included in the very first post and would save us both (and any other reviewer) a lot of time.

I'll post the log here so that nobody would need to repeat these steps to figure out what you mean:
```python
In [1]: import re

In [2]: re.compile('[\w-.]+')
[...]
error: bad character range \w-. at position 1
```

randomir:
@webknjaz, sure I always do that for non-obvious bugs. This one seemed like an obvious typo to me. Sorry.

pradyunsg:
Ahahaha. @di beat me to the merge button. ;)

Issue #619, opened 2019-06-03
Missing directory layout for tests


guettli:
I am missing a directory layout recommendation for tests here:

  https://packaging.python.org/tutorials/packaging-projects/

I know that there are several ways to put you tests. It would be very nice, if there would be an official recommendation for this.

Thank you.

pradyunsg:
Hi @guetti! Thanks for filing this issue!

We just had a long and fairly polarizing discussion on the src/ vs flat layout over at #320, so I'd prefer that we defer this discussion to a later time. Folks like myself are a little less tired of "how to layout your project" discussions right now.

While I do think we should look into this at some point, the aim here too should probably be to document the trade-offs and only after that is done, should we even visit whether we need to provide a recommendation.

ncoghlan:
Similar to flat vs src tree loyouts, in-package vs out-of-package tests is a question of trade-offs. I think the structure we're now proposing for the tree layout documentation in general will also work there - separate guides for both options, paired with a discussion that covers the trade-offs between them.

(If we ever add a testing tutorial though, I'd suggest we use a separate `tests` directory for that - including your tests in your sdist is good, but including them in your wheel archives is dubious, and a separate directory is much easier to leave out of the build process. Tests in a separate directory are also typically easier to run against a fully installed package, rather than only running against a source checkout)

guettli:
@ncoghlan 

> separate guides for both options, paired with a discussion that covers the trade-offs between them.

I think new comers want a to follow a sane default without reading a discussion that covers the trade-offs.

A footnote which links to the details makes sense. 

If you look at this question: https://stackoverflow.com/questions/56426027/pycharm-create-test-target-directory

Imagine you coursor in your IDE/editor is in a python method. Now you want to create a test for this methos. Wouldn't it be great if the IDE/editor could automatically create a basic skeleton (a file at the matching directory tree). An IDE/editor can't parse a discussion that covers the trade-offs. You need a simple and clear rule.

ncoghlan:
Nobody has the authority to dictate a particular answer as the "one true way", because all the options have downsides. 

However, narrowing it to a couple of alternatives is enough for tool creators to pick one that *they* favour, and then they can either document that it's the only approach they support, or provide an option to enable their non-default layout.

Pushing for "there can be only one" at the community turns what could have been productive collaboration on documentation of alternatives and trade-offs into an acrimonious contest of ideas that focuses more on keeping the "wrong" alternative from being selected as the default.

[...]

PR #698, opened 2020-02-18
Add unit test folder

davaya:
Include an empty test folder in the tutorial package layout.  Addresses Issue #619.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -44,9 +44,15 @@ them in the following steps.
     packaging_tutorial/
       example_pkg/
         __init__.py
+      test/
       setup.py
       LICENSE
       README.md
+
+
+Creating a test folder
+----------------------
+:file:`test/` is a placeholder for unit test files. Leave it empty for now.
 
 
 Creating setup.py
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -44,7 +44,7 @@ them in the following steps.
     packaging_tutorial/
       example_pkg/
         __init__.py
-      test/
+      tests/
       setup.py
       LICENSE
       README.md
@@ -52,7 +52,8 @@ them in the following steps.
 
 Creating a test folder
 ----------------------
-:file:`test/` is a placeholder for unit test files. Leave it empty for now.
+
+:file:`tests/` is a placeholder for unit test files. Leave it empty for now.
 
 
 Creating setup.py


source/tutorials/packaging-projects.rst
```diff
@@ -44,11 +44,17 @@ them in the following steps.
...
         __init__.py
+      test/
       setup.py
       LICENSE
       README.md
 
 
+Creating a test folder
+----------------------
+:file:`test/` is a placeholder for unit test files. Leave it empty for now.
```
pradyunsg:
```suggestion

:file:`tests/` is a placeholder for unit test files. Leave it empty for now.
```

source/tutorials/packaging-projects.rst
```diff
@@ -44,11 +44,17 @@ them in the following steps.
     packaging_tutorial/
       example_pkg/
         __init__.py
+      test/
```
pradyunsg:
```suggestion
      tests/
```

davaya:
> This will instantly turn into a massive bikeshedding exercise ...

I envisioned https://www.cbsnews.com/news/brooklyn-bike-race-red-hook-criterium-sees-large-pileup-at-start/, but discovered Parkinson's law of triviality.  Thanks for a new word for the day.

pfmoore:
> Thanks for a new word for the day.

Cool - you're one of today's [lucky 10000](https://xkcd.com/1053/) 🙂

webknjaz:
Should this mention that there's also a legitimate reason to put `tests/` under the package for distributing as an sdist?

pfmoore:
IMO no. If we're not willing to agree on a placeholder location in the very initial example, then so be it, we should do nothing and keep the status quo. But @davaya made the point in https://github.com/pypa/packaging.python.org/issues/619#issuecomment-587567723 that it would be good to encourage newcomers to include some sort of tests from the start.

webknjaz:
@pfmoore FWIW I agree that most of the time `tests/` should be in the root folder. Just wanted to explore an opportunity to improve the wording maybe.

pfmoore:
Do you have a suggestion for what to add? I'm open to consider specific wording, I just don't want to get into broad debates (and I don't know how I'd word such a comment myself).

source/tutorials/packaging-projects.rst
```diff
@@ -44,11 +44,18 @@ them in the following steps.
...
+      tests/
       setup.py
       LICENSE
       README.md
 
 
+Creating a test folder
+----------------------
+
+:file:`tests/` is a placeholder for unit test files. Leave it empty for now.
```
webknjaz:
```suggestion
:file:`tests/` is a placeholder for unit test files. Leave it empty for now.

  .. note::

     In *rare* cases, you may meet some projects that put the ``tests/`` folder under
     ``example_pkg/``. They probably use `pytest's pyargs`_. While there may be a
     legitimate reason to do so, you'll likely never need it. So for the sake of
     simplicity, this guide will stick with the most common case of a top-level
     ``tests/`` dir (in plural).

     .. _pytest's pyargs:
     https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code
```

@pfmoore ^

webknjaz: Nitpick: I'd probably remove "unit" from there. Because other types of tests exist and can be put there.

pfmoore: That seems too wordy to me, in that it obscures the key point. Maybe just 

```
:file:`tests/` is a placeholder for unit test files (other test layouts are possible, but this is a good starting point). Leave it empty for now.
```

?

webknjaz: Yeah, that sounds good. Should we keep the link?

```suggestion
:file:`tests/` is a placeholder for test files (`other test layouts`_ are possible, but this is a good starting point). Leave it empty for now.

.. _other test layouts:
   https://docs.pytest.org/en/latest/goodpractices.html#tests-as-part-of-application-code
```

@pfmoore ^

pfmoore: I'd argue no. If we link to a discussion of that layout, we should also link to a discussion of the top-level `tests` layout, and once again we spiral into bikeshedding.

ncoghlan:
I went ahead and merged the simple version (like @pf_moore, I think @davaya made a very solid point in https://github.com/pypa/packaging.python.org/issues/619#issuecomment-587567723).

We can continue discussion in #619 on how we can build on this to introduce the topics of:
* ensuring sdists include tests, not just the installed package
* ensuring the installed package works, not just the files in the source repo
* keeping built wheel files efficient by omitting unnecessary files (such as tests)

PR #699, opened 2020-03-01
Update version sourcing example from pip setup.py

lucatrv:


di:
Hi @lucatrv, thanks for the PR. What's the motivation for this change? Just removing the regular expression?

lucatrv:
That's the main reason (so it does not have the "disadvantage of having to deal with complexities of regular expressions"). The second reason is that it is the current implementation in `pip` `setup.py`.

di:
Ah, gotcha, sorry I missed that this example was taken from the `pip` project!

source/guides/single-sourcing-package-version.rst
```diff
@@ -8,33 +8,31 @@ Single-sourcing the package version
...
+                if line.startswith('__version__'):
+                    delim = '"' if '"' in line else "'"
+                    return line.split(delim)[1]
+            else:
+                raise RuntimeError("Unable to find version string.")
 
         setup(
            ...
-           version=find_version("package", "__init__.py")
+           version=get_version("src/package/__init__.py")
```
di:
```suggestion
           version=get_version("package/__init__.py")
```
Let's keep this aligned with the structure from the previous example.

lucatrv: Sure!

lucatrv:
OK @di, I amended the commit with your suggestion.

di:
Thanks!

PR #706, opened 2020-03-12
Describe using importlib-metadata as an alternative to pkg_resources for single-sourcing versions

jwodder:
Item 5 in ["Single-sourcing the package version"](https://packaging.python.org/guides/single-sourcing-package-version/) currently presents its approach using `pkg_resources`.  However, `pkg_resources` is inefficient (and possibly deprecated?), and more efficient & modern alternatives to it are being developed in the form of `importlib.resources` and `importlib.metadata`.  This PR extends item 5 to also show how to achieve the same thing using `importlib.metadata`.  It also adds a note that the `pkg_resources` version requires `setuptools` as a runtime dependency.

I expect there will be some debate over whether to present both methods side by side or whether to delete the `pkg_resources` method in favor of `importlib.metadata`.  I'm open to both possibilities.

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -86,7 +86,37 @@ number of your project:
     installation metadata, which is not necessarily the code that's currently
     imported.
 
+    Note that if the project uses ``pkg_resources`` to fetch its own version at
+    runtime, then ``setuptools`` (the project that provides ``pkg_resources``)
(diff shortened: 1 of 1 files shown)


jwodder:
@xavfernandez @pfmoore `importlib-metadata` section moved before the `pkg_resources` section.

pradyunsg:
Thanks for the PR @jwodder, and the reviews @pfmoore and @xavfernandez!

PR #708, opened 2020-03-13
Organize index of guides into 3 categories

pastelmind:
Split the toctree of guides into three broad categories:

- Installing Packages
    - Includes guides for installing and managing dependencies.
- Building and Publishing Projects
    - Includes guides for project configuration, distribution
- Miscellaneous
    - Guides that do not clearly fit into any of the above

This commit primarily affects the Guides index page; it does not create nested lists for each category in the sidebar.

### DIFF
--- a/source/guides/index.rst
+++ b/source/guides/index.rst
@@ -7,13 +7,18 @@ introduction to packaging, see :doc:`/tutorials/index`.
 
 .. toctree::
    :maxdepth: 1
+   :caption: Installing Packages:
 
-   tool-recommendations
    installing-using-pip-and-virtual-environments
    installing-stand-alone-command-line-tools
    installing-using-linux-tools
    installing-scientific-packages
    multi-version-installs
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Building and Publishing Projects:
+
    distributing-packages-using-setuptools
    using-manifest-in
    single-sourcing-package-version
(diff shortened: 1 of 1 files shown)


webknjaz (approved):

source/guides/index.rst
```diff
@@ -23,10 +28,16 @@ introduction to packaging, see :doc:`/tutorials/index`.
...
    publishing-package-distribution-releases-using-github-actions-ci-cd-workflows
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Miscellaneous:
+
+   tool-recommendations
+   analyzing-pypi-package-downloads
+   index-mirrors-and-caches
+   hosting-your-own-index
```
di, line 43: I think these two should be under "installing".

pastelmind: Good point. If I update the PR, do I have to wait another 9 months to get another review?

pradyunsg: No.

pastelmind:
@di Moved the two items. I'm sorry about making a snarky comment earlier.

pradyunsg:
Comparing the sidebar between https://packaging.python.org/ and https://python-packaging-user-guide--708.org.readthedocs.build/, I'd prefer to keep the sidebar unchanged.

Is there some change that I'm missing?

di:
@pradyunsg Looks like that was due to the branch being out of date.

di:
Thanks @pastelmind!

PR #710, opened 2020-03-13
Update flit description & add link to rationale

### LINKED ISSUE #702: Add Flit rationale to key projects entry
This is a good first issue for someone: add a link to https://flit.readthedocs.io/en/latest/rationale.html within https://packaging.python.org/key_projects/#flit .

https://github.com/pypa/packaging.python.org/blob/master/source/key_projects.rst#flit is the file you would edit to do this. Perhaps the wording could be like:

> Flit is a simple way to put Python packages and modules on PyPI ([the maintainer's rationale for the tool](https://flit.readthedocs.io/en/latest/rationale.html)).

cc @takluyver

pastelmind:
Rewrite Flit's description to reorganize sentences, improve wording, increase readability. The 1st paragraph describes what Flit is and what it can do. The 2nd paragraph describes how it works and what is needed (or not) to use Flit.

Also add a link to Flit's rationale ("[Why use Flit?](https://flit.readthedocs.io/en/latest/rationale.html)"), as requested by #702. (This was the original goal, but I figured a full rewrite was needed anyway.)

Fixes #702.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -373,8 +373,7 @@ flit
 Flit provides a simple way to upload pure Python packages and modules to PyPI.
 It focuses on `making the easy things easy <flit-rationale_>`_ for packaging.
 Flit can generate a configuration file to quickly set up a simple project, build
-source distributions and wheels, and upload them to PyPI. Flit can also produce
-wheels in a reproducible manner, which provides added verifiability.
+source distributions and wheels, and upload them to PyPI.
 
 Flit uses ``pyproject.toml`` to configure a project. Flit does not rely on tools
 such as :ref:`setuptools` to build distributions, or :ref:`twine` to upload them
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -370,19 +370,18 @@ flit
 `Issues <https://github.com/takluyver/flit/issues>`__ |
 `PyPI <https://pypi.org/project/flit>`__
 
-Flit is a simple way to put Python packages and modules on PyPI. Flit packages
-a single importable module or package at a time, using the import name as the
-name on PyPI. All subpackages and data files within a package are included
-automatically. Flit requires Python 3, but you can use it to distribute modules
-for Python 2, so long as they can be imported on Python 3.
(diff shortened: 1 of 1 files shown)


source/key_projects.rst
```diff
@@ -370,19 +370,18 @@ flit
...
-upload functionality. You can use it to quickly set up package
-configuration files for simple packages (instead of manually writing a
-file that works with :ref:`setuptools`), and build and upload packages to
-PyPI. All wheels built by flit are reproducible, which provides added
-verifiability.
+Flit provides a simple way to upload pure Python packages and modules to PyPI.
+It focuses on `making the easy things easy <flit-rationale_>`_ for packaging.
+Flit can generate a configuration file to quickly set up a simple project, build
+source distributions and wheels, and upload them to PyPI. Flit can also produce
+wheels in a reproducible manner, which provides added verifiability.
```
takluyver: I view the reproducible builds thing as (for now) a fairly niche feature, which I wouldn't mention in a summary like this. Maybe one day everyone will want reproducible builds, but today I don't think it's a priority.

If you do prefer to leave it in, 'provides added verifiability' sounds a bit corporate - maybe something like 'making it easier to be sure what they contain'? To be honest, I can't come up with a simple sentence I like for this, which is part of why I think it's best left out.

pastelmind: The sentence on reproducible wheels was added in #691 (170d7b9b4fd6d0744826720b5f733c76cfedda1c). I didn't feel qualified to decide whether to remove it, as I am unfamiliar with reproducible builds in general.

I agree that it is a niche feature; the intended audience of Flit (=beginners) probably won't be looking for it. Can I delete the line?

takluyver:
Ah, right, I'd looked too quickly and missed that it was already there. No
objections from me to removing that bit, but the packaging guide isn't my
project.

source/key_projects.rst
pastelmind: I removed the line.

brainwane:
I liked my wording but am fine with yours. :-)

Issue #696, opened 2020-02-12
Packaging instructions are not optimal about README.MD


vanrein:
Thank you for providing feedback on Python packaging!

To help us help you, please fill out as much of the following as you can. If a question is not relevant, feel free to skip it.

1. What is your operating system and version?

MacOS 10.15.3

2. What is your Python version?

Python 3.7.3

3. What version of pip do you have?

pip 19.0.3

4. If following an online tutorial or guide, please provide a link to the page or section giving you trouble:

https://packaging.python.org/tutorials/packaging-projects/#creating-setup-py

5. Could you describe your issue in as much detail as possible?

The instructions provide an adagio for a long readme and claim it is normal:

```
with open("README.md", "r") as fh:
    long_description = fh.read()
```

This however, will only run from the directory containing this file.  It is better to construct a path:

```
here = path.dirname (path.realpath (__file__))
readme = open (path.join (here, 'README.MD')).read ()
```

Both forms still leave a problem.  When `setup.py` is run during `pip install`, it will try to open `README.md` once more, while it is being installed.  (This was after uploading an `sdist`, not a `bdist`.)

```
Collecting arpa2.wsgi

  Downloading https://files.pythonhosted.org/packages/71/87/fe38499da24d4bfcd04d301aa20a03af35bfdf379e8b739069b8a726af2a/arpa2.wsgi-0.4.0.tar.gz

    Complete output from command python setup.py egg_info:

    Traceback (most recent call last):

      File "<string>", line 1, in <module>

      File "/private/var/folders/5_/dsls_kvx7tq171bf6chfm3x80000gn/T/pip-install-zcq3d6v9/arpa2.wsgi/setup.py", line 19, in <module>

        readme = open (path.join (here, 'README.MD')).read ()

    FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/5_/dsls_kvx7tq171bf6chfm3x80000gn/T/pip-install-zcq3d6v9/arpa2.wsgi/README.MD'
```

As far as I can see, I am not supposed to include the `README.MD` and it is only needed because of the packaging instruction -- which may be off.  Not sure what a good fix would be, otherwise I'd suggest it.

di:
I agree that a proper path to the README would be idea. Here's what we suggest on https://github.com/pypa/sampleproject:

```python
here = path.abspath(path.dirname(__file__))

# Get the long description from the README file
with open(path.join(here, 'README.md'), encoding='utf-8') as f:
[...]

PR #711, opened 2020-03-13
Explain why README.md should be added to sdist

pastelmind:
~~Update sample setup.py to correctly load README.md when cwd is not the same directory that setup.py lives in (discussed in #696). The updated README-loading code is for Python 3+. Since Python 2 has been deprecated, it is probably safe to assume that anyone reading this guide is most likely using Python 3.~~ The guide has been updated significantly since the inception of this PR, which obviates the `setup.py` changes. Also see https://github.com/pypa/packaging.python.org/pull/711#issuecomment-783835643

Also explain why README.md should be added to the package (discussed in #696).

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -310,6 +310,13 @@ if you'd like.
     to write your content.
 
 
+Because our build script loads :file:`README.md` to provide a ``long_description``
+for :func:`setup`, the :file:`README.md` must be included along with your code
+when you `generate a source distribution <generating archives>`_.
+:ref:`setuptools` 36.4.0 or above will automatically include :file:`README.md`
+if it exists.
+
+
 Creating a LICENSE
 ------------------
 


webknjaz:
@pastelmind hey, do you mind synching this PR with the `master`? It now has conflicts :(

pastelmind:
I'm busy atm. I'll see what I can do after today.

pastelmind:
@webknjaz I merged `master` into my branch and resolved the conflicts

source/tutorials/packaging-projects.rst
```diff
@@ -67,10 +67,14 @@ Open :file:`setup.py` and enter the following content. Update the package name t
...
+    import os
     import setuptools
 
-    with open("README.md", "r", encoding="utf-8") as fh:
-        long_description = fh.read()
+    here = os.path.abspath(os.path.dirname(__file__))
+
+    # Get the long description from the README file
+    with open(os.path.join(here, "README.md"), encoding="utf-8") as f:
+        long_description = f.read()
```
merwok, line 77: IMO this is more complicated than it should be.

- `setup.py` is intended to always run from its parent directory (other things break if not), so using a relative path is fine
- but these days we can have setupptols use a file as long description directly, supporting reST or markdown, so that should be the easy option documented in tutorials.

pastelmind: * Good point. I will drop the `abspath()` call.
* I didn't know that was possible. Can I find documentation for this feature? When was it introduced?

pfmoore: https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#metadata

pastelmind:
@pfmoore Thank you. The `file:` directive greatly simplifies the code.

webknjaz:
@pastelmind could you please resolve the merge conflict?

source/tutorials/packaging-projects.rst
```diff
@@ -69,16 +69,13 @@ Open :file:`setup.py` and enter the following content. Update the package name t
...
-        long_description = fh.read()
-
     setuptools.setup(
         name="example-pkg-YOUR-USERNAME-HERE", # Replace with your own username
         version="0.0.1",
         author="Example Author",
         author_email="author@example.com",
         description="A small example package",
-        long_description=long_description,
+        long_description="file:README.md",
```
webknjaz, line 78: This should be left unchanged: directives like `file:` only work in declarative config (`setup.cfg`) and this doesn't work in Python script (`setup.py`).

merwok: Right!  The document that was linked has most info in setup.cfg and a near-empty setup.py

pastelmind: @webknjaz You are right. I checked by uploading a [package on TestPyPI](https://test.pypi.org/project/cookiecutter-test/)...and `long_description="file:README.rst"` does not do what I thought it would do. I will revert the changes and fix the merge conflicts as well.

pastelmind:
Because `master`/`main` has advanced significantly since this PR was created, I rebased the original commit (8f7d424b6f7c6c53b145baf44112ef8777930209) onto the current `main`. In doing so, I excluded the `setup.py` changes, leaving only the explanation for `README.md`. As such, it also drops any changes made in the following commits:

- 1a739da5d01d613eacdc2eb0cd9f03dd161d63d7 Use relative path to get module directory
- 0e1d9346ab5b53844c2227c4588ff42278fc373e Remove unnecessary os import
- f7106c92765e0c581ebeb61ab41917afab9855f9 Use file: directive to specify README path

webknjaz (approved):
Perfect, thanks!

PR #712, opened 2020-03-16
Suggest read_version library for single-sourcing option #1

jwodder:
This PR adds a note to the regex-based option for single-sourcing project versions that the `read_version` library provides a function that takes care of the logic for you.

Full disclosure: I am the author of `read_version`.


jwodder:
Withdrawn in favor of #758

PR #718, opened 2020-04-16
packaging-projects tutorial: Update find_packages call to exclude tests folder

Kwpolska:
Without this, the test suite could be globally installed in site-packages, as `import tests`.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -81,7 +81,7 @@ Open :file:`setup.py` and enter the following content. Update the package name t
         long_description=long_description,
         long_description_content_type="text/markdown",
         url="https://github.com/pypa/sampleproject",
-        packages=setuptools.find_packages(),
+        packages=setuptools.find_packages(exclude=('tests', 'tests.*')),
         classifiers=[
             "Programming Language :: Python :: 3",
             "License :: OSI Approved :: MIT License",


di:
I'm not sure this is necesary: the `tests` directory shouldn't be a module (shouldn't have an `__init__.py` file in it's route. Without that, `find_packages` won't include that directory.

Kwpolska:
> (shouldn't have an `__init__.py` file in it's route)

Sometimes it’s necessary, or can help with better code organization, and there are some examples in [pytest docs](https://docs.pytest.org/en/latest/goodpractices.html) that use this scheme.

jwodder:
Cf. this similar PR that was rejected in favor of using the `include` keyword instead: https://github.com/pypa/packaging.python.org/pull/641

> I'm not sure this is necesary: the `tests` directory shouldn't be a module (shouldn't have an `__init__.py` file in it's route).

There's a big, big gulf here between what `tests` *shouldn't* be and what it actually *is* for many projects.  The wheelodex.org database (as of 2020-06-22) lists 4405 projects that include a `tests/__init__.py` file in their wheel.

pfmoore:
A tutorial *should* be opinionated and direct people into a particular usage. The position of the tutorial is that you shouldn't have an `__init__.py` file in a tests directory  (a view I agree with, FWIW). Once people get beyond the point where they are using tutorials, *then* they can look at the trade-offs and form their own opinions.

jwodder:
> The position of the tutorial is that you shouldn't have an `__init__.py` file in a tests directory (a view I agree with, FWIW).

Then perhaps the tutorial should actually state that somewhere in lieu of this PR.

danuker:
> you shouldn't have an __init__.py

I can't import my module from the tests if I don't have an `__init__.py` file in their directory (which is side-by-side with the `example_pkg` directory). Am I doing something wrong?

jwodder:
@danuker The presence of an `__init__.py` in `tests/` should have no effect on the ability to do `import mymodule` in test code.  What test framework are you using?  Can you link your repository?

danuker:
@jwodder I use `pytest`. Removing `tests/__init__.py` makes the tests un-runnable (at least on Python 3.8.6). Here is the repo: https://github.com/danuker/symreg

They fail to run even without `pytest`; `python tests/test_symreg.py` also fails, without using `pytest`.

```
$ python tests/test_symreg.py 
  File "tests/test_symreg.py", line 4, in <module>
    from symreg.regressor import Regressor
ModuleNotFoundError: No module named 'symreg'
```

danuker:
If I correctly packaged [my project](https://github.com/danuker/symreg), I would advise that you also add an `__init__.py` in the `tests` dir, for the "Create the new files listed below" list. It was very difficult to find out what was wrong.

pfmoore:
@danuker It looks like you're trying to run your tests without having installed your project. That's why the import of your project is failing.

danuker:
That is correct, I only got to the "arrange into directory structure" section of the guide. Why would I want to install my project in order to develop it?

danuker:
@pfmoore Also, would that mean that I would be testing the installed version, instead of the local directory? Additional confusion.

jwodder:
@danuker Yes, you *should* test against an installed version of your package rather than the local version.  This allows you to catch mistakes in packaging.  The standard tool for running tests against an installed version of your project is [tox](https://tox.readthedocs.io/en/latest/).

danuker:
@jwodder I see. Now it makes sense. Thank you very much for explaining. I believe the guide should talk about this. This area of Python is not straightforward (including the import errors).

bhrutledge:
Since https://github.com/pypa/packaging.python.org/issues/619 is already a discussion of if/how to include tests in the packaging projects tutorial, in favor of reviving that discussion.

PR #734, opened 2020-05-30
Add .pypirc specification

### LINKED ISSUE #730: Missing documention for .pypirc format
Would this be the correct place for this issue? https://github.com/pypa/twine/issues/638

> ```console
> u@h:~/pak$ twine upload .tox/dist/* -r test-pypi
> InvalidConfiguration: Missing 'test-pypi' section from the configuration file
> or not a complete URL in --repository-url.
> Maybe you have a out-dated '~/.pypirc' format?
> more info: https://docs.python.org/distutils/packageindex.html#pypirc
> 
> ```
> 
> Last v3 url that's valid is https://docs.python.org/3.6/distutils/packageindex.html#pypirc (and Google offers https://docs.python.org/3.3/distutils/packageindex.html#pypirc)

bhrutledge:
Closes #730.

I'm starting this out as a draft PR, because I think a complete PR would include updates to other documents that reference `.pypirc`. However, before spending time on that, I'd like get the content of this document approved.

### DIFF
--- a/source/specifications/pypirc.rst
+++ b/source/specifications/pypirc.rst
@@ -1,0 +1,122 @@
+
+.. _pypirc:
+
+========================
+The :file:`.pypirc` file
+========================
+
+**TODO** Link this document from references to :file:`.pypirc` in other docs
+
+**TODO** Normalize format of :file:`.pypirc` examples in other docs
+
+Creating a :file:`$HOME/.pypirc` file allows you to define the configuration for
+:term:`package indexes <Package Index>` (referred to here as "repositories"), so
+that you don't have to enter the URL, username, or password whenever you upload
+a package with :ref:`twine` or :ref:`flit`.
+
+The format (originally defined by the :ref:`distutils` package) is:
+
+.. code-block:: ini
+
+    [distutils]
+    index-servers =
+        first-repository
+        second-repository
+
+    [first-repository]
+    repository = <first-repository URL>
+    username = <first-repository username>
+    password = <first-repository password>
+
+    [second-repository]
+    repository = <second-repository URL>
+    username = <second-repository username>
+    password = <second-repository password>
+
+The ``distutils`` section defines an ``index-servers`` setting that lists the
+name of all sections describing a repository.
+
+Each section describing a repository defines three settings:
+
+- ``repository``: The URL of the repository.
+- ``username``: The registered username on the repository.
+- ``password``: The password that will used to authenticate the username.
+
+.. warning::
+
+    Be aware that this stores your password in plain text. For better security,
+    consider an alternative like `keyring`_, setting environment variables, or
+    providing the password on the command line.
+
+.. _keyring: https://pypi.org/project/keyring/
+
+Using :file:`.pypirc` with Twine
+--------------------------------
+
+Twine's default configuration defines repository settings for PyPI and TestPyPI:
+
+.. code-block:: ini
+
+    [distutils]
+    index-servers =
+        pypi
+        testpypi
+
+    [pypi]
+    repository = https://upload.pypi.org/legacy/
+
+    [testpypi]
+    repository = https://test.pypi.org/legacy/
+
+This means that you can set your `API token`_ for PyPI by creating a
+:file:`$HOME/.pypirc` similar to:
+
+.. code-block:: ini
+
+    [pypi]
+    username = __token__
+    password = <PyPI token>
+
+To configure an additional repository, you will need to redefine the
+``index-servers`` setting to include the repository name. Here is a complete
+example of a :file:`$HOME/.pypirc` for PyPI, TestPyPI, and a private repository:
+
+.. code-block:: ini
+
+    [distutils]
+    index-servers =
+        pypi
+        testpypi
+        private-repository
+
+    [pypi]
+    username = __token__
+    password = <PyPI token>
+
+    [testpypi]
+    username = __token__
+    password = <TestPyPI token>
+
+    [private-repository]
+    repository = <private-repository URL>
+    username = <private-repository username>
+    password = <private-repository password>
+
+.. warning::
+
+    Instead of using the ``password`` setting, consider saving your API tokens
+    and passwords securely using `keyring`_ (which is installed by Twine):
+
+    .. code-block:: bash
+
+        keyring set https://upload.pypi.org/legacy/ __token__
+        keyring set https://test.pypi.org/legacy/ __token__
+        keyring set <private-repository URL> <private-repository username>
+
+Twine defaults to uploading to PyPI. To upload to the private repository, run:
+
+.. code-block:: bash
+
+    twine upload --repository private-repository dist/*
+
+.. _API token: https://pypi.org/help/#apitoken
--- a/source/specifications/pypirc.rst
+++ b/source/specifications/pypirc.rst
@@ -5,10 +5,6 @@
 The :file:`.pypirc` file
 ========================
 
-**TODO** Link this document from references to :file:`.pypirc` in other docs
-
-**TODO** Normalize format of :file:`.pypirc` examples in other docs
-
 A :file:`.pypirc` file allows you to define the configuration for :term:`package
 indexes <Package Index>` (referred to here as "repositories"), so that you don't
 have to enter the URL, username, or password whenever you upload a package with
--- a/source/specifications/pypirc.rst
+++ b/source/specifications/pypirc.rst
@@ -47,13 +47,14 @@ Each section describing a repository defines three fields:
 .. _keyring: https://pypi.org/project/keyring/
 
 Common configurations
----------------------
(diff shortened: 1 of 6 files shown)


source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+.. warning::
+
+    Be aware that this stores your password in plain text. For better security,
+    consider an alternative like `keyring`_, setting environment variables, or
+    providing the password on the command line.
+
+.. _keyring: https://pypi.org/project/keyring/
+
+Using :file:`.pypirc` with Twine
+--------------------------------
```
bhrutledge, line 54: I'm not convinced that this is the right place for Twine-specific instructions, but I decided to start by documenting everything that felt necessary in one place, with the notion that it could be split out later if desired

stdedos:
LGTM from user's perspective.

Also clears up how the "don't save your token in plaintext" wizardry works

bhrutledge:
To see the rendered version without building locally, check out:

https://bhrutledge-ppug.bhrutledge.now.sh/specifications/pypirc.html

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+    [testpypi]
+    username = __token__
+    password = <TestPyPI token>
+
+    [private-repository]
+    repository = <private-repository URL>
+    username = <private-repository username>
+    password = <private-repository password>
+
+.. warning::
```
stdedos, line 105: Might've changed this to Info, and not "a Warning".

I was warned already a couple of lines ago.

stdedos:
Thank you for the extra mile!

Also in-line rendering looks good enough to get rendered text display
image

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+- ``username``: The registered username on the repository.
+- ``password``: The password that will used to authenticate the username.
+
+.. warning::
+
+    Be aware that this stores your password in plain text. For better security,
+    consider an alternative like `keyring`_, setting environment variables, or
+    providing the password on the command line.
+
+.. _keyring: https://pypi.org/project/keyring/
```
bhrutledge, line 51: Should Keyring be added to https://packaging.python.org/key_projects/#non-pypa-projects?

bhrutledge:
@di I hope you'll forgive the nudge, but I'd appreciate your thoughts on this before I link to it from other parts of the guide. And, ultimately, it'd be nice to resolve https://github.com/pypa/twine/pull/340.

di:
@bhrutledge On my list for today.

source/specifications/pypirc.rst
di: It's not really a "packaging" project, so I'd say no.

source/specifications/pypirc.rst
di: I think maybe we can think of this as "Behavior of popular tools that use the `.pypirc` format" instead and describe how Twine uses it here.

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+
+========================
+The :file:`.pypirc` file
+========================
+
+**TODO** Link this document from references to :file:`.pypirc` in other docs
+
+**TODO** Normalize format of :file:`.pypirc` examples in other docs
+
+Creating a :file:`$HOME/.pypirc` file allows you to define the configuration for
```
di, line 12: Is `$HOME/.pypirc` more clear than `~/.pypirc`? Or just a "Location of `.pypirc` file" section that says "expected to be in your home directory (`$HOME` or `~/`)?

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+    password = <PyPI token>
+
+    [testpypi]
+    username = __token__
+    password = <TestPyPI token>
+
+    [private-repository]
+    repository = <private-repository URL>
+    username = <private-repository username>
+    password = <private-repository password>
```
di, line 103: These parts seem to probably not be unique to twine? I would expect this to work for any tool that implements the original distutils `.pypirc`.

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. warning::
+
+    Instead of using the ``password`` setting, consider saving your API tokens
```
di, line 107:
```suggestion
    Instead of using the ``password`` field, consider saving your API tokens
```

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,122 @@
...
+
+        keyring set https://upload.pypi.org/legacy/ __token__
+        keyring set https://test.pypi.org/legacy/ __token__
+        keyring set <private-repository URL> <private-repository username>
+
+Twine defaults to uploading to PyPI. To upload to the private repository, run:
+
+.. code-block:: bash
+
+    twine upload --repository private-repository dist/*
```
di, line 120: This does seem too twine-specific.

source/specifications/pypirc.rst
stdedos: That could even follow norm configuration inheritance rules (`cwd`, `cwd/..`, `~`, `/etc/), so that APIkey-per-project makes sense

source/specifications/pypirc.rst
bhrutledge: It doesn't look like "info" is an option. It could be "note", but I feel like the security implications warrant a second warning.

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,128 @@
...
+.. warning::
+
+    Be aware that this stores your password in plain text. For better security,
+    consider an alternative like `keyring`_, setting environment variables, or
+    providing the password on the command line.
+
+.. _keyring: https://pypi.org/project/keyring/
+
+Common configurations
+---------------------
```
bhrutledge, line 50: Following up on https://github.com/pypa/packaging.python.org/pull/734#pullrequestreview-421397624

> I think maybe we can think of this as "Behavior of popular tools that use the `.pypirc` format" instead and describe how Twine uses it here.

I tried to split the difference by renaming the section and adding a note about other tools, but still giving concrete examples of what I think is the most common current use case (i.e., using Twine). That includes the `twine upload` at the end.

I also discovered that flit only [defines defaults](https://github.com/takluyver/flit/blob/3e6a9a1431ecb50ba839e199e3eda31fbbcfbc7d/flit/upload.py#L29-L59) for PyPI (i.e., not TestPyPI).

source/specifications/pypirc.rst
bhrutledge: I reworked this a bit in https://github.com/pypa/packaging.python.org/pull/734/commits/55ac803ecb375f33e45de1a3f1e45cf0a4dbc6b4, so that the specification is location-agnostic, and `$HOME/.pypirc` is left to the concrete examples later.

Other docs in this repo already used `$HOME/.pypirc`, so I left it there.

Also, @stdedos, Twine doesn't do configuration inheritance; see https://github.com/pypa/twine/issues/349.

source/specifications/pypirc.rst
```diff
@@ -0,0 +1,128 @@
...
+    repository = https://upload.pypi.org/legacy/
+
+    [testpypi]
+    repository = https://test.pypi.org/legacy/
+
+Twine will add additional configuration from :file:`$HOME/.pypirc`, the command
+line, and environment variables to this default configuration.
+
+So, to set your `API token`_ for PyPI, you can create a :file:`$HOME/.pypirc`
+similar to:
```
di, line 78:
```suggestion
To set your `API token`_ for PyPI, you can create a :file:`$HOME/.pypirc`
similar to:
```

I think this part on. API tokens probably deserves it's own header/subsection as well.

source/specifications/pypirc.rst
di: Going to reiterate that I think this is starting to document `twine`, not `.pypirc`. Maybe we can link to twine docs here instead?

merwok:
For reference, here is the distutils documentation about .pypirc before it was deleted from the CPython repo: https://github.com/python/cpython/blob/v3.4.10/Doc/distutils/packageindex.rst#the-pypirc-file

source/specifications/pypirc.rst
bhrutledge: Added in https://github.com/pypa/packaging.python.org/pull/734/commits/0ea358904a2b7990ad339f5faaa6eab82185d020, which also tweaks the "Note" in attempt to avoid documenting Twine usage.

bhrutledge:
> For reference, here is the distutils documentation about .pypirc before it was deleted from the CPython repo

Thanks! I actually started with that, then updated/expanded it.

di:
Any other @pypa/packaging-user-guide-editors want to review?

PR #747, opened 2020-06-24
Clarify wording about length of Project-URL

sersorrel:
It's not clear what a "sign" is meant to be – "character" isn't unambiguous, but it at least matches the text of the PEP.


di:
Thanks!

PR #750, opened 2020-06-25
Updated if condition as per ticket

### LINKED ISSUE #742: Wrong if condition to establish if commit is tagged
here, https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/, in the section: Publishing the distribution to PyPI and TestPyPI,
the line **if: startsWith(github.event.ref, 'refs/tags')** should be **if: startsWith(github.ref, 'refs/tags')**

SaltyCatFish:
Removed `event` from `github.event.ref` as per ticket.

Fixes #742.


di:
Thanks!

PR #755, opened 2020-07-16
Update "project" to mention pyproject.toml, not just setuptools

encukou:
Also, add the pyproject.toml and setup.cfg markers as other possible "marker" files for a "practical way to define projects".

---

The [project](https://packaging.python.org/glossary/#term-project) entry is currently outdated. Here's my suggestion for updating it.
(Is a PR is the right way to suggest the change, or should I discuss on Discourse first?)

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -102,11 +102,10 @@ Glossary
         packaged into a :term:`Distribution <Distribution Package>`.
 
         Since most projects create :term:`Distributions <Distribution Package>`
-        using :ref:`distutils` or :ref:`setuptools`, another practical way to
-        define projects currently is something that contains a :term:`setup.py`
-        at the root of the project src directory, where "setup.py" is the
-        project specification filename used by :ref:`distutils` and
-        :ref:`setuptools`.
+        using either :pep:`518` ``build-system``, :ref:`distutils` or
+        :ref:`setuptools`, another practical way to define projects currently
+        is something that contains a :term:`pyproject.toml`, :term:`setup.py`,
+        or :term:`setup.cfg` file at the root of the project source directory.
 
         Python projects must have unique names, which are registered on
         :term:`PyPI <Python Package Index (PyPI)>`. Each project will then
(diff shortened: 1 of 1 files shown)


source/glossary.rst
```diff
@@ -102,11 +102,10 @@ Glossary
         packaged into a :term:`Distribution <Distribution Package>`.
 
         Since most projects create :term:`Distributions <Distribution Package>`
-        using :ref:`distutils` or :ref:`setuptools`, another practical way to
-        define projects currently is something that contains a :term:`setup.py`
-        at the root of the project src directory, where "setup.py" is the
-        project specification filename used by :ref:`distutils` and
-        :ref:`setuptools`.
+        using either :pep:`518` ``build-system``, :ref:`distutils` or
+        :ref:`setuptools`, another practical way to define projects currently
```
webknjaz, line 106: I'd also swap distutils/setuptools to keep have list in the order of priority:
```suggestion
        using either :pep:`518` ``build-system``, :ref:`setuptools` or
        :ref:`distutils`, another practical way to define projects currently
```

encukou:
I'm not sure who should do the final approval/merge.
@webknjaz's change looks good; if you agree, please commit/squash/apply it as necessary when merging.

source/glossary.rst
```diff
@@ -102,11 +102,10 @@ Glossary
...
+        is something that contains a :term:`pyproject.toml`, :term:`setup.py`,
+        or :term:`setup.cfg` file at the root of the project source directory.
```
pfmoore, line 108: Is it valid to have `setup.cfg` without `setup.py`? That seems like an unrelated change, even if it is true. Maybe we should stick to just adding `pyproject.toml`?

webknjaz: @pfmoore it is valid to have `setup.cfg` w/o `setup.py` if you have a modern enough setuptools version. Like in case with `setuptools` mentioned as a build dep in `pyproject.toml`.

webknjaz:
@encukou I don't have merge powers in this repo so it's not up to me.

source/glossary.rst
pfmoore: OK, but in that case you need a `pyproject.toml`, so it seems a bit redundant. No big deal, though.

PR #758, opened 2020-07-19
Mention enhanced attr: directive in single-sourcing option #1

jwodder:
This PR adds a note to the `setup.py`-parsing option for single-sourcing project versions that setuptools' `attr:` directive does the same thing (without importing dependencies) starting in version 46.4.0.

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -42,12 +42,13 @@ number of your project:
        .. code-block:: ini
 
            [metadata]
-           version = attr:package.__version__
+           version = attr: package.__version__
 
        Earlier versions of setuptools implemented the ``attr:`` directive by
        importing the module, but setuptools 46.4.0 added rudimentary AST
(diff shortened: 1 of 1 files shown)


source/guides/single-sourcing-package-version.rst
```diff
@@ -33,6 +33,21 @@ number of your project:
...
+    .. note::
+
+       As of the release of setuptools 46.4.0, one can accomplish the same
+       thing by instead placing the following in the project's ``setup.cfg``
+       file (replacing "package" with the import name of the package):
+
+       .. code-block:: ini
+
+           [metadata]
+           version = attr:package.__version__
```
webknjaz, line 45: I believe that a whitespace after the colon is necessary
```suggestion
           version = attr: package.__version__
```
(even if not, it reads better this way)

jwodder: It's not necessary, but I added it anyway.

pradyunsg:
Thank you! ^>^

PR #761, opened 2020-08-10
Guide to including & accessing package data

jwodder:


source/guides/using-package-data.rst
```diff
@@ -0,0 +1,239 @@
...
+and import ``importlib-resources`` in your code as follows:
+
+.. code-block:: python
+
+    import sys
+
+    if sys.version_info < (3, 9):
+        import importlib_resources
+    else:
+        import importlib.resources as importlib_resources
```
webknjaz, line 169: Why not use the typical try-except idiom?

jwodder: Because `try: import importlib.resources` would succeed on Python 3.7 and 3.8, but we don't want to use `importlib.resources` there, as it doesn't have the `files()` function.  If you're suggesting `try: import importlib_resources` instead, that's really a stylistic decision.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,239 @@
...
+To access a package data file in your project, start by calling
+``importlib_resources.files()`` on the name of your package:
+
+.. code-block:: python
+
+    pkg = importlib_resources.files("packagename")
+    # The argument can optionally refer to a subpackage in the form
+    # "packagename.subpackage".
+
+This gives you an object with a subset of `pathlib.Path`'s methods for
```
webknjaz, line 184: I believe that no explicit role for single backticks is set and so you should use it explicitly like
```suggestion
This gives you an object with a subset of :py:class:`pathlib.Path`'s methods for
```

source/guides/using-package-data.rst
webknjaz: Good point. This should be noted explicitly...

jwodder: Updated.

source/guides/using-package-data.rst
jwodder: Updated.

webknjaz:
Could you rebase this, please? There's a lot more strictness in the sphinx setup now.

jwodder:
@webknjaz Rebased.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+when building from an sdist).  There are two ways to do this: a blanket
+"include all" method and a fine-grained method.
+
+
+Blanket Method
+--------------
+
+There are two steps to this method:
+
+1. In your :file:`setup.py`, pass ``include_package_data=True`` to the
```
webknjaz, line 35: Hm... Should this mention that this file may also be `setup.cfg`?

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+2. Create or edit your project's :file:`MANIFEST.in` file with the proper
+   commands so that all of your package data files are included in the
+   project's sdists.  See :ref:`Using MANIFEST.in` for information on how to do
+   this.
+
+   In the simplest case — in which you want all non-Python files in your import
+   package to be treated as package data — you can simply include your entire
+   import package in your sdist with the :file:`MANIFEST.in` command
+   ":samp:`graft {packagename}`" (or ":samp:`graft src`" if you're using a
+   ``src/`` layout).  If you use this command, you should also follow it with
```
webknjaz, line 49: I think, it'd be great to link src layout docs here

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+   are not included.
+
+
+Fine-Grained Method
+-------------------
+
+The fine-grained method allows you to specify package data file patterns on a
+per-package basis through the ``package_data`` argument to ``setup()``.
+``package_data`` must be a :py:class:`dict` mapping import package or
+subpackage names (or the empty string, to specify all packages & subpackages)
```
webknjaz, line 60: This would be an asterisk for `setup.cfg` IIRC

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+        # subpackages):
+        "hello": ["*.msg"],
+        # Include any *.csv files found in the "hello.utils" package:
+        "hello.utils": ["*.csv"],
+        # Include any *.dat files found in the "data" subdirectory of the
+        # "mypkg" package:
+        "mypkg": ["data/*.dat"],
+    }
+
+Note that glob patterns only select files located directly within the given
```
webknjaz, line 82: Maybe use something like this? 
```suggestion
.. tip::

   Note that glob patterns only select files located directly within the given
```

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+directory path); e.g., ``"hello": ["*.msg"]`` selects ``*.msg`` files in the
+``hello`` package but not in any of its subpackages.  To select files in
+subpackages, you must either include an entry for each subpackage or else use
+the empty string key to specify a pattern for all packages & subpackages.
+
+If a pattern contains any directory components, the forward slash (``/``) must
+be used as the directory separator, even on Windows.
+
+If a package data file is located in a directory that does not have an
+:file:`__init__.py` file (say, a ``data/`` directory inside
```
webknjaz, line 93: Is this also true when `find_namespace_packages()` is used?

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+There have been multiple ways to access package data over the years, from
+|pkg_resources' ResourceManager API|__ to |pkgutil.get_data()|__, but the most
+recent and currently-recommended way is with the `importlib-resources`__
+package.
+
+.. |pkg_resources' ResourceManager API| replace:: ``pkg_resources``' ``ResourceManager`` API
+.. __: https://setuptools.readthedocs.io/en/latest/pkg_resources.html
+       #resourcemanager-api
+
+.. |pkgutil.get_data()| replace:: ``pkgutil.get_data()``
```
webknjaz, line 128: Isn't this linkable via intersphinx?

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,243 @@
...
+
+There are two versions of ``importlib-resources`` available:
+
+- `The one on PyPI`__ that is installed with ``pip install
+  importlib-resources`` and imported with ``import importlib_resources`` (note
+  underscore)
+
+  .. __: https://pypi.org/project/importlib-resources/
+
+- `The one in the Python standard library`__ starting with Python 3.7 that is
```
webknjaz, line 145: Since I've upgraded the intersphinx links, this and some other entries should be linkable through normal roles/refs

webknjaz (approved):
LGTM overall, though addressing the comments may slightly improve it.

webknjaz:
@pradyunsg @pganssle @jaraco I think you may want to review this guide. Pinging you since I don't have sufficient privileges to "request review" in this repo.

source/guides/using-package-data.rst
jwodder: What src layout docs?

source/guides/using-package-data.rst
jwodder: I believe so.  I'm pretty sure `find_namespace_packages()` has no interaction with `package_data`.

source/guides/using-package-data.rst
jwodder: Linked (though it goes to the Python 2 docs for some reason).

source/guides/using-package-data.rst
jwodder: I tried rewriting this link in the following ways:

```
:py:mod:`The one in the Python standard library <importlib.resources>`
:ref:`The one in the Python standard library <python:importlib.resources>`
:ref:`The one in the Python standard library <python:module:importlib.resources>`
```

The first link was styled wrong.  The other two produced "undefined label" warnings, treated as errors.

source/guides/using-package-data.rst
jwodder: Seeing as nothing else in packaging.python.org references putting configuration in `setup.cfg` (I think), this would probably be more confusing than helpful.  People who already know configuration can be placed in `setup.cfg` should be able to figure out that `include_package_data` can be placed there as well.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+  imported with ``import importlib.resources`` (note period)
```
wkschwartz, line 199:
```suggestion
- The Python standard library's starting with Python 3.7 that is
  imported with ``import`` :py:mod:`python:importlib.resrouces` (note period)
```

source/guides/using-package-data.rst
wkschwartz: I [suggested a change](https://github.com/pypa/packaging.python.org/pull/761/files#r477557831) to handle the linking problem you're having.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+
+    with pkg_data_file.open() as fp:
+        # Do things with fp
+
+- To get the file's contents as :py:class:`bytes`, call the ``read_bytes()``
+  method:
+
+  .. code-block:: python
+
+    b = pkg_data_file.read_bytes()
```
wkschwartz, line 263: This can also be done without the `files` API using [`importlib.resources.read_binary`](https://docs.python.org/3/library/importlib.html#importlib.resources.read_binary)

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+
+- To get the file's contents as a :py:class:`str`, call the ``read_text()``
+  method, optionally with an ``encoding`` argument:
+
+  .. code-block:: python
+
+    s = pkg_data_file.read_text(encoding="utf-8")
```
wkschwartz, line 270: This can also be done without the `files` API using [`importlib.resources.read_text`](https://docs.python.org/3/library/importlib.html#importlib.resources.read_text)

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+
+- To get the path to the file, call ``importlib_resources.as_file()`` on it and
```
wkschwartz, line 272: This can also be done without the `files` API using [`importlib.resources.path`](https://docs.python.org/3/library/importlib.html#importlib.resources.path)

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+  use the ``iterdir()`` method.  You can test whether a resource is a directory
+  or a file with the ``is_dir()`` and ``is_file()`` methods, and you can get a
+  resource's basename via the ``name`` property:
+
+  .. code-block:: python
+
+    for entry in (pkg / "data").iterdir():
+        if entry.is_dir():
+            print(entry.name, "DIR")
+        else:
```
wkschwartz, line 295: This can be done without the `files` API using [`importlib.resources.is_resource`](https://docs.python.org/3/library/importlib.html#importlib.resources.is_resource)

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,296 @@
...
+  start of the ``with`` block and remove it at the end of the block.
+
+- To iterate through a directory (either a package or a non-package directory),
+  use the ``iterdir()`` method.  You can test whether a resource is a directory
+  or a file with the ``is_dir()`` and ``is_file()`` methods, and you can get a
+  resource's basename via the ``name`` property:
+
+  .. code-block:: python
+
+    for entry in (pkg / "data").iterdir():
```
wkschwartz, line 292: This can be done without the `files` API using [`importlib.resources.contents`](https://docs.python.org/3/library/importlib.html#importlib.resources.contents)

jwodder:
> This guide promotes the `importlib_resources.files` API without mentioning the equivalent, backward-compatible API available in `importlib.resources`. I've left comments in places where alternatives can be offered.

The older API is not equivalent; it does not support package data files inside directories without `__init__.py` files.  For that reason, I would prefer to not mention it in the guide at this point in time.  I would suggest you wait until this PR has been merged and then create your own PR to mention the older API if you so desire.

source/guides/using-package-data.rst
webknjaz: I was thinking about https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

jwodder: That's about `setup.cfg`, not `src/`.

webknjaz: Ah, sorry, I meant https://setuptools.readthedocs.io/en/latest/setuptools.html#using-a-src-layout

jwodder: That's just about how to edit `setup.cfg` for a `src/` layout.  I don't think it's worth linking here.

webknjaz: How about https://hynek.me/articles/testing-packaging/

source/guides/using-package-data.rst
webknjaz: You may explicitly prefix the ref with the intersphinx identifier to force linking a certain docs site

jwodder: Done.

source/guides/using-package-data.rst
jwodder: Done.

source/guides/using-package-data.rst
jwodder: A passage on performing this configuration in `setup.cfg` has been added.

jwodder:
@webknjaz What issues does this PR still need to address before it can be merged?

webknjaz:
@jwodder not from my side. FWIW I cannot merge PRs in this repo. You'll have to wait for somebody like @di or @pradyunsg to do this.

pradyunsg:
I'm happy to merge this if another PyPA member gives it a green tick. I've not looked at this myself yet, but definitely can do early next week.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -447,7 +447,8 @@ the directory containing the package.
 
 For more information, see `Including Data Files
 <https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files>`_
```
imba-tjd, line 449:
```suggestion
<https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html>`_
```

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+
+
+Including package data
+======================
+
+.. note::
+
+    The following instructions only pertain to projects built with setuptools.
+    Users of flit, Poetry, and other tools must consult the respective
+    backend's documentation to see how to include package data.
```
jaraco, line 20: I would not recommend documenting setuptools-specific behaviors in the packaging docs. These docs should present general packaging concerns and then defer to the setuptools docs for setuptools-specific implementation details (and other tools for their related details). By attempting to document here, you create a duplication of effort with what the setuptools docs attempt to provide. Since we're already short-staffed on maintenance support, I'm reluctant to have more documentation that might become out of sync with the implementation.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+
+The first requirement for being able to include package data in your package is
+for the files to be located somewhere inside the directory that is your
+project's :term:`import package <Import Package>`.  If your project is a
```
jaraco, line 24: Here or above, I think it would be useful to link to [Python's documentation for packages](https://docs.python.org/3/tutorial/modules.html#packages) and then describe the relationship between a Python Package (I wouldn't call it an Import Package) and a Distribution Package and indicate which form of 'package' is used in the documentation.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+single-file ``.py`` module, you must convert it to a directory/package layout
+before you can store package data in it.
+
+Once your package data is inside the import package, you must tell setuptools
+to include the files in both the :term:`wheel <Wheel>` (so that the files will
+be installed when your package is installed) and :term:`sdist <Source
+Distribution (or "sdist")>` (so that the files can be included in the wheel
```
jaraco, line 31: Here, I would focus on the different between a source distribution (aka sdist, i.e. a copy of the source suitable for another person to build, so includes the packaging scaffolding and testing and other ancillary features) and a built distribution (aka bdist or wheel, which is a distribution that's built and ready for installation). The concepts for a built distribution also apply to a package that's built and directly installed (from source with no intermediary built distribution, i.e. `setup.py install`), although this workflow is largely discouraged.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+1. In your :file:`setup.py`, pass ``include_package_data=True`` to the
+   ``setup()`` function.   If you are placing your project configuration in
+   :file:`setup.cfg` instead, set ``include_package_data = True`` in the
+   ``[options]`` section of :file:`setup.cfg`.
+
+   Doing this tells setuptools that any non-Python files found inside your
+   import package that are included in the sdist should also be included in the
+   wheel.
+
+2. Create or edit your project's :file:`MANIFEST.in` file with the proper
```
jaraco, line 53: There are other ways to ensure the files are included in the sdist. My preferred way and the way I recommend is to use the [setuptools_scm project](https://pypi.org/project/setuptools-scm) to discover all sources from SCM metadata. This approach is simple and maintenance free.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+
+Including files via setuptools plugins
+--------------------------------------
+
+As an alternative to the above methods, you can use a plugin for setuptools
+that automatically recognizes & includes package data in sdists & wheels,
+usually based on what files in the project directory are under verson control.
+One example of such a plugin is setuptools_scm_, which automatically finds all
+files under version control in a Git or Mercurial repository and augments the
+project's :file:`MANIFEST.in` (if any) with the found files.  This eliminates
```
jaraco, line 166: This phrasing may give the reader the impression that the file is somehow altered. I'd instead say that it affects the setuptools source file discovery mechanism, augmenting the list of files discovered (in addition to those found through the manifest template).

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+
+In a :file:`setup.cfg`, ``exclude_package_data`` becomes an
+``[options.exclude_package_data]`` section whose contents have the same
+structure as ``[options.package_data]``.
+
+
+Including files via setuptools plugins
+--------------------------------------
+
+As an alternative to the above methods, you can use a plugin for setuptools
```
jaraco, line 161: The use of plugins is only an alternative to the manifest template (MANIFEST.in). I instead would suggest breaking down the problem into two phases: file discovery for sources (handled by plugin or manifest template) and file inclusion for installation (and wheels), handled by 'package_data' or 'include_package_data' directive.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+.. __: https://setuptools.readthedocs.io/en/latest/pkg_resources.html
+       #resourcemanager-api
+
+.. __: http://importlib-resources.readthedocs.io
+
+
+Installing & importing ``importlib-resources``
+----------------------------------------------
+
+There are two versions of ``importlib-resources`` available:
```
jaraco, line 193: Again, here the documentation is repeating a lot of information that's authoritative elsewhere. It may be more sustainable to give a high-level introduction and leave the details to the projects to document (and maybe contribute there).

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+  .. __: https://pypi.org/project/importlib-resources/
+
+- `The one in the Python standard library`__ starting with Python 3.7 that is
+  imported with ``import importlib.resources`` (note period)
+
+  .. __: https://docs.python.org/3/library/importlib.html
+         #module-importlib.resources
+
+Development of the PyPI version tends to be ahead of whatever's in the latest
+Python version.  In particular, the new ``files()``-based API described here
```
jaraco, line 208: This API is recommended and intended to supersede the other API functions.

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+
+    import sys
+
+    if sys.version_info < (3, 9):
+        # importlib.resources either doesn't exist or lacks the files()
+        # function, so use the PyPI version:
+        import importlib_resources
+    else:
+        # importlib.resources has files(), so use that:
+        import importlib.resources as importlib_resources
```
jaraco, line 228: I'd recommend use a `from importlib` import for brevity.

```suggestion
        import importlib_resources as resources
    else:
        # importlib.resources has files(), so use that:
        from importlib import resources
```

source/guides/using-package-data.rst
```diff
@@ -0,0 +1,299 @@
...
+To access a package data file in your project, start by calling
+``importlib_resources.files()`` on the name of your package:
+
+.. code-block:: python
+
+    pkg = importlib_resources.files("packagename")
+    # The argument can optionally refer to a subpackage in the form
+    # "packagename.subpackage".
+
+This gives you an object with a subset of :py:class:`pathlib.Path`'s methods
```
jaraco, line 243: This object now has a formal Protocol, [Traversable](https://github.com/python/importlib_resources/blob/eb0554035b499e4a5349dcc56a2bda51835f4fde/importlib_resources/abc.py#L61).

source/guides/using-package-data.rst
jaraco: This API, while still supported, is legacy and discouraged.

source/guides/using-package-data.rst
jaraco: discouraged

source/guides/using-package-data.rst
jaraco: discouraged

source/guides/using-package-data.rst
jaraco: discouraged

source/guides/using-package-data.rst
jaraco: discouraged

jaraco (approved):
Thanks jwodder for working on this. I've made a run through and added a critique. Overall, this documentation is really good and provides a good high-level overview. There were a few tweaks and nitpicks I would make.

My main concern is with duplicating the documentation from more authoritative sources. I worry that duplicating the guidance from Setuptools and CPython and backports modules is going to lead to a dilution of effort around clear guidance. My recommendation would be to instead provide a high-level overview here and then link to suitable references in the relevant projects and where those references could use improved guidance, contribute those improvements there.

jwodder:
> My recommendation would be to instead provide a high-level overview here and then link to suitable references in the relevant projects

I see that your vision for this PR is incompatible with mine.  I thereby withdraw it.

PR #763, opened 2020-08-11
⚙ Refactor the sphinx setup

webknjaz:
This effort implements a set of changes that make the sphinx setup more robust.

* it bumps the sphinx version to the latest stable v3.2.0
* it enables strict nitpicky mode so that the source doesn't link against non-existing or broken things
* it replaces the internal `:role:` craft with a built-in `extlinks` sphinx extension
* it updates some outdated bits in the sphinx config, like the copyright end date and the repo hyperlinks
* it updates the intersphinx mappings to pull-in the latest Python docs
* it fixes all the errors that happened because of updating sphinx and implementing the strictness

P.S. Each commit in this branch is atomic and so it's better to use the "merge" or "rebase" strategy when accepting the PR in order to preserve them.


webknjaz:
@di mind reviewing?

pradyunsg:
Mered since I think upgrading to the newer sphinx version is a good idea. :)

PR #771, opened 2020-09-13
specifications: Requires-Python doesn't accept an environment marker

xavfernandez:
Cf https://github.com/pypa/setuptools/issues/1876 for context and https://github.com/pypa/setuptools/issues/1876#issuecomment-545642504 specifically for the reasoning.

TLDR: PEP-0345 currently allows environment markers for `Requires-Python` but most tools do not expect environment markers to be present in this field: let's update the specification/PEP to match the reality.

This would also require (?) an update to PEP-0345 (simply remove https://github.com/python/peps/blob/5a0ab0c2b71441f839914df6e7b996bfc4850cab/pep-0345.txt#L499).

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -477,14 +477,13 @@ picking which version of a project to install.
 
 The value must be in the format specified in :doc:`version-specifiers`.
 
-This field may be followed by an environment marker after a semicolon.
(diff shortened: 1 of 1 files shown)


pfmoore (approved):

xavfernandez:
PEP PR posted: https://github.com/python/peps/pull/1611

PR #772, opened 2020-09-22
Fix broken link in user guide

wimglenn:
_Note:_ I actually don't know what pylauncher is, so if Vinay's project is not the right one then apologies
The setuptools docs also have the same broken link here https://setuptools.readthedocs.io/en/latest/easy_install.html#natural-script-launcher

### DIFF
--- a/source/discussions/pip-vs-easy-install.rst
+++ b/source/discussions/pip-vs-easy-install.rst
@@ -66,4 +66,4 @@ Here's a breakdown of the important differences between pip and easy_install now
 .. [1] https://setuptools.readthedocs.io/en/latest/easy_install.html#natural-script-launcher
 
 
-.. _pylauncher support: https://bitbucket.org/pypa/pylauncher
+.. _pylauncher support: https://bitbucket.org/vinay.sajip/pylauncher


source/discussions/pip-vs-easy-install.rst
```diff
@@ -66,4 +66,4 @@ Here's a breakdown of the important differences between pip and easy_install now
 .. [1] https://setuptools.readthedocs.io/en/latest/easy_install.html#natural-script-launcher
 
 
-.. _pylauncher support: https://bitbucket.org/pypa/pylauncher
+.. _pylauncher support: https://bitbucket.org/vinay.sajip/pylauncher
```
webknjaz, line 69: @wimglenn I think [it's already in the stdlib](https://docs.python.org/3/using/windows.html#launcher). And the Python Docs website is integrated into this project via intersphinx.
So I guess you should just drop this line and use ``:ref:`python:launcher` `` as a reference right in the table instead.

@vsajip ^

pfmoore: pylauncher is a different project than the launcher that's in the core python distribution.

vsajip: To be clear, `pylauncher` was the reference implementation for [PEP-397](https://www.python.org/dev/peps/pep-0397/) and that implementation was moved to become part of the Python distribution, with minor changes. Since the launcher now ships with Python, the `pylauncher` repository is really only of historical interest / perhaps worth looking at if there are regressions in the in-Python launcher. I do occasionally look at how much the in-Python version has diverged from `pylauncher` and have sometimes synced up changes, but it's not a priority or anything.

wimglenn: The topic of this particular page, "pip vs easy install", may also be primarily historical interest - so perhaps the reference implementation is actually the best place for the link target after all?

pfmoore: Reading the page, I think it's all obsolete. I don't even know what "pylauncher support" means in this context. Certainly not as something pip doesn't have...

pradyunsg: Looks like this pages needs updating! :)

pradyunsg (approved):
Let's merge this in, as it's fixing a broken link, and we'll come back to the rest of this page afterwards. :)

webknjaz (approved):

PR #775, opened 2020-09-25
Remove Requires-Python warning from piwheels project description

bennuttall:
piwheels now honours the Requires-Python attribute on files, so I've removed the warning from the key projects page. 🚀 

https://blog.piwheels.org/requires-python-support-new-project-page-layout-and-a-new-json-api/

I also changed "Raspbian" to "Raspberry Pi OS" as the project name has [changed](https://www.raspberrypi.org/downloads/raspberry-pi-os/).


di:
Thanks! Nice work!

PR #781, opened 2020-10-25
Clarify .dist-info naming to match reality

uranusjr:
Related discussions on [Discourse](https://discuss.python.org/t/clarify-naming-of-dist-info-directories/5565) and pipxproject/pipx#528.

The writeup about .dist-info naming rules did not match reality. Additional text is added to clarify that

1. Tools should replace dashes with underscores in the file name.
2. Existing tools do not follow PEP 503 normalization, so parsing tools should not expect this.
3. New tools are encouraged to perform PEP 503 normalization.


jaraco:
I'm okay with this specification. I'd prefer to see that such a spec could be implemented in [importlib_metadata](https://github.com/python/importlib_metadata) before making it official, in case there are performance or complexity concerns that emerge in a concrete implementation. That said, this isn't written in stone, so if the implementation turns out to be infeasible, the spec can be adjusted as well.

jaraco (approved):

source/specifications/recording-installed-packages.rst
```diff
@@ -41,7 +41,16 @@ packages (commonly, the ``site-packages`` directory).
 This directory is named as ``{name}-{version}.dist-info``, with `name` and
 `version` fields corresponding to :ref:`core-metadata`.
 The name field must be in normalized form (see `PEP 503 <https://www.python.org/dev/peps/pep-0503/#normalized-names>`_
```
jaraco, line 43: This sentence conflicts with the following one. Should you remove the reference to PEP 503 since the normalization is described below?

uranusjr: The normalisation logic is actually

* For `name`, do PEP 503 normalisation, and then replace dashes with underscores
* For `version`, replace dashes with underscores

I probably expressed that very badly.

uranusjr: I’ve opened a PR at python/importlib_metadata#253

source/specifications/recording-installed-packages.rst
```diff
@@ -41,7 +41,16 @@ packages (commonly, the ``site-packages`` directory).
...
-for the definition of normalization).
+for the definition of normalization). Both the `name` and `version` fields
+should replace dash (``-``) characters with underscore (``_``) chacaters, so
+the ``.dist-info`` directory always has exactly one dash (``-``) character in
+its stem, separating the `name` and `version` fields.
+
+For backwards compatibility reasons, tools are allowed to not replace dot
```
xavfernandez, line 49: Do we really want to allow that ?
I think we should push tools to follow the strict normalization rules while still acknowledging (like you did in the following sentence) that this is currently not always the case and that tools consuming `.dist-info` will likely have to support "invalid" `.dist-info` directories.

In summary, _Be conservative in what you do, be liberal in what you accept from others_.

source/specifications/recording-installed-packages.rst
```diff
@@ -41,7 +41,16 @@ packages (commonly, the ``site-packages`` directory).
...
+(``.``) chacaters or keep upper-case ASCII alphabets in the `name` field.
```
xavfernandez, line 50:
```suggestion
(``.``) characters or keep upper-case ASCII alphabets in the `name` field.
```

source/specifications/recording-installed-packages.rst
uranusjr: > Be conservative in what you do, be liberal in what you accept from others

I thought the network stack (the original proponent of this idea) actually has largely accepted [this principle does not actually work](https://en.wikipedia.org/wiki/Robustness_principle#Criticism)? If enough senders generate the liberal format, that format effectively becomes the de-facto standard, and the receivers are now forced to follow an undocumented rule because everyone else is doing it. The approach generally considered nowadays (as I understand it) is to explicitly specify where people are allowed to be liberal, so the receiving side knows what to expect instead of making wrong assumptions and creating fragmentation.

We are more or less in the same situation here. There are countless existing `.dist-info` directories in the wild (including wheels that cannot be retroactively fixed), parsers already need to carry a compatibility fix until practically forever. It would be more beneficial if we accept this fact, and codify the situation explicitly so parser maintainers and creators (of new ones) can correctly assess the situation and implement the logic that works, instead of having to find out these compatibility edge cases later (and have to fix that anyway).

pfmoore: I don't think the issues with the "be liberal in what you consume" issue are quite like that. The failure case you quote suggests that the response should be to be *strict* in what you consume, to avoid standard-violating variations proliferating. We don't have the option to do that because we're *already* in the situation of having non-standard forms "in the wild". So what we need to do is to clearly define what non-standard variations consumers should allow, so that new variations don't catch on, as well as being very clear that new *producers* must normalise. Tools can always violate the spec, we can't do anything about that, but we can make the intent clear.

How about saying

> Historically, tools have failed to replace dot characters or normalise case in the `name` field, so parsers must be prepared to see these types of unnormalised names (which they must treat as equivalent to the normalised version). New tools which write `.dist-info` directories MUST use normalised names, and existing tools are encouraged to start normalising names if they do not already do so.

uranusjr: I’ve rewritten the paragraph.

source/specifications/recording-installed-packages.rst
```diff
@@ -41,7 +41,17 @@ packages (commonly, the ``site-packages`` directory).
...
+should replace dash (``-``) characters with underscore (``_``) chacaters, so
+the ``.dist-info`` directory always has exactly one dash (``-``) character in
+its stem, separating the `name` and `version` fields.
+
+Historically, tools have failed to replace dot characters or normalize case in
+the `name` field. Tools consuming ``.dist-info`` directories should expect a
+name field to be unnormalized, and treat it as equivalent to its normalized
+counterpart. New tools that write ``.dist-info`` directories MUST normalize
+the name field using the rules described above, and existing tools are
+encouraged to start producing normalizing `name` fields.
```
pfmoore, line 54:
```suggestion
encouraged to start normalizing `name` fields.
```

uranusjr:
Following [the discussion on Discourse](https://discuss.python.org/t/5605), I’ve also added clarification on the version field. This does not need a POC in importlib-metadata since it does not read the ``.dist-info`` directory’s version field.

source/specifications/recording-installed-packages.rst
```diff
@@ -40,8 +40,21 @@ packages (commonly, the ``site-packages`` directory).
 
 This directory is named as ``{name}-{version}.dist-info``, with `name` and
 `version` fields corresponding to :ref:`core-metadata`.
-The name field must be in normalized form (see `PEP 503 <https://www.python.org/dev/peps/pep-0503/#normalized-names>`_
-for the definition of normalization).
+Both the name and version fields must be in normalized form (see
+`PEP 503 <https://www.python.org/dev/peps/pep-0503/#normalized-names>`_ and
+`PEP 440 <https://www.python.org/dev/peps/pep-0440/#normalization>`_ for the
+definition of normalization for each field respectively),
```
webknjaz, line 46: Let's use a ```:pep``` role
```suggestion
:pep:`PEP 503 <503#normalized-names>` and :pep:`PEP 440 <440#normalization>`
for the definition of normalization for each field respectively),
```

source/specifications/recording-installed-packages.rst
```diff
@@ -40,8 +40,21 @@ packages (commonly, the ``site-packages`` directory).
...
+and replace dash (``-``) characters with underscore (``_``) chacaters, so
+the ``.dist-info`` directory always has exactly one dash (``-``) character in
+its stem, separating the `name` and `version` fields.
```
webknjaz, line 49: Single quotes don't have a role defined by default:
```suggestion
its stem, separating the ``name`` and ``version`` fields.
```

source/specifications/recording-installed-packages.rst
```diff
@@ -40,8 +40,21 @@ packages (commonly, the ``site-packages`` directory).
...
+
+Historically, tools have failed to replace dot characters or normalize case in
+the `name` field, or not perform normalization in the `version` field. Tools
```
webknjaz, line 51:
```suggestion
the ``name`` field, or not perform normalization in the ``version`` field. Tools
```

source/specifications/recording-installed-packages.rst
```diff
@@ -40,8 +40,21 @@ packages (commonly, the ``site-packages`` directory).
...
+consuming ``.dist-info`` directories should expect those fields to be
+unnormalized, and treat them as equivalent to their normalized counterparts.
+New tools that write ``.dist-info`` directories MUST normalize both `name` and
+`version` fields using the rules described above, and existing tools are
```
webknjaz, line 55:
```suggestion
New tools that write ``.dist-info`` directories MUST normalize both ``name`` and
``version`` fields using the rules described above, and existing tools are
```

source/specifications/recording-installed-packages.rst
uranusjr: Existing text uses single quotes and I assumed it’s a conscious decision and followed the same format.

webknjaz: It's rendered as italics: https://python-packaging-user-guide--781.org.readthedocs.build/specifications/recording-installed-packages/#the-dist-info-directory which is probably not what you want. The problem is that Sphinx is permissive by default. Somebody has to contribute setting it to "any" I guess.

uranusjr: > which is probably not what you want.

I’m not sure about that. A lot of stdlib documentation uses this style.

merwok: Really?  I remember a rule (even a warning?) about not using the default role.

pradyunsg: Let's fix the single backticks stuff in a follow up -- it's a bit broader than necessary and we can make it stricter throughout in one big sweep, to avoid holding this PR to a different standard that we're not yet enforcing everywhere.

webknjaz: @uranusjr stdlib doesn't address warnings. When I was helping @jaraco to adopt the distutils docs into setuptools after enabling all the strictness the other day, it was so painful: there's a lot broken references and incorrect syntax all over the place. So I'd not recommend looking up at their setup.
But yeah, this could be done separately.
@merwok this repo doesn't have default_role=any and so sphinx does not attempt to interpret it

jaraco (changes requested):
On further consideration, I'm not a fan of this proposal at all.  It proposes that the name `jaraco.clipboard` be normalized to `jaraco-clipboard` and then further normalized to `jaraco_clipboard` for `dist-info` naming. ~~But `jaraco_clipboard` and `jaraco.clipboard` are different package names in pypi.~~ I'll explain more in discourse.

uranusjr:
How about something like this?

> The `.dist-info` directory’s name is formatted to unambigiously represent a distribution in a filesystem path. Tools intending to inspect a distribution’s name and version should instead read the respective `METADATA` fields, since values listed there are unescaped and accurately reflect the distribution.

jaraco:
Unfortunately, that still doesn't address the situation where no dist-info is present. How about this:

> The `.dist-info` directory’s name is formatted to unambigiously represent a distribution in a filesystem path. Tools presenting a distribution name to a user or consuming API should avoid presenting the normalized name and should instead present the specified name (when needed prior to resolution to an installed package) or read the respective fields in Core Metadata, since values listed there are unescaped and accurately reflect the distribution.

pfmoore:
> or consuming API

I don't like this, as it makes it the norm for all APIs to have to work with non-normalised names, which means that everything ends up re-normalising. I'm fine with saying that data p[resented to the user should be unnormalised (which implies that APIs taking a name that they intend to display to the user should take unnormalised names) but I think APIs should be free to take (or even require) normalised names if they are expecting to only use them for non-display uses.

For example, I'd prefer something like resolvelib to work solely in terms of normalised names. (In practice, resolvelib avoids this issue another way, but that's not relevant here).

uranusjr:
How about

> The `.dist-info` directory’s name is formatted to unambigiously represent a distribution in a filesystem path. Tools presenting a distribution name to a user should avoid presenting the normalized name and instead present the specified name (when needed prior to resolution to an installed package) or read the respective fields in Core Metadata, since values listed there are unescaped and accurately reflect the distribution. Libraries should provide API for such tools to consume so they have access to the unnormalized name when they display distrubution information.

This separates the responsibility on this issue more cleanly IMO. Libraries that read metadata should provide access to the information; an application is fee to not use the information if it does not need to display anything to the user. If it does not use the information when it should, that’s on the application’s maintainers.

jaraco (approved):

uranusjr:
I have added the text as a `.. note::` block (and also changed the fields to all use double ticks).

xavfernandez (approved):

webknjaz:
Looks like the review comments have been addressed long ago. Merging.

PR #785, opened 2020-10-31
WIP: Add sdist format specification and metadata field (PEP 643)

pfmoore:


source/specifications/source-distribution-format.rst
```diff
@@ -0,0 +1,48 @@
...
+defined by the metadata specification.
+
+Source distributions are also known as *sdists* for short.
+
+Source distribution file name
+=============================
+
+The file name of a sdist is not currently standardised, although the *de facto*
+form is ``{name}-{version}.tar.gz``, where ``{name}`` is the canonicalized form
+of the project name (see :pep:`503` for the canonicalization rules) and
```
xavfernandez, line 27: canonicalized form *with "-" replaced by "_"* ?

pfmoore: Thanks, good catch. I always get that mixed up 🙁

source/specifications/binary-distribution-format.rst
```diff
@@ -0,0 +1,8 @@
+
+.. _binary-distribution-format:
+
+==========================
+Binary distribution format
+==========================
+
+The binary distribution format (``wheel``) is defined in :pep:`427`.
```
webknjaz, line 8: There's a glossary term: https://github.com/pypa/packaging.python.org/blob/b5f7f5d/source/glossary.rst#L239

```suggestion
The binary distribution format (:term:`wheel <Wheel>`) is defined in :pep:`427`.
```

source/specifications/source-distribution-format.rst
```diff
@@ -0,0 +1,48 @@
...
+==========================
+
+The current standard format of source distribution format is identified by the
+presence of a ``pyproject.toml`` file in the distribution archive.  The layout
+of such a distribution was originally specified in :pep:`517` and is formally
+documented here.
+
+There is also the legacy source distribution format, implicitly defined by the
+behaviour of ``distutils`` module in the standard library, when executing
+``setup.py sdist``. This document does not attempt to standardise this format,
```
webknjaz, line 15: mark it as a command?
```suggestion
:command:`setup.py sdist`. This document does not attempt to standardise this format,
```

source/specifications/source-distribution-format.rst
```diff
@@ -0,0 +1,48 @@
+
+.. _source-distribution-format:
+
+==========================
+Source distribution format
+==========================
+
+The current standard format of source distribution format is identified by the
+presence of a ``pyproject.toml`` file in the distribution archive.  The layout
```
webknjaz, line 9: How about using a file role?
```suggestion
presence of a :file:`pyproject.toml` file in the distribution archive.  The layout
```

source/specifications/source-distribution-format.rst
```diff
@@ -0,0 +1,48 @@
...
+The name and version components of the filename MUST match the values stored
+in the metadata contained in the file.
+
+Source distribution file format
+===============================
+
+A .tar.gz source distribution (sdist) contains a single top-level directory
+called ``{name}-{version}`` (e.g. ``foo-1.0``), containing the source files of
+the package. The name and version MUST match the metadata stored in the file.
+This directory must also contain a ``pyproject.toml`` in the format defined in
```
webknjaz, line 39:
```suggestion
This directory must also contain a :file:`pyproject.toml` in the format defined in
```

source/specifications/source-distribution-format.rst
```diff
@@ -0,0 +1,48 @@
...
+of the project name (see :pep:`503` for the canonicalization rules) with ``-``
+characters replaced with ``_``, and ``{version}`` is the project version.
+
+The name and version components of the filename MUST match the values stored
+in the metadata contained in the file.
+
+Source distribution file format
+===============================
+
+A .tar.gz source distribution (sdist) contains a single top-level directory
```
webknjaz, line 36: this is probably worth highlighting
```suggestion
A ``.tar.gz`` source distribution (sdist) contains a single top-level directory
```

source/specifications/core-metadata.rst
```diff
@@ -4,10 +4,6 @@
 Core metadata specifications
 ============================
 
-The current core metadata file format, version 2.1, is specified in :pep:`566`.
```
webknjaz, line 7: Old PEP removed but no mention of a new PEP 643? Is this intentional?

pfmoore: Yes, because PEP 643 does not specify the whole format, just the addition. I didn't think it made sense to say "Version 2.1 was specified in 566, which also made this document canonical, then 643 updated it to 2.2" with the result that the paragraph becomes an extended audit trail of all the PEPs that change the spec.

IMO, the whole point of making this document canonical is that we don't need to link back to the PEPs. 2.1 linking back to 566 was OK, as that was the version that changed the canonical source of the spec. But now we've moved to the next version, I think it's time to drop the (essentially historical, now) note.

eduardsmith:
Nice

sbidoul (approved):
@pfmoore was this ready to be merged ?

pfmoore:
@sbidoul Drat, yes. Thanks for the reminder, I don't know how I forgot this 🙁

PR #789, opened 2020-11-11
Update link to egg format

jgerityneurala:
The current link to the _Internal Structure of Python Eggs_ page is broken.


di:
Thanks!

PR #800, opened 2020-12-04
fix orphaned link

wimglenn:
The link on https://packaging.python.org/guides/distributing-packages-using-setuptools/#entry-points for "Dynamic Discovery of Services and Plugins" is going to the wrong place


di:
Thanks!

PR #801, opened 2020-12-06
Fix references with wrong link text

jodal:
The "easy_install" references links to a section with "setuptools" as
the title. This title is then used as the link text, making the section
incomprehensible for people not familiar with the relationship between
pip, easy_install, and setuptools.


di:
Thanks!

PR #805, opened 2020-12-19
DOC,WIP: distributing-packages-using-setuptools.rst

westurner:


source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -659,24 +659,27 @@ For example::
...
-both installed and editable in project form.
+You can install a project in "editable"
+or "develop" mode while you're working on it.
+When installed as editable, a project can be
+edited in-place without reinstallation: 
+changes to Python source files in projects installed as editable will be reflected the next time an interpreter process is started.
 
-Assuming you're in the root of your project directory, then run:
+To install a Python package in "editable"/"development" mode
+Change directory to the root of the project directory and run ``pip install -e .``:
```
webknjaz, line 669: This duplicates the code block. I think ``pip install -e .`` should be removed from this line.

PR #806, opened 2020-12-21
Add warning on support of attr: directive

ajnelson-nist:
This patch is part of addressing setuptools issue 2492:
https://github.com/pypa/setuptools/issues/2492

The code in this Note block was found to not apply to equivalently-named
features in setup.py.  After discussion on the above issue, a project
maintainer noted this behavior was expected.  This patch adds
documentation of expected behavior close to where it would be discovered
by somebody who stumbled upon the attr: feature, but might not know the
background on directives and setup.cfg vs. setup.py.


ajnelson-nist:
@jaraco I welcome any revisions to this patch.  One thing I wasn't sure about is where there might be a statement of the class of features that confused me not being supported in setup.py.

jaraco:
This looks fine to me. The note is in a particular corner of the docs (single-source version), so will be missed by others, but since you were the affected user, I'm guessing this is where the note would need to be to have the intended effect. Ultimately, guidance like this should probably go in the documentation for parameters to `setup()` and maybe in the section on `declarative config`, although given the complexities due to distutils docs overlapping with setuptools docs, there may not be a good place for that right now.

jaraco (approved):

source/guides/single-sourcing-package-version.rst
```diff
@@ -49,6 +49,8 @@ number of your project:
        analysis so that ``attr:`` can function without having to import any of
        the package's dependencies.
 
+       Also, please be aware that declarative config indicators, including the
+       ``attr:`` directive, are not supported in parameters to setup.py.
```
webknjaz, line 53: `setup.py` should be marked as inline source for consistency:
```suggestion
       ``attr:`` directive, are not supported in parameters to ``setup.py``.
```

di (approved):

PR #807, opened 2020-12-22
update setuptools link

imba-tjd:


gaborbernat (approved):

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -446,7 +446,7 @@ that should be copied into the package. The paths are interpreted as relative to
 the directory containing the package.
 
 For more information, see `Including Data Files
-<https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files>`_
+<https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html>`_
 from the `setuptools docs <https://setuptools.readthedocs.io>`_.
```
webknjaz, line 450: How about we use intersphinx?
```suggestion
For more information, see :std:doc:`Including Data Files
<setuptools:userguide/datafiles>`
from the :std:doc:`setuptools docs <setuptools:index>`.
```
This would require adding a `setuptools` entry to https://github.com/pypa/packaging.python.org/blob/c3d4b22/source/conf.py#L377.

webknjaz (approved):

PR #808, opened 2020-12-29
Use `python -m pip` instead of just `pip` in pip invocations

scop:
Stdlib has done that since 2014, and pip's own docs do that too.

Note: python vs python3 is used inconsistently between files in the
tree; this attempts to follow the existing per-file convention and does
not modify/unify it.

Note #2: this change doesn't intentlonally wrap any lines that became "long" as a result of these changes for easier reviewability. I can do that too if that's preferred, let me know.


gaborbernat (approved):

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -64,7 +64,7 @@ Requirements files
 ------------------
 
 :ref:`Requirements Files <pip:Requirements Files>` described most simply, are
-just a list of :ref:`pip:pip install` arguments placed into a file.
+just a list of :ref:`pip:python -m pip install` arguments placed into a file.
```
webknjaz, line 67: It's not a command but a reference to a label in pip's Sphinx docs. You must use an existing label for it to work. Here's a list of what's available: https://webknjaz.github.io/intersphinx-untangled/pip.pypa.io/.

```suggestion
just a list of :ref:`pip:pip install` arguments placed into a file.
```

source/discussions/pip-vs-easy-install.rst
```diff
@@ -27,13 +27,13 @@ Here's a breakdown of the important differences between pip and easy_install now
...
-|Uninstall Packages            |Yes (``pip uninstall``)               |No                             |
+|Uninstall Packages            |Yes (``python -m pip uninstall``)     |No                             |
 +------------------------------+--------------------------------------+-------------------------------+
 |Dependency Overrides          |Yes (:ref:`Requirements Files         |No                             |
 |                              |<pip:Requirements Files>`)            |                               |
 +------------------------------+--------------------------------------+-------------------------------+
-|List Installed Packages       |Yes (``pip list`` and ``pip           |No                             |
-|                              |freeze``)                             |                               |
+|List Installed Packages       |Yes (``python -m pip list`` and       |No                             |
+|                              |``python -m pip freeze``)             |                               |
```
webknjaz, line 36: Not sure if it's worth doing this in tables...

source/guides/index-mirrors-and-caches.rst
```diff
@@ -34,11 +34,11 @@ cached copies of :term:`packages <Distribution Package>`:
    by downloading all the requirements for a project and then pointing pip at
    those downloaded files instead of going to PyPI.
 2. A variation on the above which pre-builds the installation files for
-   the requirements using `pip wheel
+   the requirements using `python -m pip wheel
```
webknjaz, line 37: I wouldn't use it in links, only in code blocks.

source/tutorials/installing-packages.rst
```diff
@@ -77,7 +77,7 @@ please install the latest 3.x version from `python.org`_ or refer to the
 .. Note:: Due to the way most Linux distributions are handling the Python 3
    migration, Linux users using the system Python without creating a virtual
    environment first should replace the ``python`` command in this tutorial
-   with ``python3`` and the ``pip`` command with ``pip3 --user``. Do *not*
+   with ``python3`` and the ``python -m pip`` command with ``python3 -m pip3 --user``. Do *not*
```
webknjaz, line 80:
```suggestion
   with ``python3`` and the ``python -m pip`` command with ``python3 -m pip --user``. Do *not*
```
There's no module called `pip3`, only an executable console script.

source/tutorials/managing-dependencies.rst
```diff
@@ -39,7 +39,7 @@ Use ``pip`` to install Pipenv:
 
 .. code-block:: python
```
webknjaz, line 40: While you're here, fix this:
```suggestion
.. code-block:: bash
```

source/tutorials/packaging-projects.rst
```diff
@@ -349,7 +349,7 @@ differences:
   for the account you registered on the real PyPI.  Now that you're uploading
   the package in production, you don't need to specify ``--repository``; the
   package will upload to https://pypi.org/ by default.
-* Install your package from the real PyPI using ``pip install [your-package]``.
+* Install your package from the real PyPI using ``python3 -m pip install [your-package]``.
```
webknjaz, line 352: Why don't we keep consistency with other places?
```suggestion
* Install your package from the real PyPI using ``python -m pip install [your-package]``.
```

webknjaz (changes requested):

source/tutorials/packaging-projects.rst
scop: This file uses `python3` for all its samples, using that here specifically to keep the consistency within the file.

source/discussions/pip-vs-easy-install.rst
scop: I don't see a reason to be inconsistent, but if you feel strongly about this, feel free to push a change to the PR branch.

source/guides/index-mirrors-and-caches.rst
scop: Same as above re using in tables.

source/discussions/pip-vs-easy-install.rst
webknjaz: I don't have access to that. Not all org members have commit bits in all the repos.

My understanding is that this compares rather abstract ways of doing things and not ways of invoking pip with a specific interpreter. It makes sense to have `python -m` in code blocks as they are what the readers would copy-and-paste most often. But in other places, the long-form seems to be overkill...

webknjaz (approved):
While I've expressed my preference to have some things done slightly differently, that is not to be considered as a blocker. LGTM regardless of whether the suggestions are incorporated or not.

di (approved):

PR #810, opened 2021-01-02
Tweak opening "important note" in installing-packages tutorial

pradyunsg:
This flips the order of "synonym" vs "meaning", to convey the meaning before using
a loaded term that'll link to the glossary. This should help make it less intimidating to
first-time readers as well as generally make it easier to follow the sentence.

PS: It's likely easier to review this PR in "rich diff" mode, since I've had to re-wrap the entire passage.

### DIFF
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -7,14 +7,15 @@ Installing Packages
 This section covers the basics of how to install Python :term:`packages
 <Distribution Package>`.
 
-It's important to note that the term "package" in this context is being used as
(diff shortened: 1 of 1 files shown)


pradyunsg:
Quick link to rich diff, for potential reviewers: https://github.com/pypa/packaging.python.org/pull/810/files?short_path=2cf5869#diff-2cf586919b869d30eb978eb3877500e2d3a6b7e1feb0e03605bb2e25b490c2a1

pradyunsg:
Merging based on a self-review -- I had a day in-between, so it's fine. :P

Issue #787, opened 2020-11-05
minor English misuse


aendie:
With reference to the following page: https://packaging.python.org/overview/

At the very bottom there is the text "Once you understand the natural boundaries between each packaging solution, you begin to realize that the varied landscape is a small price Python programmers pay for using one of the most balanced, flexible language available."

The expression "one of the most" always refers to a plural, as in "one of the most frequented streets in a city". So it would be correct IMHO to use the following text: "one of the most balanced, flexible **languages** available"

webknjaz:
Feel free to send a PR

PR #811, opened 2021-01-03
Fix typos

hugovk:
Fixes https://github.com/pypa/packaging.python.org/issues/787 and removes an unnecessary, unbalanced quote mark.


pradyunsg (approved):

kaycoinsUSA (commented):
Done

PR #818, opened 2021-01-29
chore: use setup.cfg as main config file

henryiii:
Followup to #817, related to #809.

---

I've split out the setup.cfg changes. I still would highly recommend setup.cfg, because:

* It is a static, easily parsable file, which is highly superior to a Python file - for example, cibuildwheel should soon gain support for reading the Requires-Python metadata from the PEP 621 location or from setup.cfg, but setup.py is impossible to reliably parse.
* It avoids many common issues, like opening files without with, encoding issues, and issues with accessing the version.
* It keeps setup.py as _just_ logic, making it clear what's "dynamic" and what's just regular metadata.
* It's actually readable, much less boilerplate
* Less chance of importing from distutils before setuptools
* The PyPA projects, like twine, build, etc, use setup.cfg.
* It's been supported since 2016
* It's really easy to just say the fields are dynamically specifiable as keywords to setup, while it is harder to say the keywords to setup might be in `[options]` or in `[metadata]` or in ...

While I'm really excited for PEP 621 support, that's quite some time away (looks like funding has not even been secured yet), and then there's another period for the tooling to mature and become common enough to be used, it won't ever support Python 2 / PyPy 2, etc, so I expect adoption to be slow. And setup.cfg is still not going away, with some tools strongly refusing (cough, flake8, cough cough) to ever add pyproject.toml support (MyPy is still setup.cfg only too).

If people are writing setup.cfg's, it will be much easier to transition to pyproject.toml than from setup.py.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -95,8 +95,8 @@ would be completely different than the setuptools configuration described
 below. See :pep:`517` and :pep:`518` for background and details.
 
 
-Creating setup.cfg and setup.py
--------------------------------
+Configuring metadata
+--------------------
 
 You can place the metadata for your project in one of two places. Static
 information should be placed in :file:`setup.cfg`. This is simpler, easier to
@@ -107,7 +107,7 @@ really want to; both methods are shown below.
 
 .. tabs::
 
-  .. tab:: Configure with setup.cfg (static)
+  .. tab:: setup.cfg (static)
 
     :file:`setup.cfg` is the configuration file for :ref:`setuptools`. It tells
     setuptools about your package (such as the name and version) as well as which
@@ -197,7 +197,7 @@ really want to; both methods are shown below.
 
         setuptools.setup()
 
-  .. tab:: Configure with setup.py (dynamic)
+  .. tab:: setup.py (dynamic)
 
     :file:`setup.py` is the build script for :ref:`setuptools`. It tells setuptools
     about your package (such as the name and version) as well as which code files
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -98,12 +98,16 @@ below. See :pep:`517` and :pep:`518` for background and details.
 Configuring metadata
 --------------------
 
-You can place the metadata for your project in one of two places. Static
-information should be placed in :file:`setup.cfg`. This is simpler, easier to
-read, and avoids many common errors, like encoding errors. Any items that are
-dynamic, as well as extension modules or extensions to setuptools, need to go
-into :file:`setup.py`. You can completely configure via :file:`setup.py` if you
-really want to; both methods are shown below.
+There are two types of metadata: static and dynamic.
+
+* Static metadata (:file:`setup.cfg`): guaranteed to be the same every time. This is
+  simpler, easier to read, and avoids many common errors, like encoding errors.
+* Dynamic metadata (:file:`setup.py`): possibly non-deterministic. Any items that are
+  dynamic or determined at install-time, as well as extension modules or
+  extensions to setuptools, need to go into :file:`setup.py`.
+
+Static metadata should be preferred and dynamic metadata should be used only as
+an escape hatch when absolutely necessary.
 
 .. tabs::
 
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -50,6 +50,7 @@ project's root directory - you will add content to them in the following steps.
     ├── example_pkg
     │   └── __init__.py
     ├── pyproject.toml
+    ├── setup.cfg
     ├── setup.py
     └── tests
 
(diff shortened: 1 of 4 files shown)


di:
Can we experiment with using https://pypi.org/project/sphinx-tabs/ or https://pypi.org/project/sphinxcontrib-contentui/ to provide both the `setup.py` and `setup.cfg` configuration options simultaneously?

henryiii:
I went with exectuablebooks' sphinx-tabs, as I'm a Jupyterbook fan. (https://henryiii.github.io/level-up-your-python ;) ), and it likely is well maintained. Here's what the current state looks like:

Screen Shot 2021-01-29 at 2 44 00 PM

source/tutorials/packaging-projects.rst
```diff
@@ -93,82 +94,183 @@ always include it. If you were to use a different build system, such as
...
 
+Creating setup.cfg and setup.py
+-------------------------------
+
+You can place the metadata for your project in one of two places. Static
+information should be placed in :file:`setup.cfg`. This is simpler, easier to
+read, and avoids many common errors, like encoding errors. Any items that are
+dynamic, as well as extension modules or extensions to setuptools, need to go
+into :file:`setup.py`. You can completely configure via :file:`setup.py` if you
+really want to; both methods are shown below.
```
di, line 106: I think we can be a bit more opinionated here that static metadata should be preferred, and that dynamic metadata should be an escape hatch. We should also structure this in a way that we can easily add `pyproject.toml` once PEP 621 lands.

Something like:

---

There are two types of metadata: static and dynamic.

* Static metadata (`setup.cfg`): guaranteed to be the same every time. This is simpler, easier to
read, and avoids many common errors, like encoding errors.
* Dynamic metadata (`setup.py`): possibly non-deterministic. Any items that are dynamic or determined at install-time, as well as extension modules or extensions to setuptools, need to go into `setup.py`.

Static metadata should be preferred and dynamic metadata should be used only as an escape hatch when absolutely necessary.

source/tutorials/packaging-projects.rst
```diff
@@ -93,82 +94,183 @@ always include it. If you were to use a different build system, such as
...
+
+.. tabs::
+
+  .. tab:: Configure with setup.cfg (static)
```
di, line 110:
```suggestion
  .. tab:: setup.cfg (static)
```

A nit, but can we make `setup.cfg` monospaced here?

source/tutorials/packaging-projects.rst
```diff
@@ -93,82 +94,183 @@ always include it. If you were to use a different build system, such as
...
+    :func:`setup()`; this enables computed values to be used. You will also need
+    :func:`setup()` for setting up extension modules for compilation.
+
+    .. code-block:: python
+
+        import setuptools
+
+        setuptools.setup()
+
+  .. tab:: Configure with setup.py (dynamic)
```
di, line 200:
```suggestion
  .. tab:: setup.py (dynamic)
```

Same here about monospacing.

source/tutorials/packaging-projects.rst
henryiii: Don't think so, opened an issue.

di (approved):
LGTM, will give it a day or so to let other folks review as well before merging.

pradyunsg:
May I suggest using [sphinx-inline-tabs](https://github.com/pradyunsg/sphinx-inline-tabs) instead? I think it looks nicer. :)

di:
Oh, look at that! Very nice @pradyunsg. Does it support markup in the tab title?

henryiii:
I'll try it soon! :)

henryiii:
No, it doesn't, though it seems to try harder:

Screen Shot 2021-02-02 at 10 43 20 AM

henryiii:
I'll push a version that uses this:

Screen Shot 2021-02-02 at 10 51 03 AM

gaborbernat (approved):

gaborbernat:
@henryiii instead of posting the image it's better to post the link to the rendered document, e.g. https://python-packaging-user-guide--818.org.readthedocs.build/tutorials/packaging-projects/#configuring-metadata

henryiii:
The link is available already, "Details" on the readthedocs check. Putting an image in shows what it looks like without having to click a link, and once I've changed it, you only see the new version with a link, so you can't compare. Though I could have made the image clickable, I suppose :)

gaborbernat:
> The link is available already, "Details" on the readthedocs check. Putting an image in shows what it looks like without having to click a link, and once I've changed it, you only see the new version with a link, so you can't compare.

That link takes you to the main page, not the section you want to view the change rendered. It's nice to have a direct link available 😄

di:
Thanks @henryiii!

henryiii:
FYI, sphinx-tabs now supports rst in tab headings, including math. https://github.com/executablebooks/sphinx-tabs/pull/102

bfox85 (commented):
V

webknjaz:
Hey @bfox85, this sort of drive-through reviews is rather useless and is considered spam. Please use words if you have something meaningful to add here.

pradyunsg:
> Does it support markup in the tab title?
> [snip]
> No, it doesn't, though it seems to try harder:

It does now!

https://github.com/pradyunsg/sphinx-inline-tabs/issues/13

PR #820, opened 2021-02-03
Replace Travis CI with GitHub Actions

hugovk:
Travis CI has a new pricing model which places limits on open source.

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Travis support has also become unresponsive to developers who have run out of free credits, and many projects are moving to GitHub Actions instead. 

And https://github.com/pypa builds can not longer build:

> Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

image

> Owner pypa does not have enough credits.

image

---

This GHA config does much the same as the deleted Travis one. Two differences: tests with current newest Python 3.9 instead of 3.6; and adds caching of pip downloads, to try and help ease the load on PyPI, and speed up builds a bit.

Sample build:

* https://github.com/hugovk/packaging.python.org/runs/1822833000?check_suite_focus=true

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           path: ~/.cache/pip
           key:
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}
+            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('noxfile.py') }}
           restore-keys: |
             ${{ matrix.os }}-${{ matrix.python-version }}-
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
           python -m pip install --upgrade nox virtualenv
 
       - name: Nox build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,33 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key:
+            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}
```
webknjaz, line 22: Don't forget about the noxfile:
```suggestion
            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('noxfile.py') }}
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
...
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.python-version }}-
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
webknjaz, line 28: I don't see why upgrading pip would be necessary. Almost everything is going to be happening in a nox-managed virtualenv.

webknjaz:
FWIW this PR isn't urgent because we have an RTD-based check in PRs that essentially does the same smoke-test.

.github/workflows/test.yml
hugovk: Sure, will ditch it 👍

.github/workflows/test.yml
hugovk: Does the noxfile define any requirements? Looks like it defers to requirements.txt?

```py
    session.install("-r", "requirements.txt")
```

Or do we want to clear the cache when the noxfile is modified anyway?

webknjaz: Not sure. My instinct is to include it in the cache key. There's also another session defined that we don't execute in CI, it installs `sphinx-autobuild` that is not present in normal requirements.

hugovk: It's not a big deal either way, let's include it. Thanks!

webknjaz (approved):

di (approved):

PR #823, opened 2021-02-06
fix: add a few missing parts to setup.cfg setup

henryiii:
Closes pypa/packaging-problems#436

I've reordered the fields to match the description order, and synced the description orders for the two tabs. (I think I matched the order that is given in the setuptools docs originally for the cfg lines). I also added a missing description for `python_requires` to both tabs.

Personally, I rather think `python_requires` should have been metadata, not options, since it fills the Requires-Python slot, but it is in options.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -128,25 +128,26 @@ an escape hatch when absolutely necessary.
         # replace with your username:
         name = example-pkg-YOUR-USERNAME-HERE
         version = 0.0.1
-        url = https://github.com/pypa/sampleproject
         author = Example Author
         author_email = author@example.com
+        description = A small example package
+        long_description = file: README.md
+        long_description_content_type = text/markdown
+        url = https://github.com/pypa/sampleproject
         classifiers =
             Programming Language :: Python :: 3
             License :: OSI Approved :: MIT License
             Operating System :: OS Independent
-        description = A small example package
-        long_description = file: README.md
-        long_description_content_type = text/markdown
 
         [options]
+        packages = find:
         python_requires = >=3.6
 
 
     There are a `variety of metadata and options
     <https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html>`_
     supported here. This is in configparser format; do not place quotes around values.
-    This example package uses a relatively minimal set of options:
+    This example package uses a relatively minimal set of metadata:
 
     - ``name`` is the *distribution name* of your package. This can be any name as
       long as only contains letters, numbers, ``_`` , and ``-``. It also must not
@@ -168,11 +169,6 @@ an escape hatch when absolutely necessary.
     - ``url`` is the URL for the homepage of the project. For many projects, this
       will just be a link to GitHub, GitLab, Bitbucket, or similar code hosting
       service.
-    - ``packages`` is a list of all Python :term:`import packages <Import
-      Package>` that should be included in the :term:`Distribution Package`.
-      Instead of listing each package manually, we can use the ``find:`` directive
-      to automatically discover all packages and subpackages. In this case, the
-      list of packages will be ``example_pkg`` as that's the only package present.
     - ``classifiers`` gives the index and :ref:`pip` some additional metadata
       about your package. In this case, the package is only compatible with Python
       3, is licensed under the MIT license, and is OS-independent. You should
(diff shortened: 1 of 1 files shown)


henryiii:
FYI, the example I've used for the last year or so is here: https://scikit-hep.org/developer/packaging#setup-configuration-medium-priority

webknjaz:
@henryiii the PR title doesn't seem to reflect what you did here

henryiii:
There are three changes; I've added `package = find:`, I've reordered the options so they match the descriptions below, and I've added the missing `python_requires` description. Maybe I should generalize the title a bit? I was initially just trying to solve pypa/packaging-problems#436, and then was looking for anything else I might have possibly missed similar to that.

webknjaz:
Honestly, I find it easier to review/approve separate unrelated changes so if one of the updates wouldn't block others from being merged.

source/tutorials/packaging-projects.rst
```diff
@@ -128,25 +128,26 @@ an escape hatch when absolutely necessary.
         # replace with your username:
         name = example-pkg-YOUR-USERNAME-HERE
         version = 0.0.1
-        url = https://github.com/pypa/sampleproject
         author = Example Author
         author_email = author@example.com
+        description = A small example package
+        long_description = file: README.md
+        long_description_content_type = text/markdown
+        url = https://github.com/pypa/sampleproject
```
webknjaz, line 136: I think it may be useful to also mention `project_urls` demonstrating links to issues/docs/CI/etc

source/tutorials/packaging-projects.rst
```diff
@@ -259,18 +266,21 @@ an escape hatch when absolutely necessary.
...
       your package will work on. For a complete list of classifiers, see
       https://pypi.org/classifiers/.
+    - ``packages`` is a list of all Python :term:`import packages <Import
+      Package>` that should be included in the :term:`Distribution Package`.
+      Instead of listing each package manually, we can use :func:`find_packages`
+      to automatically discover all packages and subpackages. In this case, the
+      list of packages will be ``example_pkg`` as that's the only package present.
+    - ``python_requires`` gives the versions of Python supported by your
+      project. Installers like pip will look back though older versions of
+      packages until it finds one that has a matching Python version.
```
webknjaz, line 283: Would it be possible to deduplicate the copy-pasted explanations? Do you think using the `.. include::` directive would help here?

henryiii:
Should I break it out? It's really just follow up cleanup to #818, and reordering doesn't go well in parallel with other changes. Happy to pull out in any order you want, though.

source/tutorials/packaging-projects.rst
henryiii: I was trying to exactly replicate the setup.py example with no changes; would be happy to add a follow-on PR that adds anything that would be useful to add to the example.

source/tutorials/packaging-projects.rst
henryiii: Only some of them match; the "packages" example right above it, for example, does not match since it has a discussion of the `find:` directive in one case, and `setuptools.find_packages()` in the other.

henryiii: Can you include from the same file? It might be fragile, though.

source/tutorials/packaging-projects.rst
henryiii: Personally, I didn't know about / start using project_urls until fairly recently, I don't I've even added it to the scikit-hep dev docs yet. But it is really nicely displayed in PyPI, so I think it's a good idea, even if not really minimal.

webknjaz: Yeah, I think I've been using it for the last 3-4 years maybe...

source/tutorials/packaging-projects.rst
webknjaz: I think you can only "include" from the same file with substitutions. Including with "start+end" of itself with the `include` directive sounds a bit wild but I don't see why it wouldn't work (at least in cases when you don't include the place with the directive itself) :exploding_head:

webknjaz:
> Should I break it out?

Usually when the title contains references to many things, has "and" in it, or is too vague within the 50 char limit, it's a sign that the PR is too broad and when you can describe its separate parts with a few specific titles, it's a good idea to have a PR per such a title. The person who would merge the PR may get stuck trying to come up with a good commit message for the merge when they'll be attempting to replace "do something related to setup.cfg" with anything that would give the reader any idea what the commit is about.

henryiii:
It would be nice to get this in, it's an edit to something I originally wrote in #818, and the missing line is tripping people up. If there's an issue with something above, I can pull out just the missing line, but there should not be anything controversial in it, I'm just matching the option order, adding the missing line, and added a missing description. I'm about to take off two weeks for the upcoming birth of my daughter, so would prefer not to make a series of very tiny PRs that have to be sequential...

webknjaz (approved):

di (approved):

PR #839, opened 2021-02-25
Move wheel spec into packaging.python.org specifications section

takluyver:
A [small change to the wheel spec](https://discuss.python.org/t/escaping-versions-for-wheel-sdist-and-dist-info-names/5605/8) is likely to better represent version numbers in wheel filenames. @pfmoore suggested that the wheel spec should be moved here as part of that.

For now, I have copied the contents of PEP 427 to here with no changes, to simplify review.

### DIFF
--- a/source/specifications/binary-distribution-format.rst
+++ b/source/specifications/binary-distribution-format.rst
@@ -5,4 +5,463 @@
 Binary distribution format
 ==========================
 
-The binary distribution format (:term:`wheel <Wheel>`) is defined in :pep:`427`.
+The binary distribution format (:term:`wheel <Wheel>`) was originally defined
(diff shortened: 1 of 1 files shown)


pfmoore (approved):
Can you also submit a PR for PEP 427 noting that it is no longer the final reference on the wheel spec, and linking here?

Also, it's possible that including the "Copyright" section from the original PEP isn't appropriate here. I don't know what's legally correct, so I'll assume it's OK as it is.

webknjaz (approved):
I think that copyright note is unnecessary but harmless since this repo's license is CC BY-SA 3.0 which is roughly the same as public domain.

pfmoore:
Thanks @webknjaz 🙂

PR #844, opened 2021-02-28
Wheel spec: Update rules on escaping components of wheel filename

takluyver:
From the discussion at https://discuss.python.org/t/escaping-versions-for-wheel-sdist-and-dist-info-names/5605

These are the same rules as for .dist-info directories, specified in [recording installed packages](https://packaging.python.org/specifications/recording-installed-packages/), but with a bit more description for names.

### DIFF
--- a/source/specifications/binary-distribution-format.rst
+++ b/source/specifications/binary-distribution-format.rst
@@ -155,10 +155,17 @@ package's basic interpreter requirements and are detailed in PEP 425.
 Escaping and Unicode
 ''''''''''''''''''''
 
-Each component of the filename is escaped by replacing runs of
-non-alphanumeric characters with an underscore ``_``::
+As the components of the filename are separated by a dash (``-``, HYPHEN-MINUS),
+this character cannot appear within any component. This is handled as follows:
 
-    re.sub("[^\w\d.]+", "_", distribution, re.UNICODE)
+- In distribution names, any run of ``-_.`` characters (HYPHEN-MINUS, LOW LINE
+  and FULL STOP) should be replaced with ``_`` (LOW LINE). This is equivalent
+  to :pep:`503` normalisation followed by replacing ``-`` with ``_``.
+- Version numbers should be normalised according to :pep:`440`, and then ``-``
+  characters should be replaced with ``_`` (LOW LINE). Normalised version
+  numbers cannot contain ``_``, so the replacement does not lose information.
+- The remaining components may not contain ``-`` characters, so no escaping
+  is necessary.
 
 The archive filename is Unicode.  It will be some time before the tools
 are updated to support non-ASCII filenames, but they are supported in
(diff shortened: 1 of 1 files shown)


uranusjr (approved):

FFY00 (approved):

source/specifications/binary-distribution-format.rst
```diff
@@ -155,10 +155,17 @@ package's basic interpreter requirements and are detailed in PEP 425.
...
-    re.sub("[^\w\d.]+", "_", distribution, re.UNICODE)
+As the components of the filename are separated by a dash (``-``, HYPHEN-MINUS),
+this character cannot appear within any component. This is handled as follows:
+
+- In distribution names, any run of ``-_.`` characters (HYPHEN-MINUS, LOW LINE
+  and FULL STOP) should be replaced with ``_`` (LOW LINE). This is equivalent
+  to :pep:`503` normalisation followed by replacing ``-`` with ``_``.
+- Version numbers should be normalised according to :pep:`440`, and then ``-``
+  characters should be replaced with ``_`` (LOW LINE). Normalised version
+  numbers cannot contain ``_``, so the replacement does not lose information.
```
pfmoore, line 166: Why the hyphen -> underscore step? As you say, PEP 440 normalisation cannot contain dashes anyway, so adding the replacement step feels like we don't believe in what we've just said. We should either drop the replacement, or add a justification for it.

source/specifications/binary-distribution-format.rst
```diff
@@ -155,10 +155,17 @@ package's basic interpreter requirements and are detailed in PEP 425.
...
+- The remaining components may not contain ``-`` characters, so no escaping
+  is necessary.
```
pfmoore, line 168: I personally feel that we should require tools generating wheels to fail with an error if the above rules result in any component still containing a dash, just as a last-ditch safety measure to protect against future standards changes failing to take the wheel naming requirements into account, but no-one else seemed too interested in that, so I'll let it drop at this point.

pfmoore (changes requested):

source/specifications/binary-distribution-format.rst
uranusjr: Personally I feel it makes the specification more future-proof. If there's ever a PEP 440 replacement that does allow `-`, that spec won't need to remember to amend the rules.

pfmoore: But won't we then just be back in this situation, with loss of information (because the only reason we don't lose information right now is that dashes aren't allowed)?

Meh, I'm not going to argue this endlessly. It's not that big a deal, and getting something that works is more important.

pfmoore (approved):

source/specifications/binary-distribution-format.rst
takluyver: It doesn't lose information because *underscores* aren't in the normalised form. I thought that dashes were, but on looking at PEP 440 again, I see that I'm wrong, so I agree with you that they should just be normalised.

takluyver:
I addressed @pfmoore's suggestions in #845

PR #850, opened 2021-03-03
fix: Remove '<4' from python_requires in packaging guide

matthewfeickert:
As discussed in the discuss.python topic ["Use of “less-than next-major-version” (e.g., `<4`) in `python_requires` (setup.py)"](https://discuss.python.org/t/use-of-less-than-next-major-version-e-g-4-in-python-requires-setup-py/1066) and other places there can be some side effects of adding `<4` to a library's `python_requires`. This PR removes it, as requested by @njsmith in the discuss.python topic.

The section above the lines discussed in the topic that uses compatible release syntax is also removed given that `~=3.3` also enforces `<4`.

My apologies in advance if this was already discussed in an Issue and decided against (I didn't see any when I scrolled through) or in a previously closed PR. Please let me know if you'd like me to open an Issue to discuss this PR topic instead.

Tagging @pfmoore and @njsmith as they were the devs in the discuss.python topic and @NickleDave and @henryiii as they might be interested in this as well.

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -397,15 +397,10 @@ versions.  For example, if your package is for Python 3+ only, write::
 
     python_requires='>=3',
 
-If your package is for Python 3.3 and up but you're not willing to commit to
-Python 4 support yet, write::
-
-    python_requires='~=3.3',
-
 If your package is for Python 2.6, 2.7, and all versions of Python 3 starting
 with 3.3, write::
 
-    python_requires='>=2.6, !=3.0.*, !=3.1.*, !=3.2.*, <4',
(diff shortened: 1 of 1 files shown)


henryiii (approved):
Keep in mind, unlike for libraries (where honestly this is still a poor idea for a general library pin), using Requires-Python causes pip to scroll back through history to find the first matching version that does not have a  Requires-Python slot that does not include the current Python. This is _never correct_ for a "future" version, but can only be used for the use case it was designed for - dropping a version that you used to support.

pfmoore (approved):

webknjaz (approved):

matthewfeickert:
Thanks very much for the fast review @pfmoore and @webknjaz!

nnadeau:
For posterity, related to: https://github.com/nedbat/coveragepy/pull/1021

PR #851, opened 2021-03-06
Describe the file format of core metadata

takluyver:
Following [this discussion on discourse](https://discuss.python.org/t/core-metadata-email-fields-unicode/7421). The file format of core metadata is not precisely defined (PEP 241 said RFC 822, but that's outdated in practice), but there is a practical standard: the major tools consuming metadata use the `email.parser` module with the (default) `compat32` policy. I think it's better to describe the situation than to leave it completely unspecified.

Hopefully this is an interim measure until someone gets around to describing the format properly. But that's a much bigger job, and so far no-one is volunteering.

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -13,6 +13,20 @@ complete and not subject to change. The required fields are:
 
 All the other fields are optional.
 
+The standard file format for metadata (including in :doc:`wheels
+<binary-distribution-format>` and :doc:`installed projects
+<recording-installed-packages>`) is based on the format of email headers.
+However, email formats have been revised several times, and exactly which email
+RFC applies to packaging metadata is not specified. In the absence of a precise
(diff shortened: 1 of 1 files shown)


pradyunsg (approved):

pyfisch:
While I've started a discussion to [specify the format in greater detail](https://discuss.python.org/t/python-metadata-format-specification-and-implementation/7550?u=pyfisch) there are two things that can be checked to ensure invalid files are rejected by tools:

1. Make sure the [`defects`](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.defects) attribute is the empty list.
2. Check if the message contains both a `Description` header and a non-empty payload. This is a unambiguous sign that the message syntax is invalid.

webknjaz (approved):

webknjaz:
I think this change is good and any additional changes can be submitted as follow-up PRs.

PR #854, opened 2021-03-07
Fix a typo in `packaging-projects.rst` under `Creating pyproject.toml`

arshadrr:
fix a little typo

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -65,7 +65,7 @@ Creating pyproject.toml
 -----------------------
 
 :file:`pyproject.toml` is the file that tells build tools (like ``pip`` 10+ and
-``build``) what system you are using and what it requires for building. The
+``build``) what system you are using and what is required for building. The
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -65,7 +65,7 @@ Creating pyproject.toml
 -----------------------
 
 :file:`pyproject.toml` is the file that tells build tools (like ``pip`` 10+ and
-``build``) what system you are using and what it required for building. The
+``build``) what system you are using and what it requires for building. The
```
webknjaz, line 68: Hm... Maybe it was supposed to be "is required"? Let's ask @henryiii.

henryiii: Yes, “is required” is what was intended here, and I think a slightly better fix. @arshadrr could you make this small tweak?

webknjaz: I can do this

webknjaz (approved):

PR #859, opened 2021-03-11
fix: Use src directory structure for packaging tutorial

### LINKED ISSUE #847: Code and tutorial not matching
Hi all,
I'm a bit confused because on 56d3120 you switched over to using the src/ layout but the [tutorial](https://packaging.python.org/tutorials/packaging-projects/) still shows this example:
```
packaging_tutorial
├── LICENSE
├── README.md
├── example_pkg
│   └── __init__.py
├── pyproject.toml
├── setup.cfg
├── setup.py
└── tests
```

I was just wondering which layout is preferred/suggested by pypa (and if src/layout is the optimal choise, perhaps the guide needs an update since it's confusing).

Thank you!

matthewfeickert:
Resolves #847

This PR adopts the use of a `src` directory for the "Packaging Python Projects" tutorial. In addition to the use of `src` it introduces `package_dir` and where to look for packages with `[options.packages.find]` in `setup.cfg` or the `where` arg for `setuptools.find_packages` in `setup.py`.

The added documentation adopts the use of the term "root package" from [the old Distutils Examples docs](https://docs.python.org/3/distutils/examples.html#pure-python-distribution-by-package), but there may be more preferred terminology for this and I'm happy to revise this as desired.

Tagging @di as he transferred Issue #847 from pypa/sampleproject.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -21,14 +21,14 @@ To create this project locally, create the following file structure:
 .. code-block:: text
 
     packaging_tutorial
-    └── example_pkg
-        └── __init__.py
-
+    └── src
+        └── example_pkg
+            └── __init__.py
 
 Once you create this structure, you'll want to run all of the commands in this
 tutorial within the top-level folder - so be sure to ``cd packaging_tutorial``.
 
-:file:`example_pkg/__init__.py` is required to import the directory as a package,
+:file:`src/example_pkg/__init__.py` is required to import the directory as a package,
 and can simply be an empty file.
 
 .. _Python documentation for packages and modules:
@@ -46,12 +46,13 @@ project's root directory - you will add content to them in the following steps.
 
     packaging_tutorial
     ├── LICENSE
-    ├── README.md
-    ├── example_pkg
-    │   └── __init__.py
     ├── pyproject.toml
+    ├── README.md
     ├── setup.cfg
     ├── setup.py  # optional, needed to make editable pip installs work
+    ├── src
+    │   └── example_pkg
+    │       └── __init__.py
     └── tests
 
 
@@ -142,9 +143,13 @@ an escape hatch when absolutely necessary.
             Operating System :: OS Independent
 
         [options]
+        package_dir =
+            = src
         packages = find:
         python_requires = >=3.6
 
+        [options.packages.find]
+        where = src
 
     There are a `variety of metadata and options
     <https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html>`_
@@ -252,7 +257,8 @@ an escape hatch when absolutely necessary.
                 "License :: OSI Approved :: MIT License",
                 "Operating System :: OS Independent",
             ],
-            packages=setuptools.find_packages(),
+            package_dir={"": "src"},
+            packages=setuptools.find_packages(where="src"),
             python_requires=">=3.6",
         )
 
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -183,11 +188,16 @@ an escape hatch when absolutely necessary.
 
     In the options category, we have controls for setuptools itself:
 
+    - ``package_dir`` is a collection of package names and directories.
```
matthewfeickert, line 191: "collection" is maybe(?) not the best term here, so I'm happy to swap this out with suggestions.

webknjaz:
FTR here are some notes from PyPA Mini-Summit (PyCon 2019):
https://docs.google.com/document/d/1Wz2-ECkicJgAmQDxMFivWmU2ZunKvPZ2UfQ59zDGj7g/edit#heading=h.2cgqnlxl8y3e.
There are action items at the end of this section.

IIRC @pganssle @ncoghlan @dstufft moderated that specific discussion — CCing you in case you want to add some comments here.

pradyunsg:
FWIW, I'm fine with this PR. I don't think anyone has spent the time to write the required docs, and this change is fine on its own.

source/tutorials/packaging-projects.rst
```diff
@@ -183,11 +188,16 @@ an escape hatch when absolutely necessary.
...
+      An empty package name represents the "root package", so in this
```
webknjaz, line 192: FYI this was debated two years ago — https://docs.google.com/document/d/1Wz2-ECkicJgAmQDxMFivWmU2ZunKvPZ2UfQ59zDGj7g/edit#heading=h.7c3gvumy3508:

> ###### Names for the non-src layout:
> - Flat *(least-disliked option)*
> - Root (potentially non-obvious to beginners; potential confusion with, e.g., ‘filesystem root’)
> - "Bad" (.....)
> - Non-src (may imply unduly that “src” is preferred)
> - Natural (may imply unduly that “src” is discouraged)

source/tutorials/packaging-projects.rst
webknjaz: iterable?

source/tutorials/packaging-projects.rst
```diff
@@ -46,12 +46,13 @@ project's root directory - you will add content to them in the following steps.
...
     ├── LICENSE
-    ├── README.md
-    ├── example_pkg
-    │   └── __init__.py
     ├── pyproject.toml
+    ├── README.md
     ├── setup.cfg
     ├── setup.py  # optional, needed to make editable pip installs work
+    ├── src
+    │   └── example_pkg
```
webknjaz, line 54: WDYT about marking the directories with a trailing slash?
```suggestion
    ├── src/
    │   └── example_pkg/
```

webknjaz (approved):
Pradyun is right. It's good, it makes the doc better and has nothing harmful. I'm going to merge this because having this change live is more important than iterating on improvements that could always be submitted as follow-ups.

matthewfeickert:
@pradyunsg @webknjaz Thanks for the fast review and merge. FWIW, I'm always thankful for reviewer suggestions like the ones @webknjaz added, so I'll open up a follow up PR after work today to implement those. :+1:

di:
Thanks @matthewfeickert!

source/tutorials/packaging-projects.rst
matthewfeickert: I'm actually strongly in favor of this. The only reason I didn't include these is to match style with the existing tutorial, but I will happily add them in my follow up PR. :slightly_smiling_face:

PR #860, opened 2021-03-12
fix: Apply formatting corrections for packaging tutorial

matthewfeickert:
This PR is a follow up to PR #859 to apply suggested revisions and formatting (cc @webknjaz).

* Adds directory trailing slashes to indicate dir structure for clarity, so abandons `tree` behavior for `ls -F` behavior :+1: 
* Revise phrasing for `package_dir` description: "collection" -> "mapping"
* Attempts to clarify the definition of "root package"

In PR #859 it was also [suggested that](https://github.com/pypa/packaging.python.org/pull/859/files#r592172904)

> An empty package name represents the "root package", so in this case the root package is in the ``src`` directory.

might get rephrased to "flat", but the links provided to the discussion at the PyCon 2019 - PyPA Sprints + Summit notes seem to indicate that this is referring to terminology related to a **non**-src directory structure. Here the phrase "root package" is referring to the directory in the project directory structure that contains all Python source files for the package and is taken from the [old Distutils Examples docs](https://docs.python.org/3/distutils/examples.html#pure-python-distribution-by-package) where this pattern still applies in `setuptools` today.

distutils

As "the directory in the project that contains all Python source files for the package" is perhaps unwieldy for someone totally new to packaging reading the tutorial I welcome suggestions for revision here!

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -20,9 +20,9 @@ To create this project locally, create the following file structure:
 
 .. code-block:: text
 
-    packaging_tutorial
-    └── src
-        └── example_pkg
+    packaging_tutorial/
+    └── src/
+        └── example_pkg/
             └── __init__.py
 
 Once you create this structure, you'll want to run all of the commands in this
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -188,9 +188,10 @@ an escape hatch when absolutely necessary.
 
     In the options category, we have controls for setuptools itself:
 
-    - ``package_dir`` is a collection of package names and directories.
-      An empty package name represents the "root package", so in this
-      case the root package is in the ``src`` directory.
+    - ``package_dir`` is a mapping of package names and directories.
+      An empty package name represents the "root package" --- the directory in
+      the project that contains all Python source files for the package --- so
```
matthewfeickert, line 193: ...actually, even this definition might be a bit rubbish as I could see a new user asking why `example_pkg/` isn't the "root package" then. :/

webknjaz: Well, we can address this when somebody actually reports the confusion.

webknjaz (approved):

PR #861, opened 2021-03-18
docs: add info how to override pip's default behavior

jugmac00:
... when installing packages from PyPI.

### DIFF
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -291,7 +291,8 @@ Source Distributions vs Wheels
 
 :ref:`pip` can install from either :term:`Source Distributions (sdist) <Source
 Distribution (or "sdist")>` or :term:`Wheels <Wheel>`, but if both are present
-on PyPI, pip will prefer a compatible :term:`wheel <Wheel>`.
+on PyPI, pip will prefer a compatible :term:`wheel <Wheel>`. You can override
+pip`s default behavior by e.g. using its `--no-binary`_ option.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/installing-packages.rst
```diff
@@ -291,7 +291,8 @@ Source Distributions vs Wheels
 
 :ref:`pip` can install from either :term:`Source Distributions (sdist) <Source
 Distribution (or "sdist")>` or :term:`Wheels <Wheel>`, but if both are present
-on PyPI, pip will prefer a compatible :term:`wheel <Wheel>`.
+on PyPI, pip will prefer a compatible :term:`wheel <Wheel>`. You can override
+pip`s default behavior by e.g. using its `--no-binary`_ option.
```
webknjaz, line 295: Since intersphinx is enabled for the pip docs site
(https://github.com/pypa/packaging.python.org/blob/main/source/conf.py#L378) we should attempt to use that instead of hardcoding the links.

The list of linkable objects can be found at https://webknjaz.github.io/intersphinx-untangled/pip.pypa.io/.

webknjaz: Looks like this would do
```suggestion
pip`s default behavior by e.g. using its :ref:`--no-binary
<pip:install_--no-binary>` option.
```

jugmac00: Thanks for the review and teaching me something new! I never heard of intersphinx before!

webknjaz (approved):

PR #868, opened 2021-03-30
tutorial: add note about files

### LINKED ISSUE #867: Add mention of manifest.in in packaging tutorial
Readers would be aided by a brief introduction to `manifest.in` for adding arbitrary files to the distribution; see https://github.com/pypa/packaging-problems/issues/292.

henryiii:
Closes #867. This provides some links for discovering more about MANIFEST.in and package data.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -376,14 +376,15 @@ MIT license:
 Including other files
 ---------------------
 
-The files listed above will be included automatically in your source
-distribution (SDist). If you want to control what goes in the SDist, see
-:ref:`Using MANIFEST.in`.
+The files listed above will be included automatically in your
+:term:`Source Distribution (or "sdist")`. If you want to control what goes in this
+explicitly, see :ref:`Using MANIFEST.in`.
 
-The final wheel will have the Python files in the discovered or listed Python
-packages. If you want to control what goes in the wheel, such as to add data
-files, see :std:doc:`Including Data Files <setuptools:userguide/datafiles>`
-from the :std:doc:`setuptools docs <setuptools:index>`.
+The final :term:`Built Distribution`  will have the Python files in the
+discovered or listed Python packages. If you want to control what goes here,
+such as to add data files, see :std:doc:`Including Data Files
+<setuptools:userguide/datafiles>` from the :std:doc:`setuptools docs
+<setuptools:index>`.
 
 .. _generating archives:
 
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -373,6 +373,18 @@ MIT license:
     SOFTWARE.
 
 
+Including other files
+---------------------
+
+The files listed above will be included automatically in your source
+distribution (SDist). If you want to control what goes in the SDist, see
```
pradyunsg, line 380: This is the first time in seeing sdist capitalised like this. Can we use the all lower case form here, and link source distribution to the glossary by using :term:`source distribution` here?

henryiii: Done, but I feel the discussion below is quite confusing about Source Distributions and Source Archives. They are clearly treated as separate, with a Source Archive coming "prior to creation of a Source Distribution", but the output of build --sdist is a "Source Archive" and everything below is called an archive, with "You should always upload a source archive and provide built archives for the platforms your project is compatible with.".

Is this making a false distinction between a distribution and an archive? All tools (build --sdist and python setup.py sdist) make a source distribution which is in an archive format, and the only non-distributable archive I know of is the one GitHub tends to make from tags, which is not something that needs discussion here.

source/tutorials/packaging-projects.rst
```diff
@@ -373,6 +373,19 @@ MIT license:
     SOFTWARE.
 
 
+Including other files
+---------------------
+
+The files listed above will be included automatically in your
+:term:`Source Distribution (or "sdist")`. If you want to control what goes in this
```
webknjaz, line 380: based on the previous commit, did you want to do something like this?
```suggestion
:term:`sdist <Source Distribution (or "sdist")>`. If you want to control what goes in this
```

source/tutorials/packaging-projects.rst
```diff
@@ -373,6 +373,19 @@ MIT license:
...
+explicitly, see :ref:`Using MANIFEST.in`.
+
+The final :term:`Built Distribution`  will have the Python files in the
```
webknjaz, line 383:
```suggestion
The final :term:`built distribution <Built Distribution>` will
have the Python files in the
```
maybe lowercase the anchor?

source/tutorials/packaging-projects.rst
henryiii: I think the part after this probably needs to be reworked or the glossary needs to be changed to clarify that Source Archive and Source Distribution are the same thing. But that can be a later pull request. I think technically for consistency for the way it is written now, it should be Source Archive.

Maybe this would be okay for the interim:

```suggestion
:term:`source distribution <Source Distribution (or "sdist")>`. If you want to control what goes in this
```

webknjaz: fair enough

webknjaz (approved):

PR #875, opened 2021-04-11
style: use file format in tab (newly supported)

henryiii:
Follow up to #818, see https://github.com/pypa/packaging.python.org/pull/818#issuecomment-817303874.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 sphinx==3.2.0
 sphinx-autobuild==0.7.1
-sphinx-inline-tabs==2021.3.19b5
+sphinx-inline-tabs==2021.4.11b9
 git+https://github.com/python/python-docs-theme.git#egg=python-docs-theme
 git+https://github.com/pypa/pypa-docs-theme.git#egg=pypa-docs-theme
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 sphinx==3.2.0
 sphinx-autobuild==0.7.1
-sphinx-inline-tabs==2021.3.19b5
+sphinx-inline-tabs==2021.4.11b9
```
webknjaz, line 3: Maybe try triple equals?
```suggestion
sphinx-inline-tabs === 2021.4.11b9
```

henryiii: The problem is that sphinx-inline-tabs requires Python 3.8+. Can that be changed on the readthedocs config in the GUI? I don't see a `.readthedocs.yml` to change it in locally.

henryiii: Actually, it looks like it can only be changed via a config file. The GUI choices are 2.x and 3.x.

henryiii:
Added a configuration file to bump to 3.8 does fix it, though @pradyunsg, I should note that readthedocs still defaults to Python 3.7, so upgrading a sphinx extension to no longer support 3.7 cuts out quite a lot of "default" users.

webknjaz (approved):

PR #877, opened 2021-04-12
Configure ReadTheDocs to use the dirhtml builder

### LINKED ISSUE #876: URLs with trailing slashes aren't working
A recent build of the Twine docs [failed with this error](https://github.com/pypa/twine/runs/2319520105?check_suite_focus=true#step:5:222):

```
/home/runner/work/twine/twine/docs/changelog.rst.rst:5:broken link: https://packaging.python.org/guides/distributing-packages-using-setuptools/#pre-release-versioning (404 Client Error: Not Found for url: https://packaging.python.org/guides/distributing-packages-using-setuptools/)
```

At the moment, that URL shows the RTD "This page does not exist yet" error.

The same thing is happening on internal links in the docs, such as "An Overview of Packaging for Python" in the table of contents. The `href` is referencing `overview.html`, but clicking on the link results in a 301 redirect to `/overview/`, which results in a 404.

pradyunsg:
Fixes #876

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,7 @@ version: 2
 
 sphinx:
    configuration: source/conf.py
+   builder: dirhtml
 
 python:
    version: 3.8
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -5,6 +5,8 @@ version: 2
 
 sphinx:
    configuration: source/conf.py
+   # The config file overrides the UI settings:
+   # https://github.com/pyca/cryptography/issues/5863#issuecomment-792343136
    builder: dirhtml
 
 python:


webknjaz:
Wow, it looks suspiciously like https://github.com/pyca/cryptography/issues/5863.

.readthedocs.yaml
```diff
@@ -5,6 +5,7 @@ version: 2
 
 sphinx:
    configuration: source/conf.py
+   builder: dirhtml
```
webknjaz, line 8:
```suggestion
   # The config file overrides the UI settings:
   # https://github.com/pyca/cryptography/issues/5863#issuecomment-792343136
   builder: dirhtml
```

webknjaz (approved):

.readthedocs.yaml
pradyunsg: Off topic: I think this comment linking to the cryptography issue was unnecessary here.

henryiii: "Hadn't", not "Hasn't"? It's https://github.com/readthedocs/readthedocs.org/issues/8092 now?

webknjaz: Yes

henryiii: Honestly, while I could update this, I think I'm with @pradyunsg and it could just be removed. This is not going to change (v2 replaces, unlike v1 which overrides), there just might be a RtD update to be clearer in the webUI. Also, contributors can't even see the RtD UI config settings, so this the only public configuration anyway.

webknjaz: The motivation behind such comments normally is that when somebody in the future will decide to refactor this file and remove the setting seeing that it's the same in the UI (provided they have access), this link should stop them from proceeding.

PR #879, opened 2021-04-12
chore: update to using importlib.metadata for entry_points

### LINKED ISSUE #870: Mention alternatives to discover plugins via entry_points
In "Creating and discovering plugins" guide, the recommendation is to use `pkg_resources.iter_entry_points`. In https://github.com/takluyver/entrypoints it is noted that:

> The pkg_resources module distributed with setuptools provides a way to discover entrypoints as well, but it contains other functionality unrelated to entrypoint discovery, and it does a lot of work at import time. Merely importing pkg_resources causes it to scan the files of all installed packages. Thus, in environments where a large number of packages are installed, importing pkg_resources can be very slow (several seconds).

I agree with this observation too. Now with the standard library `importlib.metadata` feature landing in Python 3.10+ and [its backport](https://github.com/jaraco/backports.entry_points_selectable), this recommendation can be updated with more lighter:

```python
import sys
if sys.version_info[:2] < (3, 10):
    from backports.entry_points_selectable import entry_points
else:
    from importlib.metadata import entry_points

...
entry_points(group='myapp.plugins')
```

henryiii:
Closes #870. I think this was the outcome of the discussion?

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -393,8 +393,8 @@ nitpicky = True
 nitpick_ignore = [
     ('envvar', 'PATH'),
     ('py:func', 'find_packages'),
-    ('py:func', 'pkg_resources.iter_entry_points'),
     ('py:func', 'setup'),
+    ('py:func', 'importlib.metadata.entry_points'),  # remove when 3.10 is released
     ('py:func', 'setuptools.find_namespace_packages'),
     ('py:func', 'setuptools.find_packages'),
     ('py:func', 'setuptools.setup'),
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -136,17 +136,19 @@ in its :file:`setup.py`:
     )
 
 Then you can discover and load all of the registered entry points by using
-:func:`pkg_resources.iter_entry_points`:
+:func:`importlib.metadata.entry_points` (or the `backport`_
+``importlib_metadata >= 3.6`` for Python 3.6-3.9):
 
 .. code-block:: python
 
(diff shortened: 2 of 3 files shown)


source/conf.py
```diff
@@ -393,8 +393,8 @@
 nitpick_ignore = [
     ('envvar', 'PATH'),
     ('py:func', 'find_packages'),
-    ('py:func', 'pkg_resources.iter_entry_points'),
     ('py:func', 'setup'),
+    ('py:func', 'importlib.metadata.entry_points'),
```
webknjaz, line 397: @henryiii is it possible to add the corresponding `importlib_metadata` docs site to intersphinx? Or is this supposed to link stdlib?

henryiii: It was supposed to be stdlib, but it's stdlib in Python 3.10, so it's not showing up here yet. Maybe I could link to `importlib_metadata` when I mention the alternative, though.

henryiii: And looks like the docs for `importlib_metadata` don't define functions.

henryiii: I've added a comment to remove when 3.10 is out (such that `/3` points at 3.10), and added a direct link to the docs for `importlib_metadata`.

webknjaz: Yeah, I noticed: https://webknjaz.github.io/intersphinx-untangled/docs.python.org/.

FWIW 3.10 doesn't seem to document this either.

P.S. There's nothing stopping us from adding 3.10 to intersphinx as a separate entry already.

webknjaz (approved):

webknjaz (approved):

PR #884, opened 2021-04-18
This commit adds the manimce deprecate documentation in RST format

dayvheed:


source/mamince_release_documentation.rst
```diff
@@ -0,0 +1,29 @@
+**Package Release**
+===================
+
+This is the reference guide for manimce. The package post manimce
+0.1.1.post0 was posted today and inform you that the package will be
+deprecated going forward.
```
naveen521kk, line 6: Aren't you writing a general guide about "how to deprecate a package"?

dayvheed: From a developer standpoint? I will love to. Please can you give me a little insight of what you would be expecting in the article

webknjaz (changes requested):
Thanks for this effort but it looks like you've misunderstood the task.

The project https://packaging.python.org aims to provide generic guides for the package maintainers on how to do things related to this topic. We don't need to publish user-facing instructions for every other project on the internet.

To specify the task clearer: we need a guide answering the question "Given that I'm a project maintainer, what steps should I take in order to deprecate it?"

dayvheed:
> Thanks for this effort but it looks like you've misunderstood the task.
> 
> The project https://packaging.python.org aims to provide generic guides for the package maintainers on how to do things related to this topic. We don't need to publish user-facing instructions for every other project on the internet.
> 
> To specify the task clearer: we need a guide answering the question "Given that I'm a project maintainer, what steps should I take in order to deprecate it?"

Alright. Thanks for the clarification. I will edit and come up another write up

webknjaz:
I think I'll close this for now. Feel free to open up a new pull request with the updated guide.

PR #885, opened 2021-04-20
Add note for using extras with console_scripts

chrisburr:
As a result of the discussion in https://github.com/pypa/pip/issues/9726 I think it would be useful to clarify that the use of `extras` with `console_scripts` has no effect beyond marking the entrypoint with some additional metadata. Any thoughts?


source/specifications/entry-points.rst
```diff
@@ -120,6 +120,10 @@ For example::
     [pytest11]
     nbval = nbval.plugin
 
+.. note::
+   Using ``extras`` with ``console_scripts`` has no has no semantic meaning
+   however some tools may choose to use this metadata at runtime.
+
```
pradyunsg, line 126:
```suggestion
... note::

    There are some non-standard extensions to this format, such as setuptool's
    `[extra, names]` syntax (which adds setuptools-specific checks when the
    scripts are generated by setuptools). These extensions are generally not
    supported by other tools such as pip.
```

pradyunsg (changes requested):

source/specifications/entry-points.rst
chrisburr: That's definitely much clearer, thanks 👍

pradyunsg:
This specific thing is actually already covered in this document!

> Using extras for an entry point is no longer recommended. Consumers should
> support parsing them from existing distributions, but may then ignore them.
> New publishing tools need not support specifying extras. The functionality of
> handling extras was tied to setuptools' model of managing 'egg' packages, but
> newer tools such as pip and virtualenv use a different model.

chrisburr:
How did I read this so many times and miss that?! 😅

I still think there is value in having this clarification alongside the example, or maybe provide an example of `extras` with a made up `group` instead of associating it with `console_scripts`.

Budda0ne (approved):

pradyunsg:
> I still think there is value in having this clarification alongside the example,

I think it'd be better to change that paragraph into a `.. important` admonition, and to have a link back to that *after* this example.

You can add a reference line just before the admonition, and reference it from the text around the example -- especially the parts that talk about now to handle extras.

```
.. _entry-points-deprecated-extras:
```

webknjaz:
@chrisburr this PR has requested changes that you haven't addressed. Are you planning to? If not, we could just close it.

webknjaz:
@chrisburr thanks for your contribution but I'm unable to accept this PR at this time because it has unresolved review requests. I'm closing this PR for now but feel free to re-open / send another one once the necessary changes have been made.

PR #888, opened 2021-05-06
Fix description of entry points mechanism

ashwinvis:
- Show how a module would be loaded. A dictionary of modules will not be loaded, unless created by the user. The example code was corrected.
- Update links to external documentation
- Hint that specifying entry points is not limited to setuptools. For  example [flit](https://flit.readthedocs.io/en/latest/pyproject_toml.html#entry-points-sections)  and [poetry](https://python-poetry.org/docs/pyproject/#plugins) support it too.

Following up on #870 and #879

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -395,7 +395,7 @@ nitpick_ignore = [
     ('py:func', 'find_packages'),
     ('py:func', 'setup'),
     ('py:func', 'importlib.metadata.entry_points'),  # remove when 3.10 is released
-    ('py:class', 'importlib.metadata.EntryPoints'),  # remove when 3.10 is released
+    ('py:class', 'importlib.metadata.EntryPoint'),  # remove when 3.10 is released
     ('py:func', 'setuptools.find_namespace_packages'),
     ('py:func', 'setuptools.find_packages'),
     ('py:func', 'setuptools.setup'),
--- a/source/conf.py
+++ b/source/conf.py
@@ -395,6 +395,7 @@ nitpick_ignore = [
     ('py:func', 'find_packages'),
     ('py:func', 'setup'),
     ('py:func', 'importlib.metadata.entry_points'),  # remove when 3.10 is released
+    ('py:class', 'importlib.metadata.EntryPoints'),  # remove when 3.10 is released
     ('py:func', 'setuptools.find_namespace_packages'),
     ('py:func', 'setuptools.find_packages'),
     ('py:func', 'setuptools.setup'),
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -120,10 +120,10 @@ a list of packages to :func:`setup`'s ``packages`` argument instead of using
 Using package metadata
 ======================
 
-`Setuptools`_ provides :setuptools:doc:`special support
-<userguide/entry_point>` for plugins. By providing the ``entry_points``
-argument to :func:`setup` in :file:`setup.py` plugins can register themselves
-for discovery.
+`Setuptools`_ provides :doc:`special support
+<setuptools:userguide/entry_point>` for plugins. By providing the
+``entry_points`` argument to :func:`setup` in :file:`setup.py` plugins can
+register themselves for discovery.
 
 For example if you have a package named ``myapp-plugin-a`` and it includes
 in its :file:`setup.py`:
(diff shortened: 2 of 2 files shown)


source/guides/creating-and-discovering-plugins.rst
```diff
@@ -150,21 +150,30 @@ Then you can discover and load all of the registered entry points by using
...
+    )
+
+Now the module of your choice can be imported by executing
+``discovered_plugins['a'].load()``.
 
 .. note:: The ``entry_point`` specification in :file:`setup.py` is fairly
     flexible and has a lot of options. It's recommended to read over the entire
     section on `entry points`_.
 
+.. note:: Since this specification is part of the `standard library`_, most
```
webknjaz, line 169: Could you use intersphinx here? If the ref is not linkable, maybe add the py3 dev docs site to the mapping in the `conf.py`

source/guides/creating-and-discovering-plugins.rst
```diff
@@ -150,21 +150,30 @@ Then you can discover and load all of the registered entry points by using
...
+   packaging tools other than setuptools provide support for defining entry
+   points.
+
 .. _Setuptools: https://setuptools.readthedocs.io
 .. _special support:
 .. _entry points:
-    https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins
+   https://setuptools.readthedocs.io/en/stable/userguide/entry_point.html
```
webknjaz, line 176: Let's use intersphinx to have better linking

ashwinvis: Tried to do it in 3242424 using your tool https://webknjaz.github.io/intersphinx-untangled/. Can you check if it is right?

webknjaz: The intersphinx alias should be inside the angle brackets, it's not a part of the role.

webknjaz: Ah, I see you fixed that

source/guides/creating-and-discovering-plugins.rst
```diff
@@ -150,21 +151,26 @@ Then you can discover and load all of the registered entry points by using
     discovered_plugins = entry_points(group='myapp.plugins')
 
 
-In this example, ``discovered_plugins`` would be:
+In this example, ``discovered_plugins`` would be a collection of type :class:`importlib.metadata.EntryPoints`:
```
webknjaz, line 154: Should it be singular? 
```suggestion
In this example, ``discovered_plugins`` would be a collection of type :class:`importlib.metadata.EntryPoint`:
```

source/conf.py
```diff
@@ -395,6 +395,7 @@
     ('py:func', 'find_packages'),
     ('py:func', 'setup'),
     ('py:func', 'importlib.metadata.entry_points'),  # remove when 3.10 is released
+    ('py:class', 'importlib.metadata.EntryPoints'),  # remove when 3.10 is released
```
webknjaz, line 398:
```suggestion
    ('py:class', 'importlib.metadata.EntryPoint'),  # remove when 3.10 is released
```
```suggestion
    ('py:class', 'importlib.metadata.EntryPoints'),  # remove when 3.10 is released
```

source/guides/creating-and-discovering-plugins.rst
ashwinvis: No it is correct. See https://docs.python.org/3.10/library/importlib.metadata.html#entry-points

> The entry_points() function returns an EntryPoints object, a sequence of all EntryPoint objects with names and groups attributes for convenience:

It is the new convenient API landing in Python 3.10 and backported via the `importlib_metadata` package.

webknjaz (approved):

source/guides/creating-and-discovering-plugins.rst
ashwinvis: Maybe I can expand the text it is less confusing.

ashwinvis:
Oh no.... you merged it too soon @webknjaz :sweat_smile: !

I wanted to add this change:

```diff
From d16106c05e1e10f2821da5bdead7dbd5c691aba6 Mon Sep 17 00:00:00 2001
From: Ashwin Vishnu <y4d71nsar@relay.firefox.com>
Date: Thu, 6 May 2021 18:56:29 +0200
Subject: [PATCH] Clarify EntryPoints vs EntryPoint

---
 source/guides/creating-and-discovering-plugins.rst | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/source/guides/creating-and-discovering-plugins.rst b/source/guides/creating-and-discovering-plugins.rst
index 8fe1e41..ebdf876 100644
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -151,7 +151,9 @@ Then you can discover and load all of the registered entry points by using
     discovered_plugins = entry_points(group='myapp.plugins')
 
 
-In this example, ``discovered_plugins`` would be a collection of type :class:`importlib.metadata.EntryPoints`:
+In this example, ``discovered_plugins`` would be an
+:class:`importlib.metadata.EntryPoints` object, which is a sequence of all
+``EntryPoint`` objects belonging to the selected group.
 
 .. code-block:: python
 
-- 
2.31.1


```

webknjaz:
@ashwinvis oh, you can send that in a follow-up PR. Also, if you want to mark a PR as not ready for review, just make it a Draft PR (normally you should be able to switch between draft and normal PR state at any point), or at least mark it as a WIP in the title — this will be a signal for the maintainers to postpone reviewing/merging it.

Issue #866, opened 2021-03-23
Clarify "Packaging Python Projects" tutorial


bhrutledge:
First, **thank you** for the work to update the tutorial to use `pyproject.toml`, `setup.cfg`, and `build`! Looks like that was mostly related to #809.

I've been noticing what seems like an uptick in issues on https://github.com/pypa/packaging-problems/issues/, particularly related to `python3 -m build`. That prompted me to work through the tutorial from scratch, doing my best to keep a beginner's perspective. The outcome is a bunch of thoughts on how it could be clearer, which can be viewed as [annotations to the live site](https://hyp.is/go?url=http%3A%2F%2Fpackaging.python.org%2Ftutorials%2Fpackaging-projects.html&group=__world__) (via the [Hypothesis web annotation tool](https://web.hypothes.is/)).

Besides a bunch of small copy-edits, there are some bigger questions about the content, e.g.

- How much detail is necessary for the intended audience?
- When/how should we explain "editable installs"?
- Should this whole thing be done in a virtual environment?

Before diving into submitting a PR for any/all of these things, I'd love some feedback on my feedback. 😉

di:
> I think that tutorials shouldn't be too low-level. They need to show one way, without any branching

+1 to this.

@bhrutledge I'm certain your feedback is valuable, please feel free to go ahead and create a PR so we can review directly.

bhrutledge:
First batch submitted. I'll wait until those are merged before working on further changes.

henryiii:
FYI, some of the extra issues have been related to a few bugs in build, several of which are being fixed in the version about to be rolled out. I think the issues will slow down a little after that.

So far the changes look great. :)

I'm wondering if we could update to using pipx in the future? Using `pipx run build`, `pipx run twine`, etc. would help with the lack of virtual environments, and would remove some of the "set the next part up by installing this" steps.

bhrutledge:
> I'm wondering if we could update to using pipx in the future? Using `pipx run build`, `pipx run twine`, etc. would help with the lack of virtual environments, and would remove some of the "set the next part up by installing this" steps.

While I love pipx, I'm wary of introducing an additional tool, especially one that's not a PyPA project. I'd rather encourage the use of a virtual environment, which folks should already be familiar with from the [Installing Packages tutorial](https://packaging.python.org/tutorials/installing-packages/).

bhrutledge:
> So far the changes look great. :)

[...]

PR #894, opened 2021-05-17
Packaging Python Projects: Restructure 'A simple project'

bhrutledge:
Towards #866. I shuffled some sentences around and removed some words to make this introduction flow a little better.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -9,14 +9,11 @@ to build the package, and how to upload it to the Python Package Index.
 A simple project
 ----------------
 
-This tutorial uses a simple project named ``example_pkg``. If you are unfamiliar
-with Python's modules and :term:`import packages <Import Package>`, take a few
-minutes to read over the `Python documentation for packages and modules`_. Even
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -56,8 +54,8 @@ project's root directory - you will add content to them in the following steps.
     └── tests/
 
 
-Creating a test folder
-----------------------
+Creating a test directory
+-------------------------
```
pradyunsg, line 58: I prefer the original heading, since "Files and folders" is significantly more common phrasing than "Files and directories" IMO.

pradyunsg (approved):
LGTM, other than one non-blocking comment.

source/tutorials/packaging-projects.rst
bhrutledge: Appreciated, though "directory" was used more often in this document. After searching around a bit, I think it's also the more accurate term. For example: https://en.wikipedia.org/wiki/Directory_%28computing%29#Folder_metaphor

webknjaz (approved):

PR #896, opened 2021-05-21
Add cibuildwheel to project list

henryiii:
@pypa/cibuildwheel-maintainers My description here is rather uninspired, feel free to expand and improve.

PR #897, opened 2021-05-21
docs: update suggestion for build and cibuildwheel

henryiii:
Builds on #896, will have an undefined label until that one goes in.

### DIFF
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -46,12 +46,14 @@ Installation tool recommendations
 Packaging tool recommendations
 ==============================
 
-* Use :ref:`setuptools` to define projects and create :term:`Source Distributions
-  <Source Distribution (or "sdist")>`. [5]_ [6]_
+* Use :ref:`setuptools` to define projects. [5]_ [6]_
 
-* Use the ``bdist_wheel`` :ref:`setuptools` extension available from the
-  :ref:`wheel project <wheel>` to create :term:`wheels <Wheel>`.  This is
-  especially beneficial, if your project contains binary extensions.
+* Use the :ref:`build` package to create  :term:`Source Distributions
+  <Source Distribution (or "sdist")>` and, if your project is pure
+  Python, also :term:`wheels <Wheel>`.
+  
+* Use the :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
+  for all the platforms that you support if your project contains binary extensions.
 
 * Use `twine <https://pypi.org/project/twine>`_ for uploading distributions
   to :term:`PyPI <Python Package Index (PyPI)>`.
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -48,12 +48,14 @@ Packaging tool recommendations
 
 * Use :ref:`setuptools` to define projects. [5]_ [6]_
 
-* Use the :ref:`build` package to create  :term:`Source Distributions
-  <Source Distribution (or "sdist")>` and, if your project is pure
-  Python, also :term:`wheels <Wheel>`.
+* Use :ref:`build` to create :term:`Source Distributions
+  <Source Distribution (or "sdist")>` and :term:`wheels <Wheel>`.
   
-* Use the :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
+* Use :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
   for all the platforms that you support if your project contains binary extensions.
+  This automates the steps required to produce highly redistributable wheels with
+  sensible defaults, combining several other tools, such as build, manylinux,
+  auditwheel, and delocate.
 
 * Use `twine <https://pypi.org/project/twine>`_ for uploading distributions
   to :term:`PyPI <Python Package Index (PyPI)>`.
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -50,12 +50,10 @@ Packaging tool recommendations
 
 * Use :ref:`build` to create :term:`Source Distributions
   <Source Distribution (or "sdist")>` and :term:`wheels <Wheel>`.
-  
-* Use :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
-  for all the platforms that you support if your project contains binary extensions.
-  This automates the steps required to produce highly redistributable wheels with
-  sensible defaults, combining several other tools, such as build, manylinux,
-  auditwheel, and delocate.
+
+If you have binary extensions and want to distribute wheels for multiple
+platforms, use :ref:`cibuildwheel` as part of your CI setup to build
+distributable wheels.
 
 * Use `twine <https://pypi.org/project/twine>`_ for uploading distributions
   to :term:`PyPI <Python Package Index (PyPI)>`.
--- a/source/conf.py
+++ b/source/conf.py
@@ -376,8 +376,8 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'python2': ('https://docs.python.org/2', None),
     'pip': ('https://pip.pypa.io/en/latest/', None),
-    'setuptools': ('https://setuptools.rtfd.io/en/latest/', None),
-    'twine': ('https://twine.rtfd.io/en/stable/', None),
+    'setuptools': ('https://setuptools.readthedocs.io/en/latest/', None),
+    'twine': ('https://twine.readthedocs.io/en/stable/', None),
 }
 
 


bhrutledge (approved):
This makes sense to me, and seems aligned with the packaging tutorial.

pradyunsg (dismissed):

source/guides/tool-recommendations.rst
```diff
@@ -46,12 +46,14 @@ Installation tool recommendations
...
-  <Source Distribution (or "sdist")>`. [5]_ [6]_
-
-* Use the ``bdist_wheel`` :ref:`setuptools` extension available from the
-  :ref:`wheel project <wheel>` to create :term:`wheels <Wheel>`.  This is
-  especially beneficial, if your project contains binary extensions.
+* Use :ref:`setuptools` to define projects. [5]_ [6]_
+
+* Use the :ref:`build` package to create  :term:`Source Distributions
+  <Source Distribution (or "sdist")>` and, if your project is pure
+  Python, also :term:`wheels <Wheel>`.
```
webknjaz, line 53: I don't like that this implies the build's inability to produce platform-specific wheels. This is just untrue.

source/guides/tool-recommendations.rst
```diff
@@ -46,12 +46,14 @@ Installation tool recommendations
...
+  
+* Use the :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
```
webknjaz, line 55: Personally, I'd still emphasize that build should be used. Just because it's a higher-level opinionated thing that limits flexibility.

source/guides/tool-recommendations.rst
henryiii:
```suggestion
  <Source Distribution (or "sdist")>` and :term:`wheels <Wheel>`.
```

Maybe this is enough? It should be "for any platform if your project is pure-python, or for your current platform if the project contains binary extensions" or something like that, but that's a little more detail than needed here, I think? Plus I don't want to give readers the impression that build automatically creates redistributable wheels - that requires special setup or more steps to do properly. (audithweel, docker containers, python.org CPython on macOS, etc).

webknjaz:
I've requested additional reviews from the folks involved with build because I believe they deserve to define how their project is presented in such an authoritative guide.

source/guides/tool-recommendations.rst
webknjaz: Maybe. I just want to make sure that the guides we publish as a group are fair wrt the projects represented in them. So let's wait for some comments by build maintainers to see what they think because I feel weird trying to decide for them. If nobody comments, I'll merge. But yes, the simplified patch suggested above looks okay to me.

source/guides/tool-recommendations.rst
```diff
@@ -46,12 +46,14 @@ Installation tool recommendations
...
+  for all the platforms that you support if your project contains binary extensions.
```
webknjaz, line 56: Maybe add something like this here?
```suggestion
  for all the platforms that you support if your project contains binary extensions.

  .. note::

    Bear in mind that this tool will automate
    performing certain steps you'd have to
    handle yourself otherwise.
```

henryiii:
Build does not do the right thing for making redistributable wheels. There are a bunch of other caveats. I'm hoping to slowly improve the guide to include those, but for now, most users who are trying to do this themselves are not doing it right. For example, macOS wheels do not support a full range of macOS versions, but instead depend on what's available in CI. build's target is not to eventually download docker containers on Linux, python.org Python copies on macOS, run auditwheel on Linux, etc, but instead to be used inside of a system like cibuildwheel or multibuild if binary redistributable wheels are needed. (Of course it can and should build binary wheels, it is already an option in the cibuildhwheel 2.0 alphas)

I'd be happy to include multibuild here, by the way, it wasn't part of the guide originally.

source/guides/tool-recommendations.rst
henryiii: I don't see why this is a "bear in mind" parenthetical comment. There are lots of (still undocumented) caveats to producing a wheel that works on someone else's machine and cibuildwheel has sensible defaults for those things.

`cibuildwheel` is not supposed to be opinionated, with the single exception that it requires a package support `pip wheel` or (2.0+) `python -m build --wheel`. Every other step is replaceable (repair, etc), it just has a sensible default. It handles the download and installation of Python on macOS, the use of manylinux images on Linux (any image can be used), applying a repair command (but that can be changed or turned off), etc. It also often ships patch fixes for PyPy wheel building before the PyPy team can release them.  It's used in drastically different packages of all sizes.

henryiii:
```suggestion
  for all the platforms that you support if your project contains binary extensions.
  This automates the steps required to produce highly redistributable wheels with
  sensible defaults, combining several other tools, such as build, manylinux,
  auditwheel, and delocate.
```

henryiii: What about that?

henryiii: PS: this is also a "Packaging tool recommendations" list, so I think we should be whole heartedly recommending tools for making redistributable wheels here. "I want to publish a binary on PyPY" should always have the answer "use cibuildwheel/multibuild"; if a user is advanced enough to not need the help of a tool like that, they don't need this guide. And if they try themselves, I can be sure they will get something wrong, which won't be caught until someone else tries to load their package. Again, hoping to improve the guide over time with details of what is required to make a package redistributable, but we don't even have the information to do it yourself available currently.

The one exception is ironically Windows, where the "default" thing pretty much works.

source/guides/tool-recommendations.rst
layday: The suggested change seems fine.  I agree that we shouldn't be implying that build is only good for pure Python wheels.

source/guides/tool-recommendations.rst
```diff
@@ -46,12 +46,14 @@ Installation tool recommendations
...
 
-* Use :ref:`setuptools` to define projects and create :term:`Source Distributions
-  <Source Distribution (or "sdist")>`. [5]_ [6]_
-
-* Use the ``bdist_wheel`` :ref:`setuptools` extension available from the
-  :ref:`wheel project <wheel>` to create :term:`wheels <Wheel>`.  This is
-  especially beneficial, if your project contains binary extensions.
+* Use :ref:`setuptools` to define projects. [5]_ [6]_
+
+* Use the :ref:`build` package to create  :term:`Source Distributions
```
layday, line 51:
```suggestion
* Use :ref:`build` to create :term:`Source Distributions
```

source/guides/tool-recommendations.rst
layday, line 55:
```suggestion
* Use :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
```

source/guides/tool-recommendations.rst
henryiii: +1 this seems to be more inline with how `setuptools` is used above.

source/guides/tool-recommendations.rst
webknjaz: The concern I have here is that this tool is high-level while others are low-level. It also mutates the outer env/OS (like installing "the right" Pythons on macOS). I don't have anything against that but we should warn the users that it may influence their machines and should probably only be used in CI. They still need to just use build if they want to test something on their machine and don't care about the compatibility with others at that point.
By including it in the same list, we sort of imply that it is of about the same level while it isn't. Do you see my point here?

P.S. As for "bear in mind", that's more of an example. I'd be okay with that reworded to align with what I've said here.

source/guides/tool-recommendations.rst
```diff
@@ -46,12 +46,16 @@ Installation tool recommendations
...
+* Use :ref:`setuptools` to define projects. [5]_ [6]_
+
+* Use :ref:`build` to create :term:`Source Distributions
+  <Source Distribution (or "sdist")>` and :term:`wheels <Wheel>`.
+  
+* Use :ref:`cibuildwheel` to create distributable :term:`wheels <Wheel>`
+  for all the platforms that you support if your project contains binary extensions.
+  This automates the steps required to produce highly redistributable wheels with
+  sensible defaults, combining several other tools, such as build, manylinux,
+  auditwheel, and delocate.
```
webknjaz, line 58: How about this?
```suggestion
  auditwheel, and delocate.

  .. tip::

     This tool may download container images and
     install Python distributions system-wide
     depending on your OS environment. This is
     why you may prefer to use it in CI rather
     than locally.
```

webknjaz:
@henryiii I think it's okay for you to be opinionated, I just want to make sure that we avoid giving false impressions to the users. Hence all those clarifications. At this point, I mostly want some note along the lines of "hey, this is a magic pill, beware that it may mutate your broader system outside the project dir, maybe only use it in CI".

source/guides/tool-recommendations.rst
pradyunsg: >  By including it in the same list, we sort of imply that it is of about the same level while it isn't. Do you see my point here?

I do, however, as a counter point, this isn't "low level tool recommendations" -- `cibuildwheel` makes sense if you have a project that has compiled assets and you need to provide distributable wheels.

source/guides/tool-recommendations.rst
layday: There is no "prefer" here, `cibuildwheel` should not be used on a local machine.

webknjaz: Agreed.

webknjaz: I've edited the suggestion. Is it better now?

source/guides/tool-recommendations.rst
webknjaz: It's not about low vs high level. It's more about the same level vs a completely different level abstraction as I see it.

source/guides/tool-recommendations.rst
layday: Looks good to me, although I'm wondering if cibuildwheel isn't taking up too much visual space now - if reader attention isn't immediately drawn to the tool they are least likely to need.

webknjaz: Yeah, I'm not sure how to keep the healthy balance across all the points here.

layday: I don't think the last sentence ("This automates the steps ...") is necessary - we don't go into that much detail for setuptools or build.  Or perhaps we should be going into more detail for them.

pfmoore: If `cibuildwheel` shouldn't be run locally, it's a very different sort of tool than `build` or `pip`. Maybe we need to be clear about that? Something like the following (sorry, I can't get Github to do a multi-line "suggestion" for this)

> * Use `setuptools` to define projects.
> * Use `build` to create source distributions and wheels
>
> If you want to distribute wheels for multiple platforms, and in particular non-windows platforms, use `cibuildwheel`  as part of your CI setup to build distributable wheels.

We don't have to explain all the nuances here - surely it's implicit that if you want to get into details, you should read the tool documentation?

webknjaz: That sounds reasonable.

henryiii: Aside: It is extremely useful to run cibuildwheel locally, which is perfectly valid as long as you are building linux wheels (from any OS that has Docker installed). For cibuildwheel 2 (upcoming), if you use pyproject.toml to configure it, it's really easy just to call `pipx run cibuildwheel --platform linux` and get your manylinux wheels without even setting any configuration variables.  I built `cmake`'s linux wheels locally when CI failed to upload a wheel (before it even used cibuildwheel in CI, actually), and @mayeut build the PowerPC wheels locally when scikit-build's Travis credits ran out. Its goal is to make redistributable wheels and not be tied to a specific CI system. For the moment, if you are building Windows or macOS wheels, then you shouldn't be running it locally because it installs Python to global locations, but not true for Linux.

Anyway, I think that suggestion sounds reasonable; I'll try to rewrite it in a little bit to look more like that.

henryiii: Here's what I put:

```
If you have binary extensions and want to distribute wheels for multiple
platforms, use :ref:`cibuildwheel` as part of your CI setup to build
distributable wheels.
```

I wanted to include a mention of binary extensions, as there is no reason to use cibuidwheel if you don't have to "build" something (we've actually had users with pure Python projects try to use it), and I don't think the "non-Windows" part was very important and makes it sound like it doesn't support Windows.

How does this sound?

henryiii:
Curious:

```
intersphinx inventory 'https://twine.rtfd.io/en/stable/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectionError'>: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
intersphinx inventory has moved: https://setuptools.rtfd.io/en/latest/objects.inv -> https://setuptools.readthedocs.io/en/latest/objects.inv
```

I think this is just a network issue, but we can move this to using the full name to avoid the warning.

source/guides/tool-recommendations.rst
webknjaz: Perfect!

webknjaz (approved):

Issue #887, opened 2021-04-27
Contradictions/Inconsistencies between packaging tutorial and guide


asabbagh:
Page 1: [Packaging Python Projects](https://packaging.python.org/tutorials/packaging-projects)
Page 2: [Packaging and distributing projects](https://packaging.python.org/guides/distributing-packages-using-setuptools/)

The first states 

> "setup.py  # optional, needed to make editable pip installs work"

The second states 

> "The most important file is setup.py which exists at the root of your project directory."

The first mentions pyproject.toml, whereas the second doesn't mention it at all.

I know that one's a "tutorial" and the other is a "guide", and the website states that guides have more detailed information. In this case it appears there's some detail in the tutorial that isn't in the guide.

webknjaz:
Agreed. That sounds reasonable.

henryiii:
"needed to make editable pip installs work" is actually not true anymore since the most recent pip release. Unless you are on an old Python unsupported by the most recent pip release. :)

webknjaz:
@henryiii do you have an issue or a PR link with the change? I don't recall seeing that.

henryiii:
Either https://pip.pypa.io/en/stable/news/#v21-1 (under features, can't really link to the header because they are auto-generated, so identical for all versions) or https://github.com/pypa/pip/pull/9547.

pradyunsg:
Note that this still doesn't cover non-setuptools build backends.

henryiii:
Well, the comment about setup.py being required is also for the setuptools backend. But yes, it's not editable PEP 517 support.

bhrutledge:
This seems related to #866.

bhrutledge:
Also, rather than update https://packaging.python.org/guides/distributing-packages-using-setuptools/, could it be replaced with links to the [setuptools docs](https://setuptools.readthedocs.io/en/latest/userguide/index.html)? They've seen some dramatic improvements.

bhrutledge:
This feels related to https://github.com/pypa/packaging.python.org/issues/792.

hugobuddel:
> "needed to make editable pip installs work" is actually not true anymore since the most recent pip release. Unless you are on an old Python unsupported by the most recent pip release. :)

[...]

Issue #874, opened 2021-04-11
packaging.python.org Windows commands


wackyblackie:
In the Python Packaging User Guide the commands for Windows are ```py```. When I run them like that (with ```py```), I get:
```cmd
C:\>py --version
'py' is not recognized as an internal or external command,
operable program or batch file.
```
However, when I use the "Unix/macOS" commands, it works just fine:
```cmd
C:\>python3 --version
Python 3.8.8
```

Maybe consider switching the Windows commands to ```python3``` or take away the Windows-specific commands.

Best regards,
@wackyblackie

pfmoore:
That looks like you have a version of Python that didn't come from the python.org installers. Where did you install Python from?

wackyblackie:
I installed Python from the Microsoft Store.

pradyunsg:
Cross linking to #863.

zooba:
All the context is there - the Store package installs `python`/`python3`/`python3.x` commands, but not install `py` (because it couldn't override the python.org `py` and would just create more conflicts).

I posted on the other linked issue with my proposal, but to repeat it, "if `py` works for you, use that anywhere you see `python3`".

pradyunsg:
Closing this since #863 covered this.

Issue #863, opened 2021-03-22
Encourage `py.exe` on Windows


naveen521kk:
In windows, there is a launcher for python called `py.exe` which should be part of the default installation. What I am suggesting is using tabs to separate for Windows and `Unix/macOS`. Something like
```rst
.. tab:: Windows

    .. code-block:: cmd

        py -m pip install --user --upgrade twine

.. tab:: Unix/macOS

    .. code-block:: bash

        python3 -m pip install --user --upgrade twine
```
will render something like 

![image](https://user-images.githubusercontent.com/49693820/111990404-414e9d80-8b39-11eb-9327-f01e06e3721e.png)
![image](https://user-images.githubusercontent.com/49693820/111990481-4e6b8c80-8b39-11eb-96a0-30fcc7908b56.png)

If this is ok, I would like to help with this.

webknjaz:
Do you intend to introduce this across all the articles or do you have something specific in mind? I think it's fine to add this (the *NIX tab should probably go first, though, as it's more common in the Python dev world).

naveen521kk:
> Do you intend to introduce this across all the articles or do you have something specific in mind?

I would simply add this everywhere because I could see in many places `python3` is specified and Windows python doesn't have a shim to python3, and typing `python3` would launch Windows store to install python.

> (the *NIX tab should probably go first, though, as it's more common in the Python dev world).

:+1:

pradyunsg:
Looking at the PR (and pip's documentation), I think I'd much rather have a block of text explaining why you should use `py -m pip` and `python -m pip` outside of virtual environments, and just use `pip` in these examples, with the assumption that the user is in a virtual environment? Using such environments is best practice anyway.

That isn't to say that the PR on its own isn't useful and an improvement over status quo, but this is me initiating a slightly broader issue at play here

pradyunsg:
More characters and additional work for CoNsIsTeNcY in invocations, in virtual envs and in the global environments.

naveen521kk:
> It's basically more characters and additional work for CoNsIsTeNcY in invocations, in virtual envs and in the global environments.

I think having `py -m pip` should be better because we don't whether they are in virtual environments or not. Also, on windows, if `Add to Path` isn't checked `python.exe` or its scripts will not be added to the path, so `py.exe` is the best to use IMO.

pfmoore:
[...]

PR #898, opened 2021-05-21
Use a PEP517 build front-end in the setuptools guide

henryiii:
This updates the build prodedure for modern best practices, using `pypa/build`. (Delayed) fix for an issue somewhere where it was pointed out this page was old, but forget where it was. Will dig around a bit for it tomorrow.

The page still needs the updates requested in #887; this page seems quite old.

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -832,7 +832,7 @@ To build the wheel:
 
         py -m build --wheel
 
-The ``wheel`` package ill detect that the code is pure Python, and build a wheel
+The ``wheel`` package will detect that the code is pure Python, and build a wheel
 that's named such that it's usable on any Python installation with the same
 major version (Python 2 or Python 3) as the version you used to build the
 wheel.  For details on the naming of wheel files, see :pep:`425`.
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -717,6 +717,20 @@ To have your project installable from a :term:`Package Index` like :term:`PyPI
 <Distribution Package>` (aka ":term:`Package <Distribution Package>`") for your
 project.
 
+Before you can build wheels and sdists for your project, you'll need to install the
+``build`` package:
+
+.. tab:: Unix/macOS
+
+    .. code-block:: bash
+
+        python3 -m pip install build
+
+.. tab:: Windows
+
+    .. code-block:: bash
+
+        py -m pip install build
 
 
 Source distributions
@@ -725,15 +739,24 @@ Source distributions
 Minimally, you should create a :term:`Source Distribution <Source Distribution (or
 "sdist")>`:
 
-::
+.. tab:: Unix/macOS
 
- python setup.py sdist
+    .. code-block:: bash
+
+        python3 -m build --sdist
+
+.. tab:: Windows
+
+    .. code-block:: bash
+
+        py -m build --sdist
 
 
 A "source distribution" is unbuilt (i.e. it's not a :term:`Built
 Distribution`), and requires a build step when installed by pip.  Even if the
 distribution is pure Python (i.e. contains no extensions), it still involves a
-build step to build out the installation metadata from :file:`setup.py`.
+build step to build out the installation metadata from :file:`setup.py` and/or
+:file:`setup.cfg`.
 
 
 Wheels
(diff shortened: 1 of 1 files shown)


henryiii:
Note: we could reduce/remove quite a bit about universal wheels, and mostly make it a note "if you need Python 2+3 support, then"...), and remove some of the Python 2 parts (like how to use 2to3). This page is old.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -829,25 +824,26 @@ To build the wheel:
...
 
 .. tab:: Windows
 
     .. code-block:: bash
 
-        py setup.py bdist_wheel
+        py -m build --wheel
 
-``bdist_wheel`` will detect that the code is pure Python, and build a wheel
+The ``wheel`` package ill detect that the code is pure Python, and build a wheel
```
FFY00, line 835:
```suggestion
The ``wheel`` package will detect that the code is pure Python, and build a wheel
```

FFY00 (approved):

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -725,15 +739,24 @@ Source distributions
...
-::
+.. tab:: Unix/macOS
+
+    .. code-block:: bash
+
+        python3 -m build --sdist
+
+.. tab:: Windows
+
+    .. code-block:: bash
```
jaraco, line 750: Personally, I'd say it's not worth the effort to create bifurcated guidance, especially since `py` is nearly available for unix systems, and because this guidance doesn't apply universally anyway. My preference would be to stick with `python`, but maybe have a note or a link to a section that explains in one place what `python` means and how to invoke the right one for a given environment. In practice, most users will already have created a virtualenv and will have access to `python` in that env.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -779,35 +788,21 @@ Universal Wheels
...
-
-        py setup.py bdist_wheel --universal
-
-You can also permanently set the ``--universal`` flag in :file:`setup.cfg`:
+You should have the following setting in :file:`setup.cfg`:
 
 .. code-block:: text
 
   [bdist_wheel]
   universal=1
```
jaraco, line 796: At this point, any consideration about Python 2 compatibility should be an extra step and not the default. I'd omit this step or move it to a section on adding Python 2 compatibility or just omit it altogether and leave it to the users to find a solution if that's something they need (chances are, they're already in an environment with lots of examples).

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -829,25 +824,26 @@ To build the wheel:
...
+        py -m build --wheel
 
-``bdist_wheel`` will detect that the code is pure Python, and build a wheel
+The ``wheel`` package will detect that the code is pure Python, and build a wheel
 that's named such that it's usable on any Python installation with the same
 major version (Python 2 or Python 3) as the version you used to build the
 wheel.  For details on the naming of wheel files, see :pep:`425`.
 
 If your code supports both Python 2 and 3, but with different code (e.g., you
 use `"2to3" <https://docs.python.org/2/library/2to3.html>`_) you can run
```
jaraco, line 841: Use of `2to3` is discouraged if not deprecated. I suggest removing mentions of it.

jaraco (approved):

henryiii:
Should I update those things in this PR or in a follow up? I was trying to keep the changes focused, but yes, this page is old!

jaraco:
Happy to accept incremental changes.

source/guides/distributing-packages-using-setuptools.rst
webknjaz: FTR this has been implemented across the whole site some time ago. Currently, all the places use tabs. So deciding to change this should be consistent across all the pages. But I think somebody else has also suggested keeping just one tab elsewhere.

Refs: https://github.com/pypa/packaging.python.org/issues/863 / https://github.com/pypa/packaging.python.org/issues/874.

webknjaz:
Looks like https://github.com/pypa/pip/pull/10009/files#diff-f35faadf148777b2d0e535cfd8e8f3e1bbc0fecdb7fda5ccd0b9d7d4a4e5471eR31 substantially changed the RST labels it exposes which means that this site needs to fix the links first. This is what hard-blocks the current PR from merging.

pradyunsg (approved):

henryiii:
Can rebase after #902

Issue #838, opened 2021-02-23
Step in packaging guide to create `setup.py' might cause confusion


layday:
The "Creating the package files" step instructs readers to create a `setup.py`, which remains empty, leading to build failures:

* https://github.com/pypa/packaging-problems/issues/453
* https://github.com/pypa/packaging-problems/issues/454
* https://github.com/pypa/packaging-problems/issues/455

The guide later mentions that, "If you include [setup.py], you must have a call to setup() in it", but this appears to be easily overlooked.

Nines9S:
Was scratching my head for ages wondering what I was doing wrong. Looks like I skimmed over that line too, it's very easily missed for sure. Finally got it working thanks to you!

webknjaz:
Would turning it into an admonition help?

PR #904, opened 2021-05-26
Packaging Python Projects: Remove references to editable installs

bhrutledge:
Towards #866 and #887.

I'm still inclined to encourage using a virtual environment for the whole thing, instead of `pip install --user`, but that feels like a separate PR.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -375,13 +375,13 @@ Make sure you have the latest versions of PyPA's ``build`` installed:
 
     .. code-block:: bash
 
-        python3 -m pip install --user --upgrade build
+        python3 -m pip install --upgrade build
 
 .. tab:: Windows
 
     .. code-block:: bash
 
-        py -m pip install --user --upgrade build
+        py -m pip install --upgrade build
 
 .. tip:: If you have trouble installing these, see the
    :doc:`installing-packages` tutorial.
@@ -452,13 +452,13 @@ distribution packages. You'll need to install Twine:
 
     .. code-block:: bash
 
-        python3 -m pip install --user --upgrade twine
+        python3 -m pip install --upgrade twine
 
 .. tab:: Windows
 
     .. code-block:: bash
 
-        py -m pip install --user --upgrade twine
+        py -m pip install --upgrade twine
 
 Once installed, run Twine to upload all of the archives under :file:`dist`:
 
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -47,7 +47,6 @@ When you're done, the project structure will look like this:
     ├── pyproject.toml
     ├── README.md
     ├── setup.cfg
-    ├── setup.py  # optional, needed to make editable pip installs work
     ├── src/
     │   └── example_pkg/
     │       └── __init__.py
@@ -202,27 +201,6 @@ an escape hatch when absolutely necessary.
 
     There are many more than the ones mentioned here. See
     :doc:`/guides/distributing-packages-using-setuptools` for more details.
-
-    If you create a :file:`setup.py` file, this will enable direct interaction
-    with :file:`setup.py` (which generally should be avoided), and editable
-    installs. This file used to be required, but can be omitted in modern
-    setuptools.
-
-    .. warning::
-
-        If you include the file, you **must have** a call to :func:`setup()` in it,
-        even if there are no arguments:
-
-        .. code-block:: python
-
-           import setuptools
-
-           setuptools.setup()
-
-    Anything you set in :file:`setup.cfg` can instead be set via keyword argument to
-    :func:`setup()`; this enables computed values to be used. You will also need
-    :func:`setup()` for setting up extension modules for compilation. Currently,
-    having this file also is required if you want to use editable installs with pip.
 
 
 .. tab:: :file:`setup.py` (dynamic)
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -5,7 +5,7 @@ This tutorial walks you through how to package a simple Python project. It will
 show you how to add the necessary files and structure to create the package, how
 to build the package, and how to upload it to the Python Package Index.
 
-Some of the commands require a recent version of :ref:`pip`, so start by making
+Some of the commands require a newer version of :ref:`pip`, so start by making
 sure you have the latest version installed:
 
 .. tab:: Unix/macOS
@@ -121,7 +121,7 @@ There are two types of metadata: static and dynamic.
 
 Static metadata (:file:`setup.cfg`) should be preferred. Dynamic metadata (:file:`setup.py`)
 should be used only as an escape hatch when absolutely necessary. :file:`setup.py` used to
-be required, but can be omitted in modern setuptools.
+be required, but can be omitted with newer versions of setuptools and pip.
 
 
 .. tab:: :file:`setup.cfg` (static)
@@ -327,7 +327,7 @@ if you'd like.
 Because our configuration loads :file:`README.md` to provide a
 ``long_description``, :file:`README.md` must be included along with your
 code when you :ref:`generate a source distribution <generating archives>`.
-Recent versions of :ref:`setuptools` will do this automatically.
+Newer versions of :ref:`setuptools` will do this automatically.
 
 
 Creating a LICENSE
@@ -385,7 +385,7 @@ The next step is to generate :term:`distribution packages <Distribution
 Package>` for the package. These are archives that are uploaded to the Python
 Package Index and can be installed by :ref:`pip`.
 
-Make sure you have the latest versions of PyPA's ``build`` installed:
+Make sure you have the latest version of PyPA's ``build`` installed:
 
 .. tab:: Unix/macOS
 
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -5,6 +5,21 @@ This tutorial walks you through how to package a simple Python project. It will
 show you how to add the necessary files and structure to create the package, how
 to build the package, and how to upload it to the Python Package Index.
 
+Some of the commands require a recent version of :ref:`pip`, so start by making
+sure you have the latest version installed:
+
+.. tab:: Unix/macOS
+
+    .. code-block:: bash
+
+        python3 -m pip install --upgrade pip
+
+.. tab:: Windows
+
+    .. code-block:: bash
+
+        py -m pip install --upgrade pip
+
 
 A simple project
 ----------------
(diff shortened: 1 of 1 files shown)


source/tutorials/packaging-projects.rst
```diff
@@ -520,8 +498,21 @@ https://test.pypi.org/project/example-pkg-YOUR-USERNAME-HERE
...
+
+    .. code-block:: bash
+
+        python3 -m pip install --upgrade pip
+
+.. tab:: Windows
+
+    .. code-block:: bash
+
+        py -m pip install --upgrade pip
```
bhrutledge, line 513: While not strictly required for this, this seemed like a good spot to encourage folks to keep `pip` up to date, especially in light of https://github.com/pypa/packaging.python.org/issues/887#issuecomment-845657331.

source/tutorials/packaging-projects.rst
```diff
@@ -520,8 +498,21 @@ https://test.pypi.org/project/example-pkg-YOUR-USERNAME-HERE
...
+
+Now create a new :ref:`virtualenv` (see :doc:`/tutorials/installing-packages` for
 detailed instructions) and install your package from TestPyPI:
```
bhrutledge, line 516: This links to the [projects listing for `virtualenv`](https://packaging.python.org/key_projects/#virtualenv), when I think it would be more helpful to link to general information about creating and activating a virtual environment. I'm planning to update this is a subsequent PR.

source/tutorials/packaging-projects.rst
```diff
@@ -397,13 +375,13 @@ Make sure you have the latest versions of PyPA's ``build`` installed:
 
     .. code-block:: bash
 
-        python3 -m pip install --upgrade build
+        python3 -m pip install --user --upgrade build
```
bhrutledge, line 378: I went with `--user` because that was the recommendation for installing Twine. I'm also happy to remove it from both installs, but TBH I'm not sure what the best practice is here (esp. outside of a virtual environment).

henryiii: Personally, I still feel pipx is the right tool here. Twine and build are _not_ being used as a library, but an application.

source/tutorials/packaging-projects.rst
```diff
@@ -203,27 +202,6 @@ an escape hatch when absolutely necessary.
     There are many more than the ones mentioned here. See
     :doc:`/guides/distributing-packages-using-setuptools` for more details.
 
-    If you create a :file:`setup.py` file, this will enable direct interaction
-    with :file:`setup.py` (which generally should be avoided), and editable
-    installs. This file used to be required, but can be omitted in modern
-    setuptools.
-
-    .. warning::
```
henryiii, line 211: I'm not sure about removing this warning. It was added because people love to add empty setup.py's, which caused a lot of issues in packaging-problems.

source/tutorials/packaging-projects.rst
henryiii: Also, modern pip will automatically install to user if it can't install to the system? Seems like it just encourages bad practice.

source/tutorials/packaging-projects.rst
bhrutledge: Maybe that was because the instructions were recommending creating it? I've removed that above.

source/tutorials/packaging-projects.rst
bhrutledge: This change is a tiny bit of scope creep, just to normalize the install of `build` and `twine`. I think substantial changes to how those tools get installed belongs in a follow-up PR. I'm happy to revert this change if it's a sticking point.

> modern pip will automatically install to user if it can't install to the system? 

That's news to me; when did that get added? Does that imply that `pip install --upgrade pip` should happen before installing `build` and `twine`?

> Seems like it just encourages bad practice.

I'm not sure what you mean by this.

source/tutorials/packaging-projects.rst
bhrutledge: Though maybe it's worth noting somewhere that `setup.py` isn't necessary, for the folks who come to this having seen it used/recommended elsewhere?

source/tutorials/packaging-projects.rst
henryiii: [Pip 20.0](https://pip.pypa.io/en/stable/news/#v20-0)

> I'm not sure what you mean by this.

It's bad practice to use `--user`. It's wrong in a virtual environment, and it's confusing outside of one. It's not easy to know what state a users's `user` directory is in, and know how to reset it. Upgrades don't upgrade dependencies, and packages often forget to test the minimum versions (build does, though!). I think it would be best just to avoid ever showing it.

Personally I'd show how to install `pipx` instead of `twine` and ` build`, then use `pipx run twine` and `pipx run build`. There is no reason twine or build should be inside your virtual development environment (setuptools and wheel, as well, due to PEP 518). You aren't importing them as a library. And their requirements could even mess with your requirements (though for those two, very unlikely). This also ensure that `twine` and `build` are never more than a week old.

henryiii: PS: do _not_ change to pipx unless others agree (probably not?), and not in this PR. I'm just saying what I think would make the most sense going forward.

source/tutorials/packaging-projects.rst
henryiii: Don't think I'm against it, though you probably will have to add a note below saying that you might have to add `--user` if pip is older than pip 20.

source/tutorials/packaging-projects.rst
henryiii: No, it was because the part above gives a directory tree, and users, going step by step, make empty files for each one of the listed items. We do need this warning. Unless you want to volunteer to field 1-3 issues per day. ;)

henryiii: Oh, you've removed that item in the tree. Nice. Hmm, _maybe_ it is okay to drop the warning. Though since you'll see "setup.py" listed other places, I'm still in favor of a warning of some sort explaining that setup.py is not needed, but if present, it needs to contain a setup() call.

source/tutorials/packaging-projects.rst
pfmoore: I agree that `pipx` is the right tool for this type of usage, and as it looks like pipx is going to become part of the PyPA organisation, it seems like a reasonable tool to promote. But it is a bigger change, and probably warrants a separate PR, where the principle of promoting pipx as the tool to use can be discussed more broadly.

webknjaz: It's quite typical to have a dedicated tox env with `build` because it is one of the things one would want to pin in their automation setup. This is why I dislike the idea of using a tool with a global env.

henryiii: If someone knows how to use nox (or tox, I guess), then do they need to be following this tutorial? At least in nox's case, you don't see the "pip" part when you install build (IIRC tox too), so it's not like you are copying "pip" or "pipx" to a nox/tox config anyway.

henryiii: And, still to my point, you'd not have `--user` in a nox/tox file anyway. Same global issue (in fact, worse).

webknjaz: True. OTOH I believe this `--user` advice appeared all over the place as a response to random tutorials and READMEs on the internet showcasing `sudo pip` a little too much.

source/tutorials/packaging-projects.rst
bhrutledge: I moved a piece of this warning to the [discussion of "static" and "dynamic" metadata](https://github.com/pypa/packaging.python.org/pull/904/files#diff-04bfc8bd25a83763d519bf1a6e9178b31618e941c4a8cda88438c7aa39940d79R114-R124), which strongly encourages using `setup.cfg`.

source/tutorials/packaging-projects.rst
```diff
@@ -5,6 +5,21 @@ This tutorial walks you through how to package a simple Python project. It will
...
+    .. code-block:: bash
+
+        python3 -m pip install --upgrade pip
+
+.. tab:: Windows
+
+    .. code-block:: bash
+
+        py -m pip install --upgrade pip
+
```
bhrutledge, line 22: @henryiii @webknjaz @pfmoore Following up on the discussions in https://github.com/pypa/packaging.python.org/pull/904#discussion_r639324954 and https://github.com/pypa/packaging.python.org/pull/904#discussion_r639326142, I moved the `pip` upgrade to the top of the tutorial, and removed the `--user` option below. I think more substantial changes should be in a subsequent PR.

source/tutorials/packaging-projects.rst
henryiii: Could you put the warning block up there too, with just the "if you include this file, you must" part?

source/tutorials/packaging-projects.rst
```diff
@@ -105,8 +119,9 @@ There are two types of metadata: static and dynamic.
   dynamic or determined at install-time, as well as extension modules or
   extensions to setuptools, need to go into :file:`setup.py`.
 
-Static metadata should be preferred and dynamic metadata should be used only as
-an escape hatch when absolutely necessary.
+Static metadata (:file:`setup.cfg`) should be preferred. Dynamic metadata (:file:`setup.py`)
+should be used only as an escape hatch when absolutely necessary. :file:`setup.py` used to
+be required, but can be omitted with newer versions of setuptools and pip.
```
henryiii, line 124:
```suggestion
be required, but can be omitted with newer versions of setuptools and pip.

.. warning::

   If you include the file, you **must have** a call to :func:`setup()` in it,
   even if there are no arguments:

   .. code-block:: python

      import setuptools

      setuptools.setup()
```

henryiii: See #838 for a taste of the trouble that ended up prompting the addition of this originally.

henryiii: Other than this, I'm happy with the changes :)

bhrutledge: I understand the desire to include it, but with the other edits, it doesn't seem necessary. #838 says:

> The "Creating the package files" step instructs readers to create a `setup.py`, which remains empty, leading to build failures:

If you read through the [tutorial from the start](https://python-packaging-user-guide--904.org.readthedocs.build/tutorials/packaging-projects/), that's no  longer the case. By this point in the tutorial, the only mention of `setup.py` is  in this section, where its use is discouraged. I think including this warning is likely to be a confusing distraction to newcomers.

Could we leave it out, and restore it if folks start reporting issues?

bhrutledge:
Can this be merged? I'd like to do more work on the guide, building on this.

@webknjaz @henryiii

henryiii:
Someone can decide about https://github.com/pypa/packaging.python.org/pull/904#discussion_r640464543 - other than that, I'm happy (and I am okay with the proposal to add it back in if someone has a problem - I expect they will, but I could be wrong. It's an easy mistake to make. The error message is better now, I think, so it might be less important than it used to be.

webknjaz (approved):

PR #908, opened 2021-05-31
Packaging Python Projects: Add/update links to other docs

bhrutledge:
Towards #866.

This is mostly adding more links glossary entries, with a couple more substantial changes that felt related.

### DIFF
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -189,7 +189,7 @@ This will create a new virtual environment in the ``tutorial_env`` subdirectory,
 and configure the current shell to use it as the default ``python`` environment.
 
 
-.. _`Creating and using Virtual Environments`:
+.. _Creating and using Virtual Environments:
 
 Creating Virtual Environments
 =============================
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -77,13 +77,13 @@ Creating a test directory
 Creating pyproject.toml
 -----------------------
 
-:file:`pyproject.toml` is the file that tells build tools (like ``pip`` and
-``build``) what system you are using and what is required for building. The
-default if this file is missing is to assume a classic setuptools build system,
+:file:`pyproject.toml` is the file that tells build tools (like :ref:`pip` and
+:ref:`build`) what system you are using and what is required for building. The
+default if this file is missing is to assume a :ref:`setuptools` build system,
 but it is better to be explicit; if you have a :file:`pyproject.toml` file, you
-will be able to rely on ``wheel`` and other packages being present.
+will be able to rely on :ref:`wheel` and other packages being present.
 
-This file should be ideal for most setuptools projects:
+This file should be ideal for most :ref:`setuptools` projects:
 
 
 .. code-block:: toml
@@ -103,8 +103,8 @@ not after it is installed.
 ``build-system.build-backend`` is technically optional, but you will get
 ``setuptools.build_meta:__legacy__`` instead if you forget to include it, so
 always include it. If you were to use a different build system, such as
-:ref:`flit` or `poetry`_, those would go here, and the configuration details
-would be completely different than the setuptools configuration described
+:ref:`flit` or :ref:`poetry`, those would go here, and the configuration details
+would be completely different than the :ref:`setuptools` configuration described
 below. See :pep:`517` and :pep:`518` for background and details.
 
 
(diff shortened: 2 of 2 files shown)


source/tutorials/installing-packages.rst
```diff
@@ -189,7 +189,7 @@ This will create a new virtual environment in the ``tutorial_env`` subdirectory,
 and configure the current shell to use it as the default ``python`` environment.
 
 
-.. _`Creating and using Virtual Environments`:
+.. _Creating and using Virtual Environments:
```
bhrutledge, line 192: This change allowed the new `:ref:` link to work, and didn't break any existing ones.

source/tutorials/packaging-projects.rst
```diff
@@ -77,14 +77,9 @@ Creating a test directory
 Creating pyproject.toml
 -----------------------
```
bhrutledge, line 78: I made some more substantial changes to this section because it's the first introduction of the setuptools (and build tools in general), and it felt like there was more information than necessary for a newcomer (esp. related to legacy configurations, which could be confusing).

source/tutorials/packaging-projects.rst
```diff
@@ -367,14 +364,14 @@ Including other files
...
+:term:`source distribution <Source Distribution (or "sdist")>`. If you want to
+control what goes in this explicitly, see :ref:`Using MANIFEST.in`.
 
 The final :term:`built distribution <Built Distribution>` will have the Python
 files in the discovered or listed Python packages. If you want to control what
 goes here, such as to add data files, see
-:std:doc:`Including Data Files <setuptools:userguide/datafiles>`
-from the :std:doc:`setuptools docs <setuptools:index>`.
+:doc:`Including Data Files <setuptools:userguide/datafiles>`
+from the :doc:`setuptools docs <setuptools:index>`.
```
bhrutledge, line 374: I don't know what `:std:` is supposed to do, but it doesn't seem necessary...

source/tutorials/packaging-projects.rst
bhrutledge: @henryiii Any thoughts on the `pyproject.toml` section?

bhrutledge:
@webknjaz Could you take a look at this?

source/tutorials/packaging-projects.rst
webknjaz: Some the roles are namespaced and this is the default namespace. It is probably coming from https://webknjaz.github.io/intersphinx-untangled/ as it outputs full role names in the examples column.

bhrutledge: > It is probably coming from https://webknjaz.github.io/intersphinx-untangled/

I don't understand this.

Looks like @henryiii [added this section](https://github.com/pypa/packaging.python.org/blame/df074307ef8e131c1f1a56857ba9dcf36b0d7f42/source/tutorials/packaging-projects.rst#L376-L387) two months ago. However, there are other intersphinx links to setuptools that don't use `:std:`, e.g.:

https://github.com/pypa/packaging.python.org/blob/main/source/guides/creating-and-discovering-plugins.rst#using-package-metadata

So I'm inclined to leave it out.

webknjaz: > I don't understand this.

That's my pet project that renders all the objects that different Sphinx sites expose. If you link against one of such sites, besides adding them to `intersphinx_mapping`, you normally need to know the names and the types of the objects documented there. One way is to do it with a CLI command that sphinx provides but the output is rather poorly structured so this page outputs the same info as a table and I've added examples of how people could link each object in their docs.

Things may be `:std:doc:` or `:py:mod:`, for example. Those are full names of the object roles. But one role may be set up as default and that is usually `std`. So if one omits that and uses just `:doc:`, sphinx will automatically assume that it's `:std:doc:` under the hood.

But when outputting the intersphinx data, the full name is always encoded internally so that's what this site shows. And my guess was that this full syntax used in the document has simply been copy-pasted from my project.

FWIW both variants work. I hope that this attempt to explain `:std:` doesn't sound too confusing... 

P.S. Maybe this https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html will shed more light on the roles..

webknjaz (approved):

source/tutorials/packaging-projects.rst
bhrutledge: Thanks for the explanation, and your pet project. I totally missed that clicking on those links would take you to something like https://webknjaz.github.io/intersphinx-untangled/setuptools.rtfd.io/; I just assumed they linked to the named sites. I think adding some explanatory text on the index would make it even more useful. 😅

webknjaz: Oh, that's a good idea. I just didn't invest into that too much time after I got what I needed for myself :)
But I'll definitely consider that or accept any improvement PRs (because I don't hope to have any time for this right now).

PR #910, opened 2021-06-01
binary-extensions: add cibuildwheel and multibuild

henryiii:
This adds a mention of pypa's cibuildwheel and multibuild here before going into the (highly outdated, currently) specifics of building wheels. For most users, this likely is the correct place to go, rather than trying to set things up themselves. For those who might want to do things themselves, the following sections still need updating.

### DIFF
--- a/source/guides/packaging-binary-extensions.rst
+++ b/source/guides/packaging-binary-extensions.rst
@@ -232,16 +232,15 @@ guide includes an introduction to writing a
 Building binary extensions
 ==========================
 
-Building extensions in CI
+Building extensions for multiple platforms
 -------------------------
 
-If you plan to distribute your modules, you should provide builds for all the
+If you plan to distribute your extension, you should provide :term:`wheels <Wheel>` for all the
 platforms you intend to support. For most extensions, this is at least one
-module per Python version times the number of OS and archetectures you support.
+package per Python version times the number of OS and architectures you support.
 These are usually built on continous integration (CI) systems. There are tools
 to help you build highly redistributable binaries from CI; these include
-`cibuildwheel <https://cibuildwheel.readthedocs.io>`__ and
-`multibuild <https://github.com/matthew-brett/multibuild>`__.
+:ref:`cibuildwheel` and :ref:`multibuild`.
 
 
 Binary extensions for Windows
--- a/source/guides/packaging-binary-extensions.rst
+++ b/source/guides/packaging-binary-extensions.rst
@@ -233,14 +233,15 @@ Building binary extensions
 ==========================
 
 Building extensions for multiple platforms
--------------------------
+------------------------------------------
 
-If you plan to distribute your extension, you should provide :term:`wheels <Wheel>` for all the
-platforms you intend to support. For most extensions, this is at least one
-package per Python version times the number of OS and architectures you support.
-These are usually built on continous integration (CI) systems. There are tools
-to help you build highly redistributable binaries from CI; these include
-:ref:`cibuildwheel` and :ref:`multibuild`.
+If you plan to distribute your extension, you should provide
+:term:`wheels <Wheel>` for all the platforms you intend to support. For most
+extensions, this is at least one package per Python version times the number of
+OS and architectures you support.  These are usually built on continuous
+integration (CI) systems. There are tools to help you build highly
+redistributable binaries from CI; these include :ref:`cibuildwheel` and
+:ref:`multibuild`.
 
 
 Binary extensions for Windows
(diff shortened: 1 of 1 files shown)


source/guides/packaging-binary-extensions.rst
```diff
@@ -232,6 +232,18 @@ guide includes an introduction to writing a
...
+Building extensions in CI
+-------------------------
+
+If you plan to distribute your modules, you should provide builds for all the
+platforms you intend to support. For most extensions, this is at least one
+module per Python version times the number of OS and archetectures you support.
+These are usually built on continous integration (CI) systems. There are tools
+to help you build highly redistributable binaries from CI; these include
+`cibuildwheel <https://cibuildwheel.readthedocs.io>`__ and
+`multibuild <https://github.com/matthew-brett/multibuild>`__.
```
bhrutledge, line 244: I think the pattern has been to link to the [project summaries](https://github.com/pypa/packaging.python.org/blob/main/source/key_projects.rst):

```suggestion
:ref:`cibuildwheel` and :ref:`multibuild`.
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -232,6 +232,18 @@ guide includes an introduction to writing a
 Building binary extensions
 ==========================
 
+Building extensions in CI
```
bhrutledge, line 235: Readers might not know what CI is, and I think focusing on the desired outcome is clearer:

```suggestion
Building extensions for multiple platforms
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -232,6 +232,18 @@ guide includes an introduction to writing a
...
+-------------------------
+
+If you plan to distribute your modules, you should provide builds for all the
+platforms you intend to support. For most extensions, this is at least one
+module per Python version times the number of OS and archetectures you support.
```
bhrutledge, line 240: I _think_ this is more accurate, based on my understanding:

```suggestion
If you plan to distribute your extension, you should provide :term:`wheels <Wheel>` for all the
platforms you intend to support. For most extensions, this is at least one
package per Python version times the number of OS and architectures you support.
```

bhrutledge (changes requested):
Thanks for starting to improve this doc. I have no direct experience with binary extensions, so it's a good read for me. There's one spelling fix, plus some suggestions for clarity.

henryiii:
Thanks, @bhrutledge!

henryiii:
There's no multibuild project summary, so this PR is failing after the change. I think the best fix would be to add a multibuild section under non PyPA project, rather than to revert to a link. Will make a PR for that that this PR will then depend on.

pradyunsg (approved):
LGTM, although this whole guide likely needs to be updated and completed. :)

source/guides/packaging-binary-extensions.rst
```diff
@@ -232,6 +232,17 @@ guide includes an introduction to writing a
 Building binary extensions
 ==========================
 
+Building extensions for multiple platforms
+-------------------------
```
pradyunsg: This underline needs to be fixed tho. :)

henryiii: Too late. ;)

source/guides/packaging-binary-extensions.rst
```diff
@@ -233,14 +233,14 @@ Building binary extensions
...
-If you plan to distribute your extension, you should provide :term:`wheels <Wheel>` for all the
-platforms you intend to support. For most extensions, this is at least one
-package per Python version times the number of OS and architectures you support.
-These are usually built on continous integration (CI) systems. There are tools
-to help you build highly redistributable binaries from CI; these include
-:ref:`cibuildwheel` and :ref:`multibuild`.
+------------------------------------------
+
+If you plan to distribute your extension, you should provide :term:`wheels
+<Wheel>` for all the platforms you intend to support. For most extensions, this
```
pradyunsg, line 239: I don't think splitting a directive across multiple lines works, but hey, I'd be willing to be surprised. :)

henryiii:
I'm working on it, in small pieces. I'll tag a few possibly interested parties when I make the next PRs (likely after PyHEP and SciPy, so in a couple of weeks at the earliest)

source/guides/packaging-binary-extensions.rst
henryiii: Ahh, reflowed that and didn't watch for directives. We can see, I guess. :)

henryiii: Nope.

PR #911, opened 2021-06-01
binary-extensions: update Windows instructions

henryiii:
About a month ago Microsoft permanently killed off https://www.microsoft.com/en-gb/download/details.aspx?id=8279 (MSVC 2008 for Python 2.7); it was officially at end-of-really-extended-life-for-python-2.7 since 2018. I don't think details about Python 3.4 are all that interesting to anyone anymore, either.

Included the caveat discovered by matplotlib for MSVC 2019.

### DIFF
--- a/source/guides/packaging-binary-extensions.rst
+++ b/source/guides/packaging-binary-extensions.rst
@@ -238,26 +238,26 @@ Binary extensions for Windows
 Before it is possible to build a binary extension, it is necessary to ensure
 that you have a suitable compiler available. On Windows, Visual C is used to
 build the official CPython interpreter, and should be used to build compatible
-binary extensions.
+binary extensions.  To set up a build environment for binary extensions, install
+`Visual Studio Community Edition <https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx>`__
+- any recent version is fine.
+
+One caveat: if you use Visual Studio 2019 or later, your extension will depend
+on an "extra" file, ``VCRUNTIME140_1.dll``, in addition to the
+``VCRUNTIME140.dll`` that all previous versions back to 2015 depend on. This
+will add an extra requirement to using your extension on versions of CPython
+that do not include this extra file. To avoid this, you can add the
+compile-time argument ``/d2FH4-``. Recent versions of Python may include this
+file.
 
 Building for Python prior to 3.5 is discouraged, because older versions of
-Visual Studio are no longer available from Microsoft. If you do
-need to build for older versions, you can set ``DISTUTILS_USE_SDK=1`` and
-``MSSdk=1`` to force a the currently activated version of MSVC to be found,
-and you should exercise care when designing your extension not to malloc/free
-memory across different libraries, avoid relying on changed data structures,
-and so on. Tools for generating extension modules usually avoid these things
-for you.
+Visual Studio are no longer available from Microsoft. If you do need to build
+for older versions, you can set ``DISTUTILS_USE_SDK=1`` and ``MSSdk=1`` to
+force a the currently activated version of MSVC to be found, and you should
+exercise care when designing your extension not to malloc/free memory across
+different libraries, avoid relying on changed data structures, and so on. Tools
+for generating extension modules usually avoid these things for you.
 
-To set up a build environment for binary extensions, install
-`Visual Studio Community Edition <https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx>`__ - any
-recent version is fine.
-
-One caveat: if you use Visual Studio 2019 or later, your extension will depend on an "extra"
-file, ``VCRUNTIME140_1.dll``, in addition to the ``VCRUNTIME140.dll`` that all previous
-verisions back to 2015 depend on. This will add an extra requirement to using
-your extension on versions of CPython that do not include this extra file. To
-avoid this, you can add the compile-time argument ``/d2FH4-``.
 
 
 Binary extensions for Linux
--- a/source/guides/packaging-binary-extensions.rst
+++ b/source/guides/packaging-binary-extensions.rst
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
 build the official CPython interpreter, and should be used to build compatible
 binary extensions.
 
-Python 2.7 used Visual Studio 2008, Python 3.3 and 3.4 used Visual Studio 2010,
-and Python 3.5+ uses Visual Studio 2015 or later. Unfortunately, older versions of
-Visual Studio are no longer easily available from Microsoft, so for versions
-of Python prior to 3.5, the compilers must be obtained differently if you do
-not already have a copy of the relevant version of Visual Studio.
+Python 2.7 used Visual Studio 2008, and Python 3.5+ uses Visual Studio 2015 or
+later; since 2015 Microsoft has kept backward compatibility (with one exception,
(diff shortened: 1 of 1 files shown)


source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
+memory across different libraries, avoid relying on changed data structures,
+and so on. Tools for generating extension modules usually avoid these things
+for you.
+
+To set up a build environment for binary extensions, install
+`Visual Studio Community Edition <https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx>`__ - any
+recent version is fine.
+
+One caveat; if you use MSVC 2019+, your extension will depend on an "extra"
+file, ``VCRUNTIME140_1.dll``, in addition to the ``VCRUNTIME140.dll`` all previous
```
bhrutledge, line 259:
```suggestion
One caveat: if you use Visual Studio 2019 or later, your extension will depend on an "extra"
file, ``VCRUNTIME140_1.dll``, in addition to the ``VCRUNTIME140.dll`` that all previous
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
+verisions back to 2015 depend on. This will add an extra requirement to using
+your extension on versions of CPython that do not include this extra file; you
+can add the compile-time argument ``/d2FH4-`` to avoid this link on 2019+.
```
bhrutledge, line 262:
```suggestion
your extension on versions of CPython that do not include this extra file. To
avoid this, you can add the compile-time argument ``/d2FH4-``.
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
-Note that from Python 3.5 onwards, Visual Studio works in a backward
-compatible way, which means that any future version of Visual Studio will
-be able to build Python extensions for all Python versions from 3.5 onwards.
+Python 2.7 used Visual Studio 2008, and Python 3.5+ uses Visual Studio 2015 or
+later; since 2015 Microsoft has kept backward compatibility (with one exception,
+listed below). Unfortunately, older versions of Visual Studio are no longer
+available from Microsoft, so building for CPython < 3.5 is discouraged. If you do
+need to build for older versions, you can set ``DISTUTILS_USE_SDK=1`` and
+``MSSdk=1`` to force a the currently activated version of MSVC to be found,
+and you should exercize care when designing your extension not to malloc/free
```
bhrutledge, line 249:
```suggestion
and you should exercise care when designing your extension not to malloc/free
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
+recent version is fine.
+
+One caveat; if you use MSVC 2019+, your extension will depend on an "extra"
+file, ``VCRUNTIME140_1.dll``, in addition to the ``VCRUNTIME140.dll`` all previous
+verisions back to 2015 depend on. This will add an extra requirement to using
+your extension on versions of CPython that do not include this extra file; you
+can add the compile-time argument ``/d2FH4-`` to avoid this link on 2019+.
 
 Building with the recommended compiler on Windows ensures that a compatible C library
 is used throughout the Python process.
```
bhrutledge, line 265: This feel unnecessary, or it could be conveyed introductory paragraph.

```suggestion
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
-           (or any later version, when these are released).
-        2. Done.
-
-Note that from Python 3.5 onwards, Visual Studio works in a backward
-compatible way, which means that any future version of Visual Studio will
-be able to build Python extensions for all Python versions from 3.5 onwards.
+Python 2.7 used Visual Studio 2008, and Python 3.5+ uses Visual Studio 2015 or
+later; since 2015 Microsoft has kept backward compatibility (with one exception,
+listed below). Unfortunately, older versions of Visual Studio are no longer
+available from Microsoft, so building for CPython < 3.5 is discouraged. If you do
```
bhrutledge, line 246: I'm not sure the extra detail is necessary here (though maybe a mention of Visual Studio 2015 would still be useful).

```suggestion
Building for Python prior to 3.5 is discouraged, because older versions of
Visual Studio are no longer available from Microsoft. If you do
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -240,43 +240,26 @@ that you have a suitable compiler available. On Windows, Visual C is used to
...
+need to build for older versions, you can set ``DISTUTILS_USE_SDK=1`` and
+``MSSdk=1`` to force a the currently activated version of MSVC to be found,
+and you should exercize care when designing your extension not to malloc/free
+memory across different libraries, avoid relying on changed data structures,
+and so on. Tools for generating extension modules usually avoid these things
+for you.
+
+To set up a build environment for binary extensions, install
+`Visual Studio Community Edition <https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx>`__ - any
+recent version is fine.
```
bhrutledge, line 256: This feels like the most important bit. Could it go before the "older versions" paragraph above?

bhrutledge (changes requested):
I'm not a Windows user, so these are just suggested copy edits for clarity (plus one for spelling).

source/guides/packaging-binary-extensions.rst
henryiii: Moved.

source/guides/packaging-binary-extensions.rst
```diff
@@ -238,48 +238,27 @@ Binary extensions for Windows
...
+binary extensions.  To set up a build environment for binary extensions, install
+`Visual Studio Community Edition <https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx>`__
+- any recent version is fine.
+
+One caveat: if you use Visual Studio 2019 or later, your extension will depend
+on an "extra" file, ``VCRUNTIME140_1.dll``, in addition to the
+``VCRUNTIME140.dll`` that all previous versions back to 2015 depend on. This
+will add an extra requirement to using your extension on versions of CPython
+that do not include this extra file. To avoid this, you can add the
+compile-time argument ``/d2FH4-``. Recent versions of Python may include this
```
henryiii, line 250: I'd like to put exact numbers, but I'm not sure which ones they are (it might have changed in patch releases, too?)

bhrutledge (approved):
I think this reads much better. Thanks!

webknjaz:
I'm not a Windows user either so I'd feel much more comfortable merging the change if one could be found and posted a review. If nobody is available for some time, we'll merge this as is.

pfmoore (approved):
(Windows user here!) This looks good to me, although I only very rarely build C extensions these days, so while the technical details of the content look accurate to me, I'm not an expert.

I'd say this can be merged as is, though - we can always iterate on the technical details if further nuances are reported.

webknjaz:
Thanks, Paul!

henryiii:
Thanks for the review! I'll try to work on updating macOS and Linux next (though likely after SciPy).

PR #916, opened 2021-06-18
Update support link

bhrutledge:


pradyunsg:
I'm not sure that we want to direct support requests to Discourse at this point. Something like pypa/packaging-problems is much better suited for support requests nowadays.

webknjaz:
Honestly, I think that having discussions enabled in this repo could serve the same purpose.

pfmoore:
I'd definitely prefer *not* to use the packaging Discourse for user support. If we need a "user help" forum it should either be a separate Discourse category, or something like discussions on this repo.

pfmoore (changes requested):
See my comments - I should probably have made them a review. We should not use the existing Packaging category on Discourse for "user help" types of query.

pradyunsg:
> Honestly, I think that having discussions enabled in this repo could serve the same purpose.

We already have a place where we have users come to ask for help - the issue tracker on packagong-problems. I don't think adding another channel for users to ask for helps, helps anyone here. We'd be stretching our already-limited volunteer energy across more spaces making it even less likely that someone responds.

webknjaz:
Not additionally but rather instead.

bhrutledge:
Thanks, folks. I've linked to https://github.com/pypa/packaging-problems.

pradyunsg:
> Not additionally but rather instead.

Ah, okay. We could, but that's still a discussion to be had with more people than just us. :)

pradyunsg (approved):

pfmoore (approved):

PR #919, opened 2021-06-18
Prune project links

bhrutledge:
After scanning through some of the projects, it seemed like most of the links to IRC and mailing lists are not reliable ways of engaging with a project. Rather than direct users to those, I thought it might be better to rely on each project's repo and documentation to point users in the right direction.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -21,8 +21,7 @@ bandersnatch
 `Mailing list <http://mail.python.org/mailman/listinfo/distutils-sig>`__ [2]_ |
 `Issues <https://github.com/pypa/bandersnatch/issues>`__ |
 `GitHub <https://github.com/pypa/bandersnatch>`__ |
-`PyPI <https://pypi.org/project/bandersnatch>`__ |
-Dev IRC:`#bandersnatch <https://web.libera.chat/#bandersnatch>`__
+`PyPI <https://pypi.org/project/bandersnatch>`__
 
 ``bandersnatch`` is a PyPI mirroring client designed to efficiently
 create a complete mirror of the contents of PyPI. Organizations thus
@@ -39,9 +38,7 @@ build
 `Docs <https://pypa-build.readthedocs.io/>`__ |
 `Issues <https://github.com/pypa/build/issues>`__ |
 `GitHub <https://github.com/pypa/build>`__ |
-`PyPI <https://pypi.org/project/build>`__ |
-User IRC:`#pypa <https://web.libera.chat/#pypa>`__ |
-Dev IRC:`#pypa-dev <https://web.libera.chat/#pypa-dev>`__
+`PyPI <https://pypi.org/project/build>`__
 
 ``build`` is a :pep:`517` compatible Python package builder. It provides a CLI to
 build packages, as well as a Python API.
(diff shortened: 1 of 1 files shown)


souldia18:
#917

souldia18 (approved):

pradyunsg (dismissed):
I'd only trim distutils-sig and #pypa/#pypa-dev IRC channels. Those other mailing lists are likely active.

webknjaz:
I'd keep them — we have a linkcheck for detecting if something disappears.

bhrutledge:
> I'd only trim distutils-sig and #pypa/#pypa-dev IRC channels. Those other mailing lists are likely active.

I don't think this is the case:

- https://groups.google.com/g/python-virtualenv
- https://groups.google.com/g/devpi-dev
- https://groups.google.com/g/scikit-build
- http://librelist.com/browser/bento

> we have a linkcheck for detecting if something disappears.

True, but that doesn't detect if it's still actively used.

bhrutledge:
Based on my last comment, are y'all okay with these changes?

pradyunsg (commented):
Personally, speaking for pip, packaging, packaging.python.org and distutils -- yes, this looks good to me. I imagine that this works for twine and warehouse too.

We should get an OK from the maintainers of the other projects for the corresponding changes here tho.

---

This PR is removing the IRC links and the link to distutils-sig:

/cc @jaraco for setuptools
/cc @gaborbernat for virtualenv
/cc @vsajip for distlib
/cc @cooperlees for bandersnatch
/cc @agronholm for wheel

And... IDK who to ping for devpi / scikit-build / buildout / bento.

agronholm:
The wheel part looks good to me, nothing out of place there.

bhrutledge:
> And... IDK who to ping for devpi / scikit-build / buildout / bento.

Me neither. Which is why I thought it would be better in general to link to consistent things like project sites and documentation, rather than changeable things like communication platforms that require maintainers to keep this page updated.

cooperlees (approved):
LGTM for bandersnatch.

jaraco (approved):

pradyunsg:
I think it's safe to assume that the remaining folks are also OK with this. If that assumption proves wrong, reverts are cheap; so if someone has concerns, we can revert parts of this later.

source/key_projects.rst
```diff
@@ -280,13 +259,9 @@ virtualenv
 ==========
 
 `Docs <https://virtualenv.pypa.io/en/stable/>`__ |
-`User list <http://groups.google.com/group/python-virtualenv>`__ |
-`Dev list <https://mail.python.org/mailman3/lists/distutils-sig.python.org/>`__ |
 `Issues <https://github.com/pypa/virtualenv/issues>`__ |
 `GitHub <https://github.com/pypa/virtualenv>`__ |
-`PyPI <https://pypi.org/project/virtualenv/>`__ |
-User IRC:`#pypa <https://web.libera.chat/#pypa>`__ |
```
gaborbernat, line 288: I guess we should have added Discord here...

bhrutledge: As with the other projects, I think having a link to Discord in virtualenv's docs might be sufficient.

https://github.com/pypa/virtualenv/issues/2134

bhrutledge:
Thanks all!

souldia1 (commented):
😈

PR #928, opened 2021-06-27
Fix executable dir in + tabify pip & virtual envs guide

### LINKED ISSUE #783: Small mistake in guide
I'm relatively new to programming so please forgive me if it really is correct, but I'm pretty sure that on https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/ in the section on _Activating a virtual environment_ the windows response to the command **where python** should be **...\env\Scripts\python.exe** instead of **.../env/bin/python.exe**. At least it is for me.

ichard26:
Fixes #783.

This patch also completes the "tabification" this document.

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -156,7 +156,7 @@ Python interpreter, it should point to the ``env`` directory.
 
 .. tab:: Windows
 
-    .. code-block:: bash
+    .. code-block:: text
 
         where python
         ...\env\Scripts\python.exe
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -159,7 +159,7 @@ Python interpreter, it should point to the ``env`` directory.
     .. code-block:: bash
 
         where python
-        .../env/Scripts/python.exe
+        ...\env\Scripts\python.exe
 
 As long as your virtual environment is activated pip will install packages into that
 specific environment and you'll be able to import and use packages in your
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -19,46 +19,43 @@ update packages. You'll need to make sure you have the latest version of pip
 installed.
 
 
-Windows
-+++++++
+.. tab:: Unix/macOS
 
-The Python installers for Windows include pip. You should be able to access
-pip using:
+    Debian and most other distributions include a `python-pip`_ package, if you
+    want to use the Linux distribution-provided versions of pip see
+    :doc:`/guides/installing-using-linux-tools`.
 
-.. code-block:: bash
+    You can also install pip yourself to ensure you have the latest version. It's
+    recommended to use the system pip to bootstrap a user installation of pip:
 
-    py -m pip --version
-    pip 9.0.1 from c:\python36\lib\site-packages (Python 3.6.1)
+    .. code-block:: bash
 
-You can make sure that pip is up-to-date by running:
+        python3 -m pip install --user --upgrade pip
 
-.. code-block:: bash
+    Afterwards, you should have the newest pip installed in your user site:
 
-    py -m pip install --upgrade pip
+    .. code-block:: bash
 
+        python3 -m pip --version
+        pip 9.0.1 from $HOME/.local/lib/python3.6/site-packages (python 3.6)
 
-Linux and macOS
-++++++++++++++++
+    .. _python-pip: https://packages.debian.org/stable/python-pip
 
-Debian and most other distributions include a `python-pip`_ package, if you
-want to use the Linux distribution-provided versions of pip see
-:doc:`/guides/installing-using-linux-tools`.
+.. tab:: Windows
 
-You can also install pip yourself to ensure you have the latest version. It's
-recommended to use the system pip to bootstrap a user installation of pip:
+    The Python installers for Windows include pip. You should be able to access
+    pip using:
 
-.. code-block:: bash
+    .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -134,33 +132,34 @@ need to *activate* it. Activating a virtual environment will put the
...
 
-.. code-block:: bash
+.. tab:: Windows
 
-    where python
-    .../env/bin/python.exe
+    .. code-block:: bash
 
+        where python
+        .../env/Scripts/python.exe
```
ichard26, line 162: This is where the `bin` to `Script` fix happened, this diff is harder to parse than I'd have hoped :/

bhrutledge: It's easier if you [hide whitespace changes](https://github.blog/2018-05-01-ignore-white-space-in-code-review/).

https://github.com/pypa/packaging.python.org/pull/928/files?diff=unified&w=1

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -19,46 +19,43 @@ update packages. You'll need to make sure you have the latest version of pip
...
-.. code-block:: bash
+    The Python installers for Windows include pip. You should be able to access
+    pip using:
 
-    python3 -m pip install --user --upgrade pip
+    .. code-block:: bash
 
-Afterwards, you should have the newest pip installed in your user site:
+        py -m pip --version
+        pip 9.0.1 from c:\python36\lib\site-packages (Python 3.6.1)
```
bhrutledge, line 52: This is a very old version of pip and Python. Can we take the opportunity to show the latest version for Windows and Unix/macOS?

source/guides/installing-using-pip-and-virtual-environments.rst
bhrutledge, line 162: Should this be:

```suggestion
        ...\env\Scripts\python.exe
```

source/guides/installing-using-pip-and-virtual-environments.rst
bhrutledge: Also, this is one of a few places in this document where the command and its output are shown in a single `code-block`, but with no distinction between them, which might lead folks to thinking that `pip 9.0.1 ...` is a command to execute.

If the output is valuable, I recommend putting into a separate `code-block`, maybe something like:

```suggestion
    .. code-block:: text

        pip 9.0.1 from c:\python36\lib\site-packages (Python 3.6.1)
```

Again, we'd want to do the same for Unix/macOS, and similarly for the "You can confirm you're in the virtual environment" instructions. But that could also be a follow-up PR.

bhrutledge (commented):
Thanks for this; I like seeing the adoption of tabs in this doc. I've got a few questions/suggestions.

source/guides/installing-using-pip-and-virtual-environments.rst
ichard26: Ah, I forgot about this feature, thank you!

ichard26: Ugh, the fact I switched from Windows to Linux since 6 months really shows. My bad.

source/guides/installing-using-pip-and-virtual-environments.rst
ichard26: I was avoiding any informational changes to the document since I was afraid I lack the domain knowledge to do them properly, but this seems reasonable and I like it! I think it's fine to include these changes (at on a first glance) since this PR already makes significant visual changes.

bhrutledge: Actually, I think the fact this PR makes significant visual changes is reason enough to do this in a subsequent PR. Let's just update the slashes for now.

bhrutledge (approved):
Thanks!

source/guides/installing-using-pip-and-virtual-environments.rst
bhrutledge: I've created https://github.com/pypa/packaging.python.org/issues/931 as a follow-up.

henryiii:
There there a way to avoid the syntax highlighting doing this with the backslashes?

Screen Shot 2021-06-28 at 11 24 44 AM

henryiii:
I think you can use the `.. code-block:: bat` lexer. This problem happens on other pages, too, not something to fix right now, but good idea for later.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -134,33 +132,34 @@ need to *activate* it. Activating a virtual environment will put the
...
-On Windows:
+        which python
+        .../env/bin/python
 
-.. code-block:: bash
+.. tab:: Windows
 
-    where python
-    .../env/bin/python.exe
+    .. code-block:: bash
```
bhrutledge, line 159: Looking at the other Windows output in this doc (e.g. line 143), I think this should address https://github.com/pypa/packaging.python.org/pull/928#issuecomment-869780161:

```suggestion
    .. code-block:: text
```

bhrutledge (changes requested):
One more small tweak, based on @henryiii's observation.

source/guides/installing-using-pip-and-virtual-environments.rst
ichard26: Committed, thanks!

bhrutledge (approved):

pradyunsg (approved):

ichard26:
Thanks @henryiii, @pradyunsg and @bhrutledge for the reviews, it's well appreciated! :slightly_smiling_face:

bhrutledge:
@ichard26 Thank for the improvements! If you feel like doing more, https://github.com/pypa/packaging.python.org/issues/931 would be a good follow-up.

Issue #804, opened 2020-12-14
GitHub Actions tutorial: Problem with equal version names when uploading to test.pypi


JaroCamphuijsen:
When following this guide I ran into a problem related to https://github.com/pypa/packaging.python.org/issues/777.

The point is that you cannot upload packages to pypi (or test.pypi) with the same (version) name. This definitely happens when uploading to test.pypi on every push. The workflow from this tutorial therefore fails even when running it for a second time with the following message:

`This filename has already been used, use a different version.`

A solution is given by @bhazelton with the script from https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/931:  https://gist.github.com/plaplant/0902f09e59166bac742bbd554f3cd2f9/raw/make_dev_version.sh

However this raises another issue (the one of filling up test pypi space) which is reported in https://github.com/pypa/packaging.python.org/issues/777.

Another solution (as implemented in the same PR mentioned above) is to only publish on test.pypi for tagged versions, same as for pypi. However this defeats the purpose of publishing to test.pypi as the same package is also directly published to pypi.

bhazelton:
Agreed. We worked hard to fix the version issue, which solved it for a time, until we filled up the space on test pypi. I don't see much purpose in using test pypi at all at this point. If it either allowed me to mark versions as temporary somehow so they were auto-deleted after a period of time or allowed me to do batch deletions it would be very helpful. I really like the idea of pushing to test pypi on every push to make sure nothing is breaking -- that's very much in line with standard CI practice, but it currently does not work.

JaroCamphuijsen:
Another possible solution I can think of would be to only publish tagged github releases with "rc" (release candidate) in the in the name. In that way you can still test your release by making release candidates with version tags like "v1.2.0rc2" and then only publish the versions without "rc" to pypi. 

Of course this all depends on the release workflow which might differ per project. However I think the way it's now shown in the instructions does not work nicely out of the box. So would be great to change this to something that does not have these issues or add some disclaimers and possible solutions.

webknjaz:
Agreed. I didn't demonstrate how to solve the version clash there. I wanted to start with something simple but clearly, the tutorial should be improved. Personally, I rely on setuptools-scm to generate versions and I apply "no local version part" strategy to upload per commit (example: https://github.com/ansible/pylibssh/blob/a648cd7/.github/workflows/build-test-n-publish.yml#L337-L341).

[...]

Issue #777, opened 2020-09-29
GitHub Actions tutorial: Warn about filling up test pypi package limit if you publish every commit as suggested 


bhazelton:
I followed the guide at https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ and got everything working so that every push to our main branch was published to test.pypi and only actual releases were published to pypi. We were very happy for several months until the push to test.pypi started erroring because we had hit the project size limit. I went to the package page on test.pypi and discovered (unsurprisingly in retrospect) hundreds of releases published on test.pypi that accumulated to put us at our package size limit.

As far as I can tell, there does not seem to be a way either to set time limits on these releases so they are auto-deleted or to batch delete them in any way. I had to go through a very laborious process of clicking on the options button for each unwanted release, selecting `Delete` and then copying the release name into the the verification box. Please add some text to the tutorial warning users of the potential ramifications of publishing every push to test.pypi, which is what is currently suggested in the tutorial. Or if there are better ways to auto-delete or otherwise manage such test releases, please add language explaining how to do better.

webknjaz:
@di @pradyunsg could you please mark this as "good first issue" / "Hacktoberfest"?

webknjaz:
If anyone is to submit the PR, I think it's worth using an "attention" box and pointing to https://pypi.org/help/#project-size-limit from the text.

webknjaz:
Another thing such a warning could suggest is using a custom PyPI-compatible server in CI just for testing.

PR #929, opened 2021-06-27
Suggest restricting to main in GHA publishing guide

### LINKED ISSUE #744: Suggest branch restriction in GitHub Actions tutorial
I was working through the guide at https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ and made a mistake by creating a workflow that published any time a push is made.

The guide currently suggests:

```yaml
on: push
```

But this means that the release job will run on any push on any branch.

I think instead, we should advise to restrict the job to the master branch:

```yaml
on:
  push:
    branches:
      - master
```

ichard26:
Previously the guide suggested workflow would run on every push which
isn't ideal (eg. a push on for an in-development feature branch doesn't need
a release).

Fixes #744.

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -1,6 +1,9 @@
 name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
-on: push
+on:
+  push:
+    branches:
+      - main
 
 jobs:
   build-n-publish:
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -1,9 +1,6 @@
 name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
-on:
-  push:
-    branches:
-      - main
+on: push
 
 jobs:
   build-n-publish:
@@ -32,6 +29,11 @@ jobs:
         .
     # Actually publish to PyPI/TestPyPI
     - name: Publish distribution 📦 to Test PyPI
+      if: >-
+        github.event_name == 'push' &&
+        github.ref == format(
+          'refs/heads/{0}', github.event.repository.default_branch
+        )
       uses: pypa/gh-action-pypi-publish@master
       with:
         password: ${{ secrets.TEST_PYPI_API_TOKEN }}
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -65,6 +65,11 @@ should make GitHub run this workflow:
    :language: yaml
    :end-before: jobs:
 
+The workflow only runs on push events for the ``main`` branch to avoid
+unnecessary publishes, like on a feature branch. If your primary
+branch is named something else, like ``master``, please replace ``main`` with
+the correct branch.
+
 
 Defining a workflow job environment
 ===================================
@@ -75,7 +80,7 @@ In this guide, we'll use Ubuntu 18.04:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: on:
+   :start-after: - main
    :end-before: steps:
 
 
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -65,11 +65,6 @@ should make GitHub run this workflow:
    :language: yaml
    :end-before: jobs:
 
-The workflow only runs on push events for the ``main`` branch to avoid
-unnecessary publishes, like on a feature branch. If your primary
-branch is named something else, like ``master``, please replace ``main`` with
-the correct branch.
-
 
 Defining a workflow job environment
 ===================================
@@ -80,7 +75,7 @@ In this guide, we'll use Ubuntu 18.04:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: - main
+   :start-after: on:
    :end-before: steps:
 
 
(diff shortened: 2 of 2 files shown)


ichard26:
Ugh, made the classic "I'm too used to Markdown" error :facepalm:

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -1,6 +1,9 @@
 name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
-on: push
+on:
+  push:
+    branches:
+      - main
```
webknjaz, line 6: I think this change is out of scope. I'd rather not assume the name of the default branch in a random user's repo.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -75,7 +80,7 @@ In this guide, we'll use Ubuntu 18.04:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: on:
+   :start-after: - main
```
webknjaz, line 83: It's probably better to do this
```suggestion
   :start-at: jobs:
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -65,6 +65,11 @@ should make GitHub run this workflow:
    :language: yaml
    :end-before: jobs:
 
+The workflow only runs on push events for the ``main`` branch to avoid
+unnecessary publishes, like on a feature branch. If your primary
+branch is named something else, like ``master``, please replace ``main`` with
```
webknjaz, line 70: Do we really want to go into teaching users more about Git/GitHub and default branches? I feel like this may shift the focus from the main goal of this guide.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
webknjaz: Also, this is overly indented and doesn't match the indentation style of the rest of this file. It shouldn't add extra indent in the sequences.

webknjaz (commented):
Looking closer at this change, it seems to break the tag publishing support. The assumption was that the tagged releases would go to both TestPyPI and PyPI but the normal pushes would only go to TestPyPI. By only allowing one hardcoded branch, it basically makes the publish step useless.
FTR I've been meaning to rewrite some aspects of this guide and even write a separate one that would be more advanced but never got to it...

webknjaz:
> which
> isn't ideal (eg. a push on for an in-development feature branch doesn't need
> a release).

I think I lean towards having an on-step if-clause. The build dist smoke test should always run. Also, the universal default branch check could look like https://github.com/ansible-community/ansible-pygments/blob/e4830fa/.github/workflows/ci-cd.yml#L155-L157.

ichard26:
Oh dear, it looks like there's a lot of work to do done here to even make this half-acceptable. Thank you for the feedback and my apologies for wasting your time.

Is this even worth pursing? I think by the point I fix the main issues, this would have lost most if not all of its value.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
ichard26: Yeah, this wasn't a good call. Agreed that's too much of a distraction.

webknjaz:
@ichard26 please don't apologize, you did nothing wrong :) Your contribution is still valuable and seeing how I constantly fail to find time to do the improvements myself, I'd say that we should follow the principle of "better now than never" and try to accept any incremental improvements where possible...

I think it'd be useful to have that if-clause on the "publish to TestPyPI" step in this form:
```yaml
if: >-
  github.event_name == 'push' &&
  github.ref == format(
    'refs/heads/{0}', github.event.repository.default_branch
  )
```
But keep the original "on:" trigger and instead of a long explanation about the branches, I'd just add a `.. tip::` admonition mentioning that the conditional is needed because other branches may trigger this + producing unique per-commit versions would require more work not currently covered in this guide — this is because if the version is hardcoded in a file, different pushes (merges to main) would attempt publishing the same version to TestPyPI effectively failing because of the immutability checks on the server side.

ichard26:
OK, that seems reasonable. Y'know I would expected to \*not\* make this many errors on such a simple issue, but honestly knowing me, this doesn't surprise me. Also, I definitely should've tested the workflow so I understood it better (that probably would've alerted me the fundamental issue with the suggested event change).

Thanks again, I appreciate the encouragement! Let's see if I can get a hold of it on attempt no. 2 ...

P.S. as a maintainer myself, isn't weird how somehow we end pouring more time we think we have when the changes are from someone else? I am acutely aware at how much time can be saved if I just fix an issue myself instead of letting and guiding someone else to do it :)

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
ichard26: Not relevant anymore but noted! Thank you!

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -29,6 +29,11 @@ jobs:
         .
     # Actually publish to PyPI/TestPyPI
     - name: Publish distribution 📦 to Test PyPI
+      if: >-
+        github.event_name == 'push' &&
```
bhrutledge, line 33: Is this redundant, given that the worfklow is already `on: push`?

```suggestion
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -29,6 +29,11 @@ jobs:
...
+        github.ref == format(
+          'refs/heads/{0}', github.event.repository.default_branch
+        )
```
bhrutledge, line 36: @webknjaz I'm not entirely following your earlier suggestions. I think I understand that this will only publish to TestPyPI on a push to `main`. However, it seems like this will still publish to PyPI on any tags of any branch, regardless of whether or not they indicate a release. That doesn't feel desirable to me, and I don't think it addresses the issue reported by @di in https://github.com/pypa/packaging.python.org/issues/744.

Without looking at the GH Actions syntax, could the `if:` be moved to a higher scope, i.e. coupled with the `on:`?

Also, this feels related to https://github.com/pypa/packaging.python.org/issues/777 and https://github.com/pypa/packaging.python.org/issues/804

PR #930, opened 2021-06-28
Translation Plan

### LINKED ISSUE #927: Translation Plan
Hi,

I would like to ask if there is any translation plan of packaging.python.org?

meowmeowmeowcat:
Fixes #927


.gitignore
```diff
@@ -3,3 +3,5 @@ build/
 .nox
 *.pyc
 __pycache__
+.DS_Store
+source/_build/
```
webknjaz, line 7: This is unnecessary — our `nox` setup uses `build/` dir. We shouldn't ignore every possible dir name contributors may use — the only supported sphinx invocation is via `nox`. And any helper commands should be added there too (to `noxfile.py`).

source/conf.py
```diff
@@ -82,7 +82,9 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
+locale_dirs = ['locales/']
 language = None
+gettext_compact = "messages"
```
webknjaz, line 87: Could you explain what this does?

webknjaz (commented):
I'm pretty sure we only need a `pot` file with the translation. Weblate will generate the target po files by itself. Let's not publish content with no translation at all. This is the process Warehouse follows too.

.gitignore
meowmeowmeowcat: OK

I will remove them

source/conf.py
meowmeowmeowcat: At first, if we run a command like this : `sphinx-intl update -p _build/gettext -l es_AR`, it will generate a few of `.po` files.

However, when we use `gettext_compact = "messages"` here, it will generate a new file called `messages.po` only.

meowmeowmeowcat: > I'm pretty sure we only need a `pot` file with the translation. Weblate will generate the target po files by itself. Let's not publish content with no translation at all. This is the process Warehouse follows too.

Okay. So now I have to remove `locales` right?

meowmeowmeowcat:
Okay. So now I have to remove locales right?

meowmeowmeowcat:
I have removed `locales` folder.

.gitignore
```diff
@@ -3,3 +3,4 @@ build/
 .nox
 *.pyc
 __pycache__
+.DS_Store
```
meowmeowmeowcat, line 6: I added .DS_Store here because I think these files are useless.

webknjaz: Sure, this one is okay.

webknjaz (changes requested):
Please don't commit `source/_build/` folder.

.gitignore
meowmeowmeowcat: And also do we need `.doctree` actually?

source/conf.py
webknjaz: Sounds like we need to embed invoking `sphinx-intl` into `noxfile.py`.

webknjaz:
> I have removed `locales` folder.

The `pot` file is still necessary, I think.

meowmeowmeowcat:
Yes, and it is located in `source/_build/` folder

.gitignore
webknjaz: No, we don't. The reason it appeared in your repo is that you used a different build directory that is not gitignored. Use the correct directory called `build/` and things under it will be ignored.

meowmeowmeowcat:
So maybe I should put `messages.pot` in  a new folder?

webknjaz:
> So maybe I should put `messages.pot` in a new folder?

I think it should probably be in `locales/`.

webknjaz:
Are you familiar with https://nox.thea.codes? I suggest you start by adding a new command there that invokes i18n-related executables. Use the same directories as in other commands, and hopefully, the rest will fall into place.

meowmeowmeowcat:
> Are you familiar with https://nox.thea.codes? I suggest you start by adding a new command there that invokes i18n-related executables. Use the same directories as in other commands, and hopefully, the rest will fall into place.

I am not really familiar with it, but I will try and see if I can make it.

noxfile.py
```diff
@@ -41,3 +41,10 @@ def build(session, autobuild=False):
 def preview(session):
     session.install("sphinx-autobuild")
     build(session, autobuild=True)
+
+@nox.session(py="3")
+def translation(session):
+    session.install("sphinx-intl")
+    session.install("sphinx-inline-tabs")
+    session.run("cd","source")
+    session.run("make","gettext")
```
meowmeowmeowcat, line 50: Is something like these?

meowmeowmeowcat (commented):
Therefore, people will be required to run these (build the new `.pot`)after they have changed the docs

noxfile.py
```diff
@@ -6,6 +6,12 @@
 import shutil
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("sphinx-intl")
+    session.install("sphinx-inline-tabs")
+    session.run("cd","source")
+    session.run("make","gettext")
```
webknjaz, line 14: Let's not use `make` but invoke the commands directly instead.

noxfile.py
```diff
@@ -6,6 +6,12 @@
 import shutil
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("sphinx-intl")
+    session.install("sphinx-inline-tabs")
+    session.run("cd","source")
```
webknjaz, line 13: I don't think we should `cd` into `source/` — just follow what other commands do. I'm pretty sure you could pass dirs as CLI args.

noxfile.py
meowmeowmeowcat: Okay

Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
webknjaz, line 1: Let's not add additional to nox command runner — we don't need make.

.gitignore
```diff
@@ -3,3 +3,5 @@ build/
...
+source/locales/template/gettext/.doctrees/
```
webknjaz, line 7: I'd move `locales` outside of `source/` maybe + could we put doctrees into the top-level `build/`?

Makefile
meowmeowmeowcat: Okay. I will remove it.

.gitignore
meowmeowmeowcat: But do I have to to commit `.doctree`?

noxfile.py
meowmeowmeowcat: OK

meowmeowmeowcat:
I also have a question

webknjaz:
> How many `.pot` files I have to generate? Join all `.pot` files in one or just let them separate out?

I'm not sure, I was about to ask you the same :) Originally, I was thinking about one but OTOH maybe having a per-article file could be useful. Is Weblate able to handle multiple files as a single project?

webknjaz:
I think, if we'll be able to make the workflow Weblate-centric, we would care less about the local commands as most folks would just use that.

meowmeowmeowcat:
> > How many `.pot` files I have to generate? Join all `.pot` files in one or just let them separate out?
> 
> I'm not sure, I was about to ask you the same :) Originally, I was thinking about one but OTOH maybe having a per-article file could be useful. Is Weblate able to handle multiple files as a single project?

I think Weblate is able to handle multiple files by [add-on](https://docs.weblate.org/en/latest/admin/addons.html#addon-weblate-discovery-discovery).

meowmeowmeowcat:
> > How many `.pot` files I have to generate? Join all `.pot` files in one or just let them separate out?
> 
> I'm not sure, I was about to ask you the same :) Originally, I was thinking about one but OTOH maybe having a per-article file could be useful. Is Weblate able to handle multiple files as a single project?

I noticed that Warehouse has a file called `messages.pot` only, which contains all contents they have to translate in a file.

meowmeowmeowcat:
So I think we should run this command in local: `xgettext locales/*.pot locales/discussions/*.pot locales/guides/*.pot locales/specifications/*.pot locales/tutorials/*.pot -o messages.pot` to generate one `.pot` file. ( It will be contained in the `noxfile.py` later)

meowmeowmeowcat:
> So I think we should run this command in local: `xgettext locales/*.pot locales/discussions/*.pot locales/guides/*.pot locales/specifications/*.pot locales/tutorials/*.pot -o messages.pot` to generate one `.pot` file. ( It will be contained in the `noxfile.py` later)

But I think this command is quite long.

source/conf.py
meowmeowmeowcat: > Sounds like we need to embed invoking `sphinx-intl` into `noxfile.py`.

Added.

meowmeowmeowcat:
And I've changed Sphinx version to `3.3.0` in `requirements.txt`

meowmeowmeowcat:
According to this [pull request](https://github.com/meowmeowmeowcat/packaging.python.org/pull/1), I think we can move to next step

meowmeowmeowcat:
Now we can go on to the next step.

meowmeowmeowcat:
I think this pull request can be merged now.

PR #933, opened 2021-06-28
chore: add multibuild

henryiii:
This adds multibuild, pre-req for #910. CC @matthew-brett - better description welcome (we don't have much of a description on cibuildwheel either, @pypa/cibuildwheel-team).

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -55,7 +55,7 @@ cibuildwheel
 `Discussions <https://github.com/pypa/cibuildwheel/discussions>`__ |
 `Discord #cibuildwheel <https://discord.com/invite/pypa>`__
 
-``cibuildwheel`` is a Python package that builds :term:`wheels <Wheels>` for all common
+``cibuildwheel`` is a Python package that builds :term:`wheels <Wheel>` for all common
 platforms and Python versions on most CI systems. Also see :ref:`multibuild`.
 
 .. _distlib:
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -55,7 +55,7 @@ cibuildwheel
 `Discussions <https://github.com/pypa/cibuildwheel/discussions>`__ |
 `Discord #cibuildwheel <https://discord.com/invite/pypa>`__
 
-``cibuildwheel`` is a Python package that builds wheels for all common
+``cibuildwheel`` is a Python package that builds :term:`wheels <Wheels>` for all common
 platforms and Python versions on most CI systems. Also see :ref:`multibuild`.
 
 .. _distlib:
@@ -464,7 +464,7 @@ multibuild
 
 `GitHub <https://github.com/matthew-brett/multibuild>`__
 
-Multibuild is a set of CI scripts for building and testing Python Wheels for
+Multibuild is a set of CI scripts for building and testing Python :term:`wheels <Wheel>` for
 Linux, macOS, and (less flexibly) Windows. Also see :ref:`cibuildwheel`.
 
 .. _pex:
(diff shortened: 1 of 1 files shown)


source/key_projects.rst
```diff
@@ -55,9 +55,8 @@ cibuildwheel
 `Discussions <https://github.com/pypa/cibuildwheel/discussions>`__ |
 `Discord #cibuildwheel <https://discord.com/invite/pypa>`__
 
-``cibuildwheel`` is a package that builds wheels for all common platforms and
-Python versions on most CI systems.
-
+``cibuildwheel`` is a Python package that builds wheels for all common
```
bhrutledge, line 58:
```suggestion
``cibuildwheel`` is a Python package that builds :term:`wheels <Wheels>` for all common
```

source/key_projects.rst
```diff
@@ -458,6 +457,16 @@ dependencies for, and publish packages to PyPI. Under the hood, it
...
 install packages.
 
+.. _multibuild:
+
+multibuild
+==========
+
+`GitHub <https://github.com/matthew-brett/multibuild>`__
+
+Multibuild is a set of CI scripts for building and testing Python Wheels for
```
bhrutledge, line 467:
```suggestion
Multibuild is a set of CI scripts for building and testing Python :term:`wheels <Wheel>` for
```

bhrutledge (commented):
This might be pedantic, but "wheel" seems worth defining.

bhrutledge:
Also, FWIW, I think it would have been okay to add this to #910.

henryiii:
> Also, FWIW, I think it would have been okay to add this to #910.

I've [been told before](https://github.com/pypa/packaging.python.org/pull/823#issuecomment-774536521) by @webknjaz  it's better to keep things separate, so I've tried to do that. If one part blocks, the other can still go in.

source/key_projects.rst
```diff
@@ -55,9 +55,8 @@ cibuildwheel
 `Discussions <https://github.com/pypa/cibuildwheel/discussions>`__ |
 `Discord #cibuildwheel <https://discord.com/invite/pypa>`__
 
-``cibuildwheel`` is a package that builds wheels for all common platforms and
-Python versions on most CI systems.
-
+``cibuildwheel`` is a Python package that builds :term:`wheels <Wheels>` for all common
```
henryiii, line 58:
```suggestion
``cibuildwheel`` is a Python package that builds :term:`wheels <Wheel>` for all common
```

bhrutledge (approved):
> it's better to keep things separate,

Agreed. In this case, though, it seems like these changes are tightly-coupled to the changes in #910, and IMHO would be more easily reviewed together. I don't think that would fall under @webknjaz's qualifications for breaking it up:

> I find it easier to review/approve separate unrelated changes so if one of the updates wouldn't block others from being merged. Usually when the title contains references to many things, has "and" in it, or is too vague within the 50 char limit, it's a sign that the PR is too broad and when you can describe its separate parts with a few specific titles, it's a good idea to have a PR per such a title. 

But I'm just being nit-picky while learning the ropes as a gardener for this project. 😄

pradyunsg:
> I've [been told before](https://github.com/pypa/packaging.python.org/pull/823#issuecomment-774536521) by @webknjaz it's better to keep things separate, so I've tried to do that.

IMO, this is a balancing act. It's possible to go too-far with breaking up PRs -- having a bias toward smaller is good (easier to review + iterate), as long as there's no dependency between the various PRs. If there's dependencies in the changes, it's much better to just combine the changes into a single PR that can be reviewed as a whole.

pradyunsg (approved):
Thanks @henryiii for the PR and @bhrutledge for the thoughtful reviews! ^>^

PR #935, opened 2021-07-03
Generate `messages.pot` as translation template

### LINKED ISSUE #927: Translation Plan

meowmeowmeowcat:
- Refs #927 
- Generate `messages.pot` in `./locales` when running `nox -s translation` in the terminal
- Change the version of sphinx from `3.2.0` to `3.3.0`

#930 is too messy and I've closed it.

Thanks for review! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ build/
 .nox
 *.pyc
 __pycache__
+.DS_Store
+locales/.doctrees/
--- a/noxfile.py
+++ b/noxfile.py
@@ -13,9 +13,8 @@ def translation(session):
     session.run(
         "sphinx-build", 
         "-b", "gettext",  # build gettext-style message catalogs (.pot file)
+        "-d", ".nox/.doctrees/", # path to put the cache
         "source/",  # where the rst files are located
-        "-d", 
-        ".nox/.doctrees/", # path to put the cache
         target_dir, # where to put the .pot file
     )
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -8,8 +8,7 @@ import nox
 
 @nox.session(py="3")
 def translation(session):
-    session.install("sphinx-intl")
-    session.install("sphinx-inline-tabs")
+    session.install("-r", "requirements.txt")
     target_dir = "locales"
     session.run(
             "sphinx-build", 
@@ -17,6 +16,8 @@ def translation(session):
             "gettext",  # build gettext-style message catalogs (.pot file)
             "source/",  # where the rst files are located
             target_dir, # where to put the .pot file
+            "-d", 
+            ".nox/", # path to put the cache
         )
 
 @nox.session(py="3")
--- a/noxfile.py
+++ b/noxfile.py
@@ -12,12 +12,11 @@ def translation(session):
     target_dir = "locales"
     session.run(
         "sphinx-build", 
-        "-b",  # select a builder
-        "gettext",  # build gettext-style message catalogs (.pot file)
+        "-b", "gettext",  # build gettext-style message catalogs (.pot file)
         "source/",  # where the rst files are located
-        target_dir, # where to put the .pot file
         "-d", 
-        ".nox/", # path to put the cache (.doctrees)
+        ".nox/.doctrees/", # path to put the cache
+        target_dir, # where to put the .pot file
     )
 
 @nox.session(py="3")
--- a/noxfile.py
+++ b/noxfile.py
@@ -6,6 +6,18 @@
 import shutil
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("sphinx-intl")
+    session.install("sphinx-inline-tabs")
+    target_dir = "locales"
+    session.run(
+            "sphinx-build", 
+            "-b",  # select a builder
+            "gettext",  # build gettext-style message catalogs (.pot file)
+            "source/",  # where the rst files are located
+            target_dir, # where to put the .pot file
+        )
 
 @nox.session(py="3")
 def build(session, autobuild=False):
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==3.2.0
+sphinx==3.3.0
 sphinx-autobuild==0.7.1
 sphinx-inline-tabs==2021.4.11b9
 git+https://github.com/python/python-docs-theme.git#egg=python-docs-theme
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,10 +26,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade nox virtualenv
-        
-      - name: Nox translation
-        run: |
-          python -m nox -s translation
 
       - name: Nox build
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade nox virtualenv
+        
+      - name: Nox translation
+        run: |
+          python -m nox -s translation
 
       - name: Nox build
         run: |
--- a/locales/messages.pot
+++ b/locales/messages.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Python Packaging User Guide \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-07-03 21:32+0800\n"
+"POT-Creation-Date: 2021-07-06 12:35+0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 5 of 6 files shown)


pradyunsg:
What is the translation workflow, with this PR now?

meowmeowmeowcat:
> What is the translation workflow, with this PR now?

Umm I suddenly think that we probably don't need this translation workflow.

meowmeowmeowcat:
1. After this pull requests is merged, the maintainers of pypa (@ewdurbin or @di) have to put the `messages.pot` on Weblate for people to translate
2. Add multiple languages version on packaging.python.org (I think it can be done in the admin page of readthedocs)

meowmeowmeowcat:
We should also write some guidelines of translation for contributors too.

meowmeowmeowcat:
yes, this title is more suitable.

meowmeowmeowcat:
> What is the translation workflow, with this PR now?

The `messages.pot` generated by sphinx-intl is a template for translations.

Therefore, people will be able to translate the docs into their languages.

meowmeowmeowcat:
This PR is just for generating `messages.pot` actually.

.gitignore
```diff
@@ -3,3 +3,5 @@ build/
 .nox
 *.pyc
 __pycache__
+.DS_Store
+locales/.doctrees/
```
webknjaz, line 7: It looks like the current automation misplaces `.doctrees`. This folder should be shared with all the other commands and probably live at `build/.doctrees/` which is already gitignored. Alternatively, it could live under `.nox/` which would make the quick-and-dirty env reset/cleanup/cache invalidation easier (by using `rm -rf .nox`).

noxfile.py
```diff
@@ -6,6 +6,18 @@
 import shutil
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("sphinx-intl")
+    session.install("sphinx-inline-tabs")
```
webknjaz, line 12: This should probably reuse the same dep versions as the rest of the commands.
```suggestion
    session.install("-r", "requirements.txt")
```

Also, maybe we'd just add `sphinx-intl` into `requirements.txt` to avoid having it in just one place? Is it unnecessary in all the other envs?

noxfile.py
```diff
@@ -6,6 +6,18 @@
...
+    target_dir = "locales"
+    session.run(
+            "sphinx-build", 
+            "-b",  # select a builder
+            "gettext",  # build gettext-style message catalogs (.pot file)
+            "source/",  # where the rst files are located
+            target_dir, # where to put the .pot file
+        )
```
webknjaz, line 20: Let's align with the style of the rest of this Python module:
```suggestion
        "sphinx-build", 
        "-b",  # select a builder
        "gettext",  # build gettext-style message catalogs (.pot file)
        "source/",  # where the rst files are located
        target_dir, # where to put the .pot file
    )
```

requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx==3.2.0
+sphinx==3.3.0
```
webknjaz, line 1: This change is unrelated, please don't include it in this PR. Feel free to send a separate PR updating the sphinx version here. Maybe even use v4 there.

pradyunsg: This is necessary, for sphinx-intl compatibility.

pradyunsg: And I'd be much more careful around the Sphinx 4 upgrade -- there's a _lot_ of changes in Sphinx 4, and it's worth double checking that things look fine after rebuilding.

.gitignore
meowmeowmeowcat: The folder `.doctrees` is generated by `sphinx-build` automatically when it is generating `messages.pot`.

meowmeowmeowcat: So I will put it into `.nox/` folder by adding `-d .nox/` arguments now.

noxfile.py
meowmeowmeowcat: I think `sphinx-intl` is unnecessary now since Weblate will generate `.po` files for us when we add new languages to it.

requirements.txt
webknjaz: Sounds good.

noxfile.py
webknjaz: What about generating `messages.pot`? We still need to extract the original messages from time to time, right?

.gitignore
webknjaz: @meowmeowmeowcat yep. And I'd also explicitly set this arg in all the other session commands just to keep it consistent and make sure those can reuse the same cache.

requirements.txt
webknjaz: So now that sphinx-intl has been removed, we don't really need to update this in the current PR, right? Also, let's update the PR title now that it's clear that this extension isn't involved in generating `messages.pot`.

noxfile.py
```diff
@@ -6,6 +6,19 @@
...
+    session.install("-r", "requirements.txt")
+    target_dir = "locales"
+    session.run(
+        "sphinx-build", 
+        "-b",  # select a builder
+        "gettext",  # build gettext-style message catalogs (.pot file)
+        "source/",  # where the rst files are located
+        target_dir, # where to put the .pot file
+        "-d", 
+        ".nox/", # path to put the cache (.doctrees)
```
webknjaz, line 20: *I think* this should be
```suggestion
        ".nox/.doctrees/", # path to put the cache
```

noxfile.py
meowmeowmeowcat: yes, when people change the docs, they will have to run `nox -s translation` locally to update `messages.pot`.

noxfile.py
```diff
@@ -6,6 +6,19 @@
...
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("-r", "requirements.txt")
+    target_dir = "locales"
+    session.run(
+        "sphinx-build", 
+        "-b",  # select a builder
+        "gettext",  # build gettext-style message catalogs (.pot file)
```
webknjaz, line 16: Let's keep the option names coupled with their values
```suggestion
        "-b", "gettext",  # build gettext-style message catalogs (.pot file)
```

noxfile.py
webknjaz, line 20: *I think* this should be
```suggestion
        "-d", ".nox/.doctrees/", # path to put the cache
        target_dir, # where to put the .pot file
```
(plus putting target as the last positional arg to avoid confusion with options having values)

requirements.txt
meowmeowmeowcat: I think we still need to update sphinx to `3.3.0`. I have just tried with `3.2.0` again and this error

```Warning, treated as error:
The config value `gettext_compact' has type `str', defaults to `bool'.
``` 
 was given by `sphinx-build` , which caused `nox -s build` failed.

noxfile.py
webknjaz: I wonder if we could make a module-level constant tuple with shared args to reuse across all sessions. OTOH, this could be explored in a separate PR anyway.

requirements.txt
webknjaz: Ah, I see. Okay.

.gitignore
meowmeowmeowcat: Okay.

noxfile.py
meowmeowmeowcat: So maybe open another PR for this will be better? :)

meowmeowmeowcat:
So now I think it's ok for us to merge this PR  :)

Some other works, for example, updating the version of sphinx, writing guidelines for translation will be solved in other PRs.

meowmeowmeowcat:
All suggestions were applied!

noxfile.py
```diff
@@ -6,6 +6,18 @@
...
 
+@nox.session(py="3")
+def translation(session):
+    session.install("-r", "requirements.txt")
+    target_dir = "locales"
+    session.run(
+        "sphinx-build", 
+        "-b", "gettext",  # build gettext-style message catalogs (.pot file)
+        "source/",  # where the rst files are located
+        "-d",
```
webknjaz, line 17: Plz add the argument to this option on the same line

noxfile.py
```diff
@@ -6,6 +6,18 @@
...
 import nox
 
+@nox.session(py="3")
+def translation(session):
+    session.install("-r", "requirements.txt")
+    target_dir = "locales"
+    session.run(
+        "sphinx-build", 
+        "-b", "gettext",  # build gettext-style message catalogs (.pot file)
+        "source/",  # where the rst files are located
```
webknjaz, line 16: Move the source dir after the options to be consistent with other sessions

noxfile.py
meowmeowmeowcat: Oh, I think I have missed this. Sorry!

meowmeowmeowcat:
It should be corrected now.

webknjaz (approved):

meowmeowmeowcat:
Thanks for @webknjaz 's review!

PR #936, opened 2021-07-08
Update the version of Sphinx to 4.0.3

meowmeowmeowcat:
Things look fine after rebuilding.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==3.3.0
+sphinx==4.0.1
 sphinx-autobuild==0.7.1
 sphinx-inline-tabs==2021.4.11b9
 git+https://github.com/python/python-docs-theme.git#egg=python-docs-theme
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==4.0.1
+sphinx==4.0.3
 sphinx-autobuild==0.7.1
 sphinx-inline-tabs==2021.4.11b9
 git+https://github.com/python/python-docs-theme.git#egg=python-docs-theme


pradyunsg:
To whomsoever reviews this -- be careful here: Sphinx 4 changed a whole bunch of things around the HTML output and you'd likely want to look at each of the pages where there's directives / tables / images etc and make sure that they look OK.

bhrutledge:
Thanks for the heads up, @pradyunsg.

@meowmeowmeowcat This sounds like a big upgrade, especially given that the primary purpose of this project is to be a documentation site. I'm not a maintainer, but I assume they'll want more information about the benefits of doing this upgrade, and potential breaking changes. It would also be helpful to link to the Sphinx changelog for quick reference.

webknjaz:
@bhrutledge upgrading things usually lets us get the latest features and bugfixes, I don't see why we wouldn't want this. Honestly, I'd add a dependabot setup to let it deal with regular autoupdates.

I've updated to Sphinx 4 in multiple places and the only issue was with a lexer init in a custom extension that got resolved easily.

FWIW I've already green-lighted the upgrade during the review of another PR.

P.S. The easiest comparison would probably be generating two docs dirs with different sphinx versions and running `diff -ur dir1 dir2` against that.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx==3.3.0
+sphinx==4.0.1
```
webknjaz, line 1: Let's use the latest bugfix release, no need to hold back:
```suggestion
sphinx==4.0.3
```

webknjaz:
@meowmeowmeowcat plz post the diff as mentioned above and this should clear the concerns folks have but I personally don't expect you to find anything problematic there.

requirements.txt
meowmeowmeowcat: @webknjaz is it the best choice to use v4.0.3? I think it is still in development.

meowmeowmeowcat:
@bhrutledge The Sphinx changelog for v4.0.3 is [here](https://www.sphinx-doc.org/en/master/changes.html?highlight=4.0.3#release-4-0-3-in-development).

meowmeowmeowcat:
> @meowmeowmeowcat plz post the diff as mentioned above and this should clear the concerns folks have but I personally don't expect you to find anything problematic there.

Do you mean that I have to build two different version (one is 3.3.0 and one is 4.0.3) and then compare the diff between them?

meowmeowmeowcat:
> @meowmeowmeowcat plz post the diff as mentioned above and this should clear the concerns folks have but I personally don't expect you to find anything problematic there.

@webknjaz I've tested and the only diff is their version number:
```
-    Created using <a href="https://www.sphinx-doc.org/">Sphinx</a> 3.3.0.
+    Created using <a href="https://www.sphinx-doc.org/">Sphinx</a> 4.0.3.
```

webknjaz (approved):

meowmeowmeowcat:
Thanks @webknjaz!

PR #937, opened 2021-07-09
Change to `code-block:: text` from `code-block:: bash` & update the version of pip

### LINKED ISSUE #931: installing-using-pip: Update command output
As noted in in https://github.com/pypa/packaging.python.org/pull/928#discussion_r659366032, there are few places in that document where the command and its output are shown in a single `code-block`, but with no distinction between them, which might lead folks to thinking that `pip 9.0.1 ...` is a command to execute. If the output of a command is valuable, I recommend putting it into a separate `.. code-block:: text`.

Also, the output from the `pip --version` is quite old. Let's take this opportunity to update it.

### LINKED ISSUE #934: The windows examples use "bash" and backslashes
This creates weird highlighting especially with backslashes in paths. I think we can use the `.. code-block:: bat` lexer instead.

_Originally posted by @henryiii in https://github.com/pypa/packaging.python.org/pull/928#issuecomment-869787000_

meowmeowmeowcat:
- Fixes #934 
- Fixes #931

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -930,7 +930,7 @@ it will be automatically created when the first release is uploaded.
 For the second and subsequent releases, PyPI only requires that the
 version number of the new release differ from any previous releases.
 
-.. code-block:: text
+.. code-block:: bash
 
     twine upload dist/*
 
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -36,7 +36,7 @@ Requirements for packaging and distributing
 
     .. tab:: Windows
 
-        .. code-block:: bash
+        .. code-block:: text
 
             py -m pip install twine
 
@@ -728,7 +728,7 @@ Before you can build wheels and sdists for your project, you'll need to install 
 
 .. tab:: Windows
 
-    .. code-block:: bash
+    .. code-block:: text
 
         py -m pip install build
 
@@ -747,7 +747,7 @@ Minimally, you should create a :term:`Source Distribution <Source Distribution (
 
 .. tab:: Windows
 
-    .. code-block:: bash
+    .. code-block:: text
 
         py -m build --sdist
 
@@ -804,7 +804,7 @@ To build the wheel:
 
 .. tab:: Windows
 
-    .. code-block:: bash
+    .. code-block:: text
 
         py -m build --wheel
 
@@ -833,7 +833,7 @@ To build the wheel:
 
 .. tab:: Windows
 
-    .. code-block:: bash
+    .. code-block:: text
 
         py -m build --wheel
 
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -36,7 +36,7 @@ Requirements for packaging and distributing
 
     .. tab:: Windows
 
-        .. code-block:: text
+        .. code-block:: bat
 
             py -m pip install twine
 
@@ -728,7 +728,7 @@ Before you can build wheels and sdists for your project, you'll need to install 
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install build
 
@@ -747,7 +747,7 @@ Minimally, you should create a :term:`Source Distribution <Source Distribution (
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m build --sdist
 
@@ -804,7 +804,7 @@ To build the wheel:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m build --wheel
 
@@ -833,7 +833,7 @@ To build the wheel:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m build --wheel
 
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -38,7 +38,7 @@ installed.
 
         python3 -m pip --version
 
-    .. code-block:: bash
+    .. code-block:: text
 
         pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
 
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -167,7 +167,7 @@ It should be in the ``env`` directory:
 
 .. tab:: Unix/macOS
 
-    .. code-block:: text
+    .. code-block:: bash
 
         .../env/bin/python
 
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -158,6 +158,9 @@ Python interpreter, it should point to the ``env`` directory.
     .. code-block:: bash
 
         which python
+    
+    .. code-block:: text
+
         .../env/bin/python
 
 .. tab:: Windows
@@ -165,6 +168,9 @@ Python interpreter, it should point to the ``env`` directory.
     .. code-block:: text
 
         where python
+    
+    .. code-block:: text
+        
         ...\env\Scripts\python.exe
 
 As long as your virtual environment is activated pip will install packages into that
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -48,7 +48,7 @@ installed.
     The Python installers for Windows include pip. You can make sure that pip is
     up-to-date by running:
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install --upgrade pip
 
@@ -85,7 +85,7 @@ using pip.
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install --user virtualenv
 
@@ -115,7 +115,7 @@ in the below commands.
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m venv env
 
@@ -144,7 +144,7 @@ virtual environment-specific
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         .\env\Scripts\activate
 
@@ -159,7 +159,7 @@ Python interpreter:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         where python
     
@@ -173,7 +173,7 @@ It should be in the ``env`` directory:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
         
         ...\env\Scripts\python.exe
 
@@ -210,7 +210,7 @@ Now that you're in your virtual environment you can install packages. Let's inst
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install requests
 
@@ -249,7 +249,7 @@ a specific version of ``requests``:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install requests==2.18.4
 
@@ -263,7 +263,7 @@ To install the latest ``2.x`` release of requests:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install requests>=2.0.0,<3.0.0
 
@@ -277,7 +277,7 @@ To install pre-release versions of packages, use the ``--pre`` flag:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install --pre requests
 
@@ -296,7 +296,7 @@ specifying the extra in brackets:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         py -m pip install requests[security]
 
@@ -318,7 +318,7 @@ pip can install a package directly from source, for example:
 
 .. tab:: Windows
 
-    .. code-block:: text
+    .. code-block:: bat
 
         cd google-auth
         py -m pip install .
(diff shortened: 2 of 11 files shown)


pradyunsg:
Hmm... This PR make me think that we should not checkin the `.pot` file, and instead generate that in a GitHub action instead.

meowmeowmeowcat:
> Hmm... This PR make me think that we should not checkin the `.pot` file, and instead generate that in a GitHub action instead.

I will open another PR for this.

ed738 (approved):

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -36,7 +36,7 @@ Requirements for packaging and distributing
 
     .. tab:: Windows
 
-        .. code-block:: bash
+        .. code-block:: text
 
             py -m pip install twine
```
webknjaz, line 41: How about
```suggestion
        .. code-block:: console

            > py -m pip install twine
```

bhrutledge: @webknjaz Are you suggesting that for just this instance, or as a pattern to follow for all commands?

Assuming it's the latter, do you know how that renders Windows commands that contain `\`?

bhrutledge: Also, including a prompt character in the command would be a stylistic change from the Linux/macOS commands. As it is, readers can copy and paste the whole code block.

webknjaz: Yeah, I know. It's just a separate lexer that highlights lines with the prompt in a different color than the ones w/o it assuming that's the command output.
It'd need to be a global change. I don't think it really parses the command contents though.
The important thing is that it doesn't attempt to highlight things with `$` as vars in shell scripts or detect keywords like `for` or `if`.

webknjaz: I know that snippets on docs.python.org have a button that removes all the prompt chars for copying. That requires some sphinx ext.

bhrutledge: Looks like `console` is specifically for Bash sessions: https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer

Besides plain `text`, it looks like the Windows options are `bat` (suggested by @henryiii) and maybe `powershell`:

https://pygments.org/docs/lexers/#pygments.lexers.shell.BatchLexer
https://pygments.org/docs/lexers/#pygments.lexers.shell.PowerShellLexer

I'm honestly not sure which is best. @meowmeowmeowcat if you want to try one of those to see how it changes the rendering, go for it, but I'm guessing `text` is sufficient.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -37,7 +37,10 @@ installed.
     .. code-block:: bash
 
         python3 -m pip --version
-        pip 9.0.1 from $HOME/.local/lib/python3.6/site-packages (python 3.6)
+
+    .. code-block:: bash
+
+        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
```
bhrutledge, line 43: This isn't a `bash` command, but rather its output, so I think it should be `text`:

```suggestion
    .. code-block:: text

        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
```

bhrutledge: Also, I think these blocks under "Activating Virtual Environments" should similarly separate the command from its output:

```rst
.. tab:: Unix/macOS

    .. code-block:: bash

        which python
        .../env/bin/python

.. tab:: Windows

    .. code-block:: text

        where python
        ...\env\Scripts\python.exe
```

bhrutledge (changes requested):
Thanks for the comprehensive update! Beyond the question of `text` vs other lexers, just a couple some small tweaks to `installing-using-pip-and-virtual-environments.rst` for consistency.

source/guides/distributing-packages-using-setuptools.rst
meowmeowmeowcat: > I know that snippets on docs.python.org have a button that removes all the prompt chars for copying. That requires some sphinx ext.

@webknjaz Do you think that we should add this sphinx ext to our docs?

meowmeowmeowcat: > Looks like `console` is specifically for Bash sessions: https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer
> 
> Besides plain `text`, it looks like the Windows options are `bat` (suggested by @henryiii) and maybe `powershell`:
> 
> https://pygments.org/docs/lexers/#pygments.lexers.shell.BatchLexer
> https://pygments.org/docs/lexers/#pygments.lexers.shell.PowerShellLexer
> 
> I'm honestly not sure which is best. @meowmeowmeowcat if you want to try one of those to see how it changes the rendering, go for it, but I'm guessing `text` is sufficient.

Thanks @bhrutledge! I will try them out.

meowmeowmeowcat: After testing `bat` ,`powershell` and `text`, I think there is no difference between them. So probably we should keep using `text`.

meowmeowmeowcat: > I know that snippets on docs.python.org have a button that removes all the prompt chars for copying. That requires some sphinx ext.

@webknjaz I think that button is for removing `>>>` in the interactive mode only.

meowmeowmeowcat: Maybe we should follow the docs of [pip](https://pip.pypa.io/en/stable/quickstart/)?

They use something like this for UNIX/macOS:

```
$ python -m pip install SomePackage
[...]
Successfully installed SomePackage
```

And this for Windows:
```
C:\> py -m pip install SomePackage
[...]
Successfully installed SomePackage
```

meowmeowmeowcat: I have an idea here.

We could use `Sphinx-copybutton` for folks to copy the command.
However, it won't copy `$` or `C:\>`.

Example:

Command:
```
$ python3 -m pip install --user --upgrade pip
```
After they copied, this command will be paste in their terminal:
```
python3 -m pip install --user --upgrade pip
```

bhrutledge: Thanks for the ideas. While I like the idea of following the conventions in other docs, I'm not sure pip is doing the right thing here. Looking at the [source for the quickstart doc](https://github.com/pypa/pip/blob/21.1.3/docs/html/quickstart.rst), it's using `console` for all tabs. Also, their copy button copies the prompt, command, and output. I think it's also worth noting that pip is in the process of migrating all of their docs to Markdown (https://github.com/pypa/pip/issues/9475), so trying to follow their patterns could be challenging.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -145,22 +149,35 @@ virtual environment-specific
...
+
+    .. code-block:: text
+
+        .../env/bin/python
+
+.. tab:: Windows
+
+    .. code-block:: text
+        
         ...\env\Scripts\python.exe
```
bhrutledge, line 178: I thought having more distinction between the command and its output would read better.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -32,30 +32,34 @@ installed.
...
 
-    .. code-block:: bash
+    Afterwards, you should have the latest version of pip installed in your
+    user site:
 
-        python3 -m pip --version
-        pip 9.0.1 from $HOME/.local/lib/python3.6/site-packages (python 3.6)
+    .. code-block:: text
+
+        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
```
bhrutledge, line 42: In retrospect, the changes to this section probably should be in a separate PR, but I took the opportunity to (hopefully) add some additional clarity.

bhrutledge (approved):
Thanks for the changes. I pushed a couple more tweaks to the `installing-using-pip` doc.

Since this would be my first merge as a maintainer, I'll leave this open for another day to allow for more comments.

source/guides/distributing-packages-using-setuptools.rst
meowmeowmeowcat: Therefore, our copy button won't copies the prompt but only command. We have also separated the output out so it won't copy the output too. Adding the prompt will make folks know that it is a command easily.

meowmeowmeowcat: So now we should just keep the same as I found that they will add the prompt when the code block includes some other text only.

source/guides/installing-using-pip-and-virtual-environments.rst
webknjaz: I'm pretty sure it's console rather than text.

henryiii: "Text" removes all semantic meaning from the highlighting, promising you will never do better than text, and making it indistinguishable from a text output. I'd think `bat` or `powershell` would be better, as that's actually the language this is in, even if there's no current visual difference. `console` is supposed to be somewhat agnostic, even if bash-like - was it rendering incorrectly? Also, if they are struggling with output vs. input, what about:

```rst
    .. code-block:: bat

        where python

    .. code-block:: text

        ...\env\Scripts\python.exe
```

? Does that break it up too much? I used to have a "input" and "output" type (gitbook plugin, IIRC) that would handle this nicely. As long as we are consistent, it's easy to refactor in the future - going to "text" makes it harder, because there may really be some "text" blocks that are not Windows shell inputs.

bhrutledge: Those are good points re: semantics and refactoring, @henryiii. Since I was the one who originally suggested `text`, I just pushed the change to `bat` for Windows inputs (4e01f46), plus a couple remaining `bash` for Unix inputs (ee40d83).

I don't think `console` is correct for outputs, because [according to the docs](https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer), it's a "lexer for Bash shell sessions, i.e. command lines, including a prompt, interspersed with output". `text` is widely-used throughout this project for outputs.

meowmeowmeowcat:
Thanks for @henryiii's suggestion and @bhrutledge's commits!

meowmeowmeowcat:
@bhrutledge Do you think we should add copy button for folks so they can copy the command directly? It's very easy to do this( just add a sphinx extension).

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -36,7 +36,7 @@ Requirements for packaging and distributing
 
     .. tab:: Windows
 
-        .. code-block:: bash
+        .. code-block:: bat
 
             py -m pip install twine
```
webknjaz, line 41: For win console, it's probably just
```suggestion
        .. code-block:: doscon

            > py -m pip install twine
```

webknjaz: Or maybe 
```suggestion
        .. code-block:: ps1con

            $ py -m pip install twine
```

webknjaz: cc @bhrutledge have you seen these?

source/guides/installing-using-pip-and-virtual-environments.rst
webknjaz: Yep, also, the console session should be coupled with the output after the prompt input.

source/guides/distributing-packages-using-setuptools.rst
bhrutledge: D'oh. I missed those. I agree those seem appropriate, _if_ there's a parallel change of all of the `bash` inputs, i.e.:

```
.. code-block:: console

    $ twine upload dist/*
```

I think that's beyond the scope of this PR. Also, as noted in an earlier discussion, I'm not sure it's a better experience, because it adds a little more work to copying a command.

source/guides/installing-using-pip-and-virtual-environments.rst
bhrutledge: > coupled with the output after the prompt input

If you mean there shouldn't be any explanatory text between the input and the output `code-block`, I disagree. That's how it was in earlier commits, and I thought it was a bit confusing. Also, there are already places in this doc (and I assume others) where they are separated, e.g.:

https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#installing-packages

bhrutledge (approved):
Thanks for the additional comments. I'm going to merge this as-is, because I think it's a nice improvement. Folks can follow up with additional issues/PRs if they like.

> Do you think we should add copy button for folks so they can copy the command directly?

@meowmeowmeowcat That sounds handy. I'd probably start with a PR that just adds the button, but _not_  the prompt characters and the configuration to remove it.

source/guides/installing-using-pip-and-virtual-environments.rst
webknjaz: @bhrutledge I think I've seen `console` perform well with windows-based prompts. My understanding that it only highlights the lines starting with some sort of a prompt differently from those that don't.
But yes, looking at https://github.com/pygments/pygments/blob/1ea5fc3/pygments/lexers/shell.py#L226 and https://github.com/pygments/pygments/blob/1ea5fc3/pygments/lexers/shell.py#L556, it seems like `doscon` is a windows counterpart of the `console` lexer.

@henryiii I do think that having clear input (with prompt) + output coupled would be beneficial

source/guides/installing-using-pip-and-virtual-environments.rst
webknjaz: It's confusing when the input line isn't marked with a prompt. If it is, that should be rendered better.

webknjaz:
re:copybutton — it'd be nice to integrate https://github.com/executablebooks/sphinx-copybutton + https://github.com/python/python-docs-theme/blob/master/python_docs_theme/static/copybutton.js

meowmeowmeowcat:
> re:copybutton — it'd be nice to integrate https://github.com/executablebooks/sphinx-copybutton + https://github.com/python/python-docs-theme/blob/master/python_docs_theme/static/copybutton.js

Thanks for this!

meowmeowmeowcat:
> https://github.com/python/python-docs-theme/blob/master/python_docs_theme/static/copybutton.js

@webknjaz Is this necessary for the docs? I think most of the code blocks in the docs are shell commands.

webknjaz:
Dunno, I think it'd be nice to have the ability to easily copy those commands w/o the prompt.

meowmeowmeowcat:
I agree. I will test sphinx-copybutton and see if it can integrate the function of the copybutton in python docs .

PR #940, opened 2021-07-10
GitHub Action for updating `messages.pot`

meowmeowmeowcat:
This Github Action runs every day automatically for updating `messages.pot`.


bhrutledge:
I didn't review https://github.com/pypa/packaging.python.org/pull/935, but based on https://github.com/pypa/packaging.python.org/pull/937/commits/ede3fed7c2f78cc756ff5b5818a62f8f7a1336c2, it seems like there's still potential for `messages.pot` to be updated during local development, and accidentally included in PR's.

I'm really not familiar with translations; is it common practice to include `*.pot` files in the repo?

meowmeowmeowcat:
@bhrutledge We have to update the `messages.pot` locally and commit the change after we changed the content of html pages in Warehouse. 
Here's example: PR [#9690](https://github.com/pypa/warehouse/pull/9690/files#diff-7cc1dd5466d4caf26bedf81acb95231b510147da1a3ed4ff5a0ae3f8fa3eeb3a) in Warehouse. 

However, our main purpose is maintaining the docs, so our docs will be changed constantly. 
This makes me to think that if we should open a new repo for translations of this docs so folks won't accidentally included `messages.pot` in PR.

meowmeowmeowcat:
> I'm really not familiar with translations; is it common practice to include *.pot files in the repo?

yes, it's really common to include *.pot files in the repo, just like Warehouse.

.github/workflows/translation.yml
```diff
@@ -0,0 +1,35 @@
+name: Translation
+
+on:
+  schedule:
+    - cron: '00 00,00 * * *'
```
webknjaz, line 5: This cron spec doesn't look right.

.github/workflows/translation.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: python -m pip install --upgrade nox virtualenv
+        
+      - name: Create local changes
+        run: python -m nox -s translation
+        
+      - name: Commit files
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git commit -m "Update messages.pot" -a
```
webknjaz, line 29: I think it'd be useful to include the current commit hash in the message (the short one would do).

webknjaz: Also, I'd not use `-a` but just point to the pot file.

.github/workflows/translation.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install Dependencies
+        run: python -m pip install --upgrade nox virtualenv
+        
+      - name: Create local changes
+        run: python -m nox -s translation
+        
+      - name: Commit files
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
webknjaz, line 27: Drop that `number+` thing

.github/workflows/translation.yml
```diff
@@ -0,0 +1,35 @@
...
+          git config --local user.name "github-actions[bot]"
+          git commit -m "Update messages.pot" -a
+        
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+         github_token: ${{ secrets.ACCESS_TOKEN }}
+         branch: ${{ github.ref }}
```
webknjaz, line 35: This is unnecessary, just do `git push --atomic`

webknjaz: (specify what to push tho)

.github/workflows/translation.yml
meowmeowmeowcat: Also, I would prefer less commits to the repo, so maybe I should set it to update automatically on every Sunday?

meowmeowmeowcat: I think this is right.

Image

meowmeowmeowcat:
Moved this PR to #945

PR #941, opened 2021-07-12
Use PyPI-published version of `python-docs-theme`

webknjaz:


henryiii (approved):

pradyunsg (approved):
Ah great. I forgot that they started publishing.

PR #946, opened 2021-07-13
Add Copy Button

meowmeowmeowcat:
- This allows folks to copy commands directly with buttons.

### DIFF
--- a/source/_static/overrides.css
+++ b/source/_static/overrides.css
@@ -1,0 +1,8 @@
+/* Overrides for theme and plugin styles */
+
+/* https://github.com/executablebooks/sphinx-copybutton/blob/v0.4.0/sphinx_copybutton/_static/copybutton.css */
+
+button.copybtn img {
+    /* Fix alignment for pypa_theme */
+    padding: 0;
+}
--- a/source/conf.py
+++ b/source/conf.py
@@ -39,7 +39,7 @@ extensions = [
 ]
 
 # config for copy button
-copybutton_prompt_text = r">>> |\.\.\. "
+copybutton_prompt_text = r">>> |\.\.\. |\$ |> "
 copybutton_prompt_is_regexp = True
 
 # Add any paths that contain templates here, relative to this directory.
--- a/source/conf.py
+++ b/source/conf.py
@@ -35,7 +35,12 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx_inline_tabs',
+    'sphinx_copybutton',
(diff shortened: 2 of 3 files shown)


source/conf.py
```diff
@@ -35,8 +35,13 @@
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx_inline_tabs',
+    'sphinx_copybutton',
 ]
 
+# config for copy button
+copybutton_prompt_text = r">>> |\.\.\. "
```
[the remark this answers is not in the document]
meowmeowmeowcat: Oh, I see. However, we won't remove prompts like `$` in the future?

source/_static/overrides.css
```diff
@@ -0,0 +1,8 @@
+/* Overrides for theme and plugin styles */
+
+/* https://github.com/executablebooks/sphinx-copybutton/blob/v0.4.0/sphinx_copybutton/_static/copybutton.css */
+
+button.copybtn img {
+    /* Fix alignment for pypa_theme */
+    padding: 0;
+}
```
bhrutledge, line 8: The icon alignment was bugging me, so I took the liberty to fix it. 😄  This (and the associated  `html_` changes in `conf.py`) seemed to follow the suggestion at https://sphinx-copybutton.readthedocs.io/en/latest/#customize-the-css.

I'd love other folks' thoughts on this.

bhrutledge (approved):
Very cool, thank you!

source/conf.py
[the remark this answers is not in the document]
bhrutledge: I just opened https://github.com/pypa/packaging.python.org/issues/947 for that, but this covers the existing use, plus any prompts that remain afterwards.

source/_static/overrides.css
pradyunsg: I'd prefer we get this fixed upstream.

pradyunsg: That said, this is not a blocking concern -- rather I'm saying that this is an issue that we should follow up upstream for.

webknjaz (approved):

source/_static/overrides.css
bhrutledge: Thanks. By upstream, do you mean `pypa_theme`, or sphinx-copybutton? I'm guessing the alignment issue is specific to `pypa_theme`; the button renders fine in the [sphinx-copybutton docs](https://sphinx-copybutton.readthedocs.io/en/latest/), which also says it's designed to work with Alabaster.

PR #950, opened 2021-07-16
fix: easy_install URL now includes /deprecated/

rffontenelle:
Tiny fix for the homepage of the deprecated easy_install. Without `/deprecated/` it ends in the 404 Not Found page of rtd.


webknjaz (commented):
Please use an intersphinx reference instead of hardcoding the URLs.

https://webknjaz.github.io/intersphinx-untangled/setuptools.rtfd.io/ should help you find the right reference.

rffontenelle:
@webknjaz Is it possible to use reference to specific section in a page? I see that ```:std:doc:`deprecated/easy_install` ``` works, but won't go to the section "Natural Script Launcher".

PR #952, opened 2021-07-17
Copy-edit pipx guide

bhrutledge:
Lots of little changes for consistency, hopefully resulting in more clarity. Notably:

- Prefer "application" over "entry point"
- Reserve `pipx` for commands, and use "pipx" when referring to the project
- Adjust line lengths
- Update some URLs


source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -1,53 +1,55 @@
...
+system has on Python packages.
 
-:ref:`pipx` solves this by creating a virtual
-environment for each package, while also ensuring that package's applications
-are accessible through a directory that is on your ``$PATH``. This allows each
-package to be upgraded or uninstalled without causing conflicts with other
-packages, and allows you to safely run the program from anywhere.
+:ref:`pipx` solves this by creating a virtual environment for each package,
+while also ensuring that its applications are accessible through a directory
+that is on your ``$PATH``. This allows each package to be upgraded or
```
webknjaz, line 17:
```suggestion
that is on your :envvar:`PATH`. This allows each package to be upgraded or
```

source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -1,53 +1,55 @@
...
+uninstalled without causing conflicts with other packages, and allows you to
+safely run the applications from anywhere.
 
 .. note:: pipx only works with Python 3.6+.
 
-``pipx`` is installed with ``pip``:
+pipx is installed with pip:
```
webknjaz, line 23: For consistency with the previous paragraph:
```suggestion
:ref:`pipx` is installed with pip:
```

source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -1,53 +1,55 @@
...
   .. code-block:: bat
 
       py -m pip install --user pipx
-      
       py -m pipx ensurepath
 
-.. Note:: You may need to restart your terminal for the path updates to take effect.
+.. note::
+
+   ``ensurepath`` ensures that the application directory is on your ``$PATH``.
```
webknjaz, line 41:
```suggestion
   :command:`ensurepath` ensures that the application directory is on your :envvar:`PATH`.
```

source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -1,53 +1,55 @@
...
+   You may need to restart your terminal for this update to take effect.
 
-Now you can install packages with ``pipx install`` and access the package's entry point(s) from anywhere.
+Now you can install packages with ``pipx install`` and run the package's
```
webknjaz, line 44:
```suggestion
Now you can install packages with :command:`pipx install` and run the package's
```

source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -67,7 +69,8 @@ For example
              (__)\       )\/       ||----w |
                  ||     ||
 
-To see a list of packages installed with pipx and which CLI applications are available, use ``pipx list``.
+To see a list of packages installed with pipx and which applications are
+available, use ``pipx list``:
```
webknjaz, line 73:
```suggestion
available, use :command:`pipx list`:
```

source/guides/installing-stand-alone-command-line-tools.rst
bhrutledge: Are you suggesting that all instances of "pipx" get this treatment? That feels excessive to me, and possibly distracting to the reader. It seems like referencing the first use is sufficient.

bhrutledge (commented):
Thanks for the suggestions! I didn't know about `:command:` and `:envvar:`.

source/guides/installing-stand-alone-command-line-tools.rst
```diff
@@ -1,53 +1,55 @@
...
   .. code-block:: bat
 
       py -m pip install --user pipx
-      
       py -m pipx ensurepath
 
-.. Note:: You may need to restart your terminal for the path updates to take effect.
+.. note::
+
+   :command:`ensurepath` ensures that the application directory is on your :envvar:`PATH`.
```
bhrutledge, line 41: Hm. The command no longer renders in monospace font, and the environment variable doesn't include the `$`. That seems less useful to the reader than the previous formatting.

https://python-packaging-user-guide--952.org.readthedocs.build/guides/installing-stand-alone-command-line-tools/

webknjaz: Yeah, I was hoping that we'd get the theme support for this. With custom CSS or Furo. But @pradyunsg wanted to have a different theme for a nested docs structure like PyPUG has.

webknjaz: Though a dollar could be added like this 
```suggestion
   :command:`ensurepath` ensures that the application directory is on your :envvar:`$PATH <PATH>`.
```

source/guides/installing-stand-alone-command-line-tools.rst
webknjaz: Maybe. It's just that Sphinx is really designed to interlink things heavily.

source/guides/installing-stand-alone-command-line-tools.rst
bhrutledge: I reverted this in https://github.com/pypa/packaging.python.org/pull/952/commits/25179048adb5fe8ce5f2dac67dea569ab9d7cd01. The [docs for `:envvar`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-envvar) indicate that it's intended for cross-referencing. I read that as something we'd want to use if we were defining custom environment variables.

[`:command:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-command) makes sense semantically, but it's not consistently used in this guide, and I would rather see that done holistically, with an update to the theme to give it monospace formatting.

webknjaz: Okay, I wanted to declare that env var in the glossary maybe

bhrutledge: In that case, I think it would be referenced with a `:term:`. Either way, I'm inclined to leave that to a separate PR.

webknjaz: Sure, this is out of the scope here, after all. But in the future, I'd like to encourage more semantic markup.

webknjaz (approved):

PR #953, opened 2021-07-17
Update `installing-using-linux-tools` guide

meowmeowmeowcat:
- Use `code-block:: bash` instead of `::`
- Add commands for installing python3-pip


source/guides/installing-using-linux-tools.rst
```diff
@@ -4,9 +4,8 @@
 Installing pip/setuptools/wheel with Linux Package Managers
 ===========================================================
 
-:Page Status: Incomplete
-:Last Reviewed: 2015-09-17
```
bhrutledge, line 8: I'm not a linux user, so I can't determine if this document is "complete" now.

source/guides/installing-using-linux-tools.rst
```diff
@@ -70,26 +68,42 @@ To install pip and wheel for the system Python, there are two options:
...
-   EPEL 6 and EPEL7, you can install pip like so::
+   EPEL 6 and EPEL7, you can install pip like so:
+
+    * Python3:
+
+      .. code-block:: bash
+
+        sudo yum install python3-pip
+
+    * Python2:
```
bhrutledge, line 79: I would assume that we don't need the Python 2 instructions any more.

meowmeowmeowcat: So should I remove all Python 2 instructions from this guide?

source/guides/installing-using-linux-tools.rst
meowmeowmeowcat: I think most of the linux distributions are listed here, so I marked it as `complete` .

source/guides/installing-using-linux-tools.rst
bhrutledge: I might wait for @webknjaz to weigh in, just to confirm.

pradyunsg: Yea, please drop the Python 2 stuff if we're updating this.

source/guides/installing-using-linux-tools.rst
pradyunsg: IIRC, the main reason this isn't marked complete is that we want to provide detailed caveats of how various distributions have broken things.

bhrutledge: @pradyunsg Could that be worded as a catch-all `.. warning` at the top of the document? Can you suggest some copy for that?

source/guides/installing-using-linux-tools.rst
webknjaz: Yep, that sounds reasonable. Although, it could be a separate effort dropping py2 across all the documents

source/guides/installing-using-linux-tools.rst
webknjaz: > how various distributions have broken things.

Is this a fancy way of saying Debian?

source/guides/installing-using-linux-tools.rst
```diff
@@ -31,32 +30,31 @@ versions.  When this is known, we will make note of it below.
...
 
-  * Python 2::
+  * Python 3:
 
-      sudo yum upgrade python-setuptools
-      sudo yum install python-pip python-wheel
+    .. code-block:: bash
 
-  * Python 3: ``sudo yum install python3 python3-wheel``
+      sudo dnf install python3 python3-wheel
```
webknjaz, line 39: Maybe instead of trying to keep the docs up-to-date here, we could link the official Fedora docs that cover a lot more:
* https://developer.fedoraproject.org/tech/languages/python/python-installation.html
* https://labs.fedoraproject.org/en/python-classroom/
* https://fedoralovespython.org/
*
cc @pradyunsg

source/guides/installing-using-linux-tools.rst
meowmeowmeowcat: Okay, I will remove the Python 2 instructions from this guide.

source/guides/installing-using-linux-tools.rst
```diff
@@ -4,9 +4,8 @@
 Installing pip/setuptools/wheel with Linux Package Managers
 ===========================================================
 
-:Page Status: Incomplete
```
webknjaz, line 7: @meowmeowmeowcat so maybe we could keep this?

source/guides/installing-using-linux-tools.rst
webknjaz: cc @meowmeowmeowcat

source/guides/installing-using-linux-tools.rst
meowmeowmeowcat: Yes, and should we change the last reviewed date?

meowmeowmeowcat: Or we should keep them both unchanged?

source/guides/installing-using-linux-tools.rst
meowmeowmeowcat: I agree with @webknjaz's opinion. Folks can discover more from these docs.

source/guides/installing-using-linux-tools.rst
pradyunsg: Let's update the date, but keeping this sounds fine!

source/guides/installing-using-linux-tools.rst
```diff
@@ -31,33 +30,21 @@ versions.  When this is known, we will make note of it below.
...
 ~~~~~~
 
-* Fedora 21:
-
-  * Python 2::
-
-      sudo yum upgrade python-setuptools
-      sudo yum install python-pip python-wheel
-
-  * Python 3: ``sudo yum install python3 python3-wheel``
```
meowmeowmeowcat, line 41: @webknjaz Do you think that the command for Fedora to install pip and wheel should be `sudo dnf install python3-pip python3-wheel`?

webknjaz: I'm not even sure if you need to do anything unless it's some sort of a slim image.
https://fedoralovespython.org/

webknjaz: Alright, I checked and looks like these are missing from the Fedora 34 image so the command looks fine as long as we point the readers to the right places to explore more (like how to provision multiple pythons etc.)

Strictly speaking, wheel is not really necessary but may be useful (well, only for sdist installs w/o PEP517).

meowmeowmeowcat: Thanks for answering! Then I'll keep the command same here.

source/guides/installing-using-linux-tools.rst
```diff
@@ -31,33 +33,16 @@ versions.  When this is known, we will make note of it below.
...
-
-
-To get newer versions of pip, setuptools, and wheel for Python 2, you can enable
-the `PyPA Copr Repo <https://copr.fedoraproject.org/coprs/pypa/pypa/>`_ using
-the `Copr Repo instructions
-<https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__, and then run::
-
-  sudo yum|dnf upgrade python-setuptools
-  sudo yum|dnf install python-pip python-wheel
+To learn more about Python in Fedora, please visit `official Fedora docs`_,
```
webknjaz, line 40:
```suggestion
To learn more about Python in Fedora, please visit the `official Fedora docs`_,
```

source/guides/installing-using-linux-tools.rst
```diff
@@ -70,28 +55,37 @@ To install pip and wheel for the system Python, there are two options:
 1. Enable the `EPEL repository <https://fedoraproject.org/wiki/EPEL>`_ using
    `these instructions
    <https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F>`__. On
-   EPEL 6 and EPEL7, you can install pip like so::
+   EPEL 6 and EPEL7, you can install pip like so:
```
webknjaz, line 58: While on it, let's also make the whitespace consistent ```suggestion
   EPEL 6 and EPEL 7, you can install pip like so:
```

File: source/guides/installing-using-linux-tools.rst [hunk 8]
```diff
@@ -70,28 +55,37 @@ To install pip and wheel for the system Python, there are two options:
...
 2. Enable the `PyPA Copr Repo
    <https://copr.fedoraproject.org/coprs/pypa/pypa/>`_ using `these instructions
    <https://fedorahosted.org/copr/wiki/HowToEnableRepo>`__ [1]_. You can install
-   pip and wheel like so::
+   pip and wheel like so:
+
+   .. code-block:: bash
 
-     sudo yum install python-pip python-wheel
+     sudo yum install python3-pip python3-wheel
```
webknjaz, line 82:
```suggestion
     sudo dnf install python3-pip python3-wheel
```

source/guides/installing-using-linux-tools.rst
```diff
@@ -70,28 +55,37 @@ To install pip and wheel for the system Python, there are two options:
...
 
-   To additionally upgrade setuptools, run::
+   To additionally upgrade setuptools, run:
 
-     sudo yum upgrade python-setuptools
+   .. code-block:: bash
+
+     sudo yum upgrade python3-setuptools
```
webknjaz, line 88: dnf?

source/guides/installing-using-linux-tools.rst
```diff
@@ -31,33 +33,16 @@ versions.  When this is known, we will make note of it below.
 Fedora
 ~~~~~~
 
-* Fedora 21:
+.. code-block:: bash
 
-  * Python 2::
+  sudo dnf install python3 python3-wheel
```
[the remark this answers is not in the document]
meowmeowmeowcat: Yes, but folks may need to install pip (according to [the official docs](https://developer.fedoraproject.org/tech/languages/python/pypi-installation.html)). Maybe changing `python3` to `python3-pip`?

source/guides/installing-using-linux-tools.rst
meowmeowmeowcat: I am not quite sure about this. [The docs](https://fedoraproject.org/wiki/EPEL#Quickstart ) mainly uses yum for installation.

meowmeowmeowcat:
In the latest commit, I used `dnf` instead of `yum` and removed EPEL 6 as it has been EOL. The latest build on [PyPA Copr Repo](https://copr.fedorainfracloud.org/coprs/pypa/pypa/builds/) is 6 years ago, so I didn't add EPEL 8 to the guide.

webknjaz (approved):

PR #956, opened 2021-07-27
Use packaging-problems templates

henryiii:
Followup to https://github.com/pypa/packaging-problems/pull/532 (should be merged after that one).


bhrutledge (approved):
Thanks for seeing this through. I took the liberty of renaming `generic.yml` to "general_issue.yml" in both places.

PR #957, opened 2021-08-15
Improve distributing packages

rffontenelle:
This PR apply the following changes in `source/guides/distributing-packages-using-setuptools.rst`:

- Make `setup()` arg names literal (e.g. `name`, `version`, `description`), so that translators know it should not be translated.
- Fix a paragraph that had full-stop character before an "and", and remove literal from placeholder pair that ought to be translated.

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -466,14 +466,14 @@ cases you may need to place data files *outside* of your :term:`packages
 It is mostly useful if you need to install files which are used by other
 programs, which may be unaware of Python packages.
 
-Each ``(directory, files)`` pair in the sequence specifies the installation
-directory and the files to install there. The ``directory`` must be a relative
+Each "(directory, files)" pair in the sequence specifies the installation
+directory and the files to install there. The "directory" must be a relative
 path (although this may change in the future, see
-`wheel Issue #92 <https://github.com/pypa/wheel/issues/92>`_).
+`wheel Issue #92 <https://github.com/pypa/wheel/issues/92>`_),
 and it is interpreted relative to the installation prefix
 (Python’s ``sys.prefix`` for a default installation;
 ``site.USER_BASE`` for a user installation).
-Each file name in ``files`` is interpreted relative to the :file:`setup.py`
+Each file name in "files" is interpreted relative to the :file:`setup.py`
 script at the top of the project source distribution.
 
 For more information see the distutils section on `Installing Additional Files
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -466,14 +466,14 @@ cases you may need to place data files *outside* of your :term:`packages
 It is mostly useful if you need to install files which are used by other
 programs, which may be unaware of Python packages.
 
-Each "(directory, files)" pair in the sequence specifies the installation
-directory and the files to install there. The "directory" must be a relative
+Each ``(directory, files)`` pair in the sequence specifies the installation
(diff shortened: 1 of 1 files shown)


source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -466,14 +466,14 @@ cases you may need to place data files *outside* of your :term:`packages
 It is mostly useful if you need to install files which are used by other
 programs, which may be unaware of Python packages.
 
-Each ``(directory, files)`` pair in the sequence specifies the installation
```
webknjaz, line 469: Looks like this one should probably be kept as inline code.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -466,14 +466,14 @@ cases you may need to place data files *outside* of your :term:`packages
...
-directory and the files to install there. The ``directory`` must be a relative
```
webknjaz, line 470: And I'd keep this as is too.

webknjaz (commented):
I like this except for those data files changes that seem incorrect.

source/guides/distributing-packages-using-setuptools.rst
webknjaz: I think, maybe we could instead mark this as translatable on Weblate?

webknjaz:
> * Fix a paragraph that had full-stop character before an "and", and remove literal from placeholder pair that ought to be translated.

@meowmeowmeowcat what are your thoughts on this? I think the translatability shouldn't depend on whether something is marked as `inline code` visually. I think I'd just apply the `read-only` flag to untranslatable things on Weblate, as you suggested earlier.

source/guides/distributing-packages-using-setuptools.rst
rffontenelle: Ok, I undo this change. I thought translating the inline code would make sphinx-build complain, but I tested it right now and it doesn't. Actually, when translating Python Docs (also reStructuredText) it is Transifex that complains. Therefore, this change is unnecessary.

I will undo this. I won't revert it thought, because in the same commit I fix a fullstop in this character.

rffontenelle:
> I think the translatability shouldn't depend on whether something is marked as inline code visually.

Correct. As I mentioned in a reply above, I thought inline code should not be translated or sphinx-build would complain, and it is not true.

> I think I'd just apply the read-only flag to untranslatable things on Weblate, as you suggested earlier.

@webknjaz This PR has 2 commits, and you quoted the description of the latter. When you mention 'read-only' flag, do you refer to the former commit (which has the setup() args)?  I would be fine not putting in inline, and having it set as read-only. Actually setting as read only was my suggestion it in the comments in Weblate.

meowmeowmeowcat:
> @meowmeowmeowcat what are your thoughts on this? I think the translatability shouldn't depend on whether something is marked as `inline code` visually. I think I'd just apply the `read-only` flag to untranslatable things on Weblate, as you suggested earlier.

@webknjaz I agree with you here. However, we should also notify translators that they should be more careful when treating strings that have `inline code`.

source/guides/distributing-packages-using-setuptools.rst
webknjaz: Perfect, thanks!

P.S. _Pro tip:_ you can add `-n` to the `git revert` command and it'll let you make some edits and choose what to commit. This way you can produce partial reverts :)

webknjaz:
> Actually setting as read only was my suggestion it in the comments in Weblate.

Oh, my bad :man_facepalming:, I keep confusing folks on the Internet :)

> Like adding explanations to them.

That's a great idea! I like it.

webknjaz (approved):

PR #959, opened 2021-08-19
bento has seen to update since 2014.

guettli:
No update since 7 years. I think "less is more", and removed it from the list.


pradyunsg (approved):
Sounds fine by me!

PR #965, opened 2021-08-24
Fix broken links

meowmeowmeowcat:


webknjaz (commented):
Please use intersphinx references instead of the hardcoded URLs to avoid such problems in the future.

meowmeowmeowcat:
FIxed!

webknjaz (approved):

meowmeowmeowcat:
@webknjaz Should I open a new PR for using intersphinx references instead of the hardcoded URLs in the docs?

webknjaz:
If you feel like doing a repo-global crusade — go for it!

PR #969, opened 2021-08-27
Switch example reading Markdown to use pathlib

tdhopper:
Use simpler pathlib example instead of `os.path`.


bhrutledge (approved):
Thanks for this. I wasn't aware of this guide. 😄 

FYI, `setup.cfg` is now the [preferred method](https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata) of configuring metadata, including the description fields. I created https://github.com/pypa/packaging.python.org/issues/971 to update this guide.

PR #973, opened 2021-09-02
direct-url: Wrap `pip` commands with backticks

meowmeowmeowcat:
Found this issue when fixing broken links.

### DIFF
--- a/source/specifications/direct-url.rst
+++ b/source/specifications/direct-url.rst
@@ -287,7 +287,7 @@ Commands that generate a ``direct_url.json``:
   (in which case, ``url`` will be the local directory where the git repository has been
   cloned to, and ``dir_info`` will be present with ``"editable": true`` and no
   ``vcs_info`` will be set)
-* ``pip install -e ./app``
+* ``pip install --editable ./app``
 
 Commands that *do not* generate a ``direct_url.json``
 
--- a/source/specifications/direct-url.rst
+++ b/source/specifications/direct-url.rst
@@ -278,18 +278,18 @@ Example pip commands and their effect on direct_url.json
 
 Commands that generate a ``direct_url.json``:
 
-* pip install https://example.com/app-1.0.tgz
-* pip install https://example.com/app-1.0.whl
-* pip install "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"
-* pip install ./app
-* pip install file:///home/user/app
-* pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"
+* ``pip install https://example.com/app-1.0.tgz``
+* ``pip install https://example.com/app-1.0.whl``
+* ``pip install "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
+* ``pip install ./app``
+* ``pip install file:///home/user/app``
+* ``pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
   (in which case, ``url`` will be the local directory where the git repository has been
   cloned to, and ``dir_info`` will be present with ``"editable": true`` and no
   ``vcs_info`` will be set)
-* pip install -e ./app
+* ``pip install -e ./app``
 
 Commands that *do not* generate a ``direct_url.json``
 
-* pip install app
-* pip install app --no-index --find-links https://example.com/
+* ``pip install app``
+* ``pip install app --no-index --find-links https://example.com/``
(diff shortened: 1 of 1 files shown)


source/specifications/direct-url.rst
```diff
@@ -278,18 +278,18 @@ Example pip commands and their effect on direct_url.json
...
-* pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"
+* ``pip install https://example.com/app-1.0.tgz``
+* ``pip install https://example.com/app-1.0.whl``
+* ``pip install "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
+* ``pip install ./app``
+* ``pip install file:///home/user/app``
+* ``pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
   (in which case, ``url`` will be the local directory where the git repository has been
   cloned to, and ``dir_info`` will be present with ``"editable": true`` and no
   ``vcs_info`` will be set)
```
bhrutledge, line 289: This is rendering strangely: https://python-packaging-user-guide--973.org.readthedocs.build/specifications/direct-url/#example-pip-commands-and-their-effect-on-direct-url-json. I wonder if it's the parenthetical clause. Maybe that could a nested list item instead? Or maybe it's not necessary at all?

source/specifications/direct-url.rst
```diff
@@ -278,18 +278,18 @@ Example pip commands and their effect on direct_url.json
...
-* pip install -e ./app
+* ``pip install -e ./app``
```
bhrutledge, line 290: For consistency with the previous command:

```suggestion
* ``pip install --editable ./app``
```

bhrutledge (changes requested):

source/specifications/direct-url.rst
meowmeowmeowcat: https://packaging.python.org/guides/installing-using-linux-tools/#centos-rhel has the same problem, but I don't have a solution for it.

meowmeowmeowcat: I prefer using a nested list item instead.

meowmeowmeowcat: Image

It doesn't work too.

meowmeowmeowcat: I guess the reason is that the command `pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"` is too long.

webknjaz: Are you talking about the justified text alignment? If yes, it's probably something to fix on the CSS level in the theme rather than in this PR.

source/specifications/direct-url.rst
```diff
@@ -278,18 +278,18 @@ Example pip commands and their effect on direct_url.json
...
+* ``pip install https://example.com/app-1.0.whl``
+* ``pip install "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
+* ``pip install ./app``
+* ``pip install file:///home/user/app``
+* ``pip install --editable "git+https://example.com/repo/app.git#egg=app&subdirectory=setup"``
   (in which case, ``url`` will be the local directory where the git repository has been
   cloned to, and ``dir_info`` will be present with ``"editable": true`` and no
   ``vcs_info`` will be set)
-* pip install -e ./app
+* ``pip install --editable ./app``
```
webknjaz, line 290: This change is not related to the PR description/title/goal. Please avoid making such changes to make sure that the review and merge happen quickly.

If it wasn't here, I would've merged this already. But now I have to wait until you either revert this or justify why it is important to have the option format change *in this specific PR*.

source/specifications/direct-url.rst
webknjaz: Not this is inconsistent with the main goal of the PR. Such a change should happen in a separate PR.

source/specifications/direct-url.rst
webknjaz: FTR this is coming from
```css
_static/classic.css:197
div.body p, div.body dd, div.body li, div.body blockquote {
    text-align: justify;
    line-height: 130%;
}
```
This is probably coming from one of the underlying themes in Sphinx. But again, this is not something we can fix in this PR.

webknjaz: P.S. If #791 ever gets merged (or replaced but a better theme), the problem should go away.

webknjaz (approved):

PR #982, opened 2021-09-07
Fix links reported as broken

meowmeowmeowcat:
This is the new one for #976.


meowmeowmeowcat:
The latest commit is the same as the one in #976.

webknjaz (approved):

PR #987, opened 2021-09-27
Add pdm to non-PyPA project list

frostming:


source/key_projects.rst
```diff
@@ -464,6 +464,21 @@ multibuild
...
+
+`Docs <https://pdm.fming.dev/>`__ |
+`GitHub <https://github.com/pdm-project/pdm/>`__ |
+`PyPI <https://pypi.org/project/pdm>`__
+
+PDM is a modern Python package manager with PEP 582 support.
+It installs and manages packages in a similar way to npm that doesn't need
+to create a virtualenv at all. It uses a ``pyproject.toml`` to store project metadata
+as defined in PEP 621. With PDM, you can experience the state of art in the Python
+packaging world.
```
bhrutledge, line 480: Let's add some helpful links. Also, since the "state of the art" is constantly evolving, I think it's best to be modest.

```suggestion
PDM is a modern Python package manager with :pep:`582` support. It installs and
manages packages in a similar way to ``npm`` that doesn't need to create a
:term:`virtual environment` at all. It also uses :term:`pyproject.toml` to store
project metadata as defined in :pep:`621`.
```

bhrutledge (approved):

PR #990, opened 2021-09-29
Argument in favor of dependency pinning

danuker:
I believe the statement about package version restriction should let the reader know the trade-off they are making.

This is my first contribution here, so I apologize for breaking any conventions. I would be very thankful for feedback!

### DIFF
--- a/source/discussions/install-requires-vs-requirements.rst
+++ b/source/discussions/install-requires-vs-requirements.rst
@@ -52,6 +52,8 @@ It is not considered best practice to use ``install_requires`` to pin
 dependencies to specific versions, or to specify sub-dependencies
 (i.e. dependencies of your dependencies).  This is overly-restrictive, and
 prevents the user from gaining the benefit of dependency upgrades.
+However, it may be useful in security-critical projects,
+to prevent malicious updates. [2]_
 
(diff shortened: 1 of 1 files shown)


source/discussions/install-requires-vs-requirements.rst
```diff
@@ -52,6 +52,8 @@ It is not considered best practice to use ``install_requires`` to pin
 dependencies to specific versions, or to specify sub-dependencies
 (i.e. dependencies of your dependencies).  This is overly-restrictive, and
 prevents the user from gaining the benefit of dependency upgrades.
+However, it may be useful in security-critical projects,
+to prevent malicious updates. [2]_
```
webknjaz, line 56: But this also prevents CVE fixes the same way 🤷‍♂️

danuker: You are right. I guess it is would be useful for companies performing code audits, but those would not be the target of this tutorial.
I no longer think this should be included, since I think the risk of missing a security update becomes much greater. And there are much more instances of unpatched software running than malicious updates.

danuker:
Not useful. Thanks @webknjaz for the feedback!

PR #975, opened 2021-09-02
Use more intersphinx references & format some links

dukecat0:

PR #998, opened 2021-10-10
Fix broken links

bhrutledge:
Discovered while looking at some open PR's, and currently failing on `main`:

https://github.com/pypa/packaging.python.org/runs/3850844441?check_suite_focus=true

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -180,7 +180,7 @@ causing dependency conflicts with other packages installed on the system.
 Python Packaging User Guide
 ===========================
 
-`Docs <https://packaging.python.org/en/latest/>`__ |
+`Docs <https://packaging.python.org/>`__ |
 `Issues <https://github.com/pypa/python-packaging-user-guide/issues>`__ |
 `GitHub <https://github.com/pypa/python-packaging-user-guide>`__
 
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -34,7 +34,7 @@ Content Delivery Network (CDN).
 build
 =====
 
-`Docs <https://pypa-build.readthedocs.io/>`__ |
+`Docs <build:index>` |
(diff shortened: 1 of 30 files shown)


source/key_projects.rst
```diff
@@ -180,7 +180,7 @@ causing dependency conflicts with other packages installed on the system.
 Python Packaging User Guide
 ===========================
 
-`Docs <https://packaging.python.org/en/latest/>`__ |
+`Docs <https://packaging.python.org/>`__ |
```
bhrutledge, line 183: This one is surprising to me. Should https://packaging.python.org/en/latest/ work?

source/specifications/core-metadata.rst
```diff
@@ -733,4 +733,4 @@ Examples::
 .. _`Python Package Index`: http://pypi.org/
 
 .. [2] RFC 822 Long Header Fields:
-   http://www.freesoft.org/CIE/RFC/822/7.htm
+   https://datatracker.ietf.org/doc/html/rfc822#section-3.1.1
```
webknjaz, line 736: For RFCs, just use the `:rfc:` role in the source.

source/key_projects.rst
webknjaz: Also, I thought it's been fixed in one of the other PRs. Let me see if it got merged...

webknjaz (commented):
Merging #975, check for conflicts afterwards.

source/key_projects.rst
meowmeowmeowcat: I've used intersphinx for this in #975:
```rst
:doc:`Docs <packaging.python.org:index>`
```

bhrutledge:
I forgot about #975. Looks like this wasn't necessary. 🤦

webknjaz:
Sorry I didn't merge that one sooner, it was updated while I was on vacation and I forgot about it for a bit...

PR #1004, opened 2021-10-27
Clarify how MANIFEST.in's "include" and "exclude" work

jwodder:
I recently found out the hard way that, if you have a file `src/foobar/stuff.txt` in your project, writing just `include stuff.txt` in your `MANIFEST.in` won't do anything; you need to write `include src/foobar/stuff.txt` instead.  This PR tries to make that clearer to readers.


meowmeowmeowcat:
See #1005.

PR #1005, opened 2021-10-29
Use Wayback Machine for pipenv docs

meowmeowmeowcat:
- See https://github.com/pypa/pipenv/issues/4805

@webknjaz


webknjaz (approved):
Thanks!

PR #1007, opened 2021-11-15
WIP Enable build translated docs

rffontenelle:
## Description

Currently there is no option in nox config file for building translated version of packaging.python.org. `nox -s build` builds English-only docs. One can run e.g. `nox -s build -- -D language=pt_BR` to build Brazilian Portuguese translated version of the docs. The idea with this PR is to implement a nox's session (`--session`/`-s`) to build translated docs.

See https://github.com/pypa/packaging.python.org/issues/927#issuecomment-968310133.

/cc @meowmeowmeowcat @webknjaz 

## Initial solution

While I don't feel this is perfect, let's use it to start discussion.

This commit added a new session named 'l10n' that receives the list of languages available as parameters. This new session relies on 'build' session to build the docs.

Build docs in only English (source language):
```
nox -s build
```

Build docs in all translated languages (English excluded):
```
nox -s l10n
```

Build docs in only Brazilian Portuguese:
```
nox -s "l10n(language='pt_BR')"
```

NOTE: I had to comment the line that removes target_build_dir because it was cleaning up the directory on every run, causing it to losing all previous builds.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -32,9 +32,9 @@ def build(session, language='en', autobuild=False):
     """
     session.install("-r", "requirements.txt")
 
-    target_build_dir = "build"
+    target_build_dir = "build/" + language
 
-    #shutil.rmtree(target_build_dir, ignore_errors=True)
+    shutil.rmtree(target_build_dir, ignore_errors=True)
 
     if autobuild:
         command = "sphinx-autobuild"
@@ -47,25 +47,21 @@ def build(session, language='en', autobuild=False):
             "--keep-going",  # don't interrupt the build on the first warning
         )
 
-    if language:
-        # add language parameter; if empty do not add because it is English
-        extra_args = extra_args + ( "-D", "language=" + language)
-        target_build_dir = target_build_dir + '/' + language
     session.run(
         command, *extra_args,
         "-j", "auto",  # parallelize the build
         "-b", "html",  # use HTML builder
         "-n",  # nitpicky warn about all missing references
         "-W",  # Treat warnings as errors.
+        "-D", "language=" + language, # build in the give language
         *session.posargs,
         "source",  # where the rst files are located
         target_build_dir,  # where to put the html output
     )
 
 @nox.session()
-@nox.parametrize("language", language_list)
-def l10n(session, language):
-    build(session, language=language)
+def l10n(session):
+    build(session, language=session.posargs.pop(0))
 
 @nox.session()
 def preview(session):
--- a/noxfile.py
+++ b/noxfile.py
@@ -26,7 +26,7 @@ def translation(session):
     )
 
 @nox.session()
-def build(session, autobuild=False):
+def build(session, language='en', autobuild=False):
     """
     Make the website.
     """
@@ -34,7 +34,7 @@ def build(session, autobuild=False):
 
     target_build_dir = "build"
 
-    shutil.rmtree(target_build_dir, ignore_errors=True)
+    #shutil.rmtree(target_build_dir, ignore_errors=True)
 
     if autobuild:
         command = "sphinx-autobuild"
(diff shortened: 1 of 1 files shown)


rffontenelle:
> NOTE: I had to comment the line that removes target_build_dir because it was cleaning up the directory on every run, causing it to losing all previous builds.

Idea: Instead of removing the cleanup command for target_build_dir (which is the directory "build"), we could change its current value to "build/html" and the translated version be "build/LANG/html", so the cleanup would be only in the respective "html" instead of the whole "build" directory.

webknjaz:
> Idea: Instead of removing the cleanup command for target_build_dir (which is the directory "build"), we could change its current value to "build/html" and the translated version be "build/LANG/html", so the cleanup would be only in the respective "html" instead of the whole "build" directory.

Yeah, maybe that's a good idea. By the way, now that the MO and POT files move to a separate branch, we may need to detect that and ask people to switch the branch..

noxfile.py
```diff
@@ -6,6 +6,8 @@
 import shutil
 import nox
 
+language_list = ["eo", "es", "pt_BR", "ro", "ru", "si", "uk", "zh_Hans", "zh_Hant"]
```
webknjaz, line 9: Is this supposed to be used to make all the langs?

noxfile.py
```diff
@@ -19,12 +21,12 @@ def translation(session):
     )
 
 @nox.session(py="3")
-def build(session, autobuild=False):
+def build(session, language='', autobuild=False):
```
webknjaz, line 24: Can we have 
```suggestion
def build(session, language='en', autobuild=False):
```
and drop the conditional block?

webknjaz:
> Build docs in only Brazilian Portuguese:
> 
> ```
> nox -s "l10n(language='pt_BR')"
> ```

Wouldn't it be cleaner as

```console
$ nox -s l10n -- pt_BR
```

webknjaz:
Since it's a WIP, I'm marking it as a draft for now.

webknjaz:
@rffontenelle do we still want this?

noxfile.py
rffontenelle: Currently, my proposal is to put built docs in different paths according to their language. English already goes to `build/html`, and other languages would go to `build/<lang>/html`. If there is no condition to treat differently English from other languages, English docs will go to `build/en/html`.

Do you see a problem in that?

rffontenelle:
> Wouldn't it be cleaner as
> 
> ```
> $ nox -s l10n -- pt_BR
> ```

@webknjaz  I love your suggestion, but I'm not sure how to implement it. I tried setting

```
@nox.session()
def l10n(session):
    build(session, language=session.posargs[0])
```

which successfully passed the desired language to `build` session. However, `build` already reads `*session.posargs`, so passing pt_BR to `l10n` session results in the following incorrect command:

```
sphinx-build --color --keep-going -D language=pt_BR -j auto -b html -n -W pt_BR source build/pt_BR
```

(Notice the "pt_BR" after `-W` flag, read from posargs)

This also makes impossible to pass another arguments to l10n, although I don't know if we even want that.

noxfile.py
rffontenelle: Yes. The way my proposal is right now, running `nox -s l10n` with the `l10n` session having `@nox.parametrize("language", language_list)`, the whole list will be ran. Alternatively, a language can be passed `nox -s "l10n(language='pt_BR')"`

While this is nice as a single command to run all, in CI I personally would prefer to run a single language per job run via [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)

webknjaz:
@rffontenelle I'm thinking about something like
```diff
 @nox.session()
 def l10n(session):
-    build(session, language=session.posargs[0])
+    build(session, language=session.posargs.pop(0))
```
Or maybe even reset the posargs via `session.posargs = []`. But yeah, it's only good if you're not going to allow other args.

rffontenelle:
I feel this is not moving forward in a way that satisfy me, so I'll be closing it.

I initially wanted a session that allow 1) build a given language code or 2) build all if no language is specified. But all I've got is `nox -s "l10n(language='pt_BR')"` which is ugly but would allow build all languages. Other prettier was `nox -s l10n -- pt_BR` that wouldn't build all languages, hence is much different from `nox -s build -- -D language=pt_BR`.

Issue #912, opened 2021-06-02
Reword, reorder or rework tool recommendations for application dependencies


astrojuanlu:
Quoting @pfmoore in https://github.com/pypa/packaging.python.org/issues/701#issuecomment-612059528:

> Very little of the discussion on this issue was *actually* about the maintenance status of pipenv, more about issues such as how alternatives should be presented, whether recommending pipenv is the right choice in the first place, etc. Those issues remain, but I think someone should open a new issue, with a concrete alternative proposal to start the discussion, rather than just continue debating on this issue.

The situation of Pipenv has changed a lot since the original concerns about the viability of the project were raised. However, I still found it weird to see it's the prominent recommendation, with pip, pip-tools, and Poetry as secondary alternatives. Why?

_Disclaimer: I used Pipenv for some time while it was still maintained by Kenneth, then abandoned it in favor of pip-tools. Never used Poetry. I don't have a horse in this race, I just care about having good recommendations for newcomers and for my students._

- All tools have pros and cons, and Pipenv is no exception. My personal opinion is that Pipenv tradeoffs are not significantly better than the ones made by the other tools - but of course, this is one guy's opinion. I wonder if there has been some analysis (updated somewhat recently) of the merits of each tool, so it can justify this ordering and presentation.
- Pipenv relies on [pypa/pipfile](https://github.com/pypa/pipfile), a format that advertises itself as "a replacement for the existing standard pip's `requirements.txt` file". However, the README has a big `WARNING: This project is under active design and development. Nothing is set in stone at this point in time.`, and the experimental nature of the spec is not in line with the privileged position of Pipenv in the tool recommendations. (This might of course be a problem with the README instead.) On the other hand, [the proposal to make pip adopt it has been frozen for almost two years](https://github.com/pypa/pip/issues/6925), and in fact, it hasn't received much attention, so it is to be expected that, for the short and medium-term at least, Pipenv will be the only tool to support this format.
  - This is a minor point for users, but it's a pain point for maintainers. If people keep using Pipenv, more people will ask applications and services to support pipfile. And if the spec is not maintained or it's experimental, it might not be very appealing to do so, in turn, frustrating Pipenv users. It's all a matter of setting the appropriate expectations.
- The Pipenv recommendation looks at odds with the `venv`/`virtualenv` recommendation. At least, [according to the Python Guide](https://docs.python-guide.org/dev/virtualenvs/), venv has to be used "in place of Pipenv", rather than alongside it.

[...]

PR #1008, opened 2021-11-22
Fix some SciPy broken links

astrojuanlu:
While I was thinking about addressing #912, I decided to start with a smaller pull request first to get familiar with the contribution workflow and the expectations from reviewers.

This amends several build warnings I found because the SciPy website has changed a lot quite recently.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -419,7 +419,6 @@ intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
     'python-guide': ('https://docs.python-guide.org', None),
     'python2': ('https://docs.python.org/2', None),
-    'scipy': ('https://www.scipy.org', None),
     'setuptools': ('https://setuptools.readthedocs.io/en/latest/', None),
     'spack': ('https://spack.readthedocs.io/en/latest/', None),
     'sphinx': ('https://www.sphinx-doc.org/en/master', None),
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -142,7 +142,7 @@ To build the guide, run the following bash command in the source folder:
    nox -s build
 
 After the process has completed you can find the HTML output in the
-``./build/html`` directory. You can open the ``index.html`` file to view the
+``./build`` directory. You can open the ``index.html`` file to view the
 guide in web browser, but it's recommended to serve the guide using an HTTP
 server.
 
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -142,7 +142,7 @@ To build the guide, run the following bash command in the source folder:
    nox -s build
 
 After the process has completed you can find the HTML output in the
-``./build`` directory. You can open the ``index.html`` file to view the
+``./build/html`` directory. You can open the ``index.html`` file to view the
 guide in web browser, but it's recommended to serve the guide using an HTTP
 server.
 
--- a/source/guides/installing-scientific-packages.rst
+++ b/source/guides/installing-scientific-packages.rst
@@ -14,8 +14,8 @@ kinds of hardware, or to interoperate with different pieces of external
 software.
 
 In particular, `NumPy <https://numpy.org/>`__, which provides the basis
-for most of the software in the :doc:`scientific Python stack
-<scipy:stackspec>` can be configured
+for most of the software in the `scientific Python stack
+<https://scientific-python.org>`_ can be configured
 to interoperate with different FORTRAN libraries, and can take advantage
 of different levels of vectorised instructions available in modern CPUs.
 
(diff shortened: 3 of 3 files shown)


source/conf.py
```diff
@@ -419,7 +419,6 @@
     'python': ('https://docs.python.org/3', None),
     'python-guide': ('https://docs.python-guide.org', None),
     'python2': ('https://docs.python.org/2', None),
-    'scipy': ('https://www.scipy.org', None),
```
webknjaz, line 422: Looks like this should be https://scipy.github.io/devdocs/ now. I've updated it here https://webknjaz.github.io/intersphinx-untangled/scipy.github.io/ a while ago.

source/conf.py
```diff
@@ -419,7 +419,6 @@
...
     'setuptools': ('https://setuptools.readthedocs.io/en/latest/', None),
```
webknjaz, line 422:
```suggestion
    'scipy': ('https://scipy.github.io/devdocs', None),
    'setuptools': ('https://setuptools.readthedocs.io/en/latest/', None),
```

source/contribute.rst
```diff
@@ -142,7 +142,7 @@ To build the guide, run the following bash command in the source folder:
    nox -s build
 
 After the process has completed you can find the HTML output in the
-``./build/html`` directory. You can open the ``index.html`` file to view the
```
webknjaz, line 145: This is unrelated, submit separately.

webknjaz (changes requested):
Please keep intersphinx references where possible.

astrojuanlu:
> Please keep intersphinx references where possible.

I don't think that's possible, but I'll leave individual comments on each of the references I changed to discuss.

source/conf.py
astrojuanlu: (Postponing replying to this until we decide whether we need it or not)

source/guides/installing-scientific-packages.rst
```diff
@@ -14,8 +14,8 @@ kinds of hardware, or to interoperate with different pieces of external
 software.
 
 In particular, `NumPy <https://numpy.org/>`__, which provides the basis
-for most of the software in the :doc:`scientific Python stack
-<scipy:stackspec>` can be configured
+for most of the software in the `scientific Python stack
+<https://scientific-python.org>`_ can be configured
```
astrojuanlu, line 18: The `stackspec` page is gone. Last time seen: https://web.archive.org/web/20211023180926/https://www.scipy.org/stackspec.html

With a note that says:

> The SciPy Stack specification was developed in 2012. As of 2017, the SciPy Stack concept is obsolete given improvements in package management and distribution.

I replaced it with a reference to `scientific-python.org`, which is not a Sphinx site (I believe it's based on Hugo?) and so I can't use an intersphinx reference here.

source/guides/installing-scientific-packages.rst
```diff
@@ -90,16 +90,16 @@ macOS installers that are compatible with the macOS CPython binaries
...
+`install SciPy on macOS <https://scipy.org/install/#mac-packages>`_.
 
 
 SciPy distributions
 -------------------
 
-The SciPy site lists :doc:`several distributions
-<scipy:install>` that provide the full SciPy stack to
+The SciPy site lists `several distributions
+<https://scipy.org/install/#distributions>`_
```
astrojuanlu, line 101: Same page as above, same reasoning.

source/guides/installing-scientific-packages.rst
webknjaz: Okay, that makes sense. (for non-sphinx sites, it's possible to create roles but this would be an overkill in our case)

source/guides/installing-scientific-packages.rst
```diff
@@ -90,16 +90,16 @@ macOS installers that are compatible with the macOS CPython binaries
 published on python.org.
 
 macOS users also have access to Linux distribution style package managers
-such as ``MacPorts``. The SciPy site has more details on using MacPorts to
-install the :ref:`scientific Python stack
-<scipy:mac>`
+such as ``Homebrew``. The SciPy site has more details on using Homebrew to
+`install SciPy on macOS <https://scipy.org/install/#mac-packages>`_.
```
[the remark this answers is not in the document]
webknjaz: Okay, looks like the Sphinx docs don't have equivalent info.

source/guides/installing-scientific-packages.rst
webknjaz: Alright, makes sense.

webknjaz (approved):
All's good. Thanks!

source/guides/installing-scientific-packages.rst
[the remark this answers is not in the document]
webknjaz: Linkcheck says there's no `#mac-packages` anchor on that page.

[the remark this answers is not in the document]
webknjaz: Aha! It's just `#mac` in HTML. `#mac-packages` is a dynamic thing that only appears in JS runtime and the linter can't catch that.
```suggestion
`install SciPy on macOS <https://scipy.org/install/#mac>`_.
```

astrojuanlu:
Thanks for the quick turnaround @webknjaz !

PR #1014, opened 2021-11-27
Make translation strings collection workflow use a dedicated branch

webknjaz:


webknjaz:
This should help us keep Weblate source strings up-to-date, finally.

.github/workflows/translation.yml
```diff
@@ -1,47 +1,77 @@
...
-           git push --atomic origin HEAD:${{ github.event.repository.default_branch }}
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Grab the repo src
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # To reach the common commit
+    - name: Setup git user as [bot]
```
rffontenelle, line 24: Consider the verb "Set up", as "setup" is technically noun #nitpicky

webknjaz: Yeah, you're right. Currently, there's a more serious problem with this: Weblate doesn't work like Pradyun and I expected and now we need to think about how to address that...
```suggestion
    - name: Set up git user as [bot]
```

rffontenelle:
@webknjaz you said it is not working on Weblate. Not sure what's up, but have you considered setting on Weblate the input POT to the new branch (TBH I don't know if you

webknjaz:
> @webknjaz you said it is not working on Weblate. Not sure what's up, but have you considered setting on Weblate the input POT to the new branch?

That's exactly what I tried and it doesn't work the way I expected, it just sets a PR branch where Weblate tries to push stuff.
I need to think about it. I think moving the MO-files to a separate branch too and pointing RTD there would work.

> TBH I don't know if you can set separate branches for POT and PO. If you can't, you might want to consider a different repository for translations. See https://docs.weblate.org/en/latest/faq.html#how-can-i-limit-weblate-access-to-only-translations-without-exposing-source-code-to-it 

I wanted to avoid this.

rffontenelle:
Consider posting the issue on [Weblate discussions](https://github.com/WeblateOrg/weblate/discussions).

webknjaz:
> Consider posting the issue on [Weblate discussions](https://github.com/WeblateOrg/weblate/discussions).

I just realized that we can keep both the MOs and POT files in the `source/translation` branch and not `main`. Then, we'd publish the translated docs from that branch and Weblate would also send translation updates there w/o us needing to have those files in `main`. And also `source/translation` would be auto-updated from `main` continuously. So probably no need to raise this discussion anymore...

meowmeowmeowcat (approved):

pradyunsg (approved):

webknjaz:
Turns out, there was a bug in this PR. Here's the hotfix: https://github.com/pypa/packaging.python.org/pull/1021.

PR #1022, opened 2021-12-05
Use Sphinx 4.3.1

meowmeowmeowcat:
Errors: https://github.com/pypa/packaging.python.org/runs/4420714180?check_suite_focus=true#step:8:12

Sphinx 4.3.1 should fix these errors.

@webknjaz


webknjaz (approved):

webknjaz:
Thanks!

PR #1028, opened 2021-12-07
Advise against using distutils directly, take 2

### LINKED ISSUE #667: Formally drop support for distutils-only projects?

astrojuanlu:
Closes #667. Continuation of #672, closes #672.

I tried to respect the original spirit of #672, even if that meant _not_ updating the rest of the page (i.e. noting the deprecation of direct `setup.py` invokations and encouragement for `pyproject.toml` for static metadata), as this would require a bit more work. I prefer to defer that to a future PR.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -333,13 +333,11 @@ be required, but can be omitted with newer versions of setuptools and pip.
 
       You may see some existing projects or other Python packaging tutorials that
       import their ``setup`` function from ``distutils.core`` rather than
-      ``setuptools``. This is a `legacy approach`_ that installers support
+      ``setuptools``. This is a legacy approach that installers support
       for backwards compatibility purposes [1]_, but using the legacy ``distutils`` API
       directly in new projects is strongly discouraged, since ``distutils`` is
       deprecated as per :pep:`632` and will be removed from the standard library
       in Python 3.12.
-
-    .. _legacy approach: https://docs.python.org/3.10/library/distutils.html
 
 Creating README.md
 ------------------
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -329,17 +329,17 @@ be required, but can be omitted with newer versions of setuptools and pip.
     There are many more than the ones mentioned here. See
     :doc:`/guides/distributing-packages-using-setuptools` for more details.
 
-.. warning::
+    .. warning::
 
-   You may see some existing projects or other Python packaging tutorials that
-   import their ``setup`` function from ``distutils.core`` rather than
-   ``setuptools``. This is a `legacy approach`_ that installers support
-   for backwards compatibility purposes [1]_, but using the legacy ``distutils`` API
-   directly in new projects is strongly discouraged, since ``distutils`` is
-   deprecated as per :pep:`632` and will be removed from the standard library
-   in Python 3.12.
+      You may see some existing projects or other Python packaging tutorials that
+      import their ``setup`` function from ``distutils.core`` rather than
+      ``setuptools``. This is a `legacy approach`_ that installers support
+      for backwards compatibility purposes [1]_, but using the legacy ``distutils`` API
+      directly in new projects is strongly discouraged, since ``distutils`` is
+      deprecated as per :pep:`632` and will be removed from the standard library
+      in Python 3.12.
 
-.. _legacy approach: https://docs.python.org/3.10/library/distutils.html
+    .. _legacy approach: https://docs.python.org/3.10/library/distutils.html
 
 Creating README.md
 ------------------
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -642,7 +642,7 @@ some things you can do:
        pre-installed, and the operators of those environments may still be
        requiring users to install packages by running ``setup.py install``
        commands, rather than providing an installer like ``pip`` that
-       automatically installes required build dependendencies. These
+       automatically installs required build dependendencies. These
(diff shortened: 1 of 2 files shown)


astrojuanlu:
In particular, I made some minimal changes to the wording and the intensity of the admonition to address @pfmoore comments here: https://github.com/pypa/packaging.python.org/pull/672#discussion_r344476273

source/tutorials/packaging-projects.rst
```diff
@@ -329,6 +329,18 @@ be required, but can be omitted with newer versions of setuptools and pip.
...
+
+   You may see some existing projects or other Python packaging tutorials that
+   import their ``setup`` function from ``distutils.core`` rather than
+   ``setuptools``. This is a `legacy approach`_ that installers [1]_ support
+   for backwards compatibility purposes, but using the legacy ``distutils`` API
+   directly in new projects is strongly discouraged, since ``distutils`` is
+   deprecated as per :pep:`632` and will be removed from the standard library
+   in Python 3.12.
+
+.. _legacy approach: https://docs.python.org/3.10/library/distutils.html
```
webknjaz, line 342: Could you use an intersphinx reference to the setuptools docs?

astrojuanlu: Which link in particular? (the line you highlighted contains a link to CPython instead)

webknjaz:
@astrojuanlu do you think we need to link https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html ?

cc @pganssle

astrojuanlu:
AFAIU, that blog post is about the deprecation of direct `setup.py` invocations, and this PR is solely about the deprecation of distutils.

webknjaz:
Ah, okay, fair enough.

source/tutorials/packaging-projects.rst
webknjaz: distutils moved out of the stdlib and the docs with updates to them moved into the setuptools project.

astrojuanlu: Right, but I expect https://docs.python.org/3.10/library/distutils.html to live on forever? (Since it's "pinned" to CPython 3.10). It also shows the warning on top:

Screenshot 2021-12-07 at 20-21-39 distutils — Building and installing Python modules — Python 3 10 1 documentation

Do you know of a better page on the setuptools docs?

webknjaz: There's https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools and https://setuptools.pypa.io/en/latest/deprecated/distutils/ (although the latter does not mention the deprecation). It's not the `ditutils` library module page from the stdlib docs, though.

FWIW if you want to link against the CPython 3.10 site, add an explicit entry to the intersphinx configuration for this additional site and reference it via Sphinx roles. I've just triggered adding per-version objects to https://webknjaz.github.io/intersphinx-untangled/docs.python.org/ for convinience.

source/tutorials/packaging-projects.rst
```diff
@@ -329,6 +329,18 @@ be required, but can be omitted with newer versions of setuptools and pip.
     There are many more than the ones mentioned here. See
     :doc:`/guides/distributing-packages-using-setuptools` for more details.
 
+.. warning::
```
bhrutledge, line 332: This warning feels distracting in the default flow of the packaging tutorial, which now defaults to `setup.cfg`. For example, I don't see a previous reference to the `setup` function.

Could we isolate it to the `setup.py` tab?

source/tutorials/packaging-projects.rst
```diff
@@ -329,6 +329,18 @@ be required, but can be omitted with newer versions of setuptools and pip.
...
+
+   You may see some existing projects or other Python packaging tutorials that
+   import their ``setup`` function from ``distutils.core`` rather than
+   ``setuptools``. This is a `legacy approach`_ that installers [1]_ support
+   for backwards compatibility purposes, but using the legacy ``distutils`` API
```
bhrutledge, line 337: I think this flows a little more clearly:

```suggestion
   ``setuptools``. This is a `legacy approach`_ that installers support
   for backwards compatibility purposes [1]_, but using the legacy ``distutils`` API
```

source/tutorials/packaging-projects.rst
bhrutledge: I don't think linking to the Python `distutils` docs is useful, esp. in this tutorial aimed at beginners. https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools seems more useful, but it might just be distracting. I think the  ":pep:\`632`" link might be sufficient.

astrojuanlu: Fair enough, thanks both for the ideas. Will change the link.

di (approved):
LGTM once comments are resolved.

source/tutorials/packaging-projects.rst
astrojuanlu: Done!

source/tutorials/packaging-projects.rst
astrojuanlu: Ended up removing the link altogether 👍🏽

astrojuanlu:
Comments addressed :+1:

di:
Thanks!

PR #1029, opened 2021-12-10
Move Flit into PyPA

henryiii:
Recent move in to the PyPA requires a few changes here. :)

Will try to work on updates to the tutorials at some point to get a Flit PEP 621 version in. Should it be a new page, or should we try to integrate it into the current tutorial?


bhrutledge (approved):
Thanks!

PR #1033, opened 2021-12-16
Upgrade expected Python version

astrojuanlu:
At the moment the documentation doesn't run on Python 3.6 because sphinx-inline-tabs is incompatible to it, so I upgraded Python to a more recent version. I was still a bit conservative, so happy to bring it up to 3.9 :) not sure if all the dependencies run on 3.10 already.

### DIFF
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -131,9 +131,9 @@ need:
 
       python -m pip install --user nox
 
-2. Python 3.6. Our build scripts are designed to work with Python 3.6 only.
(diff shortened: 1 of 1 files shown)


webknjaz (approved):

webknjaz:
> so happy to bring it up to 3.9 :) not sure if all the dependencies run on 3.10 already.

RTD currently runs 3.8, it's set in the config. I'm pretty sure it'd work with 3.10 too but I don't think there's a compelling case to force the contributors to use the very latest interpreter version.

PR #1037, opened 2022-01-05
When publishing to Pipy,  your username should be __token__

danfunk:
I found I had to add this for the tutorial to work properly.  Otherwise flawless, thank you!

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -31,10 +31,12 @@ jobs:
     - name: Publish distribution 📦 to Test PyPI
       uses: pypa/gh-action-pypi-publish@master
       with:
+        username: __token__
         password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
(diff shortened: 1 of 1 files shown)


webknjaz (changes requested):
Thanks for this but I prefer only specifying the required fields. The reason is that the action defaults to `__token__` and it was designed to only need an explicit username when the end-users actually need to use passwords and not tokens (for example, when they want to upload to something like [devpi].

Are there any reasons not to keep the example simple as it is now?

[devpi]: https://devpi.net/docs/devpi/devpi/6.0/+d/

webknjaz:
> for the tutorial to work properly

If you've faced problems, I don't believe that they were caused by an implicit [`__token__`](https://github.com/pypa/gh-action-pypi-publish/blob/bea5cda/action.yml#L8). But my guess would be that you may have copied the token into the secrets incorrectly (with trailing/leading whitespace or something else missing) — this happens sometimes and is a user error.

danfunk:
I agree that if the default value is __token__, it would better not to set it.
I went back, and everything worked fine this time when I did not set the user name.   Sorry for the trouble.

webknjaz:
No problems :) Voicing your concerns is no trouble at all!

If you want to contribute some tutorial improvements, it may be a good idea to update `@master` to be `@release/v1`, plus maybe bump other action versions, and drop `--sdist`+`--wheel` args from the build command to get the latest improvements it has related to them (or even drop all the args for simplicity).

Ririe91 (commented):
48808377

PR #1042, opened 2022-01-20
Tiny fixup to wording in MAINFEST.in guide

madphysicist:
Seems like a weird place to have "&" instead of "and".


bhrutledge (approved):
Agreed! Thanks!

PR #1047, opened 2022-02-16
Update a stale link to Debian's pip package

brettcannon:


bhrutledge (approved):

brettcannon:
FYI the [linter failure](https://github.com/pypa/packaging.python.org/runs/5209293937?check_suite_focus=true#step:6:730) is due to a 500; [the link](https://github.com/pypa/setuptools/blob/main/setuptools/version.py) does work.

PR #1055, opened 2022-03-04
Update the "Application dependency management" recommendations

pradyunsg:
This is a more detailed recommendation that breaks up the recommendation
into multiple parts, to reflect that multiple tools need to be used in
workflows today to maximise compatibility with existing platforms.

It also recommends that deployed applications be isolated from the
system packages, as was implied but not explicitly noted.

The move away from Pipenv as the primary recommendation is a reflection
of the change that application development workflows have a greater
diversity in mature tooling available for them.

Toward #912

### DIFF
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -14,8 +14,8 @@ Application dependency management
 * Use :ref:`pip` in a `secure manner`_ to install a Python application and its
   dependencies during deployment.
 
-* Use :ref:`virtualenv`, or :doc:`venv <python:library/venv>` to isolate application
-  specific dependencies from a shared Python installation. [4]_
+* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
+  application-specific dependencies from a shared Python installation. [4]_
 
 * Use `pip-tools`_, :ref:`pipenv`, or `poetry`_ to generate the fully-specified
   application-specific dependencies, when developing Python applications.
@@ -32,8 +32,8 @@ Installation tool recommendations
   is installed, you may need to also install :ref:`wheel` to get the benefit
   of wheel caching. [3]_
 
-* Use :ref:`virtualenv`, or :doc:`venv <python:library/venv>` to isolate application
-  specific dependencies from a shared Python installation. [4]_
+* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
+  application-specific dependencies from a shared Python installation. [4]_
 
 * If you're looking for management of fully integrated cross-platform software
   stacks, consider:
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -11,17 +11,18 @@ what tools are currently recommended, then here it is.
 Application dependency management
 =================================
 
-Use :ref:`pipenv` to manage library dependencies when developing Python
-applications. See :doc:`../tutorials/managing-dependencies` for more details
-on using ``pipenv``.
+* Use :ref:`pip` in a `secure manner`_ to install a Python application and its
+  dependencies during deployment.
 
-When ``pipenv`` does not meet your use case, consider other tools like:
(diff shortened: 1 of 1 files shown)


source/guides/tool-recommendations.rst
```diff
@@ -11,17 +11,18 @@ what tools are currently recommended, then here it is.
...
-applications. See :doc:`../tutorials/managing-dependencies` for more details
-on using ``pipenv``.
+* Use :ref:`pip` in a `secure manner`_ to install your distribution
+  packages, when deploying Python applications.
 
-When ``pipenv`` does not meet your use case, consider other tools like:
-
-* :ref:`pip`
+* Use :ref:`virtualenv`, or :doc:`venv <python:library/venv>` to isolate application
+  specific dependencies from a shared Python installation. [4]_
```
merwok, line 18: Now that virtualenv uses venv when possible, maybe mentioning only virtualenv here would be a clearer guideline?

pradyunsg: This mirrors the phrasing in the next section of recommendations. If we want to change this, I'd prefer we do it in a follow up that updates both.

source/guides/tool-recommendations.rst
```diff
@@ -11,17 +11,18 @@ what tools are currently recommended, then here it is.
...
+* Use `pip-tools`_, :ref:`pipenv`, or `poetry`_ to generate the fully-specified
+  application-specific dependencies.
 
-When ``pipenv`` does not meet your use case, consider other tools like:
+* Use :ref:`pip` in a `secure manner`_ to install a Python application's
+  dependencies during deployment.
 
-* :ref:`pip`
+* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
+  application-specific dependencies from a shared Python installation. [4]_
```
bhrutledge, line 21: This item is duplicated in "Installation tool recommendations". Is it necessary here?

bhrutledge (commented):
Thanks for this. I took the liberty of pushing some copy edits, and I have one content question.

source/guides/tool-recommendations.rst
```diff
@@ -11,14 +11,14 @@ what tools are currently recommended, then here it is.
 Application dependency management
 =================================
 
-* Use :ref:`pip` in a `secure manner`_ to install your distribution
-  packages, when deploying Python applications.
+* Use :ref:`pip` in a `secure manner`_ to install a Python application's
```
pradyunsg, line 17: This changed meaning, since this now excludes the application itself instead of including it in this description.

source/guides/tool-recommendations.rst
pradyunsg: Yes. From the commit message:

> It also recommends that deployed applications be isolated from the
system packages, as was implied but not explicitly noted.

source/guides/tool-recommendations.rst
```diff
@@ -11,17 +11,18 @@ what tools are currently recommended, then here it is.
...
 
-Use :ref:`pipenv` to manage library dependencies when developing Python
-applications. See :doc:`../tutorials/managing-dependencies` for more details
-on using ``pipenv``.
+* Use `pip-tools`_, :ref:`pipenv`, or `poetry`_ to generate the fully-specified
+  application-specific dependencies.
 
-When ``pipenv`` does not meet your use case, consider other tools like:
+* Use :ref:`pip` in a `secure manner`_ to install a Python application and its
+  dependencies during deployment.
```
bhrutledge, line 18: @pradyunsg How's this?

pradyunsg: Works! Thanks!

pradyunsg:
@bhrutledge I've rebased this on `main` and squashed your copy edits to the new content (dropping the reordering). You can click on "force-pushed" text in the GitHub event above, to review what exactly has changed across the force-push.

bhrutledge (approved):

PR #1059, opened 2022-03-09
Avoid negative wording when describing Cython.

scoder:


source/guides/packaging-binary-extensions.rst
```diff
@@ -133,8 +133,8 @@ also be considered:
   compile most Python code to C extension modules. The initial compilation
   provides some speed increases (by bypassing the CPython interpreter layer),
   and Cython's optional static typing features can offer additional
-  opportunities for speed increases. Using Cython still has the disadvantage
-  of increasing the complexity of distributing the resulting application,
+  opportunities for speed increases. Using Cython still has the disadvantages
+  that binary extensions generally impose on distributing the resulting application,
```
bhrutledge, line 137: I think this might be slightly clearer:

```suggestion
  opportunities for speed increases. Using Cython still carries the
  `disadvantages`_ associated with using binary extensions,
```

source/guides/packaging-binary-extensions.rst
```diff
@@ -160,10 +160,14 @@ but they *can* significantly reduce the maintenance burden of keeping
...
-  hand, however, and is very repetitive, so may not be a good choice for
-  wrapping large APIs.
+  `Cython <https://cython.org/>`__ is also widely used for creating wrapper
+  modules for C or C++ APIs. It involves wrapping the interfaces by
+  hand, which gives a wide range of freedom in designing and optimising
+  the wrapper code, but may not be a good choice for wrapping very
+  large APIs quickly. See the
+  `list of third-party tools <https://github.com/cython/cython/wiki/AutoPxd>`_
+  for automatic wrapping with Cython. Several Python implementations are
+  supported that provide a CPython-like C-API, most notably PyPy and Pyston.
```
bhrutledge, line 170: Assuming this is accurate, I think this is clearer:

```suggestion
  for automatic wrapping with Cython. It also supports performance-oriented
  Python implementations that provide a CPython-like C-API, such as PyPy
  and Pyston.
```

Also, maybe it's worth linking to PyPy and Pyston?

scoder: > Also, maybe it's worth linking to PyPy and Pyston?

Well, there is a whole paragraph on PyPy right in the section above. People who don't bother looking up the acceleration section will probably already know the options, or will not care.

bhrutledge (approved):
Thanks for taking the time to do this.

henryiii:
Distributing a Cython application does increase the complexity of distribution further than a vanilla extension or pybind11, since it requires a pre-process step or requires vending a multithousand line c/c++ file in that will likely need to be regenerated on the next Python version with a new Cython release, and requires the whole oldest-support-numpy issues. Cython was really not designed for wrapping code.

I'm not against the change at all, but just guessing why the originally language was the way it was, and maybe the new language is just a tiny bit too nice. :P

henryiii (approved):

scoder:
I think the main difference between Cython and pybind11 is just that one requires users to properly know Python and a bit of C (more or less, depending on what you want to do with it), whereas the other requires proper C++ knowledge. For the rest, they pretty much do the same thing, and we shouldn't try to push one over the other. In a Python world, you'd naturally expect more users to have a good understanding of Python than a good (enough) understanding of C++. But that's really just a matter of personal preference and available skills. Not of capabilities (or "complexity" or "designed-for-ness") of the tools. I really fail to see a difference there.

PR #1061, opened 2022-03-11
Including pep658 in the simple API description

Wouterkoorn:
The relevant PEP's to the Simple Repository API are listed, but I'm under the impression that pep658 should also be included.

I'm not too proud about the phrasing, so happy for it to be changed

### DIFF
--- a/source/specifications/simple-repository-api.rst
+++ b/source/specifications/simple-repository-api.rst
@@ -8,5 +8,6 @@ Simple repository API
 The current interface for querying available package versions and
 retrieving packages from an index server is defined in :pep:`503`,
 with the addition of "yank" support (allowing a kind of file deletion)
-as defined in :pep:`592` and specifying the interface version provided
-by an index server in :pep:`629`.
+as defined in :pep:`592`, specifying the interface version provided
+by an index server in :pep:`629` and optionally providing metadata 
+separated from the distrobution package as stated in :pep:`658`.


source/specifications/simple-repository-api.rst
```diff
@@ -8,5 +8,6 @@ Simple repository API
 The current interface for querying available package versions and
 retrieving packages from an index server is defined in :pep:`503`,
 with the addition of "yank" support (allowing a kind of file deletion)
-as defined in :pep:`592` and specifying the interface version provided
-by an index server in :pep:`629`.
+as defined in :pep:`592`, specifying the interface version provided
+by an index server in :pep:`629` and optionally providing metadata 
+separated from the distrobution package as stated in :pep:`658`.
```
bhrutledge, line 13: How about something like this?

```suggestion
in :pep:`592`, specifying the interface version provided
by an index server in :pep:`629`, and retrieving packaging metadata 
in :pep:`658`.
```

bhrutledge (changes requested):
Thanks for adding this. I didn't know this was possible. 😉

source/specifications/simple-repository-api.rst
Wouterkoorn: A lot better!
I do question,  with it being so simple, does it properly reflect the scope of PEP 658?
Also, aren't the other PEP summaries given from a server-side perspective, and you describe PEP 658 from a client side's perspective (receiving instead of providing)?

bhrutledge (approved):

PR #1066, opened 2022-03-28
Have sdists normalize the version in the file name

brettcannon:
Brought up as part of https://github.com/pypa/packaging/issues/527 .


bhrutledge (approved):

Samitha211 (commented):
6

woodruffw:
As a note: does PyPI or any other PEP 503-compliant host currently do the `-` to `_` normalization? The standard here suggests that `foo-bar==1.2.3` should be normalized to `foo_bar-1.2.3.tar.gz`, but here's how PyPI's simple index supplies `pip-audit`'s source distributions ([link](https://pypi.org/simple/pip-audit/)):

```
pip-audit-2.1.0.tar.gz
pip-audit-2.1.1.tar.gz
pip-audit-2.2.0.tar.gz
pip-audit-2.2.1.tar.gz
```

(The current behavior is clearly working since `pip` has a workaround for the "vexing parse" case in https://github.com/pypa/packaging/issues/527, but I wonder if it makes sense to amend this documentation again to emphasize that the normalization suggested currently isn't practiced.)

woodruffw:
I just realized that my above comment is slightly off topic from the PR (since it's about the name component, not the version), so I'll create a separate issue to follow up.

Issue #1084, opened 2022-05-31
Is it time to update the packaging tutorial to remove setup.py and setup.cfg


venthur:
The [tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/) still [mentions](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata) that either have to use setup.py or setup.cfg to store your metadata. 

> Configuring metadata
>
> There are two types of metadata: static and dynamic.
>
>    Static metadata (setup.cfg): guaranteed to be the same every time. This is simpler, easier to read, and avoids many common errors, like encoding errors.
>
>    Dynamic metadata (setup.py): possibly non-deterministic. Any items that are dynamic or determined at install-time, as well as extension modules or extensions to setuptools, need to go into setup.py.
>
> Static metadata (setup.cfg) should be preferred. Dynamic metadata (setup.py) should be used only as an escape hatch when absolutely necessary. setup.py used to be required, but can be omitted with newer versions of setuptools and pip.

I believe this is not true anymore and all of those can now be stored in pyproject.yaml. I wonder if this can be fixed now or if there's a timeline or a required milestone before setup.py and setup.cfg can be removed in favour for pyproject.toml.

PR #1031, opened 2021-12-12
feat: add PEP 621 config only

henryiii:
This is a simpler, PEP 621 _only_ version of #1030. Might not be ready to go this route, but this would be what I think we could do.

Also, I'd love to move this over to using pipx. That would really simplify the ending commands (building and publishing), and would remove the unix/windows split from everything except the first line, where pipx would be installed.

_Update from @bhrutledge_:

Merging is currently blocked on changing the example package name to support backend discovery; see discussion in https://github.com/pypa/packaging.python.org/pull/1031#discussion_r891096802. Until that's fixed, the tutorial is broken. Further content improvements can be made in future PRs. Closes #1085.

PR #1085, opened 2022-05-31
remove setup.py and setup.cfg in packaging tutorial in favor of pyproject.toml

venthur:
This PR, removes `setup.py` and `setup.cfg` and uses `pyproject.toml` for storing the meta data in the packaging tutorial.

closes: #1084

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -182,8 +182,8 @@ uses a relatively minimal set of ``metadata``:
 - ``project.urls`` lets you list any number of extra links to show on PyPI.
   Generally this could be to documentation, issue trackers, etc.
 
-There are many more than the ones mentioned here. See
-:doc:`/guides/distributing-packages-using-setuptools` for more details.
+There are many more options than the ones mentioned here. See
+https://peps.python.org/pep-0621/ for more details.
 
 
 Creating README.md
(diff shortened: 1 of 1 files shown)


bhrutledge:
Thanks, @venthur! This looks like a good start. However, it conflicts with #1031, which goes further to demonstrate how folks can use different build backends. There's a lively unresolved discussion about that approach, and which backend should be listed first. I think that PR is waiting on input/edits from myself and the author; I haven't been able to prioritize that recently.

FYI @abravalheri @henryiii @ofek. I'd like to avoid philosophical discussions on this PR, and focus on its technical accuracy, if we decide to just use setuptools. That's a big "if"; I'm tempted to resolve it via the [packaging discussion forum](https://discuss.python.org/c/packaging/14).

bhrutledge:
In an attempt to get more feedback, I've started [a discussion on the packaging forum](https://discuss.python.org/t/removing-setup-cfg-and-setup-py-from-the-packaging-tutorial/16096).

CAM-Gerlach:
Just to note, while I personally prefer the #1031 approach for the reasons others have articulated on the thread and it seems that's the way things are going, if this does get accepted, you'll need to require at least `setuptools>=61.0` in the `build-system.requires` section above, if not higher (possibly not yet released) to ensure full stability.

venthur:
@CAM-Gerlach thanks for the heads-up. I updated the requirements to setuptools>=61.

bhrutledge:
Closing this in favor of #1031.

PR #1090, opened 2022-06-16
fix: pip broke a intersphinx reference

henryiii:
CC @pradyunsg. Current `main` is broken because pip broke references to `pip:repeatability`. Using `pip:0-repeatability` as a fix here, as that's the only thing that seems to work.

### DIFF
--- a/source/discussions/install-requires-vs-requirements.rst
+++ b/source/discussions/install-requires-vs-requirements.rst
@@ -72,7 +72,7 @@ the requirements for a complete Python environment.
 
 Whereas ``install_requires`` requirements are minimal, requirements files
 often contain an exhaustive listing of pinned versions for the purpose of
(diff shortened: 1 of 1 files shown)


pradyunsg:
Please don’t merge this as-is. You’re linking to a rubric that states that the content has been moved.

Instead, We should update the link to point to the “repeatable installs” page instead. Alternatively, we can add a `repeatability` intersphinx reference to the top of that document in pip’s end, which is probably fine as well.

x-ref: https://github.com/pypa/pip/pull/10913

pradyunsg:
CI passed the build in https://github.com/pypa/packaging.python.org/pull/1031, and the link check seems to be unrelated to pip's documentation. Closing this since the underlying cause is fixed.

Issue #926, opened 2021-06-25
Promoting Brian Rutledge as maintainer of the PUG


webknjaz:
@bhrutledge has been exceptionally active in this repository in recent months and sent some contributions a year ago too.

I propose to promote Brian to a maintainer because I believe he can be a huge help in moving the PUG forward. His past PRs have always been well-thought and he is very collaborative when something needs discussing or polishing.

He's also mentioned that he is interested in stepping up @ https://discuss.python.org/t/whats-the-best-way-to-help-with-packaging-python-org/9387/2?u=webknjaz.

cc @pypa/packaging-user-guide-editors

bhrutledge:
Thanks for the vote of confidence, @webknjaz.  😊

theacodes:
👍

brainwane:
Yeah, this sounds good to me!

pradyunsg:
Added Brian to the relevant GitHub team!

PR #1091, opened 2022-06-21
Fix tutorial 'frontend' typo

bhrutledge:


henryiii (approved):

bhrutledge:
@pypa/packaging-user-guide-editors It looks like there's some configuration that prevents me (as a maintainer) from merging this PR without an approving review (which I can't give because I'm the PR author). I've requested a review from the group, but I'm wondering if this is the desired configuration. IE, would it be acceptable for maintainers to merge their own PRs?

cc @pradyunsg, because you seem to have some authority in here (per https://github.com/pypa/packaging.python.org/issues/926#issuecomment-876691583).

pradyunsg (approved):

pradyunsg:
I'll take a closer look in the evening, after my work day. Could you file an issue for this, to give this a clear place for discussion?

PR #1092, opened 2022-06-21
chore: update GHA guide

henryiii:
This updates the GHA guide a bit with non-controversial changes (hopefully). I think using pipx instead of the current setup would be simpler and is better practice, I might do that in a followup.

Changes:
* Point out this works only for pure Python, link to cibuildwheel for binary packages
* Use Python 3.x
* Drop mention about needed pyproject.toml to use build and links to PEPs (it is backward compatible if it's missing)
* Use more recent versions of actions and the tagged versions of all actions

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -8,11 +8,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@master
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v3
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.x"
     - name: Install pypa/build
       run: >-
         python -m
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -14,6 +14,8 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
 
    This guide *assumes* that you already have a project that
    you know how to build distributions for and *it lives on GitHub*.
+   This guide also assumes you have a pure Python project. If you
+   have binary compoents, check out :ref:`cibuildwheel`.
 
 Saving credentials on GitHub
 ============================
@@ -94,9 +96,7 @@ This will download your repository into the CI runner and then
 install and activate Python 3.10.
 
 And now we can build dists from source. In this example, we'll
-use ``build`` package, assuming that your project has a
-``pyproject.toml`` properly set up (see
-:pep:`517`/:pep:`518`).
+use ``build`` package.
 
 .. tip::
 
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -12,10 +12,10 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
 
 .. attention::
 
-   This guide *assumes* that you already have a project that
-   you know how to build distributions for and *it lives on GitHub*.
-   This guide also assumes you have a pure Python project. If you
-   have binary compoents, check out :ref:`cibuildwheel`.
+   This guide *assumes* that you already have a project that you know how to
+   build distributions for and *it lives on GitHub*.  This guide also avoids
+   details of building platform specific projects. If you have binary
+   components, check out :ref:`cibuildwheel`'s GitHub Action examples.
 
 Saving credentials on GitHub
 ============================
@@ -102,13 +102,15 @@ use ``build`` package.
 
    You can use any other method for building distributions as long as
    it produces ready-to-upload artifacts saved into the
-   ``dist/`` folder.
+   ``dist/`` folder. You can even use ``actions/upload-artifact`` and
+   ``actions/download-artifact`` to tranfer files between jobs or make them
+   accessable for download from the web CI interface.
 
 So add this to the steps list:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: version: "3.10"
+   :start-after: version: "3.x"
    :end-before: Actually publish to PyPI/TestPyPI
 
 
@@ -124,7 +126,9 @@ Finally, add the following steps at the end:
 These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
 Action: the first one uploads contents of the ``dist/`` folder
 into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged.
+PyPI, but only if the current commit is tagged. It is recommended
+you use the latest release tag; a tool like GitHub's dependabot can keep
+these updated regularly.
 
 
 That's all, folks!


source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -90,12 +92,10 @@ Then, add the following under the ``build-n-publish`` section:
    :end-before: Install pypa/build
 
 This will download your repository into the CI runner and then
-install and activate Python 3.9.
+install and activate Python 3.10.
 
 And now we can build dists from source. In this example, we'll
-use ``build`` package, assuming that your project has a
-``pyproject.toml`` properly set up (see
-:pep:`517`/:pep:`518`).
```
webknjaz, line 98: I'd like to keep these.

webknjaz (commented):
I've put some suggestions below plus a portion of these changes has been merged via another recent PR making conflicts resolution necessary too.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -14,6 +14,8 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
 
    This guide *assumes* that you already have a project that
    you know how to build distributions for and *it lives on GitHub*.
+   This guide also assumes you have a pure Python project. If you
+   have binary compoents, check out :ref:`cibuildwheel`.
```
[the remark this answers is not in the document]
henryiii: I wanted to clarify that you can't follow this guide as is if you depend on the system architecture or if you depend on Python version. You can't follow this guide if you have binaries - you literally can't upload a `linux` wheel, it has to be processed into manylinux/musllinux for PyPI to accept it. The best PyPA example I have is the cibuildwheel GitHub Actions examples. If it's meant to be general, I think it should at split into two jobs, a build job and a publish job using artifacts. (This is excatly what I do in https://scikit-hep.org/developer/gha_pure for pure python wheels). Then the build job could be replaced with anything else if it's not a pure Python example, and the rest would still be applicable. This simple example doesn't generalize well to that.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -29,12 +29,12 @@ jobs:
         .
     # Actually publish to PyPI/TestPyPI
     - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@v1.5.0
```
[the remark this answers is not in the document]
henryiii: I wish this was just "v1", since that's a standard across the official actions. :(

In scikit-hep.org/developer I've set up nox jobs to automatically update the specific pins weekly for GHA (and pre-commit), if that's interesting.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
henryiii: The statement was wrong as written (`build` builds wheels just fine if there is no `pyproject.toml` too). And I thought we were discouraging linking to PEP's?

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -29,12 +29,12 @@ jobs:
...
     - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@release/v1
       with:
         password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@v1.5.0
```
henryiii, line 38:
```suggestion
      uses: pypa/gh-action-pypi-publish@release/v1
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
henryiii: I tried a second rewrite, still in the same vein, but maybe better?

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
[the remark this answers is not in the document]
webknjaz: They do use this but I think that it's a huge misuse of Git tags. GH basically encourages force-pushing tags when they are not designed for this.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
webknjaz: > I think it should at split into two jobs, a build job and a publish job using artifacts.

Yes, this is what I've been recommending in the action's repository. In fact, it's been on my list to create a separate guide explaining a more complex setup with this in mind.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: Ah, I must've missed that.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -94,15 +96,15 @@ This will download your repository into the CI runner and then
 install and activate Python 3.10.
 
 And now we can build dists from source. In this example, we'll
-use ``build`` package, assuming that your project has a
-``pyproject.toml`` properly set up (see
-:pep:`517`/:pep:`518`).
+use ``build`` package.
```
webknjaz, line 99: I feel like it could still be important to have a troubleshooting pointer here. WDYT?
```suggestion
use ``build`` package which may require ``pyproject.toml`` configuration depending on your build back-end choice.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -94,15 +96,15 @@ This will download your repository into the CI runner and then
...
 
 .. tip::
 
    You can use any other method for building distributions as long as
    it produces ready-to-upload artifacts saved into the
-   ``dist/`` folder.
+   ``dist/`` folder. You can even use ``actions/upload-artifact`` and
+   ``actions/download-artifact`` to tranfer files between jobs or make them
+   accessable for download from the web CI interface.
```
webknjaz, line 107:
```suggestion
   accessible for download from the web CI interface.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -124,7 +126,9 @@ Finally, add the following steps at the end:
 These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
 Action: the first one uploads contents of the ``dist/`` folder
 into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged.
+PyPI, but only if the current commit is tagged. It is recommended
+you use the latest release tag; a tool like GitHub's dependabot can keep
+these updated regularly.
```
webknjaz, line 131: This is untrue wrt the pypi-publish action tho.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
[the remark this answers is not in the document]
henryiii: Why can’t it be a branch? Is there anything stopping a `v1` branch? Dependabot might not handle a branch correctly? Otherwise it would work.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
henryiii: Why? I do this all the time. Like https://github.com/tomerfiliba/plumbum/pull/615

I’d generally recommend this on all non official actions (official ones shouldn’t break often & would be noisy to continually update - depends on how important always working is).

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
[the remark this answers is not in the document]
webknjaz: It was a conscious choice to use distinct branch naming that is separate from the tag naming traditions. The version tags follow that pattern, so Dependabot should be able to send updates. But it's different for release branches. If one opted for a moving pointer, they clearly don't want pinning.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: Hm... I guess I've misread the explanation here. Do you think we could make it a bit clearer? Maybe inject some "for example, `@v1.5.1`"?

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
henryiii: It doesn't require anything that pip doesn't require. It falls back to `setup.py` in the same way. This seems to imply to me that it does require `pyproject.toml`.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
henryiii:
```suggestion
you use the latest release tag like `@v1.5.1`; a tool like GitHub's dependabot
can keep these updated regularly.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: That's why I used *may*, noting that it may be the case but also may be not.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -124,7 +126,9 @@ Finally, add the following steps at the end:
 These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
 Action: the first one uploads contents of the ``dist/`` folder
 into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged.
+PyPI, but only if the current commit is tagged. It is recommended
+you use the latest release tag like `@v1.5.1`; a tool like GitHub's dependabot
+can keep these updated regularly.
```
webknjaz, line 131: I think this would be better:
```suggestion
is able to keep these updated regularly.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -124,7 +126,9 @@ Finally, add the following steps at the end:
 These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
 Action: the first one uploads contents of the ``dist/`` folder
 into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged.
+PyPI, but only if the current commit is tagged. It is recommended
+you use the latest release tag like `@v1.5.1`; a tool like GitHub's dependabot
```
henryiii, line 130:
```suggestion
you use the latest release tag like ``@v1.5.1``; a tool like GitHub's dependabot
```

😳

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -8,9 +8,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v3
     - name: Set up Python 3.10
```
hugovk, line 12: Future proofing:

```suggestion
    - name: Set up Python
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -8,9 +8,9 @@ jobs:
...
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
         python-version: "3.10"
```
hugovk, line 15: Future proofing:

```suggestion
        python-version: "3.x"
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -94,15 +96,15 @@ This will download your repository into the CI runner and then
 install and activate Python 3.10.
 
 And now we can build dists from source. In this example, we'll
-use ``build`` package, assuming that your project has a
-``pyproject.toml`` properly set up (see
-:pep:`517`/:pep:`518`).
+use ``build`` package which may require ``pyproject.toml`` configuration depending on your build back-end choice.
```
hugovk, line 99:
```suggestion
use the ``build`` package which may require ``pyproject.toml`` configuration depending on your build back-end choice.
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -8,11 +8,11 @@ jobs:
...
     steps:
-    - uses: actions/checkout@master
-    - name: Set up Python 3.10
-      uses: actions/setup-python@v3
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
       with:
-        python-version: "3.10"
+        python-version: "3.x"
```
hugovk, line 15: To fix the CI, need to update line 113 in `source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst` to match:

```diff
-:start-after: version: "3.10"
+:start-after: version: "3.x"
```

henryiii: Thanks!

PR #1094, opened 2022-06-22
chore: remove obsolete pages from the ToC

henryiii:
Discussed at https://discuss.python.org/t/cleaning-up-packaging-python-org-guides/16717 - if everyone there and here is happy with it, then these pages can retire.

This removes four pages that don't have an obvious path forward from the navigation. The URLs stay active (and we have to keep the links alive, etc).

### DIFF
--- a/source/guides/index.rst
+++ b/source/guides/index.rst
@@ -13,7 +13,6 @@ introduction to packaging, see :doc:`/tutorials/index`.
    installing-stand-alone-command-line-tools
    installing-using-linux-tools
    installing-scientific-packages
-   multi-version-installs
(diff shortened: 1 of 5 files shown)


ofek (approved):

bhrutledge (approved):
Good start to tidying things up. Thanks!

webknjaz:
@henryiii should those pages be marked as orphaned, then?

henryiii:
I thought I did mark them as orphaned?

webknjaz:
Oh, I guess I missed that in the diff. Sorry.

PR #1098, opened 2022-06-23
Declare license file in packaging tutorial

### LINKED ISSUE #1097: Declare license file in packaging tutorial
In the [packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata), a license file is included in the example (and rice classifiers), but not declared in the metadata (`license.file` key)

EpicWink:
Set and document `license.file` in project metadata in the packaging tutorial

Fixes #1097

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -217,7 +217,7 @@ following this tutorial.
   In this case, the description is loaded from :file:`README.md` (which is a
   common pattern). There also is a more advanced table form described in the
   :ref:`project metadata specification <declaring-project-metadata>`.
-- ``license.file`` USA path to the license file :file:`LICENSE`, described
+- ``license.file`` is the path to the license file :file:`LICENSE`, described
   below. See the project metadata specification for more advanced usage.
 - ``requires-python`` gives the versions of Python supported by your
   project. Installers like :ref:`pip` will look back through older versions of
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -217,8 +217,7 @@ following this tutorial.
   In this case, the description is loaded from :file:`README.md` (which is a
   common pattern). There also is a more advanced table form described in the
   :ref:`project metadata specification <declaring-project-metadata>`.
-- ``license.file`` is the path to the license file :file:`LICENSE`, described
-  below. See the project metadata specification for more advanced usage.
+- ``license`` is the path to the :file:`LICENSE` file, described below.
 - ``requires-python`` gives the versions of Python supported by your
   project. Installers like :ref:`pip` will look back through older versions of
   packages until it finds one that has a matching Python version.
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -188,6 +188,7 @@ following this tutorial.
     ]
     description = "A small example package"
     readme = "README.md"
+    license = { file="LICENSE" }
     requires-python = ">=3.7"
     classifiers = [
         "Programming Language :: Python :: 3",
@@ -216,6 +217,8 @@ following this tutorial.
   In this case, the description is loaded from :file:`README.md` (which is a
   common pattern). There also is a more advanced table form described in the
   :ref:`project metadata specification <declaring-project-metadata>`.
+- ``license.file`` USA path to the license file :file:`LICENSE`, described
+  below. See the project metadata specification for more advanced usage.
 - ``requires-python`` gives the versions of Python supported by your
   project. Installers like :ref:`pip` will look back through older versions of
   packages until it finds one that has a matching Python version.


source/tutorials/packaging-projects.rst
```diff
@@ -188,6 +188,7 @@ following this tutorial.
     ]
     description = "A small example package"
     readme = "README.md"
+    license = { file="LICENSE" }
```
bhrutledge, line 191: I can see that this is defined in the [project metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/). However, it looks like the specification is likely to change with [PEP 639](https://peps.python.org/pep-0639/) (currently a draft), which includes an [appendix](https://peps.python.org/pep-0639/#appendix-license-documentation-in-python) that notes the inconsistencies.

Out of curiosity, I built the same package with a) different backends and b )with and without `license.files`. In the snippets below, the patch version identifies the backend (`0.0.1` for Hatchling, `0.0.2` for setuptools, `0.0.3` for Flit, and `0.0.4` for PDM), `rc1` indicates the absence of `license.files`, and `rc2` indicates its presence.

In this snippet, we see that all backends include `LICENSE` in the sdist and wheel, _except_ for Flit when `license.files` is absent:

```
% fd LICENSE

./example_package_bhrutledge-0.0.1rc1/LICENSE
./example_package_bhrutledge-0.0.1rc1.dist-info/license_files/LICENSE
./example_package_bhrutledge-0.0.1rc2/LICENSE
./example_package_bhrutledge-0.0.1rc2.dist-info/license_files/LICENSE

./example_package_bhrutledge-0.0.2rc1/LICENSE
./example_package_bhrutledge-0.0.2rc1.dist-info/LICENSE
./example_package_bhrutledge-0.0.2rc2/LICENSE
./example_package_bhrutledge-0.0.2rc2.dist-info/LICENSE

./example_package_bhrutledge-0.0.3rc2/LICENSE
./example_package_bhrutledge-0.0.3rc2.dist-info/LICENSE

./example-package-bhrutledge-0.0.4rc1/LICENSE
./example_package_bhrutledge-0.0.4rc1.dist-info/license_files/LICENSE
./example-package-bhrutledge-0.0.4rc2/LICENSE
./example_package_bhrutledge-0.0.4rc2.dist-info/license_files/LICENSE
```

In this snippet, we see that _only_ Hatchling and setuptools include the `LICENSE` text in the package metadata:

```
% rg -l 'License: Copyright'

example_package_bhrutledge-0.0.1rc2/PKG-INFO
example_package_bhrutledge-0.0.1rc2.dist-info/METADATA

example_package_bhrutledge-0.0.2rc2/PKG-INFO
example_package_bhrutledge-0.0.2rc2.dist-info/METADATA
```

So, I'm not sure what the right thing is to do here. On the one hand, it seems harmless include, but it doesn't consistently include the text, and at present there doesn't seem to be a backend-agnostic way to do that.

@CAM-Gerlach @brettcannon Any suggestions?

Also FYI backend maintainers @ofek @abravalheri @takluyver

source/tutorials/packaging-projects.rst
```diff
@@ -216,6 +217,8 @@ following this tutorial.
   In this case, the description is loaded from :file:`README.md` (which is a
   common pattern). There also is a more advanced table form described in the
   :ref:`project metadata specification <declaring-project-metadata>`.
+- ``license.file`` is the path to the license file :file:`LICENSE`, described
+  below. See the project metadata specification for more advanced usage.
```
bhrutledge, line 221: I think it's okay to be keep this succinct:

```suggestion
- ``license`` is the path to the :file:`LICENSE` file, described below.
```

source/tutorials/packaging-projects.rst
abravalheri: From the point of view of `setuptools` this change should be fine. In end of the day, more license information cannot harm, right?

Keeping an eye on PEP 639 on the horizon, does it make sense to introduce it in the tutorial now considering that in some months PEP 639 might be out?

ofek: > we see that only Hatchling and setuptools include the LICENSE text in the package metadata

I'm confused; what do the others write if not the text?

source/tutorials/packaging-projects.rst
EpicWink: I agree we can drop the second sentence, as it will be replaced with PEP 639 reference when it comes out.

I don't like not referring to the `files` subkey

source/tutorials/packaging-projects.rst
EpicWink: Even just for Flint's benefit, explicitly specifying the license is important to ensure the license file is included in the package

source/tutorials/packaging-projects.rst
bhrutledge: Since all of the other items in this list refer just to the metadata attribute, rather than its contents, and that  `authors` doesn't refer directly to the dictionary keys, I'm going to leave it as `license`.

bhrutledge (approved):
Thanks for filling this gap!

PR #1106, opened 2022-07-09
Match the actual status quo in the wheel spec

dstufft:


dstufft:
Closing this as Paul has asked for any future changes to go through the PEP process, and I think having an open PR without an attached PEP, when I don't currently intend to write said PEP is confusing for everyone involved.

Issue #1093, opened 2022-06-21
Ensure that all current Python packaging interoperability standards are on packaging.python.org


pradyunsg:
This is a matter of someone going through https://peps.python.org/topic/packaging/, identifying all accepted and final PEPs, and converting their content into [reference material](https://diataxis.fr/reference/) on this documentation.

This would be content that lives under https://packaging.python.org/en/latest/specifications/, although it will also need to be updated to be the canonical reference -- with consistent structure across all these pages as well as no "see PEP XYZ for details" from these.

pradyunsg:
x-ref: https://github.com/pypa/packaging.python.org/issues/955
x-ref: https://github.com/pypa/packaging.python.org/issues/679
x-ref: https://github.com/pypa/packaging.python.org/issues/764

pradyunsg:
x-ref: https://github.com/pypa/packaging.python.org/issues/765

CAM-Gerlach:
I've begun work on and hope to submit a PR this week with a draft of the migrated specifications defined in PEP 517/518/660 following the [outline I proposed](https://github.com/pypa/packaging.python.org/issues/955#issuecomment-1161046484) in #955 , presuming no one has major objections to the overall approach (though I've given due consideration to such alternatives, it is relatively flexible to adapt if reviewers feel that, e.g. "Declaring a project's build system" should be a subsection of "Build backend interface specification", or that the hooks in the latter should be organized by required/optional rather than build type).

That seems to be the highest priority missing chunk of specs/PEPs, but there are a few other major ones:

* [ ] [Simple repository API](https://packaging.python.org/en/latest/specifications/simple-repository-api/) in PEP 503, PEP 592, PEP 629 and PEP 658, though hopefully part of the implementation of the JSON API PEP 691 can be moving over the specification of the existing simple API as well as the new JSON API to packaging.python.org.
* [ ] [Dependency specifiers](https://packaging.python.org/en/latest/specifications/dependency-specifiers/) in PEPs 345, 440, 508 and now 685
* [ ] [Version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/) in PEP 345 and PEP 440
Those are the pages with current stubs and brief summaries on the PyPA specs page.

Other potentially relevant packaging PEPs that are lower priority but still declare a formal specification and are possible candidates for migration here include:

[...]

Issue #1101, opened 2022-06-28
No clear list of permitted keys in "Declaring project metadata" specification


JDLH:
The [*Declaring project metadata* specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata) has an important task of listing the field names which are permitted in the `[project]` table of the `pyproject.toml` file. (It also has other tasks, such as defining the permitted types for values of those fields, the specific semantics of the key-value entries, and references to other specifications such as PEP508.) However, as I read the document text, it does not contain a clear and straightforward list of those keys. 

**Observed behaviour**: the document defines 10 field names by formatting those strings as monospace (` `` `) and headings (``===``), for example "`readme`".  The document defines two pairs of field names by formatting those strings as a pair of monospace strings, separated by a non-monospace slash ("/") character, all formatted as a heading (for example, "`authors`/`maintainers`"). Unless one reads carefully, it easy to mistake these pairs as a single field name which includes a slash character, i.e. the name `authors/maintainers`.  The document defines three names — actually subtable names rather than field names — by listing them within the text of a line which is labelled as "TOML type" (for example, "`scripts`"). 
 
**Desired enhancement**: add a section to the document with a list of field names. This could be a bullet point list. It could also be a table, with a row for each permitted field, and columns to give the field name, the TOML type, and perhaps a summary. This section would be located above the sections describing each field's semantics, i.e. before the [`name` section](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#name).  

**Discussion**: 

I believe this is a "readability improvements that [doesn’t] affect software interoperability", and so it is appropriate to "implement[] using the Python Packaging User Guide’s standard pull request based workflow". I think this means I don't need to write a PEP. Hence this GitHub Issue, which I can follow up with a Pull Request if there is interest.

The two pairs of names, defined in a single section heading, were probably combined for brevity. Their types are identical, and their semantics are very similar. It would also be an improvement to reword those headings, replacing the slash by clearer wording. For instance, change heading text "`authors`/`maintainers`" to read "`authors` and `maintainers` fields". Better yet, make the existing section be about just the first name of the pair, and add a second section with the second name of the pair, and brief body text saying "this is just like the previous section, but instead of X use Y." This improvement could be added to the scope of this Issue, or saved for a separate Issue. 

[...]

PR #1111, opened 2022-07-25
Add PEP 517, PEP 518 & PEP 660 to specs and glossary

### LINKED ISSUE #955: Bring over PEPs 517, 518, and 660 to the specs section
As of right now various specifications reference those PEPs but there is no complete view of what  they contain (e.g. the sdist spec references PEP 518 for what `pyproject.toml` although technically it should  also reference PEP 517 for being to actually build an sdist from a source tree). We should probably have a single specification that covers how to build sdists and wheels from a source tree which the sdist and wheel format specs can reference on how to end up with those binary artifacts.

CAM-Gerlach:
@pfmoore @brettcannon @pradyunsg 

As discussed on #955 , this (finally!) migrates the content of PEP 517, PEP 518 and PEP 660 to the specifications section and glossary, and updates the various references throughout the site, specs and glossary to the PEPs, their contents and the terms they define to refer to, and accurately reflect, the newly canonical content spec and glossary content.

Overall, while I did put a lot of effort into diligently taking advantage of the appropriate reST syntax/directives/roles and Sphinx cross referencing for consistent source formatting and output rendering, implementing editorial corrections for organization, clarity and consistency, and making some limited and motivated additions and updates to the non-normative portions (mostly added introductory/connecting material and updating/fleshing out examples), I attempted to be conservative and avoid any substantive changes to the actual content of the normative specification, which can be discussed and applied subsequent to this initial PR.

On the high level, they end up most neatly fitting into three separate specs:
* A brief new summary meta spec covering the `pyproject.toml` file as a whole, containing the high-level content from PEP 518 (as well as a bit from PEP 621/the declaring project metadata spec), since it services multiple purposes beyond just the original build-system table, so it wouldn't make much sense nor be reader-friendly to bury it inside a build interface spec. It links to the individual specs for each table (which also helps address the confusion and concerns of multiple recent readers who had difficulty finding it and its subtables, since the previous names were highly non-obvious to them).
* Broadening the existing "Declaring build system dependencies" specification to cover the `[build-system]` table as a whole, containing the `[build-system]` content from PEP 518 and the build-system top-level section from PEP 517. This section includes a lot more than just the `requires` key so it should not be limited to it, but can fairly cleanly be separated from the actual build interface specification itself, and also helps partition the content of potential interest to regular package authors versus build frontend/backend developers
* Finally, creating a new Build Frontend-Backend Interface specification that covers the hooks and the specified behavior of frontends and backends themselves from PEP 517 and PEP 660, as would be consulted by frontend and backend developers.

I've made some relatively modest adjustments to the [outline originally proposed in #955](https://github.com/pypa/packaging.python.org/issues/955#issuecomment-1161046484):

<details>

<summary>Final outline w/ mapping from existing PEP sections (click to expand)</summary>

* _The `pyproject.toml` configuration file_
  * FIle format and location ([PEP 518 section](https://peps.python.org/pep-0518/#file-format))
  * `build-system` table (high-level statement of purpose from the [PEP 518 section](https://peps.python.org/pep-0518/#build-system-table); links to _Specifying a project's build system_ spec for the actual details)
  * `project` table (very brief statement of purpose, which links to the [_Declaring project [source] metadata_ spec](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/))
  * `tool` table ([PEP 518 section](https://peps.python.org/pep-0518/#tool-table))
* _Declaring a project's build system_
  * `requires` key ([PEP 518 build-system section](https://peps.python.org/pep-0518/#build-system-table) and [PEP 517 Build Requirements section](https://peps.python.org/pep-0517/#build-requirements)
  * `build-backend` key ([PEP 517 Source Trees section](https://peps.python.org/pep-0517/#source-trees) [which only contains ≈1 sentence about source trees])
  * `backend-path` key ([PEP 517 In-tree Build Backends section](https://peps.python.org/pep-0517/#in-tree-build-backends))
* _Build front-end backend interface_
  * Wheel hooks
    * `build_wheel` (mandatory) ([PEP 517 section](https://peps.python.org/pep-0517/#build-wheel))
    * `get_requires_for_build_wheel` ([PEP 517 section](https://peps.python.org/pep-0517/#get-requires-for-build-wheel))
    * `prepare_metadata_for_build_wheel` ([PEP 517 section](https://peps.python.org/pep-0517/#get-requires-for-build-wheel))
  * Sdist hooks
    * `build_sdist` (mandatory) ([PEP 517 section](https://peps.python.org/pep-0517/#build-sdist))
    * `get_requires_for_build_sdist` ([PEP 517 section](https://peps.python.org/pep-0517/#get-requires-for-build-sdist))
  * Editable hooks
    * `build_editable` ([PEP 660 section](https://peps.python.org/pep-0660/#build-editable))
    * `get_requires_for_build_editable` ([PEP 660 section](https://peps.python.org/pep-0660/#get-requires-for-build-editable))
    * `prepare_metadata_for_build_wheel` ([PEP 660 section](https://peps.python.org/pep-0660/#prepare-metadata-for-build-editable))
  * Config settings ([PEP 517 section](https://peps.python.org/pep-0517/#config-settings))
  * Build environment ([PEP 517 build environment section](https://peps.python.org/pep-0517/#build-environment)) and semi-normative portion of ([PEP 517 forntend recommendations section](https://peps.python.org/pep-0517/#recommendations-for-build-frontends-non-normative))
  * Frontend requirements for editable installs ([PEP 660 Frontend Requirements section](https://peps.python.org/pep-0660/#frontend-requirements))
* [_PyPA Glossary_](https://packaging.python.org/en/latest/glossary/) (for terminology, linked in the above documents on first use)
    * Build backend, Build backend, Integration frontend, Source tree ([PEP 517 terminology section](https://peps.python.org/pep-0517/#terminology-and-goals))
    * Editable installation ([PEP 660 terminology section](https://peps.python.org/pep-0660/#terminology-and-goals))
    * Updates to other related terminology to cross-link and reflect changes accordingly

</details>

After a fair amount of thought and experimentation, I've tweaked the current `Package Distribution Formats` heading to read `Building and Installing` heading and added/moved the new specs there, alongside the "Recording Installed Distributions" spec, given it is a much better fit for the new section than the crowded, core-metadata-focused "Package Distribution Metadata" section, particularly given it and the wheel spec that unpacks to it and the hooks that build it go hand in hand, and the new section now describes the build and install process from end to end.

On a procedural note, I made every effort to break this up into separate PRs or at the very least separate commits, as I normally would (and all the future specs should be). However, I ended up having to make this as a single atomic, squashed commit for practicality's sake, because otherwise there would be intractable circular dependencies between each PR, given the new specs, existing specs, glossary entries and other documents all heavily cross-reference and depend upon one another, and that would furthermore require a heavy cost in both time and quality/consistency, when this will be squashed down anyway, has been blocked on such for years and is much better spent migrating other specs and making further improvements to the packaging site in followup PRs.

I'll post a thread in the Packaging Discourse to announce this and solicit reviews of this, to both check for editorial correctness and ensure it accurately reflects the PEPs on which it is based.

Part of #1093 

Resolves #955

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -12,19 +12,46 @@ Glossary
         extensions.
 
 
+    Build Frontend
+
+        A tool that users interact with directly to trigger a build of
+        a :term:`Project`, which in turn invokes the project's
+        :term:`Build Backend` in a suitable environment
+        to generate a :term:`Built Distribution` (i.e. a :term:`Wheel`),
+        from a :term:`Source Tree` or :term:`Source Distribution`.
+        Examples include :ref:`build`, as well as :ref:`pip`
+        (when running a command such as ``pip wheel some-directory/``).
+        Compare to :term:`Integration Frontend`.
+        For more details,
+        see the :ref:`build-frontend-backend-interface` specification.
+
+
+    Build Backend
+
+        A tool directly responsible for transforming a
+        :term:`Source Tree` or :term:`Source Distribution`
+        into a :term:`Built Distribution` (i.e. a :term:`Wheel`).
+        Typically invoked by a :term:`Build Frontend` rather than directly.
+        Examples include :ref:`flit`, :ref:`hatch` and :ref:`setuptools`.
+        For more details,
+        see the :ref:`build-frontend-backend-interface` specification.
+
+
     Built Distribution
 
-        A :term:`Distribution <Distribution Package>` format containing files
-        and metadata that only need to be moved to the correct location on the
-        target system, to be installed. :term:`Wheel` is such a format, whereas
-        distutil's :term:`Source Distribution <Source Distribution (or
-        "sdist")>` is not, in that it requires a build step before it can be
-        installed.  This format does not imply that Python files have to be
-        precompiled (:term:`Wheel` intentionally does not include compiled
-        Python files).
+        A :term:`Distribution` format containing files and metadata
+        that only need to be moved to the correct location
+        on the target system to be installed.
+        :term:`Wheel` is such a format, whereas a :term:`Sdist` is not,
+        in that it requires processing by the :term:`Project`'s
+        :term:`Build Backend` before it can be installed.
+        This format does not imply that Python files have to be precompiled
+        (:term:`Wheel` intentionally does not include compiled Python files).
 
 
     Distribution Package
+    Distribution
+    Package
 
         A versioned archive file that contains Python :term:`packages <Import
         Package>`, :term:`modules <Module>`, and other resource files that are
@@ -38,6 +65,27 @@ Glossary
         another kind of distribution (e.g. a Linux distribution or the Python
         language distribution), which are often referred to with the single term
         "distribution".
+
+
+    Editable Installation
+    Editable Mode
+
+        An installation mode implying that the :term:`Source Tree` of the
+        :term:`project` being installed is available in a local directory,
+        in which users expect that changes to its *Python* code
+        become effective without the need of a new installation step.
+
+        When a project is installed in "editable mode",
+        users expect it to behave as identically as practical
+        to a non-editable installation
+        (though some minor differences might be visible).
+        In particular, the code must be importable by other code,
+        and metadata must be available by standard mechanisms
+        such as ``importlib.metadata``.
+
+        Formally specified in :pep:`660` and now defined in the
+        :ref:`build-frontend-backend-interface` specification.
+
 
     Egg
 
@@ -75,6 +123,22 @@ Glossary
         is needed to prevent confusion with a :term:`Distribution Package` which
         is also commonly called a "package".
 
+
+    Integration Frontend
+
+        A tool that users run directly
+        that takes a set of :term:`Requirement`\s,
+        such as from a :term:`Project`'s :ref:`core-metadata`,
+        a :term:`Requirements File` or specified manually,
+        and attempts to update a working environment to satisfy them.
+        This may require locating, building and installing
+        a combination of :term:`Built Distribution`\s
+        and :term:`Source Distribution`\s,
+        including acting as a :term:`Build Frontend` in the latter case.
+        In a command like ``pip install lxml==2.4.0``,
+        :ref:`pip` is acting as an integration frontend.
+
+
     Module
 
         The basic unit of code reusability in Python, existing in one of two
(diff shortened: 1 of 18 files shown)


CAM-Gerlach:
I didn't realize I had never actually contributed a PR of my own here, despite several reviews on others. If the review process is going to be intensive, perhaps worth going ahead with another PR doing another task I'd planned to, like #1101 , to avoid having to wait for workflow approval each time I change something (though I've of course tested and previewed locally).

JDLH:
Great! Thank you. I have started to read this.

source/glossary.rst
```diff
@@ -12,19 +12,46 @@ Glossary
...
+        :term:`Wheel` is such a format, whereas a :term:`Sdist` is not,
+        in that it requires processing by the :term:`Project`'s
+        :term:`Build Backend` before it can be installed.
+        This format does not imply that Python files have to be precompiled
+        (:term:`Wheel` intentionally does not include compiled Python files).
 
 
     Distribution Package
+    Distribution
+    Package
```
pradyunsg, line 54: Please don't add these -- we explicitly don't use the short forms here, to avoid ambiguity when skimming through the page. That's also called out in the definition here.

If there's terms that use the short form to direct at this, I'd rather that we update those references to be more explicit.

pradyunsg (changes requested):

pradyunsg:
@CAM-Gerlach Could you break this PR up into smaller [atomic commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/), that each changes a single thing at a time? 

Right now, this PR is fairly difficult to review in a reasonable amount of time. There is a single commit that updates multiple pages for adding glossary references, adding new terms to the glossary, restructures the specifications section, adds the new documentation, and perhaps more that I didn't spot before realising that reviewing this will take way too long as-is.

Beyond that, I've only skimmed this PR so far, and I'm feeling a bit of "uhh... I don't think this is good idea" about the addition of the "integration frontend" term -- but I've not looked into the PR for long enough to check if that instinctual response is justified. :)

pfmoore:
I agree with @pradyunsg - this is *way* too big to practically review. I read the summary of your approach above, as well as briefly skimming the structure of the PR, and my view is you're going about this the wrong way.

I should note that my key concern here is that we avoid *any* change in meaning between what ends up in the specs and what was agreed in the PEPs. That seems fundamental to me.

To that end, I think we should follow the same model as all other work to include PEPs in the spec pages.

1. Copy the text of the PEP *unchanged* into the specs. PEP 518 can easily be migrated into "[Declaring build system dependencies](https://packaging.python.org/en/latest/specifications/declaring-build-dependencies/)". PEP 517 and PEP 660 can go into a new section, with PEP 517 being the body of the section and maybe PEP 660 as a linked sub-page. See "[Declaring project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/)" (PEP 621) for how this works. That's three PRs, one for each. This shouldn't need significant understanding of the specs to create or review, as the only real question here is "is the included copy identical to the original?"
2. Once the PEPs are in the spec, we're in the normal situation of revising the packaging guide for clarity and readabilliy.
3. Fixing links in the rest of the guide to point to the new spec sections should be straightforward. These can be one or more simple PRs, requiring no deep understanding of the specs to create or review.
4. Cross-linking in the specs can also be done as follow-up PRs, with care taken to not change the meaning but otherwise these should be simple PRs.
5. Adding summary sections and/or overviews can be done easily enough - but note that these should be in the *guide*, IMO, not in the *specifications* area, as they would be user-facing explanatory material, not specifications as such. To do this properly (and to review it) needs a good understanding of the specs, to make sure that the summaries don't misrepresent the detail. But as it's summaries, the text should be relatively short (if a summary is too long to review, it's too long to be a summary, IMO 🙂)
6. Revising the wording of the spec should be done *extremely* cautiously, and following the PyPA spec change process (which I personally interpret pretty strictly these days, as I feel we've been burned rather too often - anything that changes the meaning should be considered as requiring a PEP). Again, this should be one change, one PR.

The key thing for me is that the wording as agreed in the PEPs is what the community agreed as the standard. Not all of the PEP content is normative, but as a community, we don't have a history of being particularly formal over specifications, and it's hard to always be precise about what was intended as significant (or what people view as significant after the fact). Maybe things would be easier if packaging specs were precise and unambiguous like (for example) RFCs. But that's not the reality, for better or worse. So preserving the PEP content as closely as possible is key here.

(Just to be clear, I'm *not* suggesting the content in this PR varies from the PEPs. I'm sure you transcribed it accurately. My suggestions above are intended simply to make it easier to handle the *process* of confirming that).

source/glossary.rst
```diff
@@ -12,19 +12,46 @@ Glossary
...
+        (when running a command such as ``pip wheel some-directory/``).
+        Compare to :term:`Integration Frontend`.
+        For more details,
+        see the :ref:`build-frontend-backend-interface` specification.
+
+
+    Build Backend
+
+        A tool directly responsible for transforming a
+        :term:`Source Tree` or :term:`Source Distribution`
```
ofek, line 32: I think we should remove sdist since backends just get a simple directory, which frontends _may_ populate with sdist contents rather than using the source.

source/glossary.rst
```diff
@@ -96,26 +160,30 @@ Glossary
...
-        is something that contains a :term:`pyproject.toml`, :term:`setup.py`,
-        or :term:`setup.cfg` file at the root of the project source directory.
-
-        Python projects must have unique names, which are registered on
-        :term:`PyPI <Python Package Index (PyPI)>`. Each project will then
-        contain one or more :term:`Releases <Release>`, and each release may
-        comprise one or more :term:`distributions <Distribution Package>`.
+        A library, framework, script, plugin, application,
+        collection of data or other resources, or some combination thereof
+        that is intended to be packaged into a :term:`Distribution`.
```
ofek, line 165: I don't think applications are necessarily packaged, often just installed in a virtual environment with locked deps.

pradyunsg: I've definitely seen deployment mechanisms where the application is being packaged like a regular Python package, and shipped with all its dependencies, as a bundle of files. This is how, for example, https://shiv.readthedocs.io/en/latest/ works.

ofek: For sure! My point was just not _always_

PR #1112, opened 2022-07-25
Document special links

patrick91:
Related issue: https://github.com/pypi/warehouse/issues/5947

Feel free to suggest any change, I'm not too familiar with rst and I wasn't sure how to introduce the fact that PyPI has custom icons based on name and URLs.

I also have left out some urls (for example circleci, codecov) let me know if I should add them :)

### DIFF
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -357,6 +357,34 @@ List additional relevant URLs about your project. This is the place to link to
 bug trackers, source repositories, or where to support package development.
 The string of the key is the exact text that will be displayed on PyPI.
 
+PyPI provides icons based on the name of the URL or the URL's domain. Currently
+the following names will have a special icon:
+
+- Homepage, home or home page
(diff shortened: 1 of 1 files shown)


bhrutledge (changes requested):
Thanks for the PR! I think this would be a nice thing to have documented, but I don't think this is the right place for it. This document is outdated, as noted in #887. However, I don't have a concrete suggestion.

Some ideas:

- A new [guide](https://packaging.python.org/en/latest/guides/)
- A new [specification](https://packaging.python.org/en/latest/specifications/)
- In the [metadata specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls) for URLs

@brettcannon As the author of https://github.com/pypi/warehouse/issues/5947, do you have any suggestions?

brettcannon:
I think it should be documented in PyPI's docs directly.

bhrutledge:
Okay, I'm going to close this, with the suggestion that discussion/work continue on https://github.com/pypi/warehouse/issues/5947

PR #1115, opened 2022-07-29
Add PEP 621 tab for entry points docs

flying-sheep:
Since #1031 is merged, let’s gradually move away from a setuptools centric view.

### [Rendered](https://python-packaging-user-guide--1115.org.readthedocs.build/en/1115/guides/creating-and-discovering-plugins/#using-package-metadata)

### DIFF
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -120,8 +120,7 @@ a list of packages to :func:`setup`'s ``packages`` argument instead of using
 Using package metadata
 ======================
 
-Packages have metadata for plugins described in the
-:doc:`/specifications/entry-points`.
+Packages have metadata for plugins described in the :doc:`entry-points`.
 By telling your package’s build backend,
 plugins can register themselves for discovery.
 
--- a/source/guides/creating-and-discovering-plugins.rst
+++ b/source/guides/creating-and-discovering-plugins.rst
@@ -127,22 +127,10 @@ plugins can register themselves for discovery.
 For example if you have a package named ``myapp-plugin-a`` and it includes
 the following in its setup file:
 
-.. tab:: :file:`pyproject.toml` (PEP 621)
+.. code-block:: toml
 
-    .. code-block:: toml
-
-         [project.entry-points.'myapp.plugins']
-         a = 'myapp_plugin_a'
-
-.. tab:: :file:`setup.py` (setuptools)
-
-    .. code-block:: toml
-
-        setup(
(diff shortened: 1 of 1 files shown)


source/guides/creating-and-discovering-plugins.rst
```diff
@@ -120,21 +120,29 @@ a list of packages to :func:`setup`'s ``packages`` argument instead of using
...
+Packages have metadata for plugins described in the :doc:`entry-points`.
+By telling your package’s build backend,
+plugins can register themselves for discovery.
 
 For example if you have a package named ``myapp-plugin-a`` and it includes
-in its :file:`setup.py`:
+the following in its setup file:
 
-.. code-block:: python
+.. tab:: :file:`pyproject.toml` (PEP 621)
```
bhrutledge, line 130: I don't think we should reference `setup.py` at all, and just show `pyproject.toml`, without a reference to PEP 621.

source/guides/creating-and-discovering-plugins.rst
```diff
@@ -120,21 +120,29 @@ a list of packages to :func:`setup`'s ``packages`` argument instead of using
 Using package metadata
 ======================
 
-:doc:`Setuptools <setuptools:index>` provides :doc:`special support
-<setuptools:userguide/entry_point>` for plugins. By providing the
-``entry_points`` argument to :func:`setup` in :file:`setup.py` plugins can
-register themselves for discovery.
+Packages have metadata for plugins described in the :doc:`entry-points`.
+By telling your package’s build backend,
+plugins can register themselves for discovery.
```
bhrutledge, line 125: I think the language could be clearer here. Specifically, "telling your package's build backend" seems vague; what does "telling" mean, and do we need to mention "build backend" at all?

Also, it looks like this intends to link to the [entry points spec](https://packaging.python.org/en/latest/specifications/entry-points/), but it's not in the [PR preview](https://python-packaging-user-guide--1115.org.readthedocs.build/en/1115/guides/creating-and-discovering-plugins/). Perhaps it needs to be a `:ref:` instead?

bhrutledge (changes requested):

source/guides/creating-and-discovering-plugins.rst
flying-sheep: both done!

Harlee1994 (approved):

flying-sheep:
Hi @bhrutledge, can this be merged?

bhrutledge (approved):
Thanks for the nudge!

PR #1116, opened 2022-07-30
Use latest python-docs-theme

bhrutledge:
On at least Chrome and Safari on macOS, the rendering of directory trees are hard to read:

Screen Shot 2022-07-30 at 7 29 21 AM

Noting that the [`pypa_theme`](https://github.com/pypa/pypa-docs-theme) inherits from [`python_docs_theme`](https://github.com/python/python-docs-theme), I poked around on the Python docs and discovered that [it's not an issue there](https://docs.python.org/3/library/__main__.html#main-py-in-python-packages). That led me to https://github.com/python/python-docs-theme/pull/87, which was released in 2021.11.1. Sure enough, upgrading fixes the output:

Screen Shot 2022-07-30 at 7 34 31 AM

This also includes "Make sidebar scrollable and sticky (on modern browsers)", which I think is a nice enhancement.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 sphinx==4.3.1
 sphinx-autobuild==0.7.1
 sphinx-inline-tabs==2021.4.11b9
-python-docs-theme==2021.5
+python-docs-theme==2022.1
 sphinx-copybutton==0.4.0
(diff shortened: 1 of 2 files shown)


webknjaz (approved):
Looks like GH doesn't let me resolve the conflict in my browser (maybe it's a Refined GitHub's bug). So you'll need to do that yourself.

PR #1127, opened 2022-08-09
Fix broken links

adrybakov:
Two broken links were causing auto merge issues for #1126 and #1124. I believe that the new links are correct in relation to the intent of the broken ones.


bhrutledge (approved):
Thanks!

PR #1128, opened 2022-08-10
chore: nicer nox help text

henryiii:
This improves the ergonomics and help text of nox slightly.

Here's the new help text:

```
Sessions defined in /Users/henryschreiner/git/webpages/packaging.python.org/noxfile.py:

- translation -> Build the gettext .pot files.
- build -> Make the website.
- preview -> Make and preview the website.
- linkcheck -> Check for broken links.

sessions marked with * are selected, sessions marked with - are skipped.
```

Also, now if you just type `nox`, instead of trying to build twice, and
previewing, then link checking, it will just produce the above help text.

Also removed `py="3"` since it doesn't really do anything anymore.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -6,8 +6,15 @@
 import shutil
 import nox
 
-@nox.session(py="3")
+
+nox.options.sessions = []
+
+
+@nox.session()
 def translation(session):
+    """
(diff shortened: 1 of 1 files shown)


bhrutledge (approved):
Thanks!

PR #1129, opened 2022-08-19
venv issues link

fluesvamp:
source/key_projects.rst

this was pointing to `bugs.python.org`

links for ensurepip and distutils issues remain unmodified


bhrutledge (approved):
Thanks!

Issue #792, opened 2020-11-17
Links to setuptools documentation need to be updated


edlanglois:
[This packaging page](https://packaging.python.org/guides/distributing-packages-using-setuptools/) (and possibly other documentation pages, I haven't checked) links to [this setuptools guide page](https://setuptools.readthedocs.io/en/latest/setuptools.html), which has largely been emptied of content.

It looks like the new documentation is [here](https://setuptools.readthedocs.io/en/latest/userguide/index.html) and links should reference this page or the appropriate userguide sub-pages.

webknjaz:
FWIW it's best to use intersphinx references where possible. Here's a list of what's available: https://webknjaz.github.io/intersphinx-untangled/setuptools.rtfd.io/

bhrutledge:
This feels related to https://github.com/pypa/packaging.python.org/issues/887.

PR #1136, opened 2022-08-31
replaced direct link with intersphinx reference

venthur:
Related to #792 -- there's some more direct links but i couldn't find the intersphinx references [here](https://webknjaz.github.io/intersphinx-untangled/setuptools.rtfd.io/)


bhrutledge (approved):
Thanks!

Issue #447, opened 2018-02-26
Notes / todo blocks use code-block markup instead of comments


merwok:
There are some weird blocks in the packaging binary extensions page, which is explained by the use of `::` instead of `..`.

jonparrott:
Thanks, @merwok. Should be an easy fix.

ncoghlan:
That was intentional, so it was visible to readers that the doc wasn't complete.

However, I'm now more inclined to just reduce the scope of the page and say that it is *not* intended to provide a comprehensive guide to writing binary packages by hand, but instead provide pointers to tools and documentation for building and publishing them.

The new "Additional Resources" section at the end of the page should help with that, while resolving #359 would provide a good recommendation for plugging arbitrary build systems in to a setuptools based build pipeline, and mentioning enscons would provide a good example of using a full build system to emit Python packages with proper metadata.

merwok:
Showing the todo lists intentionally is alright, but they should not be highlighted as code IMO :)

PR #1137, opened 2022-08-31
Don't highlight TODO lists as code

venthur:
closes: #447


bhrutledge (approved):
Thanks! Especially for following the existing convention:

https://github.com/pypa/packaging.python.org/search?q=fixme

Issue #1107, opened 2022-07-16
Core Metadata specification missing several fields


dstufft:
They're not really super useful fields nor are they used, but the core metadata specification is missing ``Requires``, ``Provides``, and ``Obsoletes`` which were introduced in PEP 314 as part of Metadata 1.1.

They were deprecated in PEP 345 however, but AFAIK it's still valid to emit them.

PR #1138, opened 2022-08-31
Added obsolete fields: requires, provides and obsoletes

venthur:
closes: #1107

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -558,6 +558,7 @@ resolution by locally caching metadata about dependencies.
 pypiserver
 ==========
 
+`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.md>`__ |
 `GitHub <https://github.com/pypiserver/pypiserver>`__ |
 `PyPI <https://pypi.org/project/pypiserver/>`__
 
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -558,7 +558,6 @@ resolution by locally caching metadata about dependencies.
 pypiserver
 ==========
 
-`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.md>`__ |
 `GitHub <https://github.com/pypiserver/pypiserver>`__ |
 `PyPI <https://pypi.org/project/pypiserver/>`__
 
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -776,7 +776,7 @@ Each entry contains a string describing some other module or package required
 by this package.
 
 The format of a requirement string is identical to that of a module or package
-name usable with the ‘import’ statement, optionally followed by a version
+name usable with the ``import`` statement, optionally followed by a version
 declaration within parentheses.
 
 A version declaration is a series of conditional operators and version numbers,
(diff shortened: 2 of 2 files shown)


Harlee1994 (approved):

source/specifications/core-metadata.rst
```diff
@@ -755,6 +755,92 @@ Examples::
...
+
+Deprecated Fields
+=================
+
+Requires
+--------
+
+.. versionadded:: 1.1
+.. deprecated:: 1.2
+   in favour of ``Requires-Dist``
```
bhrutledge, line 767: Given that all of these are deprecated fields, what value do the detailed descriptions provide? At first glance, it seems like they mostly duplicate the current fields. Could we just limit it to these first 5 lines per section?

venthur: They are deprecated but still work, see https://github.com/pypa/packaging.python.org/issues/1107 -- so it makes sense to document them properly.

source/specifications/core-metadata.rst
```diff
@@ -755,6 +755,92 @@ Examples::
...
+
+Each entry contains a string describing some other module or package required
+by this package.
+
+The format of a requirement string is identical to that of a module or package
+name usable with the ‘import’ statement, optionally followed by a version
```
bhrutledge, line 773: Nit: let's use monospace for this:

```suggestion
name usable with the ``import`` statement, optionally followed by a version
```

source/specifications/core-metadata.rst
bhrutledge: @dstufft Is this what you had in mind when you submitted #1107? Are the descriptions accurate/helpful?

dstufft: Sorry, I missed this PR somehow.

I don't personally feel strongly, though the "Rarely Used" fields are probably the closest analogy to these and they have their full documentation, so I guess I'd lean towards including it?

dstufft (approved):

venthur:
@webknjaz the linkcheck is failing and preventing the auto-merge. I didn't add any links in this PR and I also can't figure out based on its output what it is complaining about.

source/key_projects.rst
```diff
@@ -558,7 +558,7 @@ resolution by locally caching metadata about dependencies.
 pypiserver
 ==========
 
-`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.rst>`__ |
+`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.md>`__ |
```
webknjaz, line 561: I'd rather have this as https://github.com/pypiserver/pypiserver#readme instead. Though, it's probably best to fix it in a dedicated PR.

venthur: I assume, nothing to do here then? I'm also not happy to fix unrelated thing in this PR, but the CI pipeline is set up in a way that prevented the auto merge when a link is broken, even if the link was not introduced by the PR. 

Maybe this setup should be addressed as well at some point, could be confusing for some contributors.

chrysle:
@venthur Could you rebase? The problem has been fixed recently.

venthur:
fixed, but now linkcheck fails due to a timeout of another url. this test is a bit fragile. can someone trigger a re-run of the pipeline? i think i can't do it myself.

PR #1141, opened 2022-09-17
Update sdist spec to reflect PEP 625

pfmoore:


pradyunsg (approved):
I wonder if it’s worth including a note about pre-standardisation format?

pfmoore:
I thought about that, but concluded that the spec isn't the right place to talk about non-compliant formats, unlike the PEP, which *is* expected to discuss transition and legacy formats. But I could be persuaded to add something if anyone feels strongly enough.

pradyunsg:
> if anyone feels strongly enough.

I don't.

PR #1143, opened 2022-09-18
Revert "Declare license file in packaging tutorial"

bhrutledge:
Reverts pypa/packaging.python.org#1098

While investigating https://github.com/pypa/twine/issues/926, I discovered that using `license = { file="LICENSE" }` will result in the contents of `LICENSE` being displayed on PyPI in the "Meta" sidebar section. For example, see:

https://test.pypi.org/project/example-package-bhrutledge/0.0.1/

When I remove that line, I get what I think is the desired behavior, which is the name of the license:

https://test.pypi.org/project/example-package-bhrutledge/0.0.2rc1/

FYI @EpicWink

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -281,6 +281,9 @@ MIT license:
     OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
     SOFTWARE.
 
+Most build backends automatically include license files in packages. See your
+backend's documentation for more details.
+
 
 Including other files
 ---------------------
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -188,7 +188,6 @@ following this tutorial.
     ]
     description = "A small example package"
     readme = "README.md"
-    license = { file="LICENSE" }
     requires-python = ">=3.7"
     classifiers = [
         "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


ofek (approved):

EpicWink:
Ah, that seems to be an oversight on the distribution end of Python packaging. Maybe it's worth starting a discussion in [Discourse](https://discuss.python.org)?

If a license file is in the repo, how does the backend know to include it in the package? The [Manifest.in tutorial](https://packaging.python.org/en/latest/guides/using-manifest-in/#how-files-are-included-in-an-sdist) suggests it should be in `license_files`.

ofek:
After PEP 639 it'll be under `licenses/`

bhrutledge:
> Ah, that seems to be an oversight on the distribution end of Python packaging. Maybe it's worth starting a discussion in [Discourse](https://discuss.python.org)?

I think that's already underway, in the form of [PEP 639](https://peps.python.org/pep-0639/).

> If a license file is in the repo, how does the backend know to include it in the package? The [Manifest.in tutorial](https://packaging.python.org/en/latest/guides/using-manifest-in/#how-files-are-included-in-an-sdist) suggests it should be in `license_files`.

I think this depends on the backend. As noted in https://github.com/pypa/packaging.python.org/pull/1098#discussion_r906695295, all backends [except Flit](https://github.com/pypa/flit/issues/379) include `LICENSE` in the sdist and wheel , even when `license.file` is absent. Also, I'm think `MANIFEST.in` is only relevant for setuptools.

EpicWink:
> As noted in [#1098 (comment)](https://github.com/pypa/packaging.python.org/pull/1098#discussion_r906695295), all backends [except Flit](https://github.com/pypa/flit/issues/379) include `LICENSE` in the sdist and wheel , even when `license.file` is absent.

If that's the case, perhaps that could be mentioned in place of including `license.file`

PR #1150, opened 2022-09-30
Add initial draft of src vs flat layout discussion

pradyunsg:
x-ref #320 

This discussion covers the main topics discussed in #320, trimming the opinions/preferences stated there and keeping things limited to the behaviour differences between the two formats. I've followed the same style as pip vs easy_install or wheel vs egg pages here.

### DIFF
--- a/source/discussions/index.rst
+++ b/source/discussions/index.rst
@@ -12,3 +12,4 @@ specific topic. If you're just trying to get stuff done, see
    pip-vs-easy-install
    install-requires-vs-requirements
    wheel-vs-egg
+   src-layout-vs-flat-layout
--- a/source/discussions/src-layout-vs-flat-layout.rst
+++ b/source/discussions/src-layout-vs-flat-layout.rst
@@ -3,8 +3,8 @@ src layout vs flat layout
 =========================
 
 The "flat layout" refers to organising a project's files in a folder or
-repository, such that the various configuration files and importable
-modules/packages are all in the top-level directory.
+repository, such that the various configuration files and
+:term:`import packages <Import Package>` are all in the top-level directory.
 
 ::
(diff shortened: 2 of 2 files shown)


pfmoore (approved):
This feels like the right sort of level to me - discuss the trade-offs but keep it simple.

I'm assuming that this section is *not* where we offer a recommendation. The fact that https://packaging.python.org/en/latest/tutorials/packaging-projects/ uses the src layout is where our recommendation is stated (implicitly).

pradyunsg:
> I'm assuming that this section is _not_ where we offer a recommendation.

FWIW, I'm personally not going to get into the recommendations rabbit hole for this. That'll be someone else's problem or future me's problem, when I come around to this site after https://github.com/pradyunsg/lutra/ is good to go.

pradyunsg:
This is good to go from my end. If I could get a couple of approvals on this, I’m happy to land this myself.

I am also more than happy to trust the judgement of currently-active maintainers on the guide, if they want to land this sooner. :)

takluyver (approved):
This looks good to me.

As I recall, a big part of the discussion around this related to test files, where people like to put them, whether they get installed as part of the package or not, and whether you ever want to test the source package as opposed to the installed package. I don't think you need to tackle that topic in this PR, but I imagine it might get added at some point. :slightly_smiling_face:

pradyunsg:
> As I recall, a big part of the discussion around this related to test files

It certainly was! I wanted to start with the unambigous bits and I'm gonna be cautious and not [take implicit dibs on the task](https://producingoss.com/en/producingoss.html#cookie-licking) of expanding this to cover test layouts. It might be an idea to have that discussion in a separate file even.

PR #1154, opened 2022-10-19
Link Python 3 doc instead of Python 2 (#1153)

### LINKED ISSUE #1153: Creating and discovering plugins guide links Python 2 doc instead of Python 3
In the ["Creating and discovering plugins" guide](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/), all links to documentation for `pkgutil.iter_modules` link the Python 2 version of the documentation instead of the Python 3 version.

To my understanding, these links should point to the latest Python 3 documentation [since Python 2 reached EOL in 2020](https://www.python.org/doc/sunset-python-2/).

pushfoo:
Closes #1153 by removing Python 2 from the interpshinx config.

If including Python 2 in the current doc is intentional, please let me know. I can look into a more precise solution in that case.

Built & previewed locally.


webknjaz (approved):
This should be fine unless Sphinx build fails.

PR #1155, opened 2022-10-20
update distlib links

partev:
distlib has been migrated from BitBucket to GitHub. update the links accordingly


bhrutledge (approved):
Thanks!

PR #1157, opened 2022-10-24
Update versions of github actions

rffontenelle:
Some of them like setup-python before v4 used node.js 12, which is deprecated (see [annotations in actions log](https://github.com/pypa/packaging.python.org/actions/runs/3306153606)).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,15 +23,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.9
-
-      - name: pip cache
-        uses: actions/cache@v3
-        with:
-          path: ~/.cache/pip
-          key:
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('noxfile.py') }}
-          restore-keys: |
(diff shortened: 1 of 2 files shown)


Adix82 (commented):
H

Adix82 (commented):
H

.github/workflows/test.yml
```diff
@@ -17,21 +17,14 @@ jobs:
...
-      - name: pip cache
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pip
-          key:
-            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}-${{ hashFiles('noxfile.py') }}
-          restore-keys: |
-            ${{ matrix.os }}-${{ matrix.python-version }}-
+          cache: 'pip'
+          cache-dependency-path: 'requirements.txt'
```
webknjaz, line 27: Looks like this doesn't take into account changes to the noxfile. So it may be not equivalent to the current implementation.

webknjaz: Aha! This https://github.com/pypa/packaging.python.org/blob/f535854/noxfile.py#L67 is the only extra package installed in the noxfile. But since it's not used in the CI, it's fine to accept this difference.

webknjaz (approved):

Issue #1069, opened 2022-04-01
Unclarity in Single-sourcing version option 1


soxofaan:
https://github.com/pypa/packaging.python.org/blob/526ee6d6563855551bfee4d967a06823044ddbd4/source/guides/single-sourcing-package-version.rst#L8-L11

states:

> Read the file in ...

but it's not clear which file (there is no preceding discussion about a file).
The meaning can be somewhat guessed from the code snippet below it, but that limits the interpretation to `package/__init__.py`, while it could be applied to other options too (e.g. `package/version.py`)

bhrutledge:
I think this whole document could/should be updated to favor methods that don't rely on `setup.py`, since the current recommendation to use `setup.cfg`. I also note that this is specific to setuptools, and there are now many other build backends, some of which probably have their own method of handling this. 

I'm tempted to re-purpose this issue as "Update single-sourcing package version recommendations", and probably close #616 as a duplicate. Off the top of my head, I might list the options as:

- `importlib.metadata` (currently 5)
- Version-updating tools (currently 2)
- `version = attr: package.__version__` (currently note in 1)
- setuptools_scm (currently 7)
- Methods from other build backends (flit, poetry, pdm, etc.)

I wonder if @jaraco and/or @henryiii have thoughts.

pradyunsg:
My 2 cents: Let's start simple, and fix the sentence ambiguity first. We shouldn't need to block minor sentence structure improvements on doing a major updates of various pages. :)

ChrisBarker-NOAA:
Here's a PR for more updates:

https://github.com/pypa/packaging.python.org/pull/1273

ncoghlan:
Resolved in #1612

Issue #616, opened 2019-05-07
Add importlib.metadata as an approach in "Single-sourcing the package version"


chrahunt:
The [current version](https://github.com/pypa/packaging.python.org/blob/50e680302eed35e2631d69f70881088e8d6124e6/source/guides/single-sourcing-package-version.rst) of "Single-sourcing the package version" has an example using `pkg_resources`. It would probably make sense to include an item using `importlib.metadata` since it will be standard in 3.8+ with a backport as `importlib_metadata` and can serve the same purpose. That may look like

```python
from importlib_metadata import version

__version__ = version(__package__)
```

soxofaan:
I think this issue can be closed as it is already handled in 3c7c07f135303896a8c13a14697e136c99d4237b  (and followup tweaks)

bhrutledge:
I think the documentation could be improved; looking at option 5 in https://packaging.python.org/en/latest/guides/single-sourcing-package-version/, I don't actually see a complete example of this technique.

pradyunsg:
I suggest that we close this out and file a separate issue for improving the option 5 to de-emphasize `pkg_resources` and prefer `importlib.metadata` on 3.8+.

ncoghlan:
Resolved in #1612

PR #1167, opened 2022-11-21
Single-sourcing the package version: Add instructions for setuptools using pyproject.toml

justinchuby:
On the `Single-sourcing the package version` page, the newer `pyproject.toml` is not mentioned for setuptools. This adds a clean way of configuring it in `pyproject.toml`.

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -52,6 +52,17 @@ number of your project:
        Also, please be aware that declarative config indicators, including the
        ``attr:`` directive, are not supported in parameters to ``setup.py``.
 
+       One can also specify the version dynamically in ``pyproject.toml``:
+
+       .. code-block:: toml
+
+            [project]
+            name = "package"
+            dynamic = ["version"]
+
+            [tool.setuptools.dynamic]
+            version = {attr = "package.__version__"}
+
 #.  Use an external build tool that either manages updating both locations, or
     offers an API that both locations can use.
 
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -36,23 +36,17 @@ number of your project:
     .. note::
 
        As of the release of setuptools 46.4.0, one can accomplish the same
-       thing by instead placing the following in the project's ``setup.cfg``
-       file (replacing "package" with the import name of the package):
+       thing by instead placing the following in the project's
+       :file:`setup.cfg` file (replacing "package" with the import name of the
+       package):
 
        .. code-block:: ini
 
            [metadata]
            version = attr: package.__version__
 
-       Earlier versions of setuptools implemented the ``attr:`` directive by
-       importing the module, but setuptools 46.4.0 added rudimentary AST
-       analysis so that ``attr:`` can function without having to import any of
-       the package's dependencies.
-
-       Also, please be aware that declarative config indicators, including the
-       ``attr:`` directive, are not supported in parameters to ``setup.py``.
-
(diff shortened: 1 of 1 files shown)


source/guides/single-sourcing-package-version.rst
```diff
@@ -52,6 +52,17 @@ number of your project:
...
+       One can also specify the version dynamically in ``pyproject.toml``:
+
+       .. code-block:: toml
+
+            [project]
+            name = "package"
+            dynamic = ["version"]
+
+            [tool.setuptools.dynamic]
+            version = {attr = "package.__version__"}
```
bhrutledge, line 64: Any idea what's the minimum version of setuptools that supports this? The [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration) say:

> Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the [tool.setuptools] table), is still in beta stage and might change in future releases.

Not sure if that applies to [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata). 

Also, as noted elsewhere (https://github.com/pypa/packaging.python.org/issues/1069#issuecomment-1086391442, https://github.com/pypa/packaging.python.org/issues/616#issuecomment-1086387067, https://github.com/pypa/packaging.python.org/issues/182), this whole document needs work. That should definitely be a separate PR, but this addition feels buried after some minutiae of `setup.cfg`. 

@abravalheri, does this seem appropriate to include here? Would you

justinchuby: > Any idea what's the minimum version of setuptools that supports this? The [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration) say:
> 
> > Support for declaring configurations not standardized by [PEP 621](https://peps.python.org/pep-0621/) (i.e. the [tool.setuptools] table), is still in beta stage and might change in future releases.
> 
> Not sure if that applies to [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata).

From my read of the PEP dynamic is specified: https://peps.python.org/pep-0621/#dynamic

I haven’t tested the min version. Support for pyproject.toml starts from 61.0.0 according to the doc.

I think this approach is worth being included since it is one of the cleaner way, no runtime parsing, and uses pyproject.toml which is what new projects are encouraged to use

bhrutledge: > I think this approach is worth being included since it is one of the cleaner way, no runtime parsing, and uses pyproject.toml which is what new projects are encouraged to use

I agree. I'm mostly wanting to double-check the accuracy, and wondering if it's worth editing the `Note` to make its inclusion more obvious. But maybe that should be saved for a bigger update to this doc.

justinchuby: That all sounds good. It's my first time contributing to the pypa doc. Let me know when there is any change I need to make, thanks!

venthur:
Thanks for spotting and fixing the issue!

source/guides/single-sourcing-package-version.rst
bhrutledge: @justinchuby I took the liberty of pushing a commit that updates this section to just show the relevant configuration, without going into details. What do you think?

bhrutledge (approved):

source/guides/single-sourcing-package-version.rst
justinchuby: They look great. Thanks!

PR #1181, opened 2022-12-11
Dependency specifiers grammar fixes

konstin:
This PR contains three changes:

---

[Dependency Specifiers: Don't require whitespace after URL](https://github.com/pypa/packaging.python.org/commit/3ab0328dbd8208c4dae3aad9a5182ef6052a0b3a)

otherwise `numpy @ https://example.org/numpy` is not a valid specifier (missing whitespace after the URL). This already what pip and packaging do and also what the parsley grammar says

----

[Dependency Specifiers: Require whitespace before in and not in](https://github.com/pypa/packaging.python.org/commit/4193e048f5bac6b5e996000e329cddc069a0a40a)

Otherwise `numpy; os_namein 'posix'` would be a valid specifier which it isn't

---

[Dependency Specifiers: Require whitespace after in and not in](https://github.com/pypa/packaging.python.org/commit/f91e32f29244fb8a9d313a9701b81036de25842e)

Otherwise `numpy; os_name in'posix'` and `numpy; os_name inos_name` would be valid. pypa/packaging 22.0 actually allows currently `numpy; os_name in'posix'`, but i argue there should be a whitespace after `in` and `not in` just like in python itself


chrysle (approved):
This looks good to me; sorry for the lengthy delay in reviewing.

PR #1188, opened 2022-12-26
Create verfolgung pypa SECURITY.md

Dadudidas:
kontrolle


webknjaz:
I don't think it's relevant to this project. Besides, we don't need a file template without actual content.

PR #1193, opened 2023-01-12
Update the flit_core lower bound

hauntsaninja:
Version 3.4 added PEP 660 support
Version 3.3 added PEP 621 support
https://flit.pypa.io/en/stable/history.html

Feels wrong to allow a version here that doesn't support those


bhrutledge (approved):
Thanks!

hauntsaninja:
The linkcheck is flaky (failing to hit archive.org); I think you'll need to redo the automerge

PR #1196, opened 2023-01-24
Add tools comparison tables

sinoroc:
Is it something that would be interesting for the guide in principle?


source/discussions/packaging-tools-comparisons.rst
```diff
@@ -0,0 +1,197 @@
...
+    Pipenv,yes,no
+    Poetry,yes,no
+    tox,yes,yes [#]_
+    venv,no,no
+    virtualenv,yes,no
+    virtualenvwrapper,yes,no
+
+Unlike all the other tools here,
+``venv`` is part of Python's own standard library,
+it is always available without having to be installed separately.
```
abravalheri, line 172: In some Linux distributions (e.g. Debian-based), you need to install an extra package for `venv` (e.g. `sudo apt install python3 python3-pip python3-venv python3-dev`)

sinoroc: Good point. Thanks.

Replaced with:

> Unlike the other tools presented in this section,
``venv`` is part of Python's own standard library,
it should be always available without having to be installed separately.
But note that on some Linux distributions (e.g. Debian, Ubuntu and derivatives)
``venv`` is packaged separately from the rest of the Python distribution
and it might be necessary to install ``venv`` explicitly
(typically with a command such as `apt install python3-venv`,
consult the documentation of the Linux distribution for exact details).

source/discussions/packaging-tools-comparisons.rst
```diff
@@ -0,0 +1,202 @@
...
+
+Build back-ends
+---------------
+
+.. csv-table::
+    :align: left
+    :header-rows: 1
+    :stub-columns: 1
+
+    ,``[build-system]`` (PEP-517),``[project]`` (PEP-621),editable (PEP-660),C extensions
```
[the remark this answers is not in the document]
sinoroc: @henryiii Thanks for the suggestions. I think I managed to integrate them all.

willingc:
@sinoroc If you fix the conflict, I would be happy to review. @jeanas If you want to give this an initial review, that would be great too.

sinoroc:
@willingc, rebased.

---

For info:
* this initially took inspiration from [Chris Warrick's "_How to improve Python packaging, or why fourteen tools are at least twelve too many_"](https://chriswarrick.com/blog/2023/01/15/how-to-improve-python-packaging/)
* since then [PyOpenSci's Python packaging guide](https://www.pyopensci.org/python-package-guide/package-structure-code/python-package-build-tools.html#package-tool-features-table) was created (or at least I found out about it after I started this PR)
* and also [Anna-Lena Popkes' "_An unbiased evaluation of environment management and packaging tools_"](https://alpopkes.com/posts/python/packaging_tools/)

I do not know which form is the most appropriate.

I do not believe our guide should go too much into details in order to avoid risking being seen as choosing one tool over another (not the Pipenv story again).

For sure what this needs is links (but those can be added at the end, because I find it quite annoying to do links in reStructuredText).

willingc:
Thanks for the rebase @sinoroc. I think that it would be helpful to add a date to give the reader when the last time this was updated and note that tools are evolving so do check their documentation as well.

PR #1198, opened 2023-01-29
Add hashes key to the Direct URL data structure

sbidoul:
The first commit adds the `hashes` key to the Direct URL data structure, as a approved at https://discuss.python.org/t/draft-pep-amendment-to-the-pep-610-direct-url-json-data-structure/22299/4.

In the second commit, I took the liberty to extract the specification of the data structure into a standalone document, and update the `Recording the Direct URL Origin of installed distributions` spec to refer to it.

This will make it easier to reference the data structure specification when it is used in different contexts.

The third commit are minor edits to the text for it to read better in a standalone manner. I isolated them in a separate commit for easier review.

### DIFF
--- a/source/specifications/direct-url-data-structure.rst
+++ b/source/specifications/direct-url-data-structure.rst
@@ -1,0 +1,277 @@
+
+.. _direct-url-data-structure:
+
+=========================
+Direct URL Data Structure
+=========================
+
+This document specifies a JSON-serializable abstract data structure that can represent
+URLs to python projects and distribution artifacts such as VCS source trees, local
+source trees, source distributions and wheels.
+
+.. contents:: Contents
+   :local:
+
+Specification
+=============
+
+The Direct URL Data Structure MUST be a dictionary, serializable to JSON according to
+:rfc:`8259`.
+
+It MUST contain at least two fields. The first one is ``url``, with
+type ``string``. Depending on what ``url`` refers to, the second field MUST be
+one of ``vcs_info`` (if ``url`` is a VCS reference), ``archive_info`` (if
+``url`` is a source archives or a wheel), or ``dir_info`` (if ``url``  is a
+local directory). These info fields have a (possibly empty) subdictionary as
+value, with the possible keys defined below.
+
+``url`` MUST be stripped of any sensitive authentication information,
+for security reasons.
+
+The user:password section of the URL MAY however
+be composed of environment variables, matching the following regular
+expression::
+
+    \$\{[A-Za-z0-9-_]+\}(:\$\{[A-Za-z0-9-_]+\})?
+
+Additionally, the user:password section of the URL MAY be a
+well-known, non security sensitive string. A typical example is ``git``
+in the case of an URL such as ``ssh://git@gitlab.com/user/repo``.
+
+When ``url`` refers to a VCS repository, the ``vcs_info`` key MUST be present
+as a dictionary with the following keys:
+
+- A ``vcs`` key (type ``string``) MUST be present, containing the name of the VCS
+  (i.e. one of ``git``, ``hg``, ``bzr``, ``svn``). Other VCS's SHOULD be registered by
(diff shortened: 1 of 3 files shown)


sbidoul:
@pfmoore

pfmoore (approved):
I'm assuming (i.e., I didn't check!) that the commit which moved the data structure spec into the separate document made no actual changes to the text, and all the content changes are in the other commits.

pfmoore:
I'll leave this a while before I hit the merge button, in case others want to make comments on the restructuring of the document.

source/specifications/direct-url-data-structure.rst
```diff
@@ -0,0 +1,292 @@
...
+
+History
+=======
+
+This data structure was originally specified in :pep:`610` and is formally documented
+here.
+
+This specification was updated in January 2023 to add the ``archive_info.hashes`` key.
+This change was formally approved at
+https://discuss.python.org/t/draft-pep-amendment-to-the-pep-610-direct-url-json-data-structure/22299/4.
```
pradyunsg, line 292:
```suggestion
- March 2020: This data structure was originally specified in :pep:`610`.
- January 2023: Added the ``archive_info.hashes`` key ([discussion](https://discuss.python.org/t/22299)).
```

source/specifications/direct-url.rst
```diff
@@ -292,3 +62,9 @@ Commands that *do not* generate a ``direct_url.json``
 
 * ``pip install app``
 * ``pip install app --no-index --find-links https://example.com/``
+
+History
+=======
+
+``direct_url.json`` was originally specified in :pep:`610` and is formally documented
+here.
```
pradyunsg, line 70:
```suggestion
- March 2020: This data structure was originally specified in :pep:`610`.
```

pradyunsg:
Can we add a toctree to `direct-url` to point to `direct-url-data-structure`, and remove it from `index`?

That way we'd avoid having multiple entries on the specifications page for the same underlying "thing", which feels more appropriate.

pradyunsg (commented):
Let's have a linear chronological/reverse-chronological history section?

sbidoul:
> Can we add a toctree to `direct-url` to point to `direct-url-data-structure`, and remove it from `index`?
> 
> That way we'd avoid having multiple entries on the specifications page for the same underlying "thing", which feels more appropriate.

@pradyunsg I'm not sure. If you look at the rendering:

image

they don't necessarily feel as being the same thing. My point with this restructuring is that the direct URL data structure has broader reach in itself, as our use in different contexts in pip shows. So it makes sense to have it stand out a little bit, so tools or future interoperability specs can refer to it independently.

source/specifications/direct-url.rst
sbidoul: In this file, it is not the data structure that was specified in pep 610, but the `direct_url.json` metadata file, which is a specific use of the data structure to record that the distribution was installed from a direct URL. So this?

```suggestion
March 2020: the ``direct_url.json`` metadata file was originally specified in :pep:`610` and is formally documented
here.
```

pfmoore:
Given the discussion, I’m inclined to say that the actual change to the spec (which I approved) should be done in a separate commit, and the reorganisation can be handled (and discussed) in a follow-up PR. @sbidoul can you do that, please?

sbidoul:
> Given the discussion, I’m inclined to say that the actual change to the spec (which I approved) should be done in a separate commit, and the reorganisation can be handled (and discussed) in a follow-up PR. @sbidoul can you do that, please?

Change to the spec in #1199.

pfmoore:
Thanks. Sorry for disrupting things, I just want to make sure we have a very clear audit of what was approved (given some of the debates over "minor" spec changes in the past 🙁)

sbidoul:
No worries, I understand perfectly :)

sbidoul:
The second part is now in #1200.

PR #1199, opened 2023-01-29
Add a hashes key to the direct URL data structure

sbidoul:
Add the hashes key to the Direct URL data structure, as a approved at https://discuss.python.org/t/draft-pep-amendment-to-the-pep-610-direct-url-json-data-structure/22299/4.

@pfmoore

### DIFF
--- a/source/specifications/direct-url.rst
+++ b/source/specifications/direct-url.rst
@@ -316,9 +316,7 @@ Commands that *do not* generate a ``direct_url.json``
 History
 =======
 
-The layout of this file was originally specified in :pep:`610` and is formally
-documented here.
-
-This specification was updated in January 2023 to add the ``archive_info.hashes`` key.
-This change was formally approved at
-https://discuss.python.org/t/draft-pep-amendment-to-the-pep-610-direct-url-json-data-structure/22299/4.
+- March 2020: the ``direct_url.json`` metadata file was originally specified in
+  :pep:`610` and is formally documented here.
+- January 2023: Added the ``archive_info.hashes`` key
+  (`discussion <https://discuss.python.org/t/22299>`__).
--- a/source/specifications/direct-url.rst
+++ b/source/specifications/direct-url.rst
@@ -7,8 +7,7 @@ Recording the Direct URL Origin of installed distributions
 
 This document specifies a :file:`direct_url.json` file in the
 :file:`*.dist-info` directory of an installed distribution, to record the
-Direct URL Origin of the distribution. The layout of this file was originally
-specified in :pep:`610` and is formally documented here.
+Direct URL Origin of the distribution.
 
 .. contents:: Contents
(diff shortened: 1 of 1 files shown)


pfmoore (approved):
Thanks, this looks good!

pradyunsg (approved):

sbidoul:
The linkcheck failure seems unrelated.

pfmoore:
Apparently I can't merge without all checks passing. @pradyunsg do you know what we do here? (You enabled auto-merge, so I'm assuming you know more about the processes here than me 🙂)

pradyunsg:
... well, this repo blocks owners from merging PRs that fails CI.

sbidoul:
:green_apple:

source/specifications/direct-url.rst
```diff
@@ -292,3 +312,11 @@ Commands that *do not* generate a ``direct_url.json``
...
 * ``pip install app``
 * ``pip install app --no-index --find-links https://example.com/``
+
+History
+=======
+
+- March 2020: the ``direct_url.json`` metadata file was originally specified in
+  :pep:`610` and is formally documented here.
+- January 2023: Added the ``archive_info.hashes`` key
+  ([discussion](https://discuss.python.org/t/22299)).
```
webknjaz, line 322: This is not a Markdown document. RST is a must.
```suggestion
  (`discussion <https://discuss.python.org/t/22299>`__).
```

It is visible how this is broken in the preview: https://python-packaging-user-guide--1199.org.readthedocs.build/en/1199/specifications/direct-url/#history

webknjaz (changes requested):
A syntax fix is needed for this to be acceptable.

webknjaz:
> do you know what we do here?

@pfmoore we normally fix such things in separate PRs to unblock the CI.

webknjaz:
> The linkcheck failure seems unrelated.

Looking into https://github.com/pypa/packaging.python.org/actions/runs/4037007410, I see that it's been cancelled because of a newer push to this PR:
> 
Canceling since a higher priority waiting request for 'Test-1199' exists

I don't see it being broken. Unless, I'm looking at the wrong log, that is (some tend to be hidden by force-pushes, so it's best to record the exact URLs in the comments for future investigation).
If a job is flaky, restarting it fixes the branch protection too.

webknjaz:
Oh, found it: https://github.com/pypa/packaging.python.org/actions/runs/4037021743/jobs/6940079785#step:5:765. So the network was flaky. The solution was indeed restarting the job.

ofek (approved):

source/specifications/direct-url.rst
sbidoul: Thanks @webknjaz. Some markdown did indeed sneak in during review. This is fixed, now.

PR #1207, opened 2023-02-11
Use `req @ url` syntax to install from remote VCS

matthewfeickert:
Use 'req @ url' syntax when using pip to install from a remote git repository over 'url#egg=req' to avoid the use of #egg= fragments with a non-PEP 508 name. This will be required in pip v25.0+.
   - c.f. pypa/pip#11617 for more details.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ sphinx-autobuild==0.7.1
 sphinx-inline-tabs==2021.4.11b9
 python-docs-theme==2022.1
 sphinx-copybutton==0.5.0
-git+https://github.com/pypa/pypa-docs-theme.git#egg=pypa-docs-theme
+pypa-docs-theme @ git+https://github.com/pypa/pypa-docs-theme.git
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -693,7 +693,7 @@ you want "bar" installed from VCS in editable mode, then you could construct a
 requirements file like so::
 
   -e .
-  -e git+https://somerepo/bar.git#egg=bar
+  -e bar @ git+https://somerepo/bar.git
 
 The first line says to install your project and any dependencies. The second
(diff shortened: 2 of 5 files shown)


matthewfeickert:
This is ready for review. Let me know if you have any questions. :+1: 

cc @pradyunsg @pfmoore given their involvement in https://github.com/pypa/pip/pull/11617

pradyunsg (changes requested):
Could you add whitespace around `@`? IMO that makes it clearer and easier to read.

matthewfeickert:
> Could you add whitespace around @? IMO that makes it clearer and easier to read.

Sure. I applied that universally, rebased, and `force-with-lease` pushed back.

pradyunsg (approved):

sinoroc:
There are issues with sub-directories.

`pip install "git+https://example.com/repo/app.git#egg=app&subdirectory=setup`

was transformed into

`pip install "app&subdirectory=setup @ git+https://example.com/repo/app.git"`

which does not seem correct to me. Is it?

matthewfeickert:
@sinoroc do you have a failing example?

sinoroc:
I might be wrong, but I do not think it has anything to do with "*PEP 610 – Recording the Direct URL Origin of installed distributions*". The issue is that this is not a valid requirement notation for the *pip CLI*.

The best I can find is [this section of *pip*'s documentation on "URL fragments" in "VCS Support"](https://pip.pypa.io/en/stable/topics/vcs-support/#url-fragments) which lists two accepted notations for VCS and subdirectory:

* `python -m pip install "pkg @ vcs+protocol://repo_url/#subdirectory=pkg_dir"`
* `python -m pip install -e "vcs+protocol://repo_url/#egg=pkg&subdirectory=pkg_dir"`

matthewfeickert:
Thanks @sinoroc. Indeed it looks like

```
python -m pip install "pkg @ vcs+protocol://repo_url/#subdirectory=pkg_dir"
```

is what should have been done. I'll open up a fix PR with this correction once I've finished making coffee.

PR #1213, opened 2023-02-27
Replace refs to PEP-503 with simple repository API

sinoroc:
Preview: https://python-packaging-user-guide--1213.org.readthedocs.build/en/1213/


bhrutledge (approved):
Thanks! Sorry for taking so long to merge this.

PR #1214, opened 2023-03-01
Replace `python` interpreter with `python3`

chrysle:
I fixed some mistakes found while reading the guide.

### DIFF
--- a/source/guides/analyzing-pypi-package-downloads.rst
+++ b/source/guides/analyzing-pypi-package-downloads.rst
@@ -311,7 +311,7 @@ Install `pypinfo`_ using pip.
 
 .. code-block:: bash
 
-    python -m pip install pypinfo
+    python3 -m pip install pypinfo
 
 Usage:
 
--- a/source/guides/distributing-packages-using-setuptools.rst
+++ b/source/guides/distributing-packages-using-setuptools.rst
@@ -72,7 +72,7 @@ sample project <https://github.com/pypa/sampleproject>`_.
 
 2. It's the command line interface for running various commands that
    relate to packaging tasks. To get a listing of available commands, run
-   ``python setup.py --help-commands``.
+   ``python3 setup.py --help-commands``.
 
 
 setup.cfg
@@ -677,7 +677,7 @@ Change directory to the root of the project directory and run:
 
 .. code-block:: bash
 
-   python -m pip install -e .
+   python3 -m pip install -e .
 
 
 The pip command-line flag ``-e`` is short for ``--editable``, and ``.`` refers
@@ -712,7 +712,7 @@ Lastly, if you don't want to install any dependencies at all, you can run:
 
 .. code-block:: bash
 
-   python -m pip install -e . --no-deps
+   python3 -m pip install -e . --no-deps
 
 
 For more information, see the
@@ -812,7 +812,7 @@ To build the wheel:
 
     .. code-block:: bash
 
-        python -m build --wheel
+        python3 -m build --wheel
 
 .. tab:: Windows
 
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -15,13 +15,13 @@ jobs:
         python-version: "3.10"
     - name: Install pypa/build
       run: >-
-        python -m
+        python3 -m
         pip install
         build
         --user
     - name: Build a binary wheel and a source tarball
       run: >-
-        python -m
+        python3 -m
         build
         --sdist
         --wheel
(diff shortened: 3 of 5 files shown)


sinoroc:
I agree with the expansion of `pip install -U` into `pip install --upgrade`. I believe that in written documentation, scripts, etc. the long forms should always be used because they are more self explanatory. Short forms should be reserved for interactive sessions (REPLs and such).

But I am really not sure about the change from `python` to `python3`.

chrysle:
> But I am really not sure about the change from `python` to `python3`.

I did this to be more consistent. Most of the guide already uses `python3` instead of `python`, so I thought this wouldn't do any damage.

sinoroc:
> > But I am really not sure about the change from `python` to `python3`.
> 
> I did this to be more consistent. Most of the guide already uses `python3` instead of `python`, so I thought this wouldn't do any damage.

Ah... I had not realized that. Personally I went through a phase of writing `python3` explicitly everywhere (mostly based on advice from [PEP 394](https://peps.python.org/pep-0394/)) and just very recently decided to get out of it and write `python`, because I felt it was the right time (as far as I know most Linux distribution do not have any Python 2 by default anymore).

So I guess that for consistency's sake this is the right move. But I wonder if we should agree on `python` instead. Not my place to say, others with more authority can make a decision.

henryiii:
FWIW, macOS doesn't have `python`, most ways to get it (including brew & Xcode command line tools) only add `python3`. Also some Linux distro's don't have `python`, only the numbered one. But on the other hand, Windows doesn't always have `python3`. I had to request it be added to GHA before they had it on Windows. `¯\_(ツ)_/¯`

For CI, you should use the long form. For the command line, the short form is fine. But in a tutorial, aren't you teaching people how to do things on the command line? How would a user know that `-U` is fine if it's not shown? Someone learning to write out commands might find it harder if they have to type `--upgrade` every time, while `-U` seems simpler to type. (Don't know, and I'm fine with the change, just playing devil's advocate - I know I'd rather be shown the short form for something I'd be typing a lot. Which I don't, since I use disposable venvs & pipx :) )

I'd honestly rather pipx be used thoughout the tutorial - it's better for beginning users and is the best long-term practice too, IMO, and reduces the chances of them breaking their system by upgrading something that can't be upgraded...

chrysle:
> For CI, you should use the long form. For the command line, the short form is fine. But in a tutorial, aren't you teaching people how to do things on the command line? How would a user know that `-U` is fine if it's not shown? Someone learning to write out commands might find it harder if they have to type `--upgrade` every time, while `-U` seems simpler to type.

But to not break consistency, only one of these forms should be used, while `--upgrade`, IMO, is more self-explanatory. 
I haven't explored this guide fully yet... would it fit to add a section about handy cmd line options and their short forms for the basic packaging tools?

sinoroc:
I go from the point of view that explicit is better, and the longer form of options is more explicit. Users should know to look for each CLI tool's documentation for the shorter forms if they want to (we can not always teach everything from the ground up, or we might end up having to teach how to use bash and the Windows command prompt, we must assume some level of basic knowledge, especially if it is out of our scope).

Regarding `python` vs. `python3`, if there is no consistency by default among the operating systems, then I would still recommend to go for `python`, because I assume that people use virtual environments and inside virtual environments there is `python`.

But well, it is just a matter of consistency, so it is not that critical. I am fine with either decision. :)

chrysle:
> I go from the point of view that explicit is better, and the longer form of options is more explicit. Users should know to look for each CLI tool's documentation for the shorter forms if they want to (we can not always teach everything from the ground up, or we might end up having to teach how to use bash and the Windows command prompt, we must assume some level of basic knowledge, especially if it is out of our scope).

I agree.

> Regarding `python` vs. `python3`, if there is no consistency by default among the operating systems, then I would still recommend to go for `python`, because I assume that people use virtual environments and inside virtual environments there is `python`.

Maybe the last decision should be made by a member, since substituting `python3` would mean a complete refresh of the guide. And there are maybe other reasons to use it we don't know about. 

Of course your point of view is reasonable. In either way, I suggest adding a paragraph to the introduction covering this topic, to better inform the end readers about the choice.

chrysle:
@pradyunsg Could you maybe have a look at this?

chrysle:
@webknjaz Would you help reach a decision on this?

webknjaz:
I don't remember the discussions of python3 vs python but if you submit changes separately, this would help merge parts of this patch faster.

chrysle:
Thanks for the feedback, I've opened #1240 and #1241 

This remains the discussion for the change from `python` to `python3`.

bhrutledge (approved):
This makes sense to me. Explicit is better than implicit. Thanks!

PR #1215, opened 2023-03-05
Clarify specs for external installer interactions

### LINKED ISSUE #1210: Direct URL installation database spec is disconnected from the rest of the installation database spec
PEP 610 should have been modifying the installation database specification at https://packaging.python.org/en/latest/specifications/recording-installed-packages/ rather than defining a new independent specification at https://packaging.python.org/en/latest/specifications/direct-url/

The fact the latter isn't a truly independent specification is indicated by its opening sentence: "The direct_url.json file MUST be created in the *.dist-info directory by installers when installing a distribution from a requirement specifying a direct URL reference (including a VCS URL)."

While the term `*.dist-info directory` isn't explained or linked on the spec page, it's referring to the installation database.

It's a sufficiently complicated part of the specification that having a dedicated page for it makes sense, but there are some editorial tweaks that would help make the relationship between the two specs clearer:

* arrange for the ToC in https://packaging.python.org/en/latest/specifications/#package-distribution-metadata to nest the direct URL spec bullet under the installation database one
* add a ToC to https://packaging.python.org/en/latest/specifications/recording-installed-packages/ indicating that the subspec page exists (this should also be the simplest way to achieve the first point while maintaining the existing direct link URL for the subspec)
* adjust the text in https://packaging.python.org/en/latest/specifications/recording-installed-packages/ to cover the new optional file and reference the subspec page for details

### LINKED ISSUE #1211: Installation database spec does not make it clear how to protect a package from uninstallation
In a long discussion thread on DPO (https://discuss.python.org/t/wanting-a-singular-packaging-tool-vision/21141/190) the question came up as to why conda isn't using the installed distribution metadata to tell `pip` to leave conda-installed packages alone in the same way that Linux system package managers do.

Attempting to add a relevant link to the thread made me realise that finding such a link is *hard* due to two problems:

* the installation database spec doesn't *clearly* explain the historical mechanism for getting `pip` and similar tools to leave an installed package alone (omit or rename the `RECORD` file, and set `INSTALLER` to the name of the tool that should be used instead)
* PEP 668 (which covers the installation level EXTERNALLY-MANAGED file) isn't linked from the specifications page

This ticket covers the first point. The second is covered in TBD.

Technically, the installation database spec *does* provide the relevant information.

The final paragraph in https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-record-file reads:

> If the RECORD file is missing, tools that rely on .dist-info must not attempt to uninstall or upgrade the package. (This does not apply to tools that rely on other sources of information, such as system package managers in Linux distros.)

The final paragraph in https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-installer-file then reads:

> This value should be used for informational purposes only. For example, if a tool is asked to uninstall a project but finds no RECORD file, it may suggest that the tool named in INSTALLER may be able to do the uninstallation.

(`pip` specifically also checks the `INSTALLER` file to decide whether or not it should run its check to see if a newer version is available from PyPI: if the `INSTALLER` file for `pip` itself exists and contains a value other than `pip`, it skips the version availability check)

This would be quite a bit clearer if there was a dedicated installer interoperability section that said something like the following:

> ### Intentionally preventing package management interoperability
>
> In some cases (such as when needing to manage external dependencies in addition to Python ecosystem dependencies), it is
> desirable for a tool that installs packages into a Python environment to ensure that other tools are not used to uninstall or
> otherwise modify that installed package, as doing so may cause compatibility problems with the wider environment.
>
> To achieve this, affected tools should take the folllowing steps:
> * Rename or remove the `RECORD` file to prevent changes via other tools (e.g. appending a suffix to create a non-standard
>   `RECORD.tool` file if the tool itself needs the information, or omitting the file entirely if the package contents are tracked
>   elsewhere)
> * Write an `INSTALLER` file indicating the name of the tool that should be used to manage the package
>
> Python runtime providers may also prevent modification of platform provided packages by modifying the default Python
> package installation scheme to use a location other than that used by platform provided packages (while also ensuring both
> locations appear on the default Python import path).
>
> In some circumstances, it may be desirable to block even installation of additional packages via Python-specific tools.
> For these cases refer to (PPO PEP 668 link TBD)

### LINKED ISSUE #1212: PEP 668 (EXTERNALLY-MANAGED file) is missing from specifications page
In a long discussion thread on DPO (https://discuss.python.org/t/wanting-a-singular-packaging-tool-vision/21141/190) the question came up as to why conda isn't using the installed distribution metadata to tell `pip` to leave conda-installed packages alone in the same way that Linux system package managers do.

Attempting to add a relevant link to the thread made me realise that finding such a link is *hard* due to two problems:

* the installation database spec doesn't *clearly* explain the historical mechanism for getting `pip` and similar tools to leave an installed package alone (omit or rename the `RECORD` file, and set `INSTALLER` to the name of the tool that should be used instead)
* [PEP 668](https://peps.python.org/pep-0668/) (which covers the installation level EXTERNALLY-MANAGED file) isn't linked from the specifications page

This ticket covers the second point. The first is covered in #1211.

PEP 668 doesn't fit cleanly in any of the existing specification categories on https://packaging.python.org/en/latest/specifications/, but it is definitely a PyPA managed specification, since it doesn't affect Python interpreters directly, but is instead metadata that a Python runtime provided can use to influence the way that Python-specific package management tools behave.

Since it should be included, but doesn't fit into any of the existing categories, my suggestion would be to add a new category for "Python Environment Metadata" and link not only to PEP 668 (the `EXTERNALLY-MANAGED` spec), but also to PEP 405 (the `pyvenv.cfg` spec)

ncoghlan:
Closes #1210 
Closes #1211 
Closes #1212

### DIFF
--- a/source/specifications/recording-installed-packages.rst
+++ b/source/specifications/recording-installed-packages.rst
@@ -206,7 +206,7 @@ The entry_points.txt file
 =========================
 
 This file MAY be created by installers to indicate when packages contain
-components intended or discovery and use by other code, including console
+components intended for discovery and use by other code, including console
 scripts and other applications that the installer has made available for
 execution.
 
--- a/source/specifications/recording-installed-packages.rst
+++ b/source/specifications/recording-installed-packages.rst
@@ -21,12 +21,14 @@ History and change workflow
 ===========================
 
 The metadata described here was first specified in :pep:`376`, and later
-amended in :pep:`627`.
+amended in :pep:`627` (and other PEPs).
 It was formerly known as *Database of Installed Python Distributions*.
-Further amendments (except trivial language or typography fixes) must be made
-through the PEP process (see :pep:`1`).
+As with other PyPA specifications, editorial amendments with no functional
+impact may be made through the GitHub pull request workflow. Proposals for
+functional changes that would require amendments to package building and/or
+installation tools must be made through the PEP process (see :pep:`1`).
 
-While this document is the normative specification, these PEPs that introduce
+While this document is the normative specification, the PEPs that introduce
 changes to it may include additional information such as rationales and
 backwards compatibility considerations.
 
@@ -66,12 +68,14 @@ encouraged to start normalizing those fields.
     provide API for such tools to consume, so tools can have access to the
     unnormalized name when displaying distrubution information.
 
-This ``.dist-info`` directory can contain these files, described in detail
-below:
+This ``.dist-info`` directory may contain the following files, described in
+detail below:
 
 * ``METADATA``: contains project metadata
 * ``RECORD``: records the list of installed files.
 * ``INSTALLER``: records the name of the tool used to install the project.
+* ``entry_points.txt``: see :ref:`entry-points` for details
+* ``direct_url.json``: see :ref:`direct-url` for details
 
 The ``METADATA`` file is mandatory.
 All other files may be omitted at the installing tool's discretion.
@@ -169,7 +173,7 @@ Here is an example snippet of a possible ``RECORD`` file::
 
 If the ``RECORD`` file is missing, tools that rely on ``.dist-info`` must not
 attempt to uninstall or upgrade the package.
-(This does not apply to tools that rely on other sources of information,
+(This restriction does not apply to tools that rely on other sources of information,
 such as system package managers in Linux distros.)
 
 
@@ -197,6 +201,18 @@ For example, if a tool is asked to uninstall a project but finds no ``RECORD``
 file, it may suggest that the tool named in ``INSTALLER`` may be able to do the
 uninstallation.
 
+
+The entry_points.txt file
+=========================
+
+This file MAY be created by installers to indicate when packages contain
+components intended or discovery and use by other code, including console
+scripts and other applications that the installer has made available for
+execution.
+
+Its detailed specification is at :ref:`entry-points`.
+
+
 The direct_url.json file
 ========================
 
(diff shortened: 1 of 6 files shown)


source/specifications/recording-installed-packages.rst
```diff
@@ -197,6 +201,18 @@ For example, if a tool is asked to uninstall a project but finds no ``RECORD``
 file, it may suggest that the tool named in ``INSTALLER`` may be able to do the
 uninstallation.
 
+
+The entry_points.txt file
+=========================
+
+This file MAY be created by installers to indicate when packages contain
+components intended or discovery and use by other code, including console
```
pfmoore, line 209:
```suggestion
components intended for discovery and use by other code, including console
```

pfmoore (approved):

sinoroc:
Maybe the page about virtual environments should link to ["_How venvs work_"](https://docs.python.org/3/library/venv.html?highlight=venv#how-venvs-work) from the Python documentation. I had understood that in general we should avoid pointing people to PEPs whenever possible (PEPs are a process, not the spec).

ncoghlan:
@sinoroc Hmm, that's a good point. I was treating it like one of the other packaging specs, where the PEP was written to be the spec, but that isn't really the case with PEP 405: since it was adding a new standard library module, it put the behaviour spec in the library reference rather than leaving it solely in the PEP. I'll put an unresolved conversation on that so we don't merge this until I've had a chance to reconsider how to structure that bit.

source/specifications/virtual-environments.rst
```diff
@@ -0,0 +1,40 @@
+
+.. _virtual-environments:
+
+===========================
+Python Virtual Environments
+===========================
+
+In Python 3.3 and later versions, :pep:`405` provides interpreter level support
```
ncoghlan, line 8: As @sinoroc pointed out, this should be pointing to the standard library docs on how virtual environments work, rather than only referring to the PEP that added the venv module.

ncoghlan: venv module docs weren't quite suitable as the *sole* reference (they're more "how `venv` specifically works" and less "this is how runtimes are required to handle `pyvenv.cfg` files in general", while PEP 405 specifically has the latter in it)

Still, useful to include the module docs references in addition to the PEP ones. Also highlights the necessity of this new page though, as `Final` PEPs are definitely not the right place to have the sole documentation of ongoing interoperability requirements!

ncoghlan:
Tweaked the wording of the virtual environments section a bit, but still ended up having to relying primarily on the PEP 405 reference (the `venv` docs are more "how this module works", and less about the underlying interoperability requirements that PEP 405 laid out in terms of Python runtimes loading `pyvenv.cfg` files)

PR #1216, opened 2023-03-06
Quote extras to guard shells with glob qualifiers

### LINKED ISSUE #1209: Quote extras to avoid problems with glob qualifiers in shells
Some shells, like zsh, have glob qualifiers that will error if an extra is not quoted

```console
$ python -m pip install black[jupyter]
zsh: no matches found: black[jupyter]
$ python -m pip install --upgrade 'black[jupyter]'  # correct way without disabling globbing
```

Globbing can of course be turned off in these shells, but some users might not think of this right away.

At the moment all the extras in the packaging guide are unquoted:

https://github.com/pypa/packaging.python.org/blob/a103b5699eba73f0266044eeb8eecc6835642a3e/source/guides/installing-using-pip-and-virtual-environments.rst?plain=1#L295

https://github.com/pypa/packaging.python.org/blob/a103b5699eba73f0266044eeb8eecc6835642a3e/source/guides/installing-using-pip-and-virtual-environments.rst?plain=1#L301

https://github.com/pypa/packaging.python.org/blob/a103b5699eba73f0266044eeb8eecc6835642a3e/source/tutorials/installing-packages.rst?plain=1#L656-L658

https://github.com/pypa/packaging.python.org/blob/a103b5699eba73f0266044eeb8eecc6835642a3e/source/tutorials/installing-packages.rst?plain=1#L664-L666

Would a contribution that quotes all these be of interest? Or as this is not necessarily _required_ for installation of an extra should this not be added?

Tagging @pradyunsg on this as the same question [applies to `pip`'s docs](https://github.com/pypa/pip/blob/852deddb9c14afdb780e77ea28c2fa6d29f8c2e1/docs/html/cli/pip_install.rst?plain=1#L383-L403).

If this is of interest, I have a branch on my fork with this done already that I can PR.

matthewfeickert:
Resolves #1209

Shells like `zsh` have glob qualifiers that will error if an extra is not quoted. While the glob qualifiers can be disabled, adding quotes guards against errors if people are copy-pasting or do not know that they can disable the behavior.

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -292,13 +292,13 @@ specifying the extra in brackets:
 
     .. code-block:: bash
 
-        python3 -m pip install requests[security]
+        python3 -m pip install 'requests[security]'
 
 .. tab:: Windows
 
     .. code-block:: bat
 
-        py -m pip install requests[security]
+        py -m pip install "requests[security]"
 
 .. _extras:
     https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -653,17 +653,17 @@ you know publishes one, you can include it in the pip installation command:
 
     .. code-block:: bash
 
-        python3 -m pip install SomePackage[PDF]
-        python3 -m pip install SomePackage[PDF]==3.0
-        python3 -m pip install -e .[PDF]  # editable project in current directory
+        python3 -m pip install 'SomePackage[PDF]'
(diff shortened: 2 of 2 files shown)


matthewfeickert:
cc @sinoroc

matthewfeickert:
I've opened a similar PR over at `pip`: https://github.com/pypa/pip/pull/11842

sinoroc (approved):

matthewfeickert:
This is ready for review, so gentle ping to @henryiii and @pradyunsg if they have time.

henryiii (approved):
I'm in favor, I always add the quotes to be safe - it's also easy to write `pip install x>3` and not know why you suddenly have a file `3` and no output. :)

bhrutledge (approved):

Issue #1223, opened 2023-03-15
Major version change does not imply breaking change in semantic versioning


sinoroc:
The following bit taken out of the ["*`install_requires` vs requirements files*" article](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/?highlight=semantic%20versioning#install-requires) is incorrect in my opinion:

> It may also be known that project A follows semantic versioning, and that v2 of ‘A’ will indicate a break in compatibility, so it makes sense to not allow v2

As far as I know, [semantic versioning (v2.0 at the time if this writing)](https://semver.org/) does not specify that a "*breaking change*" is a condition necessary for incrementing the major version number. It is perfectly possible to increment the major version number without making any breaking change. It seems to be a rather widespread misunderstanding. This leads to a lot of libraries placing unnecessary upper bounds on their dependencies, which seems to break quite a lot of things in the Python packaging ecosystem nowadays.

I believe the phrasing should be corrected. There is not point in trying to predict the future. Upper bounds should be set once it is actually known that there is a breaking change, not before the breaking change happens.

I guess the following should section be fixed as well:

* https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#semantic-versioning-preferred

***Related***:

* https://iscinumpy.dev/post/bound-version-constraints/

PR #1224, opened 2023-03-15
Remove reference to SemVer to justify upper bound

sinoroc:
GitHub: fixes https://github.com/pypa/packaging.python.org/issues/1223


henryiii (approved):
I like this wording much better. :)

sinoroc:
@henryiii Is there anything I can or have to do to get this moving forward? I do not know what the process is for this particular project.

webknjaz (approved):

PR #1225, opened 2023-03-15
Quote guard version requirements for pip installs

matthewfeickert:
As pointed out by @henryiii in https://github.com/pypa/packaging.python.org/pull/1216#pullrequestreview-1342428923, adding quotes to version information on packages provides safeguards against common mistakes. Additionally this matches style with https://github.com/pypa/pip/pull/11842.

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -245,13 +245,13 @@ a specific version of ``requests``:
 
     .. code-block:: bash
 
-        python3 -m pip install requests==2.18.4
+        python3 -m pip install 'requests==2.18.4'
 
 .. tab:: Windows
 
     .. code-block:: bat
 
-        py -m pip install requests==2.18.4
+        py -m pip install "requests==2.18.4"
 
 To install the latest ``2.x`` release of requests:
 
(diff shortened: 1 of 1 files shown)


matthewfeickert:
This is ready for review.

webknjaz (approved):

webknjaz:
This is great improvement! I tend to add quotes myself to avoid such problems... Didn't realize it wasn't documented.

As a related improvement, we could also show that it's possible to use sparse identifiers with actual whitespaces when quoted.
Like `pip ~= 23.1`, or `resolvelib > 0.5, != 0.6.0`. I think, this makes them more readable and this style should be used in human-maintained files (as opposed to the tool-generated ones, like constraint pins made with `pip-tools`).

matthewfeickert:
> As a related improvement, we could also show that it's possible to use sparse identifiers with actual whitespaces when quoted. Like `pip ~= 23.1`, or `resolvelib > 0.5, != 0.6.0`. I think, this makes them more readable and this style should be used in human-maintained files (as opposed to the tool-generated ones, like constraint pins made with `pip-tools`).

Agreed. @pradyunsg had pointed this out in https://github.com/pypa/packaging.python.org/pull/1207#pullrequestreview-1295269111 and I think it improved the readability of the PR a lot.

PR #1236, opened 2023-04-09
inconsistent usage of double quotes

DrejT:
Double quotes are used when installing the package under the Installing from PyPi section,and not when upgrading the package. Both the methods work the same (according to my limited knowledge), but this could lead to misconceptions regarding upgrades and installs of python packages.

Instead add a separate note where the usage of quotes is acknowledged.


bhrutledge:
Thanks for PR, but the quotes are recommended to avoid shells interpreting special characters like `<` and `[`. See https://github.com/pypa/packaging.python.org/pull/1216 and https://github.com/pypa/packaging.python.org/pull/1225.

Issue #1238, opened 2023-04-23
Broken link to `pypa.io` specifications page


sinoroc:
On [`https://packaging.python.org/en/latest/specifications/`](https://packaging.python.org/en/latest/specifications/):

> This is a list of currently active interoperability specifications maintained by the Python Packaging Authority. The process for updating these standards, and for proposing new ones, is documented on [pypa.io](https://www.pypa.io/en/latest/specifications/).

The link is broken. Which fails `nox -s linkcheck` and blocks all pull requests.

Not sure when it happened, my guess is [when *Sphinx* was updated on `pypa.io`](https://github.com/pypa/pypa.io/commit/0088b8029b638a03b0c8ae387332ba34c9adaadd).

The new URL seems to be: [`https://www.pypa.io/en/latest/specifications.html`](https://www.pypa.io/en/latest/specifications.html).

PR #1239, opened 2023-04-23
Fix link to specifications page on `pypa.io`

sinoroc:
GitHub: fixes https://github.com/pypa/packaging.python.org/issues/1238


chrysle (approved):

bhrutledge (approved):
Thanks!

PR #1241, opened 2023-04-24
Improve heading and remove space in commands

chrysle:
This are some low-level improvements for readability.

### DIFF
--- a/source/guides/installing-using-linux-tools.rst
+++ b/source/guides/installing-using-linux-tools.rst
@@ -114,8 +114,8 @@ openSUSE
  
     sudo zypper install python3-pip python3-setuptools python3-wheel
 
-Debian/Ubuntu
-~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


source/guides/installing-using-linux-tools.rst
```diff
@@ -114,8 +114,8 @@ openSUSE
  
     sudo zypper install python3-pip python3-setuptools python3-wheel
 
-Debian/Ubuntu
-~~~~~~~~~~~~~
+Debian/Ubuntu and derivatives
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pradyunsg, line 118: This should get a new anchor as well.

chrysle: Thanks for reviewing. I have added an anchor. It doesn't seem to work though, can you help with the syntax?

bhrutledge (approved):

PR #1243, opened 2023-04-28
Fix typos in overview

cnpryer:
Not necessarily typos, so if you'd like for me to update the commit I'd be happy to.

### DIFF
--- a/source/overview.rst
+++ b/source/overview.rst
@@ -300,7 +300,7 @@ Bringing your own Python executable
 
 Computing as we know it is defined by the ability to execute
 programs. Every operating system natively supports one or more formats
-of the program they can natively execute.
+of programs they can natively execute.
 
 There are many techniques and technologies which turn your Python
 program into one of these formats, most of which involve embedding the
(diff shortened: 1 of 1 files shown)


source/overview.rst
```diff
@@ -300,7 +300,7 @@ Bringing your own Python executable
 
 Computing as we know it is defined by the ability to execute
 programs. Every operating system natively supports one or more formats
-of program they can natively execute.
+of the program they can natively execute.
```
chrysle, line 303:
```suggestion
of programs they can natively execute.
```

bhrutledge (approved):
Thanks!

chrysle:
@cnpryer You should still make the change.

cnpryer:
> @cnpryer You should still make the change.

I actually think I agree, but I mainly just wanted to point this out. If @bhrutledge agrees I'll push it (also have given maintainers push access).

chrysle:
I think he has already agreed by enabling auto-merge (this won't be merged before all conversations are resolved).

PR #1244, opened 2023-05-02
Update extras example to use unioned dependencies

### LINKED ISSUE #1194: Extras example for dependency specification lacks an example with multiple extras
https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras

There's one at https://packaging.python.org/en/latest/specifications/dependency-specifiers/#examples , but if you happened to be deep-linked into the "Extras" part of that page you probably would miss that example at the top.

savannahostrowski:
Saw this low hanging fruit hanging around - let me know if you were looking for something more elaborate!

Closes #1194


bhrutledge (approved):
Thank you!

PR #1247, opened 2023-05-06
Replace link to bottle project page with PyPi link

chrysle:
The bottle project page is not reachable at the moment, so I have replaced the link temporarily with a reference to PyPi.

I'll open a ticket soon.


bhrutledge (approved):

chrysle:
I opened an issue at bottlepy/bottle#1419

However, development to bottle seems to have been rather sparse lately so this could take some time.

PR #1248, opened 2023-05-07
CI: Don't run cron for forks

hugovk:
The test workflow triggers on a daily cron:

```yml
on: 
  push:
  pull_request:
  schedule:
    - cron: "0 6 * * *" # daily at 6am
```

This is good for upstream, but not necessary for contributors' forks, as it sends a daily email to contributors when it fails:

image

https://github.com/hugovk/packaging.python.org/actions

image

Let's add a condition so the `schedule` trigger only runs for upstream, and all other triggers run for both upstream and forks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ concurrency:
 jobs:
   build:
     name: ${{ matrix.noxenv }}
-    if: ${{ (github.repository_owner == 'pypa' && github.event_name == 'schedule') || github.event_name != 'schedule' }}
+    if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ concurrency:
 jobs:
   build:
     name: ${{ matrix.noxenv }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -13,6 +13,7 @@ concurrency:
 jobs:
   build:
     name: ${{ matrix.noxenv }}
+    if: ${{ (github.repository_owner == 'pypa' && github.event_name == 'schedule') || github.event_name != 'schedule' }}
```
bhrutledge, line 16: I think this could be shortened to:

```suggestion
    if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
```

bhrutledge (commented):
Good call. One suggestion on the logic.

.github/workflows/test.yml
hugovk: Looks good, let's try it :)

bhrutledge (approved):

hugovk:
Verification the daily `Test` cron still works for upstream: 

image

https://github.com/pypa/packaging.python.org/actions

And `Test` cancelled for me, which is good:

image

https://github.com/hugovk/packaging.python.org/actions

However the `Translation` one still triggers for forks, on the `workflow_run: types: completed` trigger:

```yml
on:
  workflow_run:
    workflows:
    - Test
    branches:
    - main
    types:
    - completed
```

It would be nice to filter that out in the same way, but I'm not sure what the `if:` would need to contain.

PR #1249, opened 2023-05-08
Homepage of bottlepy is running again

chrysle:
Looks like this has been fixed faster than expected. Reverts part of #1247

### DIFF
--- a/source/overview.rst
+++ b/source/overview.rst
@@ -66,8 +66,8 @@ library.
 This is great for sharing simple scripts and snippets between people
 who both have compatible Python versions (such as via email,
 StackOverflow, or GitHub gists). There are even some entire Python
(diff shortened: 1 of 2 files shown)


source/overview.rst
```diff
@@ -67,7 +67,7 @@ This is great for sharing simple scripts and snippets between people
 who both have compatible Python versions (such as via email,
 StackOverflow, or GitHub gists). There are even some entire Python
 libraries that offer this as an option, such as `bottle.py
-<https://pypi.org/project/bottle/>`_ and :doc:`boltons
+<https://bottlepy.org/docs/dev/>`_ and :doc:`boltons
```
webknjaz, line 70: It's Sphinx-powered. Can we link it via intersphinx instead?

webknjaz (approved):

PR #1258, opened 2023-05-22
removed references to PEP 582 which has been rejected.

venthur:
The [PEP](https://peps.python.org/pep-0582/#relationship-to-virtual-environments) has been rejected, so we shouldn't mention its support as a feature for tools anymore.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -486,10 +486,10 @@ pdm
 `GitHub <https://github.com/pdm-project/pdm/>`__ |
 `PyPI <https://pypi.org/project/pdm>`__
 
-PDM is a modern Python package manager with :pep:`582` support. It installs and
-manages packages in a similar way to ``npm`` that doesn't need to create a
-:term:`virtual environment` at all. It also uses :term:`pyproject.toml` to store
-project metadata as defined in :pep:`621`.
+PDM is a modern Python package manager. It installs and manages packages in a
+similar way to ``npm`` that doesn't need to create a :term:`virtual
+environment` at all. It also uses :term:`pyproject.toml` to store project
+metadata as defined in :pep:`621`.
 
 .. _pex:
 
--- a/source/tutorials/managing-dependencies.rst
+++ b/source/tutorials/managing-dependencies.rst
@@ -167,10 +167,8 @@ and techniques, listed in alphabetical order, to see if one of them is a better 
   wrapper around pip that supports ``requirements.txt``, Pipenv and Poetry lock files,
   or converting them to pip-tools compatible output. Designed for containerized
   Python applications, but not limited to them.
-* `PDM <https://github.com/pdm-project/pdm>`_ for a modern Python package management
-  tool supporting :pep:`582` (replacing virtual environments with ``__pypackages__``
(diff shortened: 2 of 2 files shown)


chrysle (approved):

source/tutorials/managing-dependencies.rst
```diff
@@ -167,10 +167,8 @@ and techniques, listed in alphabetical order, to see if one of them is a better
   wrapper around pip that supports ``requirements.txt``, Pipenv and Poetry lock files,
   or converting them to pip-tools compatible output. Designed for containerized
   Python applications, but not limited to them.
-* `PDM <https://github.com/pdm-project/pdm>`_ for a modern Python package management
-  tool supporting :pep:`582` (replacing virtual environments with ``__pypackages__``
-  directory for package installation) and relying on standards such as :pep:`517` and
-  :pep:`621`.
+* `PDM <https://github.com/pdm-project/pdm>`_ for a modern Python package
+  management relying on standards such as :pep:`517` and :pep:`621`.
```
bhrutledge, line 171: Does PDM still use `__pypackages__`? If so, it might be worth noting that it as a distinction from virtual environments. And even though PEP 582 is rejected, it might still be relevant as a reference.

chrysle: [This discussion](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/430) provides more information. It looks like the PEP isn't needed.

bhrutledge: I note that the [PDM docs](https://pdm.fming.dev/latest/usage/project/#virtualenv-or-not) mention the option of using a virtualenv or `__pypackages__`, but I can see how that's a detail that's less relevant to the readers of this doc.

source/key_projects.rst
```diff
@@ -486,10 +486,10 @@ pdm
 `GitHub <https://github.com/pdm-project/pdm/>`__ |
 `PyPI <https://pypi.org/project/pdm>`__
 
-PDM is a modern Python package manager with :pep:`582` support. It installs and
-manages packages in a similar way to ``npm`` that doesn't need to create a
-:term:`virtual environment` at all. It also uses :term:`pyproject.toml` to store
-project metadata as defined in :pep:`621`.
+PDM is a modern Python package manager. It installs and manages packages in a
+similar way to ``npm`` that doesn't need to create a :term:`virtual
+environment` at all. It also uses :term:`pyproject.toml` to store project
```
bhrutledge, line 491: The note about not needing to create a virtual environment seems to be at odds with removing the reference `__pypackages__` in the other doc. I'd recommend aligning those.

bhrutledge (changes requested):

source/key_projects.rst
venthur: good catch! i removed the npm reference.

source/tutorials/managing-dependencies.rst
venthur: I agree, it is probably not very useful information for a user that a particular PEP which was rejected is supported.

bhrutledge (approved):

PR #1251, opened 2023-05-14
Replace outdated reference to Python 3.10 as the latest 3.x release

chriskuehl:

PR #1261, opened 2023-05-24
Add guidance for trusted publishing

chrysle:
See https://github.com/pypa/packaging.python.org/pull/1251#issuecomment-1555881672 for context.

##### Preview

https://python-packaging-user-guide--1261.org.readthedocs.build/en/1261/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -3,15 +3,10 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 on: push
 
 jobs:
-  build-n-publish-pypi:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI
+  build:
+    name: Build the source package
     runs-on: ubuntu-latest
-    environment:
-      name: pypi
-      url: https://pypi.org/p/<package-name>
-    permissions:
-      id-token: write
-
+    
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python
@@ -28,11 +23,25 @@ jobs:
       run: >-
         python3 -m
         build
-        --sdist
-        --wheel
-        --outdir dist/
-        .
-    # Actually publish to PyPI
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
+      with:
+        name: python-package-distributions     
+  build-n-publish-pypi:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/<package-name>
+    permissions:
+      id-token: write
+
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
@@ -46,25 +55,11 @@ jobs:
       id-token: write
 
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python
-      uses: actions/setup-python@v4
+    - name: Download all the dists
+      uses: actions/download-artifact@v3
       with:
-        python-version: "3.x"
-    - name: Install pypa/build
-      run: >-
-        python3 -m
-        pip install
-        build
-        --user
-    - name: Build a binary wheel and a source tarball
-      run: >-
-        python3 -m
-        build
-        --sdist
-        --wheel
-        --outdir dist/
-        .
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to Test PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -3,9 +3,12 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 on: push
 
 jobs:
-  build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+  build-n-publish-pypi:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
     runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/<package-name>
     permissions:
       id-token: write
 
@@ -29,11 +32,40 @@ jobs:
         --wheel
         --outdir dist/
         .
-    # Actually publish to PyPI/TestPyPI
+    # Actually publish to PyPI
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+  build-n-publish-testpypi:
+    name: Build and publish Python 🐍 distributions 📦 to TestPyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/p/<package-name>
+    permissions:
+      id-token: write
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.x"
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python3 -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
     - name: Publish distribution 📦 to Test PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         repository-url: https://test.pypi.org/legacy/
-    - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
-      uses: pypa/gh-action-pypi-publish@release/v1
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -6,6 +6,8 @@ jobs:
   build-n-publish:
     name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
 
     steps:
     - uses: actions/checkout@v3
@@ -31,10 +33,7 @@ jobs:
     - name: Publish distribution 📦 to Test PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository-url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -36,8 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      # Fill in your project (e.g. repository) name
-      # for <package-name>
+      # Fill in your project (e.g. repository) name for <package-name>
       url: https://pypi.org/p/<package-name>
     permissions:
       id-token: write  # IMPORTANT: mandatory for trusted publishing
@@ -64,21 +63,24 @@ jobs:
       id-token: write  # IMPORTANT: mandatory for sigstore
 
     steps:
-     - name: Sign the dists with Sigstore
-       uses: sigstore/gh-action-sigstore-python@v1.2.3
-       with:
-         inputs: >-
-           ./dist/*.tar.gz
-           ./dist/*.whl
-     - name: Upload artifact signatures to GitHub Release
-       # Confusingly, this action also supports updating releases, not
-       # just creating them. This is what we want here, since we've manually
-       # created the release above.
-       uses: softprops/action-gh-release@v1
-       with:
-         # `dist/` contains the built packages, and the
-         # sigstore-produced signatures and certificates.
-         files: dist/**
+    - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Sign the dists with Sigstore
+      uses: sigstore/gh-action-sigstore-python@v1.2.3
+      with:
+        inputs: >-
+          ./dist/*.tar.gz
+          ./dist/*.whl
+    - name: Upload artifact signatures to GitHub Release
+      env:
+        GH_TOKEN: ${{ github.token }}
+      # Upload to GitHub Release using the `gh` CLI.
+      # `dist/` contains the built packages, and the
+      # sigstore-produced signatures and certificates.
+      run: gh release upload "${{ github.ref_name }}" dist/** --repo "${{ github.repository }}"
 
   publish-to-testpypi:
     name: Publish Python 🐍 distribution 📦 to TestPyPI
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -1,4 +1,4 @@
-name: Publish Python 🐍 distribution packages 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
 on: push
 
@@ -29,7 +29,7 @@ jobs:
 
   publish-to-pypi:
     name: >-
-      Publish Python 🐍 distribution packages 📦 to PyPI
+      Publish Python 🐍 distribution 📦 to PyPI
       and sign them with Sigstore
     needs:
     - build
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -2,13 +2,11 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
 on: push
 
-# Only trigger this for tag changes.
-if: startsWith(github.ref, 'refs/tags/')
-
 jobs:
   build:
     name: Build the source package
     runs-on: ubuntu-latest
+
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python
@@ -22,15 +20,19 @@ jobs:
         build
         --user
     - name: Build a binary wheel and a source tarball
-      run: >-
-        python3 -m
-        build
+      run: python3 -m build
     - name: Store the distribution packages
       uses: actions/upload-artifact@v3
       with:
         name: python-package-distributions
+        path: dist/
+
   publish-to-pypi:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    name: >-
+      Publish Python 🐍 distributions 📦 to PyPI
+      and sign them with Sigstore
+    needs:
+    - build
     runs-on: ubuntu-latest
     environment:
       name: pypi
@@ -46,8 +48,18 @@ jobs:
         path: dist/
     - name: Publish distribution 📦 to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Sign the dists with Sigstore
+      uses: sigstore/gh-action-sigstore-python@v1.2.3
+      with:
+        inputs: >-
+          ./dist/*.tar.gz
+          ./dist/*.whl
+
   publish-to-testpypi:
     name: Build and publish Python 🐍 distributions 📦 to TestPyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    needs:
+    - build
     runs-on: ubuntu-latest
     environment:
       name: testpypi
@@ -61,7 +73,7 @@ jobs:
       with:
         name: python-package-distributions
         path: dist/
-    - name: Publish distribution 📦 to Test PyPI
+    - name: Publish distribution 📦 to TestPyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         repository-url: https://test.pypi.org/legacy/
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      # Fill in your project (e.g. repository) name for <package-name>
+      # Replace <package-name> below with your PyPI project name:
       url: https://pypi.org/p/<package-name>
     permissions:
       id-token: write  # IMPORTANT: mandatory for trusted publishing
@@ -76,11 +76,11 @@ jobs:
           ./dist/*.whl
     - name: Upload artifact signatures to GitHub Release
       env:
-        GH_TOKEN: ${{ github.token }}
+        GITHUB_TOKEN: ${{ github.token }}
       # Upload to GitHub Release using the `gh` CLI.
       # `dist/` contains the built packages, and the
       # sigstore-produced signatures and certificates.
-      run: gh release upload "${{ github.ref_name }}" dist/** --repo "${{ github.repository }}"
+      run: gh release upload "${{ github.ref_name }}" dist/**
 
   publish-to-testpypi:
     name: Publish Python 🐍 distribution 📦 to TestPyPI
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -40,16 +40,16 @@ jobs:
       # for <package-name>
       url: https://pypi.org/p/<package-name>
     permissions:
-      id-token: write
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
 
     steps:
-     - name: Download all the dists
-       uses: actions/download-artifact@v3
-       with:
-         name: python-package-distributions
-         path: dist/
-     - name: Publish distribution 📦 to PyPI
-       uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Publish distribution 📦 to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
 
   github-release:
     name: >-
@@ -57,6 +57,12 @@ jobs:
       and upload them to GitHub Release
     needs:
     - publish-to-pypi
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write  # IMPORTANT: mandatory for making GitHub Releases
+      id-token: write  # IMPORTANT: mandatory for sigstore
+
     steps:
      - name: Sign the dists with Sigstore
        uses: sigstore/gh-action-sigstore-python@v1.2.3
@@ -70,8 +76,8 @@ jobs:
        # created the release above.
        uses: softprops/action-gh-release@v1
        with:
-         # dist/ contains the built packages, which smoketest-artifacts/
-         # contains the signatures and certificates.
+         # `dist/` contains the built packages, and the
+         # sigstore-produced signatures and certificates.
          files: dist/**
 
   publish-to-testpypi:
@@ -79,11 +85,13 @@ jobs:
     needs:
     - build
     runs-on: ubuntu-latest
+
     environment:
       name: testpypi
       url: https://test.pypi.org/p/<package-name>
+
     permissions:
-      id-token: write
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
 
     steps:
     - name: Download all the dists
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -1,10 +1,10 @@
-name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distribution packages 📦 to PyPI and TestPyPI
 
 on: push
 
 jobs:
   build:
-    name: Build the source package
+    name: Build distribution packages
     runs-on: ubuntu-latest
 
     steps:
@@ -29,7 +29,7 @@ jobs:
 
   publish-to-pypi:
     name: >-
-      Publish Python 🐍 distributions 📦 to PyPI
+      Publish Python 🐍 distribution packages 📦 to PyPI
       and sign them with Sigstore
     needs:
     - build
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -2,6 +2,9 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
 on: push
 
+# Only trigger this for tag changes.
+if: startsWith(github.ref, 'refs/tags/')
+
 jobs:
   build:
     name: Build the source package
@@ -42,7 +45,6 @@ jobs:
         name: python-package-distributions
         path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
   publish-to-testpypi:
     name: Build and publish Python 🐍 distributions 📦 to TestPyPI
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -79,7 +79,10 @@ jobs:
       # Upload to GitHub Release using the `gh` CLI.
       # `dist/` contains the built packages, and the
       # sigstore-produced signatures and certificates.
-      run: gh release upload "${{ github.ref_name }}" dist/**
+      run: >-
+        gh release upload
+        '${{ github.ref_name }}' dist/**
+        --repo '${{ github.repository }}'
 
   publish-to-testpypi:
     name: Publish Python 🐍 distribution 📦 to TestPyPI
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -36,8 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      # Replace <package-name> below with your PyPI project name:
-      url: https://pypi.org/p/<package-name>
+      url: https://pypi.org/p/<package-name>  # Replace <package-name> with your PyPI project name
     permissions:
       id-token: write  # IMPORTANT: mandatory for trusted publishing
 
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -6,7 +6,6 @@ jobs:
   build:
     name: Build the source package
     runs-on: ubuntu-latest
-    
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python
@@ -26,8 +25,8 @@ jobs:
     - name: Store the distribution packages
       uses: actions/upload-artifact@v3
       with:
-        name: python-package-distributions     
-  build-n-publish-pypi:
+        name: python-package-distributions
+  publish-to-pypi:
     name: Build and publish Python 🐍 distributions 📦 to PyPI
     runs-on: ubuntu-latest
     environment:
@@ -45,7 +44,7 @@ jobs:
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
-  build-n-publish-testpypi:
+  publish-to-testpypi:
     name: Build and publish Python 🐍 distributions 📦 to TestPyPI
     runs-on: ubuntu-latest
     environment:
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -1,10 +1,10 @@
-name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
 
 on: push
 
 jobs:
   build:
-    name: Build distribution packages
+    name: Build distribution 📦
     runs-on: ubuntu-latest
 
     steps:
@@ -30,34 +30,52 @@ jobs:
   publish-to-pypi:
     name: >-
       Publish Python 🐍 distribution 📦 to PyPI
-      and sign them with Sigstore
+    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
     needs:
     - build
     runs-on: ubuntu-latest
     environment:
       name: pypi
+      # Fill in your project (e.g. repository) name
+      # for <package-name>
       url: https://pypi.org/p/<package-name>
     permissions:
       id-token: write
 
     steps:
      - name: Download all the dists
-      uses: actions/download-artifact@v3
-      with:
-        name: python-package-distributions
-        path: dist/
-    - name: Publish distribution 📦 to PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-    - name: Sign the dists with Sigstore
-      uses: sigstore/gh-action-sigstore-python@v1.2.3
-      with:
-        inputs: >-
-          ./dist/*.tar.gz
-          ./dist/*.whl
+       uses: actions/download-artifact@v3
+       with:
+         name: python-package-distributions
+         path: dist/
+     - name: Publish distribution 📦 to PyPI
+       uses: pypa/gh-action-pypi-publish@release/v1
+
+  github-release:
+    name: >-
+      Sign the Python 🐍 distribution 📦 with Sigstore
+      and upload them to GitHub Release
+    needs:
+    - publish-to-pypi
+    steps:
+     - name: Sign the dists with Sigstore
+       uses: sigstore/gh-action-sigstore-python@v1.2.3
+       with:
+         inputs: >-
+           ./dist/*.tar.gz
+           ./dist/*.whl
+     - name: Upload artifact signatures to GitHub Release
+       # Confusingly, this action also supports updating releases, not
+       # just creating them. This is what we want here, since we've manually
+       # created the release above.
+       uses: softprops/action-gh-release@v1
+       with:
+         # dist/ contains the built packages, which smoketest-artifacts/
+         # contains the signatures and certificates.
+         files: dist/**
 
   publish-to-testpypi:
-    name: Build and publish Python 🐍 distributions 📦 to TestPyPI
-    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
     needs:
     - build
     runs-on: ubuntu-latest
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -17,61 +17,39 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
    details of building platform specific projects. If you have binary
    components, check out :ref:`cibuildwheel`'s GitHub Action examples.
 
-Saving credentials on GitHub
-============================
+Configuring trusted publishing
+==============================
 
-In this guide, we'll demonstrate uploading to both
-PyPI and TestPyPI, meaning that we'll have two separate sets
-of credentials. And we'll need to save them in the GitHub repository
-settings.
+This guide relies on PyPI's `trusted publishing`_ implementation to connect
+to `GitHub Actions CI/CD`_. This is recommended for security reasons, since 
+the generated tokens are created for each of your projects
+individually and expire automatically. Otherwise you'll need to generate an
+`API token`_ or provide a username/password combination for both PyPI and
+TestPyPI. 
+
+Since this guide will demonstrate uploading to both
+PyPI and TestPyPI, we'll need two trusted publishers configured. 
+The following steps will lead you through creating the "pending" publishers.
 
 Let's begin! 🚀
 
-1. Go to https://pypi.org/manage/account/#api-tokens and
-   create a new `API token`_. If you have the project on PyPI
-   already, limit the token scope to just that project.
-   You can call it something like
-   ``GitHub Actions CI/CD — project-org/project-repo``
-   in order for it to be easily distinguishable in the token
-   list.
-   **Don't close the page just yet — you won't see that token
-   again.**
-2. In a separate browser tab or window, go to the ``Settings``
-   tab of your target repository and then click on `Secrets`_
-   in the left sidebar.
-3. Create a new secret called ``PYPI_API_TOKEN`` and copy-paste
-   the token from the first step.
-4. Now, go to https://test.pypi.org/manage/account/#api-tokens
-   and repeat the steps. Save that TestPyPI token on GitHub
-   as ``TEST_PYPI_API_TOKEN``.
+1. Go to https://pypi.org/manage/account/publishing/
+2. Fill in the name you wish to publish your new project under,
+   your repository data and the name of the release workflow file 
+   under the ``.github/`` folder, see :ref:`workflow-definition`. 
+   Finally add the name of the GitHub Actions environment
+   running under your repository. 
+   Register the trusted publisher.
+3. Now, go to https://test.pypi.org/manage/account/publishing/ and repeat
+   the second step.   
+4. Your "pending" publishers are now ready for their first use and will 
+   create your projects automatically once you use them 
+   for the first time.
 
    .. attention::
 
       If you don't have a TestPyPI account, you'll need to
       create it. It's not the same as a regular PyPI account.
-
-
-Using trusted publishing
-------------------------
-
-It is also possible to authenticate to PyPI without having to provide 
-an `API token`_. This can be done using 
-PyPI's `trusted publishing`_ implementation. This is recommended
-also for security reasons, since the generated tokens are created for each of your projects
-individually and expire automatically. 
-
-The following steps will lead you through creating a "pending" publisher.
-
-1. Go to https://pypi.org/manage/account/publishing/
-2. Now fill in the name you wish to publish your new project under,
-   your repository data and the name of the release workflow file 
-   under the ``.github/`` folder, see :ref:`workflow-definition`. 
-   Finally add the name of the GitHub Actions environment
-   running under your repository. 
-   Add the trusted publisher.   
-3. Your "pending" publisher is no ready for its first use and will 
-   create your project automatically once you use it 
-   for the first time.
 
 
 .. _workflow-definition:
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -24,19 +24,23 @@ This guide relies on PyPI's `trusted publishing`_ implementation to connect
 to `GitHub Actions CI/CD`_. This is recommended for security reasons, since 
 the generated tokens are created for each of your projects
 individually and expire automatically. Otherwise you'll need to generate an
-`API token`_ or provide a username/password combination for both PyPI and
-TestPyPI. 
+`API token`_ for both PyPI and TestPyPI. In case of publishing to third-party
+indexes like :doc:`devpi <devpi:index>`, you will need to provide a
+username/password combination.
 
 Since this guide will demonstrate uploading to both
 PyPI and TestPyPI, we'll need two trusted publishers configured. 
-The following steps will lead you through creating the "pending" publishers.
+The following steps will lead you through creating the "pending" publishers
+for your new project. However it is also possible to add `trusted publishing`_
+to any pre-existing project, if you are its owner.
 
 Let's begin! 🚀
 
-1. Go to https://pypi.org/manage/account/publishing/
+1. Go to https://pypi.org/manage/account/publishing/.
 2. Fill in the name you wish to publish your new project under,
-   your repository data and the name of the release workflow file 
-   under the ``.github/`` folder, see :ref:`workflow-definition`. 
+   your GitHub username and repository name and 
+   the name of the release workflow file under 
+   the ``.github/`` folder, see :ref:`workflow-definition`. 
    Finally add the name of the GitHub Actions environment
    running under your repository. 
    Register the trusted publisher.
@@ -74,21 +78,24 @@ should make GitHub run this workflow:
 Defining a workflow job environment
 ===================================
 
-Now, let's add initial setup for our job. It's a process that
-will execute commands that we'll define later.
+We will have to define two jobs to publish to PyPI 
+and TestPyPI respectively.
+
+Now, let's add initial setup for our job that will publish to PyPI.
+It's a process that will execute commands that we'll define later.
 In this guide, we'll use the latest stable Ubuntu LTS version
 provided by GitHub Actions:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
    :start-after: on:
-   :end-before: steps:
+   :end-before: environment:
 
 
 Checking out the project and building distributions
 ===================================================
 
-Then, add the following under the ``build-n-publish`` section:
+Then, add the following under the ``build-n-publish-pypi`` section:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
@@ -96,7 +103,10 @@ Then, add the following under the ``build-n-publish`` section:
    :end-before: Install pypa/build
 
 This will download your repository into the CI runner and then
-install and activate the newest available Python 3 release.
+install and activate the newest available Python 3 release. It 
+also defines the package index to publish to, PyPI, and grants
+a permission to the action that is mandatory for trusted 
+publishing.
 
 And now we can build dists from source. In this example, we'll
 use ``build`` package.
@@ -114,25 +124,36 @@ So add this to the steps list:
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
    :start-after: version: "3.x"
-   :end-before: Actually publish to PyPI/TestPyPI
+   :end-before: Actually publish to PyPI
 
 
-Publishing the distribution to PyPI and TestPyPI
-================================================
+Publishing the distribution to PyPI
+===================================
 
 Finally, add the following steps at the end:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: Actually publish to PyPI/TestPyPI
+   :start-after: Actually publish to PyPI
+   :end-before: build-n-publish-testpypi
 
-These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
-Action: the first one uploads contents of the ``dist/`` folder
-into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged. It is recommended
-you use the latest release tag; a tool like GitHub's dependabot can keep
+This step uses the `pypa/gh-action-pypi-publish`_ GitHub
+Action: It uploads the contents of the ``dist/`` folder
+into PyPI unconditionally, but only if the current commit 
+is tagged. It is recommended you use the latest release 
+tag; a tool like GitHub's dependabot can keep
 these updated regularly.
 
+Separate workflow for publishing to TestPyPI
+============================================
+
+Now, repeat these steps and create another job for
+publishing to the TestPyPI package index under the ``jobs``
+section:
+
+.. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
+   :language: yaml
+   :start-after: uses: pypa/gh-action-pypi-publish@release/v1
 
 That's all, folks!
 ==================
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -79,7 +79,8 @@ Defining a workflow job environment
 ===================================
 
 We will have to define two jobs to publish to PyPI 
-and TestPyPI respectively.
+and TestPyPI respectively, and an additional job to 
+build the distribution packages.
 
 Now, let's add initial setup for our job that will publish to PyPI.
 It's a process that will execute commands that we'll define later.
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -51,6 +51,31 @@ Let's begin! 🚀
       create it. It's not the same as a regular PyPI account.
 
 
+Using trusted publishing
+------------------------
+
+It is also possible to authenticate to PyPI without having to provide 
+an `API token`_. This can be done using 
+PyPI's `trusted publishing`_ implementation. This is recommended
+also for security reasons, since the generated tokens are created for each of your projects
+individually and expire automatically. 
+
+The following steps will lead you through creating a "pending" publisher.
+
+1. Go to https://pypi.org/manage/account/publishing/
+2. Now fill in the name you wish to publish your new project under,
+   your repository data and the name of the release workflow file 
+   under the ``.github/`` folder, see :ref:`workflow-definition`. 
+   Finally add the name of the GitHub Actions environment
+   running under your repository. 
+   Add the trusted publisher.   
+3. Your "pending" publisher is no ready for its first use and will 
+   create your project automatically once you use it 
+   for the first time.
+
+
+.. _workflow-definition:
+
 Creating a workflow definition
 ==============================
 
@@ -150,3 +175,4 @@ sure that your release pipeline remains healthy!
    https://github.com/marketplace/actions/pypi-publish
 .. _Secrets:
    https://docs.github.com/en/actions/reference/encrypted-secrets
+.. _trusted publishing: https://docs.pypi.org/trusted-publishers/
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -33,13 +33,25 @@ username/password combination.
 Since this guide will demonstrate uploading to both
 PyPI and TestPyPI, we'll need two trusted publishers configured. 
 The following steps will lead you through creating the "pending" publishers
-for your new project. However it is also possible to add `trusted publishing`_
-to any pre-existing project, if you are its owner.
+for your new :term:`PyPI project <Project>`.
+However it is also possible to add `trusted publishing`_ to any
+pre-existing project, if you are its owner.
+
+   .. attention::
+
+      If you followed earlier versions of this guide, you will
+      have created the secrets ``PYPI_API_TOKEN`` and ``TEST_PYPI_API_TOKEN``
+      for direct PyPI and TestPyPI access. These are obsolete now and
+      you should remove them from your GitHub repository and revoke
+      them in your PyPI and TestPyPI account settings.
+
 
 Let's begin! 🚀
 
 1. Go to https://pypi.org/manage/account/publishing/.
-2. Fill in the name you wish to publish your new project under,
+2. Fill in the name you wish to publish your new
+   :term:`PyPI project <Project>` under
+   (the ``name`` value in your ``setup.cfg`` or ``pyproject.toml``),
    your GitHub username and repository name and
    the name of the release workflow file under
    the ``.github/`` folder, see :ref:`workflow-definition`.
@@ -53,10 +65,16 @@ Let's begin! 🚀
    create your projects automatically once you use them 
    for the first time.
 
-   .. attention::
+   .. note::
 
       If you don't have a TestPyPI account, you'll need to
       create it. It's not the same as a regular PyPI account.
+
+
+   .. hint::
+
+      For security reasons, you should require manual approval
+      on each run for the ``pypi`` environment.
 
 
 .. _workflow-definition:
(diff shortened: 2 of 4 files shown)


source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -51,6 +51,31 @@ Let's begin! 🚀
       create it. It's not the same as a regular PyPI account.
 
 
+Using trusted publishing
```
webknjaz, line 54: The idea of the guide is to be a simple and straightforward starting point. It should demonstrate a flow without any deviations, it should be linear.
Also, OIDC is more important as the default for anybody using GitHub. We can mention that it's possible to use login+password when publishing to third party indexes but it mustn't be the central part of this document.

webknjaz (commented):
Let's not add it as a second-class citizen but instead replace it with OIDC entirely.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -150,3 +175,4 @@ sure that your release pipeline remains healthy!
    https://github.com/marketplace/actions/pypi-publish
 .. _Secrets:
    https://docs.github.com/en/actions/reference/encrypted-secrets
+.. _trusted publishing: https://docs.pypi.org/trusted-publishers/
```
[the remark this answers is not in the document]
chrysle: The PyPI docs haven't configured that. They'd need the [mkdocstrings](https://mkdocstrings.github.io/usage/#cross-references-to-other-projects-inventories) plugin.

I consider adding it there.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
chrysle: I agree and have adapted the documentation and release workflow file.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
webknjaz: I'm not sure. This doc describes referencing other inventories, not producing one..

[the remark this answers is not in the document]
chrysle: See the last example.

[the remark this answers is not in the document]
webknjaz: Oh, I see now. Okay.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -6,6 +6,8 @@ jobs:
   build-n-publish:
     name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
     runs-on: ubuntu-latest
+    permissions:
```
webknjaz, line 9: Since this adds extra lines, please verify that it doesn't break included fragments in RST, especially line highlights.

webknjaz, line 9: It's strongly recommend to use environments to further lock it down:
```suggestion
    environment:
      name: pypi
      url: https://pypi.org/p/<package-name>
    permissions:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -31,10 +33,7 @@ jobs:
     - name: Publish distribution 📦 to Test PyPI
```
webknjaz, line 33: I've been actively discouraging publishing from the same job as building and publishing to PyPI and TestPyPI. We need to rewrite this workflow to split them. The latter would have `testpypi` as the environment name, allowing people to use differing protections on the GH side + the URL in the UI would be different as well. Also, the job building dists mustn't have access to `id-token` since it may potentially run untrusted code.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +17,43 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
-Saving credentials on GitHub
-============================
+Configuring trusted publishing
+==============================
+
+This guide relies on PyPI's `trusted publishing`_ implementation to connect
+to `GitHub Actions CI/CD`_. This is recommended for security reasons, since 
+the generated tokens are created for each of your projects
+individually and expire automatically. Otherwise you'll need to generate an
+`API token`_ or provide a username/password combination for both PyPI and
```
webknjaz, line 27: I wouldn't suggest using username and password for PyPI or TestPyPI. Instead, I'd only mention such a possibility in the context of publishing to third party indexes like devpi.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +17,43 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+TestPyPI. 
 
-In this guide, we'll demonstrate uploading to both
-PyPI and TestPyPI, meaning that we'll have two separate sets
-of credentials. And we'll need to save them in the GitHub repository
-settings.
+Since this guide will demonstrate uploading to both
+PyPI and TestPyPI, we'll need two trusted publishers configured. 
+The following steps will lead you through creating the "pending" publishers.
```
webknjaz, line 32: Let's mention that it's also possible to configure trust for pre-existing projects, if one is an owner.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +17,43 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
-   again.**
-2. In a separate browser tab or window, go to the ``Settings``
-   tab of your target repository and then click on `Secrets`_
-   in the left sidebar.
-3. Create a new secret called ``PYPI_API_TOKEN`` and copy-paste
-   the token from the first step.
-4. Now, go to https://test.pypi.org/manage/account/#api-tokens
-   and repeat the steps. Save that TestPyPI token on GitHub
-   as ``TEST_PYPI_API_TOKEN``.
+1. Go to https://pypi.org/manage/account/publishing/
```
webknjaz, line 36: We probably need a trailing period for style consistency here:
```suggestion
1. Go to https://pypi.org/manage/account/publishing/.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +17,43 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+2. Fill in the name you wish to publish your new project under,
+   your repository data and the name of the release workflow file
```
webknjaz, line 38: Since it's a tutorial, we shouldn't send people scroll the page back and forth. Instead, we have specific values in the examples and should explicitly tell the reader what to enter.

webknjaz (commented):
Here's some initial feedback. But I might've missed something reviewing from the phone..

chrysle:
Thanks! I tried to apply your review feedback best as possible.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
webknjaz: @chrysle I think you might've missed this comment.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -27,14 +32,40 @@ jobs:
...
       with:
-        password: ${{ secrets.PYPI_API_TOKEN }}
+        python-version: "3.x"
+    - name: Install pypa/build
+      run: >-
+        python3 -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
```
webknjaz, line 60: As I wrote in https://github.com/pypa/packaging.python.org/pull/1261/files#r1208671692, building should happen in a separate job, not each job. The publishing jobs **must not** have a build step at all.

I've got an example of the publishing jobs showing how to get dists out of the GHA artifacts at https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/098c781/.github/workflows/ci-cd.yml#L623-L682. There's a build job there too, but it wraps commands with tox. Here, we'll want to keep the pypa/build invocation, just look up how to store the artifacts, maybe.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -27,14 +32,40 @@ jobs:
...
+      run: >-
+        python3 -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
```
webknjaz, line 67: When moving the build step, plz drop all these arguments. They are not needed in the simple case this document describes. Also, declaring `--sdist --wheel` on CLI can be harmful in some cases.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
webknjaz: Marking this thread as resolved.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: Marking this thread as resolved.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -3,10 +3,10 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 on: push
 
 jobs:
-  build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+  build:
+    name: Build the source package
     runs-on: ubuntu-latest
-
+
```
webknjaz, line 9:
```suggestion

```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +23,44 @@ jobs:
...
-        .
-    # Actually publish to PyPI/TestPyPI
-    - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
+        name: python-package-distributions
```
webknjaz, line 29:
```suggestion
        name: python-package-distributions
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +23,44 @@ jobs:
...
-    # Actually publish to PyPI/TestPyPI
-    - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
+        name: python-package-distributions     
+  build-n-publish-pypi:
```
webknjaz, line 30:
```suggestion
  publish-to-pypi:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +23,44 @@ jobs:
...
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
+  build-n-publish-testpypi:
```
webknjaz, line 48:
```suggestion
  publish-to-testpypi:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +23,44 @@ jobs:
...
+      id-token: write
+
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
```
webknjaz, line 46: Such conditionals should be moved from the step level to the job level. No need to spawn an entire VM only to do nothing useful later on.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -8,7 +8,9 @@ popular choice is having a workflow that's triggered by a
 ``push`` event.
 This guide shows you how to publish a Python distribution
 whenever a tagged commit is pushed.
-It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
+It will use the `pypa/gh-action-pypi-publish GitHub Action`_ for
+publishing and `upload-artifact`_ and `download-artifact`_ actions
```
webknjaz, line 12: That's a long sentence. Let's try splitting it.
```suggestion
publishing. It also uses `upload-artifact`_ and `download-artifact`_ actions 
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,47 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
-3. Create a new secret called ``PYPI_API_TOKEN`` and copy-paste
-   the token from the first step.
-4. Now, go to https://test.pypi.org/manage/account/#api-tokens
-   and repeat the steps. Save that TestPyPI token on GitHub
-   as ``TEST_PYPI_API_TOKEN``.
+1. Go to https://pypi.org/manage/account/publishing/.
+2. Fill in the name you wish to publish your new project under,
+   your GitHub username and repository name and 
+   the name of the release workflow file under 
+   the ``.github/`` folder, see :ref:`workflow-definition`.
```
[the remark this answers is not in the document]
chrysle: I only have [Anaconda](http://damnwidget.github.io/anaconda/) in use, that doesn't care about trailing spaces in a RST file... Can you suggest me anything?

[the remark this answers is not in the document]
webknjaz: I think we should integrate pre-commit and use https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace... I'm not sure about other projects. I know that Vim/NeoVim can be configured to show trailing spaces visually, there's probably some plugins for that. Most projects just use pre-commit for this sort of linters/formatters.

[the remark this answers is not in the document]
chrysle: Thanks for suggesting, I've opened #1268 for that.

webknjaz:
Hm.. looks like I have a pending review that I didn't submit...

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
-        .
-    # Actually publish to PyPI/TestPyPI
-    - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
+        name: python-package-distributions
```
webknjaz, line 31: We have to specify the path to the artifacts we want to upload:
```suggestion
        name: python-package-distributions
        path: dist/
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+  publish-to-testpypi:
```
webknjaz, line 49:
```suggestion

  publish-to-testpypi:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
-    # Actually publish to PyPI/TestPyPI
-    - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
+        name: python-package-distributions
+  publish-to-pypi:
```
webknjaz, line 32: let's have visual separators between the job definitions
```suggestion

  publish-to-pypi:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
```
webknjaz, line 33: This job shouldn't duplicate building, it should just publish.
```suggestion
    name: >-
      Publish Python 🐍 distributions 📦 to PyPI
      and sign them with Sigstore
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+  publish-to-testpypi:
+    name: Build and publish Python 🐍 distributions 📦 to TestPyPI
```
webknjaz, line 50:
```suggestion
    name: Publish Python 🐍 distributions 📦 to TestPyPI
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+    permissions:
+      id-token: write
+
+    steps:
+    - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Publish distribution 📦 to Test PyPI
```
webknjaz, line 64:
```suggestion
    - name: Publish distribution 📦 to TestPyPI
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
-      uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Store the distribution packages
+      uses: actions/upload-artifact@v3
       with:
-        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
+        name: python-package-distributions
+  publish-to-pypi:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    runs-on: ubuntu-latest
```
webknjaz, line 34: Publishing has to wait until build produces the dists:
```suggestion
    needs:
    - build
    runs-on: ubuntu-latest
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+  publish-to-testpypi:
+    name: Build and publish Python 🐍 distributions 📦 to TestPyPI
+    runs-on: ubuntu-latest
```
webknjaz, line 51:
```suggestion
    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
    needs:
    - build
    runs-on: ubuntu-latest
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +77,77 @@ should make GitHub run this workflow:
...
+   :lines: 41-48
 
-These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
-Action: the first one uploads contents of the ``dist/`` folder
-into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged. It is recommended
-you use the latest release tag; a tool like GitHub's dependabot can keep
-these updated regularly.
+This step uses the `pypa/gh-action-pypi-publish`_ GitHub
+Action: After the stored distribution package has been
```
webknjaz, line 137:
```suggestion
Action: after the stored distribution package has been 
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -23,18 +25,43 @@ jobs:
...
+
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
```
webknjaz, line 48: I figured, while we're on it, let's show people how to sign the dists. Especially, as GPG support is being phased out o PyPI.
```suggestion
      uses: pypa/gh-action-pypi-publish@release/v1
    - name: Sign the dists with Sigstore
      uses: sigstore/gh-action-sigstore-python@v1.2.3
      with:
        inputs: >-
          ./dist/*.tar.gz
          ./dist/*.whl
```
Only add this to one job since we don't need to sign the same artifacts twice. We don't do this in the build job since it doesn't have the OIDC privilege.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +77,77 @@ should make GitHub run this workflow:
...
-These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
-Action: the first one uploads contents of the ``dist/`` folder
-into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged. It is recommended
-you use the latest release tag; a tool like GitHub's dependabot can keep
-these updated regularly.
+This step uses the `pypa/gh-action-pypi-publish`_ GitHub
+Action: After the stored distribution package has been 
+downloaded by the `download-artifact`_ action, it uploads 
+the contents of the ``dist/`` folder into PyPI unconditionally.
```
webknjaz, line 139:
```suggestion
the contents of the ``dist/`` folder into PyPI unconditionally.
This job also signs the artifacts with Sigstore right after publishing them to PyPI.
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -2,11 +2,13 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
 
 on: push
 
+# Only trigger this for tag changes.
+if: startsWith(github.ref, 'refs/tags/')
+
```
webknjaz, line 7: We can't have it on the workflow level since it'd contradict the statement we have at the very bottom, in the summary paragraph, that the dists to TestPyPI are published on non-tagged pushes.
```suggestion
```

chrysle:
Thanks a lot for the extensive review! Diving into this tomorrow.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,48 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
-2. In a separate browser tab or window, go to the ``Settings``
-   tab of your target repository and then click on `Secrets`_
-   in the left sidebar.
-3. Create a new secret called ``PYPI_API_TOKEN`` and copy-paste
-   the token from the first step.
-4. Now, go to https://test.pypi.org/manage/account/#api-tokens
-   and repeat the steps. Save that TestPyPI token on GitHub
-   as ``TEST_PYPI_API_TOKEN``.
+1. Go to https://pypi.org/manage/account/publishing/.
+2. Fill in the name you wish to publish your new project under,
```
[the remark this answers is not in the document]
sinoroc: Name of project? Name of distribution package?

* https://packaging.python.org/en/latest/glossary/#term-Project
* https://packaging.python.org/en/latest/glossary/#term-Distribution-Package

If using the terms as defined in glossary (with link) then it should be unambiguous.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,48 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+username/password combination.
 
-In this guide, we'll demonstrate uploading to both
-PyPI and TestPyPI, meaning that we'll have two separate sets
-of credentials. And we'll need to save them in the GitHub repository
-settings.
+Since this guide will demonstrate uploading to both
+PyPI and TestPyPI, we'll need two trusted publishers configured. 
+The following steps will lead you through creating the "pending" publishers
+for your new project. However it is also possible to add `trusted publishing`_
```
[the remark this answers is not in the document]
sinoroc: Maybe we should use project with link to glossary: https://packaging.python.org/en/latest/glossary/#term-Project

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -2,11 +2,13 @@ name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
...
 jobs:
-  build-n-publish:
-    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+  build:
+    name: Build the source package
```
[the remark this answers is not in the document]
sinoroc: Build distribution packages (source distribution and wheel)

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
sinoroc: Publish distribution packages to TestPyPI

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
chrysle: Can we use a relative link for that anchor?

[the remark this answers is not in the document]
webknjaz: Usually, glossary terms are linkable through the `:term:` role.
And I see it's available indeed: https://webknjaz.github.io/intersphinx-untangled/packaging.python.org/.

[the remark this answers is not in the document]
webknjaz: I'd still like that extra context so how about both a reference and a hint?
```suggestion
2. Fill in the name you wish to publish your new :term:`PyPI project <Project>` under (the ``name`` value in your ``setup.cfg`` or ``pyproject.toml``),
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
webknjaz: I'd still like that extra context so how about both a reference and a hint?
```suggestion
2. Fill in the name you wish to publish your new :term:`PyPI project <Project>` under (the ``name`` value in your ``setup.cfg`` or ``pyproject.toml``),
```

chrysle:
Updated!

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,60 @@ jobs:
...
+    steps:
+     - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
     - name: Publish distribution 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Sign the dists with Sigstore
```
webknjaz, line 51: I just realized that this is not enough. It doesn't publish the signatures.
This is another step that is needed:
```yaml
    - name: Upload artifact signatures to GitHub Release
      # Confusingly, this action also supports updating releases, not
      # just creating them. This is what we want here, since we've manually
      # created the release above.
      uses: softprops/action-gh-release@v1
      with:
        # dist/ contains the built packages, which smoketest-artifacts/
        # contains the signatures and certificates.
        files: dist/**
```

But since it turns into a multi-step job, I propose moving signing+uploading signatures into a separate job called `github-release`.

webknjaz (commented):
I've had this in draft for a while, and it's probably incomplete, but I'm posting the partial review to keep this going. It's better than blocking for a long time. I'll try to review again later, after the next batch of updates...

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
webknjaz: No problem! You're right, it might be getting complicated, but I'm trying to still keep it small. I was hoping that by providing people with something to copy-and-paste is going to cancel out most of the complexity. Though, this means that whatever we end up publishing shouldn't have silly mistakes, as it'd confuse newbies even more.
Honestly, I'd love to come up with a reusable workflow that people could make use of, but due to the limitations of the current implementation, it's not possible to use trusted publishing in such an environment. So maybe later.

webknjaz: @chrysle did you think of getting back to this? Do you need me to take over?

chrysle:
Sure, thanks for the ping! I'll send in a patch shortly.

chrysle:
Will get back to this tomorrow!

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
 
-   .. attention::
+
+Let's begin! 🚀
+
+1. Go to https://pypi.org/manage/account/publishing/.
+2. Fill in the name you wish to publish your new
+   :term:`PyPI project <Project>` under
+   (the ``name`` value in your ``setup.cfg`` or ``pyproject.toml``),
+   your GitHub username and repository name and
```
miketheman, line 55: **suggestion:**  Since users may add publishing from an org repository, telling them to add their own username may be misleading.
‏
```suggestion
   the GitHub repository owner's name and repository name and
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,68 +95,81 @@ should make GitHub run this workflow:
...
-will execute commands that we'll define later.
-In this guide, we'll use the latest stable Ubuntu LTS version
-provided by GitHub Actions:
-
-.. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
-   :language: yaml
-   :start-after: on:
-   :end-before: steps:
-
+This will also ensure that the release workflow is only triggered
```
miketheman, line 98: **nitpick:** ‏the updated comment states that this will only work on tagged, but the current yml example shows `on: push` - is that correct?

A later example shows `if: startsWith(github.ref, 'refs/tags/')` - that's only relevant to the testpypi workflow, not the pypi one.

Would it be better expressed as something like:

```yaml
on:
  push:
    tags:
      - '*'
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,68 +95,81 @@ should make GitHub run this workflow:
...
+and TestPyPI respectively, and an additional job to 
+build the distribution packages.
+
+First, we'll define the job for building the dist packages of 
+your project and storing them for later use:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: runs-on:
+   :start-after: jobs:
```
miketheman, line 114: **nitpick:** ‏this change makes `jobs:` never appear in the rendered document, so there's no place for the user to know that there needs to be a `jobs:` key.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,68 +95,81 @@ should make GitHub run this workflow:
...
+provided by GitHub Actions. This also defines a GitHub Environment
+for the job to run in its context and a URL to be displayed in GitHub's
+UI nicely. Additionally, it allows aqcuiring an OpenID Connect token
+which is mandatory that the ``pypi-publish`` actions needs to
+implement secretless trusted publishing to PyPI.
+
+.. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
+   :language: yaml
+   :start-after: path: dist/
+   :end-before: steps:
```
miketheman, line 144: This step shows the user the example `url: https://pypi.org/p/<package-name>` but makes no mention of which name to place in `package-name`

miketheman (commented):
Very cool to see this update! I added some comments inline, mostly in the style of https://conventionalcomments.org/ . Let me know if any are unclear or could use more details!

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: How about adding `(org or user)` clarification?

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: No, this is an incorrect assumption — the workflow can run on normal pushes and the build step would be a smoke test while the publishing would be skipped. Honestly, I'd rather refactor this to trigger on `workflow_dispatch:` but that would be too complex — maybe it's better to have a separate guide for a more complicated / comprehensive scenario.
I usually expect this to be a part of the same main CI/CD workflow that also includes tests. This way, you can also test those built artifacts before publishing.

webknjaz:
In https://github.com/python-jsonschema/check-jsonschema/commit/a9a3504ef9bcd19d410a0b9f5ac25a1f184e496e#r125535746 @sirosen suggested that it may be a good idea to have the entire workflow included in the document.
I think we can make another snippet inclusion with the entire YAML file, probably wrapped as a collapsible block.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
miketheman: Okay - I'll rely on your experience here. I think I've seen the build/publish steps as part of a distinct workflow, which would trigger on tags, so if the intent to keep this all in one workflow, then the refs/tags approach is fine!

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
chrysle: Thanks, maybe the `start-before` directive is better here.

chrysle:
> I think we can make another snippet inclusion with the entire YAML file, probably wrapped as a collapsible block.

Done, I hope the `sphinx-toolbox` dependency is fine.

sirosen:
Looking at the preview to see the changes; I have a few notes.

First, thanks for adding the complete example! It looks great to me.

Is there some explanation of sigstore which could be linked? Even as someone who tries to stay on top of things and reads the python discourse, this one isn't familiar to me. Which suggests that most readers of this guide won't know what it is. Plus, with pypi phasing out gpg signatures, anything that sounds like "signing a package" probably deserves some extra explanation.

I see the guide recommends using `softprops/action-gh-release`. I only looked into it minimally when I learned that github had deprecated their release artifact action, but some of that repo made our team a bit wary. (In particular, the github release stream vs the `v1` tag sends very mixed messages.)
I've been tinkering with using the `gh` CLI to do this instead, which should be something like
```yaml
      - name: Publish to GitHub Release
        env:
          GH_TOKEN: ${{ github.token }}
          GH_REPO: ${{ github.repository }}
        run: gh release upload "${{ github.ref_name }}" dist/*
```
(Caveat emptor: I think that's right, but I haven't tested it yet.)
If I get this working, would it be interesting to put into the guide instead, or no? I can make sure to share back the results of my tinkering/testing when it's done if so.

Last thing: I've used `sphinx-design` and been pretty happy with it. `sphinx-toolbox` looks to be similarly popular, but if there's any issue with `sphinx-toolbox`, there's another nice alternative. :slightly_smiling_face:

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,99 @@ should make GitHub run this workflow:
...
 
-And now we can build dists from source. In this example, we'll
-use ``build`` package.
+This step uses the `pypa/gh-action-pypi-publish`_ GitHub
+Action: after the stored distribution package has been
+downloaded by the `download-artifact`_ action, it uploads
+the contents of the ``dist/`` folder into PyPI unconditionally.
 
-.. tip::
+Signing the distribution packages
```
sethmlarson, line 161: Do we want to have Sigstore signing in the guide for publishing with Trusted Publishers? These are different concepts and the current reading makes it seem like you need both to accomplish the task?

chrysle:
> Is there some explanation of sigstore which could be linked? Even as someone who tries to stay on top of things and reads the python discourse, this one isn't familiar to me. Which suggests that most readers of this guide won't know what it is. Plus, with pypi phasing out gpg signatures, anything that sounds like "signing a package" probably deserves some extra explanation.

I already linked the [marketplace site ](https://github.com/marketplace/actions/gh-action-sigstore-python) of the `sigstore-python` GitHub Action (which originally is a command line tool) which contains all the documentation I have found.

> I see the guide recommends using `softprops/action-gh-release`. I only looked into it minimally when I learned that github had deprecated their release artifact action, but some of that repo made our team a bit wary. (In particular, the github release stream vs the `v1` tag sends very mixed messages.) 

Could you elaborate on this, probably provide an example? Maybe it's also a good idea to raise an issue.

sirosen:
> I already linked the [marketplace site ](https://github.com/marketplace/actions/gh-action-sigstore-python) of the `sigstore-python` GitHub Action (which originally is a command line tool) which contains all the documentation I have found.

Maybe I'm just out of the loop on this, but this is my first encounter with Sigstore. The result is that even with a link to the GitHub Action docs, I was missing all of the important context of "what is Sigstore?", "what value does signing with Sigstore provide my users?", "is this different from GPG signed packages?", etc.

I'm not sure exactly what linkage or explanation would have been best.
Maybe a link going straight to https://docs.sigstore.dev/ is the right place?

Mostly, as an outsider providing feedback, I wanted to make sure there's no assumption that readers have any prior notion of what Sigstore is or how to use it. Telling users to use the action, but without the context to know what it does, naturally raises these questions.

I guess you could see this as a fault of the sigstore action docs? What matters most, IMO, is that as a reader is that I get relatively quickly to some place where I know whether or not I can or should omit that part of the workflow.

> > I see the guide recommends using `softprops/action-gh-release`. I only looked into it minimally when I learned that github had deprecated their release artifact action, but some of that repo made our team a bit wary. (In particular, the github release stream vs the `v1` tag sends very mixed messages.)
> 
> Could you elaborate on this, probably provide an example? Maybe it's also a good idea to raise an issue.

There were basically two things that made us slightly uncomfortable.
- a growing backlog, including a backlog of PRs, with infrequent recent touches from the repo owner[^1]
- a "v1" tag, but [releases](https://github.com/softprops/action-gh-release/releases) numbered under 0.1.x

The versioning is strange and is what I referred to as "mixed messages" -- is this a "v1 stable" project, or a "v0.x unstable" project?

[^1]: I'm guilty of letting PRs lag on some projects too, so I understand how that can happen. But it often is a sign of burnout or disinvestment in the project by the author, and it can become a feed-forward process. It might be best to try to find a new maintainer, but that's a difficult process in its own right. None of this is meant as criticism. It's an assessment of the maintenance and health of the project.

Looking for another path for this functionality, the GitHub CLI seems more appropriate because GitHub and their (large, non-volunteer) team is responsible for making sure that `gh release upload` works.

The fact that GitHub has dropped their own action (https://github.com/actions/upload-release-asset) and then pointed at someone else's version of the same looks strange to me. There's a sympathetic reading of this -- that they wanted to boost someone else's project -- but I have a more critical take. It's not directly relevant to assessing the long-term stability of these tools though, so I'll refrain from wandering too far OT.

webknjaz:
@chrysle looks like https://github.com/pypa/packaging.python.org/pull/1275/files#diff-6d2963726fc7c77dce51b2199325c2951882227016bf62bdd0340d84bdd6af11R106-R107 introduced a small merge conflict that needs addressing.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: > Okay - I'll rely on your experience here. I think I've seen the build/publish steps as part of a distinct workflow, which would trigger on tags, so if the intent to keep this all in one workflow, then the refs/tags approach is fine!

Yeah, I've seen different workflows too. I just didn't want to change too many things in this PR. But later, I'd definitely rethink this. Though, I was considering a separate more in-depth document rather then extending this with multiple branches.

webknjaz:
> Is there some explanation of sigstore which could be linked? Even as someone who tries to stay on top of things and reads the python discourse, this one isn't familiar to me. Which suggests that most readers of this guide won't know what it is.

Not off the top of my head.

> Plus, with pypi phasing out gpg signatures, anything that sounds like "signing a package" probably deserves some extra explanation.

This is exactly why I wanted to have a ready-to-copy example. I've also been watching the development of the security changes around PyPI. GPG has been considered discouraged for like a decade, and Warehouse (the PyPI web app project) was working on the TUF among other things. Sigstore started emerging over the past few years and seems to be getting most traction. It is also targeting things outside the Python ecosystem, like containers.

I've seen some mentions of the possibility of integrating Sigstore but for now, the demo just shows how to publish signatures via GH releases.
I'm sure @di or @woodruffw could provide better insight, being more involved with it. My understanding is that they at least contributed to the Python implementation:
* https://blog.sigstore.dev/announcing-the-1-0-release-of-sigstore-python-4f5d718b468d/
* https://blog.trailofbits.com/2023/01/13/sigstore-python/

By the way, CPython itself started signing the python.org artifacts with Sigstore: https://www.python.org/download/sigstore/.
It's also been mentioned in the context of PyPI @ https://pbarker.dev/posts/2023-05-24/thoughts-on-pypi-pgp-and-sigstore/.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+      id-token: write
+
+    steps:
+     - name: Download all the dists
+       uses: actions/download-artifact@v3
+       with:
+         name: python-package-distributions
+         path: dist/
+     - name: Publish distribution 📦 to PyPI
+       uses: pypa/gh-action-pypi-publish@release/v1
```
webknjaz, line 52: Looks like this adds one-space indents for some reason.
```suggestion
    - name: Download all the dists
      uses: actions/download-artifact@v3
      with:
        name: python-package-distributions
        path: dist/
    - name: Publish distribution 📦 to PyPI
      uses: pypa/gh-action-pypi-publish@release/v1
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+
+  github-release:
+    name: >-
+      Sign the Python 🐍 distribution 📦 with Sigstore
+      and upload them to GitHub Release
+    needs:
+    - publish-to-pypi
+    steps:
+     - name: Sign the dists with Sigstore
```
webknjaz, line 61: This also has one-space indents which is inconsistent with no-indent sequences in other places in this file.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+           ./dist/*.tar.gz
+           ./dist/*.whl
+     - name: Upload artifact signatures to GitHub Release
+       # Confusingly, this action also supports updating releases, not
+       # just creating them. This is what we want here, since we've manually
+       # created the release above.
+       uses: softprops/action-gh-release@v1
+       with:
+         # dist/ contains the built packages, which smoketest-artifacts/
+         # contains the signatures and certificates.
```
webknjaz, line 74: Looks like I copied this from somewhere but it doesn't read well:
```suggestion
         # `dist/` contains the built packages, and the
         # sigstore-produced signatures and certificates.
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+     - name: Publish distribution 📦 to PyPI
+       uses: pypa/gh-action-pypi-publish@release/v1
+
+  github-release:
+    name: >-
+      Sign the Python 🐍 distribution 📦 with Sigstore
+      and upload them to GitHub Release
+    needs:
+    - publish-to-pypi
+    steps:
```
webknjaz, line 60: The first step should be the same call to `actions/download-artifact` as in the previous job — sigstore needs the files it'll sign to be on disk, in the locations mentioned as inputs.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+         path: dist/
+     - name: Publish distribution 📦 to PyPI
+       uses: pypa/gh-action-pypi-publish@release/v1
+
+  github-release:
+    name: >-
+      Sign the Python 🐍 distribution 📦 with Sigstore
+      and upload them to GitHub Release
+    needs:
+    - publish-to-pypi
```
webknjaz, line 59: The job needs these privileges. The target GHA VM is also mandatory for every job definition.
```suggestion
    needs:
    - publish-to-pypi
    runs-on: ubuntu-latest
    permissions:
      contents: write  # IMPORTANT: mandatory for making GitHub Releases
      id-token: write  # IMPORTANT: mandatory for trusted publishing & sigstore
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+  publish-to-testpypi:
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/p/<package-name>
+    permissions:
+      id-token: write
```
webknjaz, line 86: Let's show the comments as this is an educational document, after all:
```suggestion
      id-token: write  # IMPORTANT: mandatory for trusted publishing
```

Plus, it's important to record the context in which certain privileges are required.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      # Fill in your project (e.g. repository) name
+      # for <package-name>
+      url: https://pypi.org/p/<package-name>
+    permissions:
+      id-token: write
```
webknjaz, line 43:
```suggestion
      id-token: write  # IMPORTANT: mandatory for trusted publishing
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+  publish-to-pypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      # Fill in your project (e.g. repository) name
```
webknjaz, line 39: In general, comments should be on the same line, if they fit. This helps prevent other lines from being injected in-between over time, which would make the comments drift away from their context.

webknjaz:
> I see the guide recommends using `softprops/action-gh-release`. I only looked into it minimally when I learned that github had deprecated their release artifact action, but some of that repo made our team a bit wary. (In particular, the github release stream vs the `v1` tag sends very mixed messages.) I've been tinkering with using the `gh` CLI to do this instead, which works something like
> 
> ```yaml
>       - name: Publish to GitHub Release
>         env:
>           GH_TOKEN: ${{ github.token }}
>           GH_REPO: ${{ github.repository }}
>         run: gh release upload "${{ github.ref_name }}" dist/*
> ```
> 
> Would it be interesting to put into the guide instead or in addition to the existing guidance?

I think it's a good idea for as long as it covers the same features being used. Although, I wouldn't use the env vars — they are likely unneeded. But if it's necessary, I'd just use CLI args. Also, this probably requires a trailing `**` to match files in subdirs.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: @sethmlarson this is not a tutorial about trusted publishing but about publishing to PyPI from GHA which happens to use OIDC instead of long-lived credentials now. Though, I get your point and it's probably hard to pick where to draw the line between what's in and out of the scope. Perhaps, a `.. note::` admonition could explain that this job is not mandatory.
But I do feel like it's important to show people a friction-free way of signing their artifacts, especially given the now-official deprecation of GPGs in Warehouse.

webknjaz:
> > I already linked the [marketplace site ](https://github.com/marketplace/actions/gh-action-sigstore-python) of the `sigstore-python` GitHub Action (which originally is a command line tool) which contains all the documentation I have found.
> 
> Maybe I'm just out of the loop on this, but this is my first encounter with Sigstore. The result is that even with a link to the GitHub Action docs, I was missing all of the important context of "what is Sigstore?", "what value does signing with Sigstore provide my users?", "is this different from GPG signed packages?", etc.
> 
> I'm not sure exactly what linkage or explanation would have been best. Maybe a link going straight to https://docs.sigstore.dev/ is the right place?
> 
> Mostly, as an outsider providing feedback, I wanted to make sure there's no assumption that readers have any prior notion of what Sigstore is or how to use it. Telling users to use the action, but without the context to know what it does, naturally raises these questions.
> 
> I guess you could see this as a fault of the sigstore action docs? What matters most, IMO, is that as a reader is that I get relatively quickly to some place where I know whether or not I can or should omit that part of the workflow.

I fully agree with this. Sigstore is known among people who are directly in touch with security-related activities but often not to ordinary people.

Though, I'd probably link the root https://www.sigstore.dev site (or both root and docs). And add some context along the lines of "This is a replacement for GPG signatures, it allows signing the artifacts generated on GitHub. In this example, the signatures are uploaded to GitHub releases."

@chrysle @sirosen WDYT?

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+       uses: sigstore/gh-action-sigstore-python@v1.2.3
+       with:
+         inputs: >-
+           ./dist/*.tar.gz
+           ./dist/*.whl
+     - name: Upload artifact signatures to GitHub Release
+       # Confusingly, this action also supports updating releases, not
+       # just creating them. This is what we want here, since we've manually
+       # created the release above.
+       uses: softprops/action-gh-release@v1
```
webknjaz, line 71: @chrysle plz replace this action per https://github.com/pypa/packaging.python.org/pull/1261#issuecomment-1712995303 and @sirosen's suggestions.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,99 @@ should make GitHub run this workflow:
...
+The whole CD workflow
+=====================
 
-These two steps use the `pypa/gh-action-pypi-publish`_ GitHub
-Action: the first one uploads contents of the ``dist/`` folder
-into TestPyPI unconditionally and the second does that to
-PyPI, but only if the current commit is tagged. It is recommended
-you use the latest release tag; a tool like GitHub's dependabot can keep
-these updated regularly.
+.. collapse:: Load file
```
webknjaz, line 187:
```suggestion
.. collapse:: Click here to display the entire GitHub Actions CI/CD workflow definition
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,99 @@ should make GitHub run this workflow:
...
+Now, repeat these steps and create another job for
+publishing to the TestPyPI package index under the ``jobs``
+section:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: Actually publish to PyPI/TestPyPI
+   :start-at: publish-to-testpypi
+
+The whole CD workflow
```
webknjaz, line 184: This section looks weird with just one collapsed line. Can we add some paragraph at the beginning? Or think of some other structure?

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+         # dist/ contains the built packages, which smoketest-artifacts/
+         # contains the signatures and certificates.
+         files: dist/**
+
+  publish-to-testpypi:
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
```
webknjaz, line 82:
```suggestion

    environment:
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,78 @@ jobs:
...
+      name: testpypi
+      url: https://test.pypi.org/p/<package-name>
+    permissions:
```
webknjaz, line 85:
```suggestion

    permissions:
```

chrysle:
Thanks, I tried to apply your review comments best as possible!

> Though, I'd probably link the root https://www.sigstore.dev site (or both root and docs). And add some context along the lines of "This is a replacement for GPG signatures, it allows signing the artifacts generated on GitHub. In this example, the signatures are uploaded to GitHub releases."

Sure! I also linked the blog post announcing that PyPI has dropped support for PGP/GPG.

> I think it's a good idea for as long as it covers the same features being used. Although, I wouldn't use the env vars — they are likely unneeded. But if it's necessary, I'd just use CLI args. Also, this probably requires a trailing `**` to match files in subdirs.

Done! However, the env var seems to be necessary for the GitHub token, since I have not found a corresponding CLI arg.

sirosen:
> > I think it's a good idea for as long as it covers the same features being used. Although, I wouldn't use the env vars — they are likely unneeded. But if it's necessary, I'd just use CLI args. Also, this probably requires a trailing `**` to match files in subdirs.
> 
> Done! However, the env var seems to be necessary for the GitHub token, since I have not found a corresponding CLI arg.

Yeah, my understanding is that the env var is the simplest way of getting the token from the workflow context into the `gh` CLI. If there's a better way, that would be great, but this is what I have tested and verified for now.
Thanks for being so receptive to this suggested change!

> Though, I'd probably link the root https://www.sigstore.dev site (or both root and docs). And add some context along the lines of "This is a replacement for GPG signatures, it allows signing the artifacts generated on GitHub. In this example, the signatures are uploaded to GitHub releases."
> 
> [...] @sirosen WDYT?

That seems like a good summary explanation.
It wasn't that hard for me to get from `sigstore.dev` to the doc site, so any linkage in this family of sites is probably sufficient.

> By the way, CPython itself started signing the python.org artifacts with Sigstore: https://www.python.org/download/sigstore/.

I think this is *really* relevant, in that it shows that there's a lot of community buy-in for this signing solution.
Clearly I need to do some reading on the topic to catch-up! What do you think of including a mild reference to this, as it helps build readers' confidence in Sigstore if (like me) they hadn't heard of it?
e.g., "The following job signs the package with Sigstore, the same artifact signing system [used to sign CPython](https://www.python.org/download/sigstore/)."

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,88 @@ jobs:
...
+  publish-to-pypi:
+    name: >-
+      Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    needs:
+    - build
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      # Fill in your project (e.g. repository) name for <package-name>
```
webknjaz, line 39: The term "repository" is quite overloaded. Let's not give the readers a change to assume it's a Git repo.. `pypi.org`, `test.pypi.org` and things like devpi deployments are also _[software]_ repositories.
```suggestion
      # Replace <package-name> below with your PyPI project name:
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
-   You can use any other method for building distributions as long as
-   it produces ready-to-upload artifacts saved into the
-   ``dist/`` folder. You can even use ``actions/upload-artifact`` and
-   ``actions/download-artifact`` to transfer files between jobs or make them
-   accessible for download from the web CI interface.
-
-So add this to the steps list:
+This additional job signs the distribution packages with `Sigstore`_,
+using the `sigstore/gh-action-sigstore-python GitHub Action`_,
+and then uploads them to GitHub Release.
```
webknjaz, line 166:
```suggestion
and then uploads them to a GitHub Release.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
+   :end-before:  publish-to-testpypi
 
 
-Publishing the distribution to PyPI and TestPyPI
-================================================
+.. note::
 
-Finally, add the following steps at the end:
+   This is a replacement for GPG signatures, for which support has been
+   `removed <https://blog.pypi.org/posts/2023-05-23-removing-pgp/>`_ by PyPI.
```
webknjaz, line 177: how about
```suggestion
   `removed from PyPI <https://blog.pypi.org/posts/2023-05-23-removing-pgp/>`_.
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,88 @@ jobs:
...
+      url: https://pypi.org/p/<package-name>
```
webknjaz, line 40: Oh, I realized that the earlier request to put it on the same line needs addressing too:
```suggestion
      url: https://pypi.org/p/<package-name>  # Replace <package-name> with your PyPI project name
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,88 @@ jobs:
...
+        path: dist/
+    - name: Sign the dists with Sigstore
+      uses: sigstore/gh-action-sigstore-python@v1.2.3
+      with:
+        inputs: >-
+          ./dist/*.tar.gz
+          ./dist/*.whl
+    - name: Upload artifact signatures to GitHub Release
+      env:
+        GH_TOKEN: ${{ github.token }}
```
webknjaz, line 79: https://cli.github.com/manual/gh_help_environment says that the environment variable `GITHUB_TOKEN` is used for auth. I think GHA used to set it as an environment variable in the past but probably doesn't do so anymore. That's what confused me, I think. I'd still use that name because it's easier to grep for it.
```suggestion
      env:
        GITHUB_TOKEN: ${{ github.token }}
```

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -20,21 +20,88 @@ jobs:
...
+      # Upload to GitHub Release using the `gh` CLI.
+      # `dist/` contains the built packages, and the
+      # sigstore-produced signatures and certificates.
+      run: gh release upload "${{ github.ref_name }}" dist/** --repo "${{ github.repository }}"
```
webknjaz, line 83: I wonder if `--repo` is necessary... https://cli.github.com/manual/gh_release_upload says "Select another repository using the [HOST/]OWNER/REPO format" so the default is the current repository, I think.

webknjaz:
> What do you think of including a mild reference to this, as it helps build readers' confidence in Sigstore if (like me) they hadn't heard of it?
> e.g., "The following job signs the package with Sigstore, the same artifact signing system [used to sign CPython](https://www.python.org/download/sigstore/)."

That sounds reasonable! @chrysle would you include this too? Maybe add another commit with a co-author per https://hynek.me/til/easier-crediting-contributors-github/?

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
-   .. attention::
+
+Let's begin! 🚀
+
+1. Go to https://pypi.org/manage/account/publishing/.
+2. Fill in the name you wish to publish your new
+   :term:`PyPI project <Project>` under
+   (the ``name`` value in your ``setup.cfg`` or ``pyproject.toml``),
+   the GitHub repository owner's name (org or user)
+   and repository name and the name of the release workflow file under
```
webknjaz, line 56: I think we should stick some commas here:
```suggestion
   the GitHub repository owner's name (org or user),
   and repository name, and the name of the release workflow file under
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+   the ``.github/`` folder, see :ref:`workflow-definition`.
+   Finally add the name of the GitHub Actions environment
```
webknjaz, line 58: This is a bit confusing but the term is GitHub Environment. There's no "GitHub Actions environments" but the GitHub Environments can be specified in the GitHub Actions job definitions which is something different.
```suggestion
   Finally, add the name of the GitHub Environment
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+   (``pypi``) we're going set up under your repository.
+   Register the trusted publisher.
+3. Now, go to https://test.pypi.org/manage/account/publishing/ and repeat
+   the second step, but now enter ``testpypi`` as the name of the
```
webknjaz, line 62: Let's avoid the repetition of "now":
```suggestion
   the second step, but this time, enter ``testpypi`` as the name of the
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+   GitHub Actions environment.
```
webknjaz, line 63: ...and fix the terminology:
```suggestion
   GitHub Environment.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -17,40 +19,66 @@ It will use the `pypa/gh-action-pypi-publish GitHub Action`_.
...
+
+   .. note::
 
       If you don't have a TestPyPI account, you'll need to
       create it. It's not the same as a regular PyPI account.
 
 
+   .. hint::
+
+      For security reasons, you should require manual approval
```
webknjaz, line 76: Many people probably wouldn't know what this is and where to set it up. We should either add this to the steps above or at least, link something like https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#deployment-protection-rules.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
webknjaz, line 1: `actions/checkout@v3` should now be updated to `actions/checkout@v4` which got released recently.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: on:
+   :start-after: path: dist/
    :end-before: steps:
 
+This will also ensure that the PyPI publishing workflow is only triggered
+if the current commit is tagged. It is recommended you use the
+latest release tag.
```
webknjaz, line 144: I'm not sure what this last sentence means. Can we rephrase it to make it more illustrative?

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
 
 
-Publishing the distribution to PyPI and TestPyPI
-================================================
+.. note::
 
-Finally, add the following steps at the end:
+   This is a replacement for GPG signatures, for which support has been
+   `removed <https://blog.pypi.org/posts/2023-05-23-removing-pgp/>`_ by PyPI.
+   However, this job is not mandatory for defining the workflow.
```
webknjaz, line 178:
```suggestion
   However, this job is not mandatory for uploading to PyPI and can be omitted.
```

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
+that the ``pypi-publish`` actions needs to implement secretless
+trusted publishing to PyPI.
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: on:
+   :start-after: path: dist/
    :end-before: steps:
 
+This will also ensure that the PyPI publishing workflow is only triggered
```
webknjaz, line 142: Let's also mention somewhere in this paragraph that it's important to set up the manual approval requirement for the `pypi` gh env.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,109 @@ should make GitHub run this workflow:
...
+
+Now, repeat these steps and create another job for
+publishing to the TestPyPI package index under the ``jobs``
+section:
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: Actually publish to PyPI/TestPyPI
+   :start-at: publish-to-testpypi
+
```
webknjaz, line 191:
```suggestion

.. tip::

   Requiring manual approvals in the `testpypi` GitHub Environment is typically unnecessary as it's designed to run on each commit to the main branch and is often used to indicate a healthy release publishing pipeline.


```

webknjaz (commented):
Hopefully, these comments are the last edits...

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
chrysle: Right, looks like it:

> GH_REPO: specify the GitHub repository in the "[HOST/]OWNER/REPO" format for commands that otherwise operate on a local repository.

https://cli.github.com/manual/gh_help_environment

chrysle:
Addressed the comments!

> That sounds reasonable! @chrysle would you include this too? Maybe add another commit with a co-author per https://hynek.me/til/easier-crediting-contributors-github/?

To me too! Nice guide! Since I have a somewhat outdated `git` version on a somewhat outdated Ubuntu LTS not providing `--trailer`, I modified the script to work with a temporary commit template.

webknjaz:
> I have a somewhat outdated `git` version on a somewhat outdated Ubuntu LTS

These days, I tend to recommend new pythonistas to start by using Fedora as it's Python-friendly and ships modern software, being a creator-oriented distribution.

webknjaz (commented):
Let's fix these and get it merged.

webknjaz:
@chrysle I'm going to commit the suggestions myself now.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz: Marking this as resolved.

webknjaz:
I'm going to merge after https://github.com/pypa/packaging.python.org/pull/1261#discussion_r1324736220 is cleared, in case anybody wants to leave last reviews.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -67,69 +95,116 @@ should make GitHub run this workflow:
...
 
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-test-pypi.yml
    :language: yaml
-   :start-after: on:
+   :start-after: path: dist/
    :end-before: steps:
 
+This will also ensure that the PyPI publishing workflow is only triggered
+if the current commit is tagged. It is recommended that you commit using the
+latest release tag.
```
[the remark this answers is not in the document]
chrysle: I haven't added this; see https://github.com/pypa/packaging.python.org/pull/1092#discussion_r929321386 for context. I see you have been struggling with that wording back then, too ;-). There is an unaccepted suggestion in the same discussion that could be added.

But I agree I understood it falsely at first, which is why the sentence is even more misleading now. 

I'm quite sure this means keeping the release tag of _gh-action-pypi-publish_ up-to-date (compare with the current version of the guide) . But since we have quite a lot of actions integrated now, I think we should either take this out or move it to a more obvious position.

chrysle:
> These days, I tend to recommend new pythonistas to start by using Fedora as it's Python-friendly and ships modern software, being a creator-oriented distribution.

Thanks for suggesting! I shall try to overcome my obsession with Ubuntu (if they still haven't modernised anything with soon-to-come 24.04).

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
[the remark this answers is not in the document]
webknjaz: Oh, that didn't occur to me at all. Agreed, moving and making it clear that it's about the actions that are integrated is a good idea.

chrysle:
Done in the latest commit!

webknjaz (approved):
Fantastic job, @chrysle! Thank you for getting this across the finish line!

chrysle:
Many thanks to you all, especially @webknjaz, for your thorough reviews!

PR #1266, opened 2023-06-09
Fix typos in Recording installed packages spec

silopolis:


chrysle:
> BTW, shouldn't this file be renamed, or its title changed so that both match, or maybe is it intentional?

Since package is not identical to project, I think the file (and label) should be renamed.

silopolis:
Really? I thought the title would better be changed

chrysle:
The term [python project](https://packaging.python.org/en/latest/glossary/#term-Project) reappears several times in the specification. So that seems right.

silopolis:
True, but my understanding is that what we are dealing with here are really installed packages, not projects.
Further, a project may produce several distribution packages, and what we're tracking is indeed installation of the latter.

chrysle:
Yes, this sounds reasonable. Since projects can not be installed directly, but the packages built from them, the heading should be changed.

source/specifications/recording-installed-packages.rst
```diff
@@ -246,7 +246,7 @@ To achieve this, affected tools should take the folllowing steps:
 
 Python runtime providers may also prevent inadvertent modification of platform
 provided packages by modifying the default Python package installation scheme
-to use a location other than that used by platform provided packages (while also
+to use a location other than that used by distribution packages (while also
```
pradyunsg, line 249: This isn't a typo, and should say "platform provided" here.

pradyunsg (changes requested):
Per above.

source/specifications/recording-installed-packages.rst
willingc, line 249:
```suggestion
to use a location other than that used by platform provided packages (while also
```

willingc (commented):
Thanks for the PR. I'm going to move this along by updating the PR via suggestion to put back the original wording. The PR would then be the typo fix.

willingc (approved):

PR #1273, opened 2023-07-10
WIP: "modernize" single-source versioning

ChrisBarker-NOAA:
The  single-sourcing-package-version page is pretty out of date.

I've (started to) update it to reflect the current standard of practice:

I know this doc deliberately doesn't make a recommendation, but I think putting the "preferred" methods first, and noting the reasons you might not want to use the older methods will help newbies -- and I did put in a "no longer recommended" section -- up to the team to decide if we want to word it that strongly.

What I did:

- Restructured it a bit with a few sections to group related methods

- Put the declarative forms up front (setup.cfg and pyproject.toml) 

- Moved `setuptools-vcs` from the bottom.
  NOTE: I'm not sure how strongly we want to recommend that at this point, I've been trying it out, and it's not all that mature -- in particular, it only works well with  "standard" setups -- which might be OK for anyone starting fresh, but it's a bit clunky for those of us with legacy code structures.

- Put the methoids that use built-time and run-time version finding into "no longer recommended"

- some copy editing

This reflects my personal opinion, so will need some editing / discussion, but the page really should be modernized!

### DIFF
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -66,7 +66,7 @@ URI is defined in :rfc:`std-66 <3986>`)::
     version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
     version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+
     version_one   = version_cmp version wsp*
-    version_many  = version_one (wsp* ',' version_one)*
+    version_many  = version_one (',' version_one)* (',' wsp*)?
     versionspec   = ( '(' version_many ')' ) | version_many
     urlspec       = '@' wsp* <URI_reference>
 
@@ -303,7 +303,7 @@ The complete parsley grammar::
     version_cmp   = wsp* <'<=' | '<' | '!=' | '==' | '>=' | '>' | '~=' | '==='>
     version       = wsp* <( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+>
     version_one   = version_cmp:op version:v wsp* -> (op, v)
-    version_many  = version_one:v1 (wsp* ',' version_one)*:v2 -> [v1] + v2
+    version_many  = version_one:v1 (',' version_one)*:v2 (',' wsp*)? -> [v1] + v2
     versionspec   = ('(' version_many:v ')' ->v) | version_many
     urlspec       = '@' wsp* <URI_reference>
     marker_op     = version_cmp | (wsp* 'in') | (wsp* 'not' wsp+ 'in')
@@ -424,6 +424,7 @@ A test program - if the grammar is in a string ``grammar``:
         "name",
         "name<=1",
         "name>=3",
+        "name>=3,",
         "name>=3,<2",
         "name@http://foo.com",
         "name [fred,bar] @ http://foo.com ; python_version=='2.7'",
@@ -481,6 +482,9 @@ History
   ``'.'.join(platform.python_version_tuple()[:2])``, to accommodate potential
   future versions of Python with 2-digit major and minor versions
   (e.g. 3.10). [#future_versions]_
+- June 2024: The definition of ``version_many`` was changed to allow trailing
+  commas, matching with the behavior of the Python implementation that has been
+  in use since late 2022.
 
 
 References


wimglenn:
Related: https://github.com/pypa/packaging.python.org/issues/182

flying-sheep:
We could just remove that page.

`__version__` is just a convention, `importlib.metadata.version('distname')` is short, standardized, and always works.

Maybe we should just recommend to not bother about `__version__`.

abravalheri:
> We could just remove that page.
> 
> `__version__` is just a convention, `importlib.metadata.version('distname')` is short, standardized, and always works.
> 
> Maybe we should just recommend to not bother about `__version__`.

Hi @flying-sheep, @wimglenn, I agree with the idea that `__version__` is just a convention and that at runtime if you want to know the version of something that was installed the best is to do `importlib.metadata.version('distname')`.

But I don't think the value of this page comes from the fact it allows you to write `__version__` in your Python file... For example, even if we simply rename `__version__` in the examples to `_my_UGLY_private_CONST_dont_USE_it_in_runtime`, the documentation page would hold have relevant information about the build process.

Some workflows imply in using a single source of truth for the version information. Maybe because it is creating a Python wrapper for an existing library written in a compiled language. Maybe because the devs want to use the VCS as single source of truth. Maybe there are other reasons (and here I am avoiding to judge if each one of these reasons is good or not)...

But my interpretation is that the value of this page comes from it giving the user options to access version information from different files during build time, not runtime. So I don't think removing it is the best option. Maybe the content needs adjustments to reflect that reality, but I see the value in the page existing somewhere.

We can argue that there is no canonical way of doing that across multiple backends and that therefore it should be moved away from the website into another one; I understand this point of view. But I don't think the best answer is "people should not be using `__version__`, so let's remove the page").

source/guides/single-sourcing-package-version.rst
```diff
@@ -6,17 +6,51 @@ Single-sourcing the package version
...
+        [project]
+        name = "package"
+        dynamic = ["version"]
+
+        [tool.setuptools.dynamic]
+        version = {attr = "package.__version__"}
+
+   Please be aware that declarative config indicators, including the
+   ``attr:`` directive, are not supported in parameters to
+   :file:`setup.py`.
```
abravalheri, line 42: How about we rename `__version__` to something less controversial. For example:

```suggestion
One way of single sourcing is to specify the version somewhere in the source code. Let's suppose you have a Python module (e.g. ``my_package/__init__.py``) that defines the project version as a constant (e.g. `VERSION = "0.42"`). Some build-backends allow you to use this value when building the package.

If you use `setuptools` as build-backend you can try one of the following:

#. Using the `version = attr:` in ``setup.cfg``:

   With recent versions of setuptools (since 46.4.0), one can add a declaration
   to the project's :file:`setup.cfg` file (replacing `my_package` with the import
   name of the package and `VERSION` with the name of the relevant attribute):

   .. code-block:: ini
       [metadata]
       version = attr: my_package.VERSION
       
#. Using a dynamic version in `pyproject.toml`:

   As of the release of setuptools 61.0.0, one can also specify the
   version dynamically in the project's :file:`pyproject.toml` file.

   .. code-block:: toml
        [project]
        name = "package"
        dynamic = ["version"]
        [tool.setuptools.dynamic]
        version = {attr = "package.__version__"}
        
Please be aware that declarative config indicators, including the
``attr:`` directive, are not supported in parameters to
:file:`setup.py`.
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -6,17 +6,51 @@ Single-sourcing the package version
...
+    with older versions of setuptools, you may need to add a version-reading
+    function to setup.py: Example adapted from (from `pip setup.py <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
+
+        from pathlib import Path
 
         def read(rel_path):
-            here = os.path.abspath(os.path.dirname(__file__))
-            with codecs.open(os.path.join(here, rel_path), 'r') as fp:
+            here = Path(__file__).parent.absolute()
+            with open(here / rel_path), 'r') as fp:
```
abravalheri, line 53: Not sure how much people need to rely on `setuptools<46.4.0` nowadays. Maybe we can omit this section?

abravalheri (commented):
Hi @ChrisBarker-NOAA, I am not a maintainer in this repo, but thank you very much for spending time trying to improve the docs.

I added some comments/suggestions, that you might agree or disagree :P

source/guides/single-sourcing-package-version.rst
abravalheri: Maybe we could replace this with the use case the dev needs to process a given file to extract the version string (e.g. the cases when you have a plain text file with several info or a file written in a different language?).

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +67,56 @@ number of your project:
...
+
+    With this approach you must make sure that the ``VERSION`` file is included in
+    all your source and binary distributions (e.g. add ``include VERSION`` to your
+    :file:`MANIFEST.in`).
+
+
+No longer recommend methods
+---------------------------
+
+These methods rely on importing of code during the build process, or dynamically genrated the version at run time. These methods are prone to errors and security issues, but you may encounter them in older code bases.
```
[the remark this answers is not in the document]
ChrisBarker-NOAA: yes, `importlib{.,_}metadata` is recommended, but that's not the same  "dynamically generated the version at run time" referred to here. Better wording suggestions are welcome :-)

ChrisBarker-NOAA:
> we rename `__version__` to something less controversial

Is `__version__` controversial?

Putting an attribute into the module for the version number is somewhat controversial -- but I *think* there's' some consensus that if you do do that, it should be called `__vresion__.

It is certainly the most common name used.

source/guides/single-sourcing-package-version.rst
flying-sheep, line 119: typo:

```suggestion
These methods rely on importing of code during the build process, or dynamically generated the version at run time. These methods are prone to errors and security issues, but you may encounter them in older code bases.
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -6,17 +6,51 @@ Single-sourcing the package version
 
 
 There are many techniques to maintain a single source of truth for the version
-number of your project:
+number of your projects:
```
sinoroc, line 9: I recommend using "*version string*" consistently (throughout this document, the whole `packaging.python.org` and in general).

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +67,56 @@ number of your project:
            ...
         )
 
-    .. note::
+Generate the version string from the SCM
+----------------------------------------
+
+Most projects use a Source Code Management System (SCM), such as git or mercurial to manage the code and also manage releases.
+In order to keep the versioning of releases in sync with your package, the version number can be kept in the tags of a version control system, and automatically extracted with a tool such as
```
sinoroc, line 74: version string

source/guides/single-sourcing-package-version.rst
sinoroc: Do we really want to have examples specific to a build back-end? Isn't there a way to stay agnostic? At the very least we could use tabs like in the packaging tutorial.

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +67,56 @@ number of your project:
...
-       As of the release of setuptools 61.0.0, one can specify the
-       version dynamically in the project's :file:`pyproject.toml` file.
+An external build tool can be used that manages the version in both the SCM and source code, either directly or via an API:
 
-       .. code-block:: toml
+A few tools you could use, in no particular order, and not necessarily complete:
+`bump2version <https://pypi.org/project/bump2version>`_,
+`changes <https://pypi.org/project/changes>`_,
+`commitizen <https://pypi.org/project/commitizen>`_,
+`versioneer <https://github.com/python-versioneer/python-versioneer>_`
```
sinoroc, line 90: Typo. The underscore should be after the backtick.

sinoroc (commented):
I refer to my [comment](https://github.com/pypa/packaging.python.org/issues/182#issuecomment-1646659063). In short, I wonder if we wouldn't be better off just mentioning `importlib.metadata.version()`, how to use it, and saying that 3rd party tools (build back-end plugins) exist without naming them explicitly. I feel like such a document as it is now as well as this new modernized version belong outside of `packaging.python.org`.

ChrisBarker-NOAA (commented):
I've addressed a number of the comments in an update to the PR.

ChrisBarker-NOAA:
I've addressed most of the comments. As I noted in #1276, I think it's important to update this page SOON -- even if it will change again in the not-too distant future. What's published now is really, really, out of date.

source/guides/single-sourcing-package-version.rst
webknjaz: > wouldn't it be nice if there were a standard way to specify how to set yourversion other than hard coding it in a toml file?

`attr:` is setuptools-specific, but `dynamic = ["version"]` allows all the PEP 517 build backends to generate the version during the build time. How this is done and how it's configured, it's always going to be specific to each build backend.
PEP 517 describes the communication interface between the parties but it doesn't mandate which backend-specific features should be implemented since it's out of the scope. What's the point of a standard that stays away from the implementation details if it'd demand each backend to implement the same mechanism? If that was the case, we wouldn't have the multitude of backends — just one.
Also, imagine a project being packaged is a wrapper around a Rust crate. And the version is in that crate. A packager uses a build backend specifically targeting Rust projects. If a backend is asked to implement loading the version from Python importables, it becomes impossible since that contradicts its design.

So the `dynamic` thing is really the extent of the standard that cares about packaging interoperability and doesn't have leaky abstractions.

source/guides/single-sourcing-package-version.rst
webknjaz: > Not sure how much people need to rely on `setuptools<46.4.0` nowadays. Maybe we can omit this section?

Pip-users? Not many. This'd mostly hit downstream packagers that might be stuck with old pinned packages in their distros.

source/guides/single-sourcing-package-version.rst
webknjaz: > > How about we rename **version** to something less controversial.
> 
> `__version__` may be controversial, but it's also a very widely used convension -- seems a bit silly to not use the most common convention as the example.

I agree — to me, anything else would be controversial. I'd keep it as it is. But we can add a `.. tip::` admonition stating that "`__version__` is widely used and is sometimes considered controversial".

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,18 +5,51 @@ Single-sourcing the package version
...
-    <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
 
-        import codecs
-        import os.path
+Auto extract the version string from the source
+-----------------------------------------------
+
+One way of single sourcing is to specify the version string somewhere in the source code, e.g. ``my_package/__init__.py``. Then it can be found at runtime with ``my_package.__version__``, and that same value can be used to set the version when building the package.
+
+#. Declare to read the version string  from the source in ``setuptools.cfg`` or ``pyproject.toml``
```
webknjaz, line 16: Typo
```suggestion
#. Declare to read the version string  from the source in :file:`setup.cfg` or `pyproject.toml`.
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,18 +5,51 @@ Single-sourcing the package version
...
+        version = {attr = "package.__version__"}
+
+   Please be aware that declarative config indicators, including the
+   ``attr:`` directive, are not supported in parameters to
+   :file:`setup.py`.
+
+
+#.  While the modern standard is to use declarative files (such as ``setup.cfg``),
+    with older versions of setuptools, you may need to add a version-reading
+    function to setup.py: Example adapted from (from `pip setup.py <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
```
webknjaz, line 46: Let's pin the URL since it may become 404 if pip ever removes this file from its main branch.
```suggestion
    function to :file`setup.py`: Example adapted from (from `pip setup.py <https://github.com/pypa/pip/blob/6f3a7181b6bfa0e95d479b3269baa0e551ff2cb2/setup.py#L11>`_)::
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,18 +5,51 @@ Single-sourcing the package version
...
+
+        from pathlib import Path
 
         def read(rel_path):
-            here = os.path.abspath(os.path.dirname(__file__))
-            with codecs.open(os.path.join(here, rel_path), 'r') as fp:
+            here = Path(__file__).parent.absolute()
+            with open(here / rel_path), 'r') as fp:
```
webknjaz, line 52: Since this is already injecting pathlib, just use the high-level API instead:
```suggestion
            return (here / rel_path).read_text(encoding='utf-8')
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
            ...
         )
 
-    .. note::
+Generate the version string from the SCM
+----------------------------------------
+
+Most projects use a Source Code Management System (SCM), such as git or mercurial to manage the code and also manage releases.
+In order to keep the versioning of releases in sync with your package, the version string can be kept in the tags of a version control system, and automatically extracted with a tool such as
```
webknjaz, line 73:
```suggestion
In order to keep the versioning of releases in sync with your package, the version string can be kept in the tags of a version control system, and automatically extracted during the build, with a tool such as
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
+`setuptools_scm <https://pypi.org/project/setuptools-scm/>`_.
 
-       As of the release of setuptools 46.4.0, one can accomplish the same
-       thing by instead placing the following in the project's
-       :file:`setup.cfg` file (replacing "package" with the import name of the
-       package):
+.. NOTE: maybe put in an example using setuptools_scm?
 
-       .. code-block:: ini
+.. NOTE2: Is setuptools_scm the only one now?
```
webknjaz, line 78: For setuptools, yes. Other backends have their own plugins. Like hatch-vcs (which actually wraps setuptools-scm). There's also dunamai but it's a generic project, that doesn't provide direct packaging integrations.

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
-           version = attr: package.__version__
+Use an external version management tool
+---------------------------------------
 
-       As of the release of setuptools 61.0.0, one can specify the
-       version dynamically in the project's :file:`pyproject.toml` file.
+An external build tool can be used that manages the version string in both the SCM and source code, either directly or via an API:
 
-       .. code-block:: toml
+A few tools you could use, in no particular order, and not necessarily complete:
```
webknjaz, line 85: Twisted has their versioning own thing too. And there's also `pbr` that I don't regard well, in comes from the OpenStack ecosystem.

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
+.. NOTE2: Is setuptools_scm the only one now?
 
-           [metadata]
-           version = attr: package.__version__
+Use an external version management tool
+---------------------------------------
 
-       As of the release of setuptools 61.0.0, one can specify the
-       version dynamically in the project's :file:`pyproject.toml` file.
+An external build tool can be used that manages the version string in both the SCM and source code, either directly or via an API:
```
webknjaz, line 83:
```suggestion
An external build tool can be used that manages the version string in both the SCM and source code, either directly or via an API.
```

source/guides/single-sourcing-package-version.rst
wimglenn: The plan from setuptools side is to [eventually deprecate `setup.cfg`](https://github.com/pypa/setuptools/issues/3214), so it should not be mentioned. There's also spurious spacing here between "string  from".

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
-            [tool.setuptools.dynamic]
-            version = {attr = "package.__version__"}
+Dedicated file for the version string
+-------------------------------------
 
-       Please be aware that declarative config indicators, including the
-       ``attr:`` directive, are not supported in parameters to
-       :file:`setup.py`.
+Place the value in a simple ``VERSION`` text file and have both
+:file:`setup.py` and the project code read it.
```
webknjaz, line 98: I wonder if `version = file: VERSION` would work. Not sure if this helper works for the `version` attribute... @abravalheri do you know?

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
-    offers an API that both locations can use.
+::
 
-    Few tools you could use, in no particular order, and not necessarily complete:
-    `bump2version <https://pypi.org/project/bump2version>`_,
-    `changes <https://pypi.org/project/changes>`_,
-    `commitizen <https://pypi.org/project/commitizen>`_,
-    `zest.releaser <https://pypi.org/project/zest.releaser>`_.
+    with open(os.path.join(mypackage_root_dir, 'VERSION')) as version_file:
+        version = version_file.read().strip()
```
webknjaz, line 103: The above example already uses pathlib for consistency, this should too:
```suggestion
    version = (mypackage_root_dir / 'VERSION').read_text(encoding='utf-8').strip()
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
 
-            [tool.setuptools.dynamic]
-            version = {attr = "package.__version__"}
+Dedicated file for the version string
+-------------------------------------
 
-       Please be aware that declarative config indicators, including the
-       ``attr:`` directive, are not supported in parameters to
-       :file:`setup.py`.
+Place the value in a simple ``VERSION`` text file and have both
```
webknjaz, line 97:
```suggestion
Place the value in a simple :file:`VERSION` text file and have both
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
-    `zest.releaser <https://pypi.org/project/zest.releaser>`_.
+    with open(os.path.join(mypackage_root_dir, 'VERSION')) as version_file:
+        version = version_file.read().strip()
 
+An advantage with this technique is that it's not specific to Python.  Any
+tool can read the version string.
+
+.. warning::
+
+    With this approach you must make sure that the ``VERSION`` file is included in
```
webknjaz, line 110:
```suggestion
    With this approach you must make sure that the :file:`VERSION` file is included in
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,18 +5,51 @@ Single-sourcing the package version
...
+#.  While the modern standard is to use declarative files (such as ``setup.cfg``),
+    with older versions of setuptools, you may need to add a version-reading
+    function to setup.py: Example adapted from (from `pip setup.py <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
+
+        from pathlib import Path
 
         def read(rel_path):
-            here = os.path.abspath(os.path.dirname(__file__))
-            with codecs.open(os.path.join(here, rel_path), 'r') as fp:
+            here = Path(__file__).parent.absolute()
```
wimglenn, line 51: Why calling absolute method? `__file__` is an absolute path already since [Python 3.4](https://docs.python.org/3.4/whatsnew/3.4.html#other-language-changes).

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
+
+    With this approach you must make sure that the ``VERSION`` file is included in
+    all your source and binary distributions (e.g. add ``include VERSION`` to your
+    :file:`MANIFEST.in`).
+
+
+No longer recommend methods
+---------------------------
+
+These methods rely on importing of code during the build process, or dynamically generating the version at build time. These methods are prone to errors and security issues, but you may encounter them in older code bases.
```
webknjaz, line 118:
```suggestion
These methods rely on importing of code during the build process and are prone to errors and security issues, but you may encounter them in older code bases.
```

I don't agree with this part of the sentence — it's perfectly fine to generate version during the build. `setuptools-scm` does just that — it makes a version by combining multiple sources — the last Git tag, distance to the last Git tag, current commit, current date, Git dirty flag, `.git_archival.txt` (for git archives and sdists).

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
            ...
         )
 
-    .. note::
+Generate the version string from the SCM
+----------------------------------------
+
+Most projects use a Source Code Management System (SCM), such as git or mercurial to manage the code and also manage releases.
```
wimglenn, line 72: VCS is a more familiar and less overloaded acronym than SCM

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
+tool can read the version string.
+
+.. warning::
+
+    With this approach you must make sure that the ``VERSION`` file is included in
+    all your source and binary distributions (e.g. add ``include VERSION`` to your
+    :file:`MANIFEST.in`).
+
+
+No longer recommend methods
```
wimglenn, line 115: Poor english.

source/guides/single-sourcing-package-version.rst
wimglenn, line 103: This method is just awful.  It totally breaks [zipapp-packaging](https://docs.python.org/3/library/zipapp.html#creating-standalone-applications-with-zipapp). Recommend removal.

source/guides/single-sourcing-package-version.rst
```diff
@@ -84,25 +130,9 @@ number of your project:
...
-
-    .. warning::
-
-        With this approach you must make sure that the ``VERSION`` file is included in
-        all your source and binary distributions (e.g. add ``include VERSION`` to your
-        :file:`MANIFEST.in`).
 
 #.  Set the value in :file:`setup.py`, and have the project code use the
-    ``importlib.metadata`` API to fetch the value at runtime.
+    ``importlib.metadata`` API to fetch the value at build time.
```
webknjaz, line 135: No, this change is incorrect. The original variant is true.

`importlib.metadata` works with an already installed dist. At build time, it's not yet installed and there's no metadata to read, vice versa — that metadata is being created so it can't be also the source. It mustn't be used during the build but in runtime (for example, via `__version__ = importlib.metadata.version(__package__)`. The original source in the described case is `setup.py`.

```suggestion
    ``importlib.metadata`` API to fetch the value at runtime.
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -120,7 +150,7 @@ number of your project:
     installation metadata, which is not necessarily the code that's currently
     imported.
 
-    If a project uses this method to fetch its version at runtime, then its
+    If a project uses this method to fetch its version string at build time, then its
```
webknjaz, line 153: Same here.
```suggestion
    If a project uses this method to fetch its version string at runtime, then its
```

source/guides/single-sourcing-package-version.rst
```diff
@@ -161,12 +191,7 @@ number of your project:
 
     .. warning::
 
-        Although this technique is common, beware that it will fail if
+        Although this technique was common, beware that it will fail if
```
webknjaz, line 194:
```suggestion
        Although this technique used to be common, beware that it will fail if
```

source/guides/single-sourcing-package-version.rst
wimglenn: This assumes your package is extracted to a filesystem. That's not a safe assumption.

webknjaz: Fair point. But then, `open()` still wouldn't work, right? This means the need for something like `importlib.resources.files()`.

Though, I think that it's pretty much guaranteed to be on the file system at build time. The runtime, OTOH, might need the importlib helpers.

source/guides/single-sourcing-package-version.rst
wimglenn: It doesn't work, you must specify in the dynamic section if you want to use that feature.

```
[tool.setuptools.dynamic]
version = {file = "VERSION"}
```

source/guides/single-sourcing-package-version.rst
webknjaz: Many confuse `absolute()` and `resolve()`: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/2.
We should use the latter:
```suggestion
            here = Path(__file__).parent.resolve()
```

@wimglenn to answer your question, `__file__` is not what's being unwound on this line — the parent directory is. Also, absolute and resolved are different concepts. Absolute just makes the path start with `/` so it's not relative. The resolution process follows symlinks, revealing actual location in the file system.

That said, I don't know if anything would change, realistically, if we were to skip the resolution...

source/guides/single-sourcing-package-version.rst
webknjaz: Honestly, I feel like it's vice versa. So this must be a subjective feeling.

source/guides/single-sourcing-package-version.rst
webknjaz: How about
```suggestion
Nearly extinct methods
```

source/guides/single-sourcing-package-version.rst
webknjaz: https://github.com/pypa/packaging.python.org/pull/1273#discussion_r1317971494

source/guides/single-sourcing-package-version.rst
wimglenn: Can you show me an example where resolving the parent of `Path(__file__)` is necessary?

source/guides/single-sourcing-package-version.rst
webknjaz: > It doesn't work, you must specify in the dynamic section if you want to use that feature.

Oh, I'm talking about `setup.cfg`, not `pyproject.toml`. It is processed differently.

source/guides/single-sourcing-package-version.rst
wimglenn: Interesting, what is VCS overloaded for?  SCM is overloaded to [config management](https://en.wikipedia.org/wiki/Software_configuration_management).

webknjaz: I meant that it's more familiar.

webknjaz: Anyway, the overloading is comparable: https://en.wikipedia.org/wiki/VCS#Technology vs. https://en.wikipedia.org/wiki/SCM#Computing.

source/guides/single-sourcing-package-version.rst
webknjaz: I would imagine that probably not in the context of Python packaging. In general, it's probably useful when paths are processed by different programs with different access privileges.

source/guides/single-sourcing-package-version.rst
wimglenn: `importlib.resources.files()` is good, but then you need the `importlib_resources` port for supporting Python < 3.9 (conditional dependency), or a compatibility fallback to stdlib [pkgutil](https://docs.python.org/3/library/pkgutil.html). To do it properly/correctly with a static file in the package it gets complicated pretty fast.

source/guides/single-sourcing-package-version.rst
wimglenn: It works in `setup.cfg`, but as previously commented the plan on setuptools side is deprecation of that file.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I think this is a overall question -- this is already under "no longer recommended" -- I left it (and others) in because I thought it would be less controversial change (or less radical anyway) -- but I'm happy to simply remove the really old and crufty ones :-)

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: That makes the point :-)

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I suppose VCS/SCM would work -- I used SCM because we later reference setuptools_scm.

source/guides/single-sourcing-package-version.rst
wimglenn: Yes, they should be removed. The examples which import `pkg_resources` are cause deprecation warnings already.

source/guides/single-sourcing-package-version.rst
wimglenn: But you also called it a version control system on line 73 :)

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +66,56 @@ number of your project:
...
+---------------------------------------
 
-       As of the release of setuptools 61.0.0, one can specify the
-       version dynamically in the project's :file:`pyproject.toml` file.
+An external build tool can be used that manages the version string in both the SCM and source code, either directly or via an API:
 
-       .. code-block:: toml
+A few tools you could use, in no particular order, and not necessarily complete:
+`bump2version <https://pypi.org/project/bump2version>`_,
+`changes <https://pypi.org/project/changes>`_,
```
wimglenn, line 87: This project seems to be dead (no recent git commits, no release on PyPI for almost 10 years, people complaining on the tracker that it doesn't work since Python 3.7)

ChrisBarker-NOAA: Clearly this list needs a refresh -- anyone up to date on what's being currently maintained? In the meantime, i've removed that ones that aren't being actively maintained.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: fixed ...

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I'm not sure exactly what part you don't agree with -- but the idea is that running *custom* code at build time is not great -- which is different than using a declaration that invokes another tool that runs -- after all, *something* has to run to build a package :-) If that's not clear, then please suggest some more clear phrasing -- I'm at a loss.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I've changed it to `.resolve` -- why not?

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I've rephrased it bit -- maybe better?

source/guides/single-sourcing-package-version.rst
wimglenn: Why not: because it should be no more complicated than necessary. I'm not convinced resolving the path is ever necessary.

ChrisBarker-NOAA:
I *think* I've addressed all the comments, one way or another.There's still a bit of philosophical question about to what extent this should document the current state of the practice, vs what is hopefully the future.

I suppose that part of resolving the question is the intended audience -- is this only for people starting from scratch with a new project? Or is it also for people maintaining older projects?

And another somewhat philisophical question -- packaging has moved in the direction of robust, best practices solutions that can handle all (most) cases. But this gets away from the KISS approach -- if someone is making a small package for internal use -- it's nice to have a simple way to simply solve the problem at hand, and I'd like to see that documented.

I'm not in a position to make these judgement calls.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+   With older versions of setuptools the declaration can be added to
+   the project's :file:`setup.cfg` file (replacing "package" with the import
+   name of the package):
+
+   .. code-block:: ini
+
+       [metadata]
+       version = attr: package.__version__
+
+    Note that the use of :file:`setup.cfg` is being depricated by setuptools -- new projects should use :file:`pyproject.toml`.
```
chrysle, line 43: This line is somehow swallowed by the INI code block.

source/guides/single-sourcing-package-version.rst
sinoroc: > The plan from setuptools side is to [eventually deprecate `setup.cfg`](https://github.com/pypa/setuptools/issues/3214),

Oh? Really? That's sad. I do not want to have a `pyproject.toml` file that is kilometers long. I like having one configuration file per tool. It is easier to find things. I just wished there was maybe a convention like `pyproject.d/setuptools.toml`.

sinoroc:
I like that there is a healthy discussion.

Maybe this should be merged as a large overhaul of the whole page. Then work can begin to refine specific points.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
-        import codecs
-        import os.path
+Auto extract the version string from the source
+-----------------------------------------------
+
+One way of single sourcing is to specify the version string somewhere in the source code, e.g. ``my_package/__init__.py``. Then it can be found at runtime with ``my_package.__version__``, and that same value can be used to set the version when building the package.
+
+#. Declare to read the version string from the source in ``pyproject.toml`` or ``setuptools.cfg``
+
+   As of the release of setuptools 61.0.0, one can also specify the
```
sinoroc, line 18: I would remove the "also" here.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+   version dynamically in the project's :file:`pyproject.toml` file.
+
+   .. code-block:: toml
+
+        [project]
+        name = "package"
+        dynamic = ["version"]
+
+        [tool.setuptools.dynamic]
+        version = {attr = "package.__version__"}
```
sinoroc, line 28: Should be `my_package` instead of `package` since that is the example used earlier in the page.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+
+   Please be aware that declarative config indicators, including the
+   ``attr:`` directive, are not supported in parameters to
+   :file:`setup.py`.
+
+   With older versions of setuptools the declaration can be added to
```
sinoroc, line 34: Why the "with older versions of setuptools"? Current versions support this perfectly well as far as I know.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+   the project's :file:`setup.cfg` file (replacing "package" with the import
```
sinoroc, line 35: `my_package`

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+   name of the package):
+
+   .. code-block:: ini
+
+       [metadata]
+       version = attr: package.__version__
```
sinoroc, line 41: `my_package`

source/guides/single-sourcing-package-version.rst
sinoroc: Reference to the deprecation?

Also typo: *deprecated* instead of *depricated*.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,51 @@ Single-sourcing the package version
...
+
+    Note that the use of :file:`setup.cfg` is being depricated by setuptools -- new projects should use :file:`pyproject.toml`.
+
+#.  While the modern standard is to use declarative files (such as ``pyproject.toml``),
+    with older versions of setuptools, you may need to add a version-reading
```
sinoroc, line 46: *setuptools* will support `setup.py` forever. Maybe `setup.cfg` will disappear (hopefully not), but I am quite sure `setup.py` will stay for a very very long time.

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +65,57 @@ number of your project:
...
-       package):
+Use an external version management tool
+---------------------------------------
 
-       .. code-block:: ini
+An external build tool can be used that manages the version string in both the SCM and source code, either directly or via an API.
 
-           [metadata]
-           version = attr: package.__version__
+A few tools you could use, in no particular order, and not necessarily complete:
```
sinoroc, line 85: the order is alphabetical, fair enough :)

and of course the list is not complete since we said "a few tools".

So I guess I would just leave it at "For example:"

source/guides/single-sourcing-package-version.rst
```diff
@@ -120,7 +152,7 @@ number of your project:
     installation metadata, which is not necessarily the code that's currently
     imported.
 
-    If a project uses this method to fetch its version at runtime, then its
+    If a project uses this method to fetch its version string at runtime, then its
     ``install_requires`` value needs to be edited to install
     ``importlib-metadata`` on pre-3.8 versions of Python like so::
```
sinoroc, line 157: I think we can leave the detail of `importlib-metadata` pre Python 3.8 out of the guide. People can figure this out on their own. And Python 3.8 is almost EOL, right?

source/guides/single-sourcing-package-version.rst
chrysle: > Maybe it would be good to have a convention like `pyproject.d/setuptools.toml`.

That sounds like a great standard to have!

However it'd require a huge push to create and support it by the various packaging tools. Could you make a proposal over at the `peps` repository?

source/guides/single-sourcing-package-version.rst
chrysle: Well, still one year to go. Maybe we could add a todo comment here.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: thanks -- fixed

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: The idea was that you should use pyproject.toml with new setuptools -- but yes, that is confusing -- and there is a lot of setup.cfg out there still. I"ll rephrase.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: well, yes the the declarative forms are still preferred -- and you can use setup.cfg or pyproject.toml alongside setup.py, yes?

ChrisBarker-NOAA: > Reference to the deprecation?

Someone said that in this PR -- if that's not a done deal, we should remove it.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: I borrowed that text from the original -- but I've simplified it -- but leaving the "not complete" in case anyone gets upset their favoratie tools wasn't included.

source/guides/single-sourcing-package-version.rst
ChrisBarker-NOAA: And I think this is going to have to get updated sooner than the last update cycle anyway ...

comptechNvidia (approved):

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,49 @@ Single-sourcing the package version
 ===================================
```
comptechNvidia, line 5:
```suggestion
===================================
```

source/guides/single-sourcing-package-version.rst
webknjaz: > _setuptools_ will support `setup.py` forever. Maybe `setup.cfg` will disappear (hopefully not), but I am quite sure `setup.py` will stay for a very very long time.

Agreed. It'd be unfortunate and these were deprecated fully. There are cases when one would want to keep metadata in a dedicated file. In case of mass-maintenance of many projects, for example.

pradyunsg:
I've moved this to a draft status given the WIP in the title and merge conflicts, to avoid accident merges before the author or a maintainer is able to address those aspects.

willingc:
@ChrisBarker-NOAA Let's see if we can unblock this PR since the proposed content is an improvement over what exists. Next step would be to remove the conflict and take this out of WIP. For modernizing the guides, I think we should:
- merge if significantly better than existing dated doc
- make a note in the text if there is something that still requires further discussion by maintainers
- make a note, if needed, at the top of the file explaining to the user that this has been modernized but is still under revision.

source/guides/single-sourcing-package-version.rst
```diff
@@ -82,24 +125,10 @@ number of your project:
             exec(fp.read(), version)
         # later on we use: version['__version__']
 
-    Example using this technique: `warehouse <https://github.com/pypa/warehouse/blob/64ca42e42d5613c8339b3ec5e1cb7765c6b23083/warehouse/__about__.py>`_.
+.. NOTE: this example is from a (very!) old version of warehouse -- at a glance, I couldn't tell how warehouse does it now, but it's not this.
 
-#.  Place the value in a simple ``VERSION`` text file and have both
-    :file:`setup.py` and the project code read it.
+..    Example using this technique: `warehouse <https://github.com/pypa/warehouse/blob/64ca42e42d5613c8339b3ec5e1cb7765c6b23083/warehouse/__about__.py>`_.
```
webknjaz, line 130: Why comment this out?

source/guides/single-sourcing-package-version.rst
webknjaz: Having reflected on this for a bit, I don't think this is right at all. Running custom code is fine, for as long as it's invoked through the standard interface. It's even documented in the official setuptools docs: https://setuptools.pypa.io/en/latest/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks.
Besides, it's quite mandatory to add some project-specific build code for building C-extensions. `setuptools` kinda _expects_ that `setup.py` would be used for this. But personally, I've experimented with wrappers for it on the PEP 517 level and that went well.

Hatchling also has hooks for running Python code: https://hatch.pypa.io/1.1/plugins/build-hook/#custom

source/guides/single-sourcing-package-version.rst
```diff
@@ -33,43 +63,56 @@ number of your project:
...
+
+    With this approach you must make sure that the ``VERSION`` file is included in
+    all your source and binary distributions (e.g. add ``include VERSION`` to your
+    :file:`MANIFEST.in`).
+
+
+Nearly extinct methods
+----------------------
+
+These methods rely on importing of code during the build process, or dynamically generating the version string at build time with custom code. These methods are prone to errors and security issues, but you may encounter them in older code bases.
```
webknjaz, line 115: Per https://github.com/pypa/packaging.python.org/pull/1273#discussion_r1407774382, I don't think that there's inherent security issue implications here. This statement is not universally true.

We're talking about the build scripts, and they run dynamic stuff by definition. Even when it's abstracted away into external libs. A more realistic attack surface is the supply chain — poisoning the deps is much more effective and simpler, then checking in things into a project since they go through code reviews much more often, then the deps updates. This is why, in the articles related to publishing from GHA to PyPI, for example, we always emphasize that building must happen in a separate CI job that wouldn't be able to elevate privileges and publish packages using legitimate credentials.

I think, this paragraph should instead focus on the UX/DX/maintainability/elegance aspects of the approach.

source/guides/single-sourcing-package-version.rst
```diff
@@ -5,19 +5,49 @@ Single-sourcing the package version
...
+
+   The declaration can alternatively be added to the project's :file:`setup.cfg` file (replacing "my_package" with the import name of the package):
+
+   .. code-block:: ini
+
+       [metadata]
+       version = attr: my_package.__version__
+
+
+   Note that the use of :file:`setup.cfg` is being deprecated by setuptools -- new projects should use :file:`pyproject.toml`.
```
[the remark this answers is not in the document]
wimglenn: Where was that established? As far as I know the plan from setuptools side is still to "eventually deprecate" setup.cfg, because it does not make sense to have two different declarative formats for the same purpose.

Ref: https://github.com/pypa/setuptools/issues/3214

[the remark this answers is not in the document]
willingc: Perhaps a good way to handle this is to put a note block in here and avoid deprecation comment here since it is not necessary for the reader.

```
In general, :file:`pyproject.toml` should be used for new projects.
```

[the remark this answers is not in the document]
flying-sheep: I don’t understand your proposed sentence. “stick with” means to keep using an old thing, no? `pyproject.toml` is the new thing, so you can’t “stick with” it.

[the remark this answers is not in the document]
webknjaz: You're right. Perhaps, I had a slightly different context in my head, thinking of a situation when there's a `pyproject.toml`, I want to separate it into a dedicated config (which would have to be `setup.cfg`) and somebody would want to keep using `pyproject.toml`.

Let's try rephrasing:
```suggestion
   Note that many people prefer :file:`pyproject.toml` over :file:`setup.cfg`, while some still choose to keep using the latter regardless.
```

jeanas:
> @ChrisBarker-NOAA Let's see if we can unblock this PR since the proposed content is an improvement over what exists. Next step would be to remove the conflict and take this out of WIP. For modernizing the guides, I think we should:
> 
>     * merge if significantly better than existing dated doc
> 
>     * make a note in the text if there is something that still requires further discussion by maintainers
> 
>     * make a note, if needed, at the top of the file explaining to the user that this has been modernized but is still under revision.

+1 to @willingc's comment. While the concerns expressed around, e.g., to what extent `setup.cfg` is obsolete, are valid, the current content of the page is very out of date, so let's try to get it updated without blocking ourselves on larger questions.

@ChrisBarker-NOAA Do you still intend to work on this PR? If not, I'd be willing to submit a redux PR (possibly more focused).

ChrisBarker-NOAA:
Sorry I dropped the ball on this -- I"M more that happy for anyone else to pick it up -- and it's been another year, so maybe can be a bit simpler:

My advice: keep to documenting the current state of practice, and don't worry about focusing on the "best" way.

Most important is to not longer be documenting clearly out of data practices -- e.g. custom code to read the version string in a setup.py file.

ChrisBarker-NOAA:
I've gotten a bit confused about the state of this -- there is this old PR of mine, and this one: #1276

And a recent discussion on Discourse.

I think we **do** have consensus that keeping old, outdated docs published is a bad idea.

That PR, concludes with:

"I've considered all the comments here but still believe removing this page is the best way forward."

My thought:

Given that this page has been around a long time, simply removing it will create dead links, etc, so
I think slightly better though is to replace this page with one that essentially says something along the lines of:

"""
It is best practice to specify the version number of your distribution in a single place, rather than trying to keep it in sync in more than one place.

Consult your build system's docs for how best to do that.
"""

And *maybe* a set of link to the most common build systems.

But we should do *something* in any case.

PR #1274, opened 2023-07-12
A better description for "Requires-Python" metadata field

wimglenn:
The common usage of this field is for [Dropping support for older Python versions](https://packaging.python.org/en/latest/guides/dropping-older-python-versions/). Adding upper bounds can be problematic.

The current description is misleading, and that was discussed extensively [here](https://discuss.python.org/t/requires-python-upper-limits/12663/1), but unfortunately no PR came out of it.  So here's an attempt!

I won't reproduce all of Henry's excellent arguments, but I will reiterate that the wording "_is guaranteed to be compatible with_" causes issues - today I've witnessed an environment failing to bootstrap when updated from Python 3.11.3 to Python 3.11.4 because one (internal) dependency had declared an upper bound on 3.11.3.

This was added because 3.11.3 was the runtime the developer's code was tested in, and if you were to take the current description literally, it's quite right that their distribution is not "guaranteed to be compatible with" 3.11.4 - how could you guarantee compatibility with an unreleased Python version? For someone unfamiliar with Python packaging, it is not obvious that such an upper bound would wreak havoc on the tooling and block an upgrade until a new release is published without over-constraining the Requires-Python metadata.


source/specifications/core-metadata.rst
```diff
@@ -545,18 +545,17 @@ Requires-Python
...
 picking which version of a project to install.
 
 The value must be in the format specified in :doc:`version-specifiers`.
 
-This field cannot be followed by an environment marker.
+For example, if a distribution uses `f-strings <https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals>`_
+then it may prevent installation on Python < 3.6 by specifying:
 
-Examples::
+    Requires-Python: >=3.6
```
ofek, line 556: Oldest non-EOL

```suggestion
    Requires-Python: >=3.8
```

ofek (approved):

source/specifications/core-metadata.rst
```diff
@@ -551,18 +551,17 @@ Requires-Python
 .. versionadded:: 1.2
 
 This field specifies the Python version(s) that the distribution is
-guaranteed to be compatible with. Installation tools may look at this when
+compatible with. Installation tools may look at this when
```
sinoroc, line 554: The way I think about this field (and the dependency version constraints) is something along the lines of "let's exclude the versions that we know we are not compatible with at the time of writing". But I do not know how to phrase this properly.

wimglenn: Agreed - the "exclude incompatible" use case is elaborated somewhat in the (new) example offered below.

webknjaz: How about something along the lines of "A typical use case for this field is allowing the packagers to declare Python versions that are known to be incompatible with the distribution, which lets the installers disregard such distribution packages during the dependency resolution under incompatible runtimes".

source/specifications/core-metadata.rst
```diff
@@ -551,18 +551,17 @@ Requires-Python
...
 picking which version of a project to install.
 
 The value must be in the format specified in :doc:`version-specifiers`.
 
-This field cannot be followed by an environment marker.
+For example, if a distribution uses `f-strings <https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals>`_
```
webknjaz, line 559: Please, use intersphinx-provided references to link other Sphinx sites:
```suggestion
For example, if a distribution uses :ref:`f-strings <whatsnew36-pep498>`
```

source/specifications/core-metadata.rst
```diff
@@ -551,18 +551,17 @@ Requires-Python
...
+then it may prevent installation on Python < 3.6 by specifying:
```
webknjaz, line 560: Code blocks w/o an explicit directive require trailing double colon:
```suggestion
then it may prevent installation on Python < 3.6 by specifying::
```

source/specifications/core-metadata.rst
```diff
@@ -551,18 +551,17 @@ Requires-Python
...
 
-Examples::
+    Requires-Python: >=3.6
 
-    Requires-Python: >=3
-    Requires-Python: >2.6,!=3.0.*,!=3.1.*
-    Requires-Python: ~=2.6
```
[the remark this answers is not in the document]
wimglenn: Disagree on these. Excluding versions from range and compatible release are well covered in PEP 440, which is referenced several times in this document already.

Neither of these are typical use-cases for Requires-Python, but more edge cases.  `Requires-Python >2.6,!=3.0.*,!=3.1.*` was useful when projects still needed to support 2.7 + recent 3.x, but those Python versions are EOL now.

The compatible release ~= includes an upper bound, and should not normally be used with Requires-Python.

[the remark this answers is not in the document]
webknjaz: I disagree. By removing such example, we'd make them tribal knowledge. The PR description implies improvements and not getting rid of examples that make possibilities clear.

source/specifications/core-metadata.rst
```diff
@@ -551,18 +551,17 @@ Requires-Python
...
+For example, if a distribution uses :ref:`f-strings <whatsnew36-pep498>`
+then it may prevent installation on Python < 3.6 by specifying::
 
-Examples::
+    Requires-Python: >=3.6
 
-    Requires-Python: >=3
-    Requires-Python: >2.6,!=3.0.*,!=3.1.*
-    Requires-Python: ~=2.6
+This field cannot be followed by an environment marker.
```
seberg, line 565: I will go two steps further now and ask if you can't just add more.  To be possibly heretic, something in the direction of:
```
.. note::
    At this time, the PyPA strongly suggests to never add upper version limits on this field, at least for libraries.
    For most packages an upper version pin is not necessary since even a Python 4.0 relase is
    not likely to break them.
    Even for libraries where an upper version pin is clearly correct, we strongly discourage adding the pin because:
    * It is often not useful in practice because current tooling means it is typically
      only useful if old versions support a wider range of Python versions.
    * Some tools like poetry and PDM force the bounds of dependencies on the package.
      That may be correct for lockfiles but is incorrect for this field for libraries:
      a library will not stop working with a new Python version because it has a
      dependency that does not *yet* work with a new Python version.
      Even for lockfiles the existing ecosystem tooling unfortunately may also lead
      to the first issue creating nuisance for users.

    Unfortunately, these issues need to be addressed in the ecosystem and by improving the
    tools before we can recommend upper version constraints. 
```
(even this is probably nonsense in the detail, and I don't know how much poetry/PDM are to blame exactly, although I doubt they are blameless.)

source/specifications/core-metadata.rst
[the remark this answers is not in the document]
wimglenn: Simplicity and conciseness is the goal, not detailing every possibility. Further, I replaced with a single example for a good reason - offering multiple examples could suggest that this field may be specified multiple times, which it may not. If you look on the [existing page](https://packaging.python.org/en/latest/specifications/core-metadata/), all the single use fields offer a single example. Apart from **Requires-Python**, only the fields marked as "(multiple use)" display multiple examples within the same code block.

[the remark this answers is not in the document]
pradyunsg: @webknjaz I'm gonna move forward with this PR for now, since it is an overall improvement over status quo -- if you feel strongly about this, let's do a follow-up issue/PR to add a Python 2.7+ & 3.6+ example here.

pradyunsg (approved):

rgommers (approved):

PR #1279, opened 2023-08-01
Fix broken links

### LINKED ISSUE #1272: Sphinx linkcheck fails for GitHub link with anchor in CI
**Problem**

Sphinx linkcheck fails for https://github.com/pypa/pip/blob/main/setup.py#L11 in [CI](https://github.com/pypa/packaging.python.org/actions/runs/5485668000/jobs/10005322198?pr=1267).

image

This is likely due to Github making the file viewer require Javascript, which breaks linkcheck's anchor checks.

**Approaches**

From specific to general:
1. Remove this link or remove the anchor portion of this link.  This might degrade clarity of documentation and future use of github links with anchor would break CI.
2. Ignore linkcheck for this single link.  However, if another github link with anchor is added to documentation, CI would break again.
3. Disable linkcheck for github links with anchor.  A regex can be used to detect github links with anchor.  See also feature request to Sphinx: https://github.com/sphinx-doc/sphinx/issues/11484.  But github links with anchor (current and in the future) can rot.
4. Ignore Github links altogether.  However, there are many github links currently in use.

For now I'm leaning towards approach 2 or 3.

di:
Fixes #1272.

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -9,7 +9,7 @@ There are many techniques to maintain a single source of truth for the version
 number of your project:
 
 #.  Read the file in :file:`setup.py` and get the version. Example (from `pip setup.py
-    <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
+    <https://github.com/pypa/pip/blob/main/setup.py#user-content-L11>`_)::
 
         import codecs
         import os.path
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -9,7 +9,7 @@ There are many techniques to maintain a single source of truth for the version
 number of your project:
 
 #.  Read the file in :file:`setup.py` and get the version. Example (from `pip setup.py
-    <https://github.com/pypa/pip/blob/main/setup.py#user-content-L11>`_)::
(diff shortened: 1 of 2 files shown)


theacodes (approved):

theacodes:
looks like it's still mad about:

> (guides/single-sourcing-package-version: line   11) broken    https://github.com/pypa/pip/blob/main/setup.py#user-content-L11 - Anchor 'user-content-L11' not found

source/guides/single-sourcing-package-version.rst
```diff
@@ -9,7 +9,7 @@ There are many techniques to maintain a single source of truth for the version
 number of your project:
 
 #.  Read the file in :file:`setup.py` and get the version. Example (from `pip setup.py
-    <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
+    <https://github.com/pypa/pip/blob/main/setup.py#user-content-L11>`_)::
```
webknjaz, line 12:
```suggestion
    <https://github.com/pypa/pip/blob/main/setup.py#L11>`_)::
```

So it looks like we'll need to extend https://github.com/pypa/packaging.python.org/blob/03b212a/source/conf.py#L388-L395 with something like
```python
r"^https://github\.com(/[^/+]){2}[^#]*#L\d+(-L\d+)?$",  # these render fully on front-end from anchors
```
(disclaimer: the regex is not tested and made up while writing this comment)

PR #1290, opened 2023-08-19
Add references to PEPs 691, 700, and 714 to the simple repository API spec page

brettcannon:


source/key_projects.rst
```diff
@@ -558,7 +558,7 @@ resolution by locally caching metadata about dependencies.
 pypiserver
 ==========
 
-`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.rst>`__ |
+`Docs <https://github.com/pypiserver/pypiserver/blob/master/README.md>`__ |
```
webknjaz, line 561: This should probably just reference the repo page.

brettcannon: That's what I thought, too, but I didn't want to be too adventurous. 😅 I'll update the PR.

brettcannon:
I also noticed some of the links in the file with the busted link point to `/en/latest/` links on Read the Docs which is probably not preferred in case the docs are translated.

source/key_projects.rst
webknjaz: thanks!

PR #1294, opened 2023-09-06
Add a gate job to GHA CI/CD

webknjaz:


webknjaz:
@pradyunsg could you change the branch protection to require the `check` job instead or the separate ones + require merge queues?

PR #1307, opened 2023-09-29
Add step to create empty GH Release before upload

### LINKED ISSUE #1304: 'Upload artifact signatures to GitHub Release' - gh release step fails
Refering to the guide [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

For the step 'Upload artifact signatures to GitHub Release' it's not clear that this step will only succeed if the release has already been created on github (i.e. manually or by another process).  If the release is not present, then ['release not found'](https://github.com/cli/cli/blob/ec4935946714159c90d9eca1acf95410d20aaa20/pkg/cmd/release/shared/fetch.go#L234) will be displayed.

It's possible that the preferable action would be to create a release using ['gh release create'](https://cli.github.com/manual/gh_release_create)

chrysle:
Fixes #1304.

As suggested, I chose adding an empty release body, but we could also demonstrate how to read from a changelog file.

Could I also include fixes for #777 and #804 to solve the issues in one piece?

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -73,6 +73,11 @@ jobs:
         inputs: >-
           ./dist/*.tar.gz
           ./dist/*.whl
+    - name: Create GitHub Release
+      run: >-
+        gh release create
+        '${{ github.ref_name }}'
+        --notes ""
     - name: Upload artifact signatures to GitHub Release
       env:
         GITHUB_TOKEN: ${{ github.token }}
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -163,8 +163,13 @@ Signing the distribution packages
 The following job signs the distribution packages with `Sigstore`_,
 the same artifact signing system `used to sign CPython <https://www.python.org/download/sigstore/>`_.
 
-It uses the `sigstore/gh-action-sigstore-python GitHub Action`_,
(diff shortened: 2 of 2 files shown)


webknjaz:
> but we could also demonstrate how to read from a changelog file.

That would be something for a separate guide, I think. Though, there's a flag for auto-generating it. Of course, it just dumps the commits into the "log" but it's probably better, than nothing.

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -73,6 +73,11 @@ jobs:
         inputs: >-
           ./dist/*.tar.gz
           ./dist/*.whl
+    - name: Create GitHub Release
+      run: >-
+        gh release create
+        '${{ github.ref_name }}'
```
webknjaz, line 79: Does this have the `refs/tags/` prefix?

webknjaz (approved):

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
chrysle: No, that is `${{ github.ref }}}`.

webknjaz: Okay, marking this as resolved.

webknjaz:
> Could I also include fixes for #777 and #804 to solve the issues in one piece?

I wouldn't. Atomic PRs are easier to review. Plus with the merge queues on, it should be possible to merge them at the same time.

webknjaz:
Also, fixing them would introduce setuptools-scm which means going into the details of the packaging tooling which is something the guide explicitly avoids right now. So it's probably something to explore in a more advanced guide.

Issue #1203, opened 2023-02-03
Add changelogs to all standards


encukou:
It would be good to add a “History” section to all [PyPA specifications](https://packaging.python.org/en/latest/specifications/), [like the one in `direct-url`](https://packaging.python.org/en/latest/specifications/direct-url/#history).
Even with a single entry, the section is useful -- to set expectations for future updates.
Please add dates if the spec isn't versioned in another way.
See [discussion on Discourse](https://discuss.python.org/t/1659/28).

jeanas:
I agree that dates throughout would be useful, since they may be relevant to users when they wonder how recent the tool must be to support feature X or Y.

IMHO, we don't need the PEP approval messages because they're already recorded on the PEP's page (or should be); OTOH, links to discussions other than PEPs are useful.

PR #1308, opened 2023-10-03
Add history section to externally managed environments spec.

serixscorpio:
Addresses a small part of #1203 by adding a history section to the external management environments specification page.

### DIFF
--- a/source/specifications/externally-managed-environments.rst
+++ b/source/specifications/externally-managed-environments.rst
@@ -20,3 +20,7 @@ installation to indicate to Python-specific tools such as ``pip`` that they
 neither install nor remove packages into the interpreter’s default installation
 environment, and should instead guide the end user towards using
 :ref:`virtual-environments`.
+
+History
+=======
+- `June 2022 <https://discuss.python.org/t/pep-668-marking-python-base-environments-as-externally-managed/10302/44>`_: ``EXTERNALLY-MANAGED`` marker file was originally specified in :pep:`668#marking-an-interpreter-as-using-an-external-package-manager`.


chrysle (approved):

webknjaz:
@chrysle it seems like that breezy website needs to be added to `linkcheck_ignore`: https://lists.ubuntu.com/archives/bazaar/2023q4/076422.html.

Also, looking into the merge queues behavior: it looks like GH generates 2 events for it — one is `merge_group` and the other is `push`. This triggers 2 CI runs — https://github.com/pypa/packaging.python.org/actions/runs/6614309433 and https://github.com/pypa/packaging.python.org/actions/runs/6614309436 respectively.
We also have concurrency groups set up in a way that results into both events sharing the same group identifier. So the `push` CI workflow cancels the `merge_group` one. We may want to adjust this too...

webknjaz:
This https://github.blog/changelog/2022-08-18-merge-group-webhook-event-and-github-actions-workflow-trigger/ says
> A `push` event is still sent when a merge group branch is created, and will trigger a GitHub Actions workflow. However, unlike a `merge_group` event, a `push` event does not include the target branch of the merge group.

Which kinda implies that we may need to drop the `push` event.

webknjaz:
> So the `push` CI workflow cancels the `merge_group` one.

Ah, and that cancellation prompts GH to consider the CI run a failure, leading to destroying of the merge queue and unqueueing the PR.

webknjaz:
@chrysle https://github.com/pypa/packaging.python.org/pull/1318 + https://github.com/pypa/packaging.python.org/pull/1319 should improve things.

PR #1312, opened 2023-10-10
Address `distutils` removal from Python stdlib

### LINKED ISSUE #1286: Update reference to distutils
https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#platform-tag needs an alternative to `distutils.util.get_platform()`

chrysle:
This fixes some links to the `distutils` section in the Python docs and updates a few references to `distutils`.

Closes #1286.


chrysle:
@webknjaz BTW, you can't actually add a pull request to the queue without a passed CI run, can you? Because this means I'll have to update some branches nevertheless before queueing, to get the CI green.

webknjaz:
This is weird. I'll need to ask @pradyunsg to clarify what the current settings are..

webknjaz:
IIRC PR approvals would still be required. Without that, there's no point in putting things on the queue..

webknjaz (approved):

webknjaz:
So it allowed me to mark it as automerge but actually added to the queue after I posted the required approval..

PR #1319, opened 2023-10-23
Temporarily ignore www.breezy-vcs.org @ linkcheck

webknjaz:
Ref:
* https://github.com/pypa/packaging.python.org/pull/1308#issuecomment-1775347690


chrysle:
The website seems to be online again.

chrysle (approved):

PR #1321, opened 2023-10-25
Update packaging-projects.rst to require Python 3.8+

finsberg:
Set lower bound on `requires-python` to 3.8 since 3.7 is EOL.


willingc (approved):
Thanks @finsberg for the update.

chrysle (approved):

PR #1323, opened 2023-10-26
Fix issue with creation of GitHub release caused by missing cli switch

aholmes:
Reference: #1307 and #1304

The command added, `gh release create`, fails without the `--repo` switch. This change adds it.

Here is an Action run that fails without it. https://github.com/uclahs-cds/BL_Python/actions/runs/6658063247/job/18094142825

And an Action run that succeeds with it. https://github.com/uclahs-cds/BL_Python/actions/runs/6658181985/job/18094515444

A secondary change in this PR is necessary so authorization will succeed when both `gh release create` and `gh release upload` are executed. The change moves the definition of the `GITHUB_TOKEN` envvar from the `gh release upload` step and adds it to the job so that the envvar is available to both steps.

See here for the failure this secondary change fixes. https://github.com/uclahs-cds/BL_Python/actions/runs/6658013371/job/18093994984


aholmes:
@webknjaz any chance to get this merged? I can make any additional changes necessary?

source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
```diff
@@ -61,6 +61,9 @@ jobs:
       contents: write  # IMPORTANT: mandatory for making GitHub Releases
       id-token: write  # IMPORTANT: mandatory for sigstore
 
+    env:
```
webknjaz, line 64: @aholmes plz move this to the specific step. We shouldn't leak such secrets to steps that don't need them, especially third-party ones. Scoping is intentional and is a result of respective security considerations.

aholmes: Hi @webknjaz,
In my testing I found that both steps require the token. Would you still prefer this only exists in each individual step?

aholmes: I went ahead with the change.

pradyunsg (approved):

pradyunsg:
I'll defer to @webknjaz to click the merge button. :)

webknjaz (approved):

webknjaz:
Thanks!

PR #1325, opened 2023-11-01
Choosing your own build backend

hauntsaninja:
53 posts and counting on https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502/ , so figured it's time to start translating that into PRs


zooba:
Haha, looks like you posted this while I was doing #1326 and I didn't notice it.

I've gone a bit further than just changing the title - think you could review mine and maybe push it through?

hauntsaninja:
Great!

PR #1326, opened 2023-11-01
Update Choosing a build backend section

zooba:
Following on from the discussion at https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502/


source/tutorials/packaging-projects.rst
```diff
@@ -159,14 +164,24 @@ Open :file:`pyproject.toml` and enter one of these ``[build-system]`` tables:
...
 
-- ``requires`` is a list of packages that are needed to build your package. You
-  don't need to install them; build frontends like :ref:`pip` will install them
-  automatically in a temporary, isolated virtual environment for use during the
-  build process.
-- ``build-backend`` is the name of the Python object that frontends will use to
-  perform the build.
+The ``requires`` key is a list of packages that are needed to build your package.
+The front end should install them automatically when building your package.
+There will always be your build backend package, and potentially other build-time
```
zooba, line 169:
```suggestion
This should always include your backend's package, and might have other build-time
```

source/tutorials/packaging-projects.rst
```diff
@@ -159,14 +164,24 @@ Open :file:`pyproject.toml` and enter one of these ``[build-system]`` tables:
...
+dependencies.
 
-.. TODO: Add note to check the tools' documentation for the current snippet?
+The ``build-backend`` key is the name of the Python object that frontends will use
+to perform the build.
+
+Both of these values will be given to you by the documentation for your build
```
zooba, line 175:
```suggestion
Both of these values will be provided by the documentation for your build
```

ketozhang:
This changes breaks the cohesion reading from _Creating the package files_ where `pyproject.toml` is first introduced—the users has to make a logical jump to _Choosing a build backend_.

I suggest adding a summary of each file/dir in _Creating the package files_. Here, `pyproject.toml` should point users to the _Choosing a build backend_ section.

zooba:
Funny, I never actually made the connection that this was going to be an entire section of creating files. I guess because all the headings look about the same size.

For the sake of not bloating this change out of scope, I'm going to defer that to another PR, which someone else can probably do. I want to prove that it's possible to contribute small improvements to the documentation without every contributor being forced into a complete rewrite every time. That's _exactly_ the sort of bad habit that leads to people spitting the dummy and never contributing.

ketozhang:
Fair enough—let's not change other sections.

source/tutorials/packaging-projects.rst
```diff
@@ -103,15 +103,17 @@ Creating a test directory
...
 
 
-Creating pyproject.toml
+Choosing a build backend
 -----------------------
 
-.. TODO: Add an intro sentence about pyproject.toml, and a sub-heading for
-   "Configuring build tools"
+Tools like :ref:`pip` and :ref:`build` do not actually convert your sources
+into a :term:`distribution package <Distribution Package>`;
```
hauntsaninja, line 110:
```suggestion
into a :term:`distribution package <Distribution Package> (like a wheel)`; 
```
nit: the glossary entry is a little heavy, useful to avoid forcing user to click

source/tutorials/packaging-projects.rst
```diff
@@ -103,15 +103,17 @@ Creating a test directory
...
 
-.. TODO: Add an intro sentence about pyproject.toml, and a sub-heading for
-   "Configuring build tools"
+Tools like :ref:`pip` and :ref:`build` do not actually convert your sources
+into a :term:`distribution package <Distribution Package>`; 
+that job is performed by a _build backend_. The build backend determines how
+your project will specify its configuration, including metadata (information
+about the project, for example, the name and tags that are displayed on PyPI)
+and input files. Build backends have different levels of functionality, and
+you should choose one that suits your needs, but also meets your preferences.
```
hauntsaninja, line 115:
```suggestion
and input files. Build backends have different levels of functionality (in particular,
when it comes to building :term:`extension modules <Extension Module`), and
you should choose one that suits your needs, but also meets your preferences.
```

source/tutorials/packaging-projects.rst
```diff
@@ -124,7 +126,10 @@ table for :ref:`metadata <configuring metadata>`.
    management, as well as building, uploading, and installing packages. This
```
hauntsaninja, line 126: (not part of this PR, but this paragraph would be aided by examples)

source/tutorials/packaging-projects.rst
```diff
@@ -124,7 +126,10 @@ table for :ref:`metadata <configuring metadata>`.
...
    tutorial uses single-purpose tools that work independently.
 
-Open :file:`pyproject.toml` and enter one of these ``[build-system]`` tables:
+The :file:`pyproject.toml` tells "front end" build tools like :ref:`pip` and
```
hauntsaninja, line 129: (not necessarily part of this PR, but frontend and backend get used enough that we should have glossary entries that we link to from here)

source/tutorials/packaging-projects.rst
```diff
@@ -159,14 +164,24 @@ Open :file:`pyproject.toml` and enter one of these ``[build-system]`` tables:
...
 
 
-- ``requires`` is a list of packages that are needed to build your package. You
-  don't need to install them; build frontends like :ref:`pip` will install them
-  automatically in a temporary, isolated virtual environment for use during the
-  build process.
-- ``build-backend`` is the name of the Python object that frontends will use to
-  perform the build.
+The ``requires`` key is a list of packages that are needed to build your package.
+The front end should install them automatically when building your package.
```
hauntsaninja, line 168: Build isolation was an important part of the discuss thread, this could be a good place to mention this.
```suggestion
The frontend should install them automatically when building your package.
Note many frontends will default to building out of a clean environment, so
undeclared build dependencies will cause build time errors.
```

hauntsaninja (commented):
nit: frontend is sometimes spelt with the space, I'd vote on consistently using "frontend" everywhere for symmetry with "backend" / because it's a common term in the post PEP 517 world

hauntsaninja (approved):
Overall looks great, thank you for doing this! No comments that I'd consider blocking, but left a few nits

source/tutorials/packaging-projects.rst
zooba:
```suggestion
into a :term:`distribution package <Distribution Package>` (like a wheel); 
```

I think the comment needs to go outside the quote. Adding my own suggestion so I can merge it in the web UI

source/tutorials/packaging-projects.rst
zooba:
```suggestion
and input files. Build backends have different levels of functionality, such as
whether they support building :term:`extension modules <Extension Module>`, and
you should choose one that suits your needs and preferences.
```

Agree, just rewording slightly and fixing the syntax.

source/tutorials/packaging-projects.rst
zooba: Agree, and already filed at #1327

source/tutorials/packaging-projects.rst
zooba: Actually, it's technically a requirement, so should probably word it more strongly. And agree, this is the place to mention it (briefly).

```suggestion
The front end should install them automatically when building your package.
Front ends usually run builds in isolated environments, so omitting dependencies
here will cause build-time errors.
```

hauntsaninja (approved):

sinoroc:
> nit: frontend is sometimes spelt with the space, I'd vote on consistently using "frontend" everywhere for symmetry with "backend" / because it's a common term in the post PEP 517 world

https://english.stackexchange.com/a/132325

There seems to be a consensus towards "front end (noun)" and "front-end (adjective)".

zooba:
(Cross posting from https://github.com/pypa/packaging.python.org/issues/1327#issuecomment-1789737664 )

I think the space-less versions have enough precedence that we should use them (certainly we're pretty good about "backend", and while "frontend" looks weird to my eyes, I'm not massively offended by it).

e.g. https://en.wikipedia.org/wiki/Frontend_and_backend

The adjective forms get the hyphen, but mostly we use the noun form (and possessive - "backend's documentation" rather than "back-end documentation").

willingc (approved):
Thanks @zooba for taking the time to make this improvement based on the user feedback on discourse. 💯

pradyunsg:
@zooba IIUC, this PR needs changes to fix the CI failures. :)

henryiii:
```
/home/runner/work/packaging.python.org/packaging.python.org/source/tutorials/packaging-projects.rst:107: WARNING: Title underline too short.

Choosing a build backend
-----------------------

/home/runner/work/packaging.python.org/packaging.python.org/source/tutorials/packaging-projects.rst:109: ERROR: Unknown target name: "backend".
```

source/tutorials/packaging-projects.rst
```diff
@@ -103,15 +103,18 @@ Creating a test directory
 :file:`tests/` is a placeholder for test files. Leave it empty for now.
 
 
-Creating pyproject.toml
+Choosing a build backend
 -----------------------
```
henryiii, line 107:
```suggestion
------------------------
```

source/tutorials/packaging-projects.rst
```diff
@@ -103,15 +103,18 @@ Creating a test directory
...
 
-.. TODO: Add an intro sentence about pyproject.toml, and a sub-heading for
-   "Configuring build tools"
+Tools like :ref:`pip` and :ref:`build` do not actually convert your sources
+into a :term:`distribution package <Distribution Package>` (like a wheel); 
+that job is performed by a _build backend_. The build backend determines how
```
henryiii, line 111:
```suggestion
that job is performed by a *build backend*. The build backend determines how
```

Underlines for emphasis are only in markdown. You have to use stars in rst. This is being interpreted as a reference. https://docutils.sourceforge.io/docs/user/rst/quickref.html

zooba:
Thanks @henryiii!

willingc (approved):

willingc:
@pradyunsg @henryiii When one of you has a moment, please merge. The automerge was disabled when a push happened. Thanks.

webknjaz (approved):

pradyunsg:
Thanks @webknjaz for the magic button clicks, @zooba for filing this and everyone else for the discussion here. :)

PR #1328, opened 2023-11-02
Remove backend choice and just use Hatchling so the tutorial can be used without making any choices.

BrenBarn:
This conflicts with #1326.  As has been discussed on Discourse, I think for a tutorial the goal should be to have it work straight through without requiring the user to decide among competing tools.  If we want to talk about choosing build backends that could be a separate section.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -1,9 +1,9 @@
 Packaging Python Projects
 =========================
 
-This tutorial walks you through how to package a simple Python project. It will
-show you how to add the necessary files and structure to create the package, how
-to build the package, and how to upload it to the Python Package Index (PyPI).
+The tools required for building and distributing packages are third-party tools not included with Python.  This tutorial uses Hatchling, which is one among several such tools.  If you are already using different build tools, or you want to for whatever reason, you will need to look at the documentation for those tools.  In many cases it will be fairly easy to adapt this example.
+
+This tutorial walks you through how to package a simple Python project --- that is, one that contains only Python code (although it may depend on libraries that include extensions written in other languages).  If you need to package a project that has parts written in a language other than Python (such as a C extension), please look for third-party tools.  But, for a simple pure Python project, this tutorial show you how to add the necessary files and structure to create the package, how to build the package, and how to upload it to the Python Package Index (PyPI).
 
 .. tip::
 
@@ -103,6 +103,12 @@ Creating a test directory
 :file:`tests/` is a placeholder for test files. Leave it empty for now.
 
 
+Installing the build backend
+----------------------------
+
+As mentioned, this tutorial uses the Hatchling build tool.  Go to https://hatch.pypa.io/latest/install/ and follow the instructions to install it.  (Or, install a different build tool and adapt the example below, following the documentation for that tool.)
+
+
 Creating pyproject.toml
 -----------------------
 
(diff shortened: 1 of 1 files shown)


henryiii:
The point of the tabs is to show that it doesn't matter what backend you use, and visually illustrate exactly what selecting a backend means. It also keeps `packaging.python.org` independent of a specific backend, rather than being a hatching tutorial. (Or a setuptools tutorial, etc). I'd recommend looking up the 6+ months of discussion on this before it was accepted before trying to remove it.

As far as I can tell, the point of the discourse thread was that selecting a backend needs to be an important step, not removed. https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502

BrenBarn:
> The point of the tabs is to show that it doesn't matter what backend you use

That is not a point that should be made in this tutorial.  This tutorial is about how to package a simple Python project.  This tutorial should show you the user to package a simple Python project, without the user even having to think about build backends *at all*.

> I'd recommend looking up the 6+ months of discussion on this before it was accepted before trying to remove it.

I did read that discussion and I don't agree with the decision so I'm making a PR with a version that I think is better.

henryiii:
IMO, it would be a better use of time to update https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/ to support modern packaging and multiple backends. **It literally does not mention pyproject.toml**. Once that is done, _then_ this tutorial probably could be simplified.

BrenBarn:
> IMO, it would be a better use of time to update https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/ to support modern packaging and multiple backends. It literally does not mention pyproject.toml. Once that is done, then this tutorial probably could be simplified.

Hmmm.  What do you see as the difference in goals or approach between that tutorial and this one?  When I look at that, it looks to me like all the setup.py stuff could just be excised and replaced with a link to this tutorial, along with a brief snippet of a `[build-system]` table using setuptools (if we want that tutorial to specifically be about how to do things with setuptools.)

henryiii:
That's not supposed to be a tutorial, it's a guide. It should be more in-depth, and not "follow-along". Currently, this tutorial is also basically a guide, since it's the one and only modern page on packaging. Unfortunately, it needs to be completely rewritten, I think.

pradyunsg:
My opinion on this is that we should flip the order here: change the tutorial to use a single backend (specifically: setuptools) and update the guide to showcase all the options available regarding the backends.

henryiii:
Yes, but I think we need to update the guide first, before removing the guide-like parts from the tutorial.

> Once that is done, then this tutorial probably could be simplified.


Why setuptools, though? That's a lot slower to setup than hatching, and produces more complex error messages; newcomers get a nicer first taste of Python packaging with hatching (which is why it's first).

pradyunsg:
Closing this out, since we've moved forward with #1326. Thanks for filing this none the less @BrenBarn!

> Why setuptools, though?

2 reasons:

1. It's the most popular backend _today_ (yes, historical reasons and all that) and being the backend picked here will be taken as a signal of recommendation. setuptools is a name and tool that people are already familiar with if they've done Python packaging stuff $years ago and are looking at the space again (or are a newcomer who is looking at this space for the first time, but been "around" the language in a manner of saying).
2. I think a tutorial, which guides users through each step and configuration, does not need to worry about the papercuts as much, just because it's less likely that users will hit them.

Issue #1327, opened 2023-11-01
Add build backend and build frontend to the glossary


sinoroc:
I believe the terms "build frontend" and "build backend" should be defined in the glossary.

sinoroc:
https://english.stackexchange.com/a/132325

There seems to be a consensus towards "front end (noun)" and "front-end (adjective)".

zooba:
I think the space-less version have enough precedence that we should use them (certainly we're pretty good about "backend", and while "frontend" looks weird to my eyes, I'm not massively offended by it).

e.g. https://en.wikipedia.org/wiki/Frontend_and_backend

The adjective forms get the hyphen, but mostly we use the noun form (and possessive - "backend's documentation" rather than "back-end documentation").

PR #1329, opened 2023-11-02
Add build backend and frontend to the glossary

sinoroc:
GitHub: fixes https://github.com/pypa/packaging.python.org/issues/1327

Preview:

* [Build Backend](https://python-packaging-user-guide--1329.org.readthedocs.build/en/1329/glossary/#term-Build-Backend)
* [Build Frontend](https://python-packaging-user-guide--1329.org.readthedocs.build/en/1329/glossary/#term-Build-Frontend)

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -12,6 +12,36 @@ Glossary
         extensions.
 
 
+    Build Backend
+
+        A library that takes a source tree or
+        :term:`source distribution <Source Distribution (or "sdist")>`
+        and builds a source distribution or :term:`wheel <Wheel>` from it.
+        The build is delegated to the backend by a
+        :term:`frontend <Build Frontend>`.
+        All backends offer a standardized interface.
+
+        Examples of build backends are
+        :ref:`flit's flit-core <flit>`,
+        :ref:`hatch's hatchling <hatch>`,
+        :ref:`maturin`,
+        :ref:`meson-python`,
+        :ref:`scikit-build-core`,
+        and :ref:`setuptools`.
+
+
+    Build Frontend
+
+        A tool that users might run
+        that takes arbitrary source trees or
+        :term:`source distributions <Source Distribution (or "sdist")>`
+        and builds source distributions or :term:`wheels <Wheel>` from them.
+        The actual building is delegated to each source tree's
+        :term:`build backend <Build Backend>`.
+
+        Examples of build frontends are :ref:`pip` and :ref:`build`.
+
+
     Built Distribution
 
         A :term:`Distribution <Distribution Package>` format containing files
--- a/source/flow.rst
+++ b/source/flow.rst
@@ -82,9 +82,12 @@ For example, here is a table for using :ref:`hatch`:
     requires = ["hatchling"]
     build-backend = "hatchling.build"
 
-With such a table in the :file:`pyproject.toml` file, a "frontend" tool like
-:ref:`build` can run your chosen build tool's "backend" to create the build
-artifacts. Your build tool may also provide its own frontend. An install tool
+With such a table in the :file:`pyproject.toml` file,
+a ":term:`frontend <Build Frontend>`" tool like
+:ref:`build` can run your chosen
+build tool's ":term:`backend <Build Backend>`"
+to create the build artifacts.
+Your build tool may also provide its own frontend. An install tool
 like :ref:`pip` also acts as a frontend when it runs your build tool's backend
 to install from a source distribution.
 
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -140,7 +140,7 @@ hatch
 
 Hatch is a unified command-line tool meant to conveniently manage
 dependencies and environment isolation for Python developers. Python
-package developers use Hatch and its build backend Hatchling to
+package developers use Hatch and its :term:`build backend <Build Backend>` Hatchling to
 configure, version, specify dependencies for, and publish packages
 to PyPI. Its plugin system allows for easily extending functionality.
 
--- a/source/specifications/declaring-project-metadata.rst
+++ b/source/specifications/declaring-project-metadata.rst
@@ -27,7 +27,7 @@ table which are not defined by this specification. For tools wishing
 to store their own settings in ``pyproject.toml``, they may use the
 ``[tool]`` table as defined in the
 :ref:`build dependency declaration specification <declaring-build-dependencies>`.
-The lack of a ``[project]`` table implicitly means the build back-end
+The lack of a ``[project]`` table implicitly means the :term:`build backend <Build Backend>`
 will dynamically provide all keys.
 
 The only keys required to be statically defined are:
--- a/source/specifications/source-distribution-format.rst
+++ b/source/specifications/source-distribution-format.rst
@@ -43,7 +43,8 @@ The name and version components of the filename MUST match the values stored
 in the metadata contained in the file.
 
 Code that produces a source distribution file MUST give the file a name that matches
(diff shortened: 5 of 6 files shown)


sinoroc:
Definitions are taken from [PEP 517](https://peps.python.org/pep-0517/#terminology-and-goals). I hope someone will jump in and make improvement suggestions either now or in a follow-up pull request.

source/glossary.rst
```diff
@@ -12,6 +12,36 @@ Glossary
...
+        :term:`source distributions <Source Distribution (or "sdist")>`
+        and builds a :term:`wheel <Wheel>` from it.
+        The build is delegated to the backend by a
+        :term:`frontend <Build Frontend>`.
+        All backends offer a standardized interface.
+
+        Examples of build backends are
+        :ref:`setuptools`,
+        :ref:`hatch's hatchling <hatch>`,
+        and :ref:`flit`.
```
henryiii, line 27:
```suggestion
        and :ref:` flit's flit-core <flit>`.
```

Technically flit-core. Might be nice to provide a few examples of compiled backends too (scikit-build-core, meson-python, and maturin).

sinoroc: @henryiii 

Oh, right it definitely should have been `flit-core` (especially since I took care of listing `hatchling`).

Regarding adding more to the list, it is tough, because where do we end the list? :D

I wanted to keep the list as short as possible because:
* It is just a glossary definition. Maybe we should remove the examples entirely?
* Maybe we should avoid falling in the trap again of giving the impression that the PyPUG recommends or blesses those.

I think I will just implement your suggestion and we'll see if someone else has an opinion.

Thanks for feedback :)

sinoroc:
This assumes https://github.com/pypa/packaging.python.org/pull/1330 gets merged first.

source/glossary.rst
```diff
@@ -12,6 +12,39 @@ Glossary
...
+        :term:`frontend <Build Frontend>`.
+        All backends offer a standardized interface.
+
+        Examples of build backends are
+        :ref:`flit's flit-core <flit>`,
+        :ref:`hatch's hatchling <hatch>`,
+        :ref:`maturin`,
+        :ref:`meson-python`,
+        :ref:`scikit-build-core`,
+        and :ref:`setuptools`.
```
henryiii, line 30:
```suggestion
        Examples of Python build backends are
        :ref:`flit's flit-core <flit>`,
        :ref:`hatch's hatchling <hatch>`,
        and :ref:`setuptools`. Compiled extensions use backends such as
        :ref:`maturin`,
        :ref:`meson-python`, and
        :ref:`scikit-build-core`.
```

I'm not sure exactly how to put it, but the idea was to have a small list of the most common Python build backends, and a small list of the most common compiled extension backends. Users (like in the thread that started this) are sometimes looking for a replacement for compiled extensions, and being pointed to something like flit isn't helpful for them. But many users don't care so listing a few examples separately helps both groups see that they are included, I think. Also, setuptools does both, but I don't think that matters here.

sinoroc: From my point of view this goes to deep in details for a glossary. What do you think? But I am fine with taking in this suggestion, for me the most important is to get something merged so that we can have the links in place for the other pull requests.

I think what we might need is a dedicated page/chapter to the whole build frontends and backends story. Maybe even with some kind of general comparison of build backends like this: https://python-packaging-user-guide--1196.org.readthedocs.build/en/1196/discussions/packaging-tools-comparisons/#build-back-ends

abravalheri: But `setuptools` happens to be both 😅

sinoroc: I would assume that all build backends can do pure Python, and only some of them can also do extensions. I really am not sure it makes sense to touch on this distinction in the glossary.

henryiii: How about:

Examples of build backends are flit's flit-core, hatch's hatchling, and setuptools. Some build backends support compiled extensions too.

henryiii:
Rebase time. :)

webknjaz:
@henryiii rebasing shouldn't be necessary with the merge queues anymore.

webknjaz:
OTOH, the conflict does need to be solved now.

sinoroc:
Rebased. And linked a few more occurrences of the terms (only one per page, the first occurrence).

I think it is ready.

@henryiii @webknjaz

source/glossary.rst
```diff
@@ -12,6 +12,39 @@ Glossary
         extensions.
 
 
+    Build Backend
+
+        A library that takes a source tree or
+        :term:`source distributions <Source Distribution (or "sdist")>`
+        and builds a :term:`wheel <Wheel>` from it.
```
webknjaz, line 19: This is not entirely true. It also builds sdists from source, not just wheels.

source/glossary.rst
```diff
@@ -12,6 +12,39 @@ Glossary
...
+        that takes arbitrary source trees or
+        :term:`source distributions <Source Distribution (or "sdist")>`
+        and build :term:`wheels <Wheel>` from them.
+        The actual building is done by each source tree's
+        :term:`build backend <Build Backend>`.
+
+        Examples of build frontends are
+        :ref:`pip`
+        and
+        :ref:`build`.
```
webknjaz, line 45: It's probably fine to have these on one line.
```suggestion
        Examples of build frontends are :ref:`pip` and :ref:`build`.
```

source/glossary.rst
```diff
@@ -12,6 +12,39 @@ Glossary
...
+        :ref:`scikit-build-core`,
+        and :ref:`setuptools`.
+
+
+    Build Frontend
+
+        A tool that users might run
+        that takes arbitrary source trees or
+        :term:`source distributions <Source Distribution (or "sdist")>`
+        and build :term:`wheels <Wheel>` from them.
```
webknjaz, line 38: Same here: this implies that sdists are built by some other kind of tooling. But the toolchain is the same, really. PEP 517 declares hooks concerning both wheels and sdists separately.

Moreover, with PEP 660, the build backend and frontend pairs interact with each other, producing a special kind of wheel and metadata. We should probably mention that somehow.

source/glossary.rst
sinoroc: Right... I wonder why it was phrased liked this in the PEP.

A bit tough to phrase correctly without making it sound awkward, though... I will give it a shot.

sinoroc: Fixed. I did not do a good job at phrasing this, it is awkward as I feared, but at least the info is here. My hope is that someone else will come to fix this in a follow-up PR.

source/glossary.rst
sinoroc: Done.

source/glossary.rst
sinoroc: > Moreover, with PEP 660, the build backend and frontend pairs interact with each other, producing a special kind of wheel and metadata. We should probably mention that somehow.

I wonder if it would be confusing to have this level of details in the glossary. I think we should rather have a dedicated page/chapter somewhere to explain this.

source/glossary.rst
webknjaz: > Right... I wonder why it was phrased liked this in the PEP.

Which PEP?

https://peps.python.org/pep-0517/#terminology-and-goals mentions wheels:

> A _build frontend_ is a tool that users might run that takes arbitrary source trees or source distributions and builds wheels from them. The actual building is done by each source tree’s _build backend_. In a command like `pip wheel some-directory/`, pip is acting as a build frontend.

I suppose this does sound ambiguous… @pradyunsg WDYT?

webknjaz (approved):

PR #1330, opened 2023-11-02
chore: add and update some backends

henryiii:
This adds binary backends scikit-build-core and maturin to the list of related packaging tools.

And changes:

* Multibuild now is in a org, the old link now is a fork.
* Distutils isn't part of the standard library anymore, so move it to a PyPI project and update the text.

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -484,7 +484,7 @@ Maturin
 `Docs <https://www.maturin.rs>`__ |
 `GitHub <https://github.com/PyO3/maturin>`__
 
-Maturin is a PEP 517 build backend for Rust extension modules, also written in
+Maturin is a build backend for Rust extension modules, also written in
 Rust. It supports building wheels for python 3.7+ on Windows, Linux, macOS and
 FreeBSD, can upload them to PyPI and has basic PyPy and GraalPy support.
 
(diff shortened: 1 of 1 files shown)


source/key_projects.rst
```diff
@@ -604,6 +641,22 @@ build requirements. To speed up and parallelize the build of large projects,
...
+.. _scikit-build-core:
+
+scikit-build-core
+=================
+
+`Docs <https://scikit-build-core.readthedocs.io/en/latest/>`__ |
+`GitHub <https://github.com/scikit-build/scikit-build-core/>`__ |
+`PyPI <https://pypi.org/project/scikit-build-core>`__
+
+Scikit-build-core is a PEP 517 backend for CPython C/C++/Fortran/Cython
```
sinoroc, line 653: We should avoid referencing PEPs, especially if they are already finalized. I am not sure we have an agreed upon name for this yet, though. I think a name was discussed but I do not remember what the conclusion of the discussion was. I usually call it "_the `[build-system]` standard_".

source/key_projects.rst
```diff
@@ -452,6 +476,19 @@ problem of installing scientific software, and making package
...
 
+.. _maturin:
+
+Maturin
+=======
+
+`Docs <https://www.maturin.rs>`__ |
+`GitHub <https://github.com/PyO3/maturin>`__
+
+Maturin is a PEP 517 build backend for Rust extension modules, also written in
```
sinoroc, line 487: Avoid references to PEPs.

sinoroc (commented):
Besides the unlucky reference to a PEP, I think it is good.

sinoroc (approved):
I think it is good.

If I do not forget: once this is merged I will rebase and update my PR https://github.com/pypa/packaging.python.org/pull/1329 to put links to the "build backend" definition.

pradyunsg (approved):

PR #1331, opened 2023-11-02
Add discussion "Is setup.py deprecated?"

sinoroc:
PR preview demo: https://python-packaging-user-guide--1331.org.readthedocs.build/en/1331/discussions/setup-py-deprecated/.

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -1,0 +1,149 @@
+.. _setup-py-deprecated:
+
+
+===========================
+Is ``setup.py`` deprecated?
+===========================
+
+No, :term:`setup.py` is not deprecated,
+it is a valid configuration file for :ref:`setuptools`
+that happens to be written in Python, instead of in *TOML* for example
+(a similar practice is used by other tools
+like *nox* and its :file:`nox.py` configuration file,
+or *pytest* and :file:`conftest.py`).
+
+And of course *setuptools* itself is not deprecated either.
+
+It is however deprecated to run ``python setup.py`` as a command line tool.
+
+This means for example that the following commands **MUST NOT** be run anymore:
+
+* ``python setup.py install``
+* ``python setup.py develop``
+* ``python setup.py sdist``
+* ``python setup.py bdist_wheel``
+
+
+What commands should be used instead?
+=====================================
+
++---------------------------------+----------------------------------------+
+| Deprecated                      | Current recommendation                 |
++=================================+========================================+
+| ``python setup.py install``     | ``python -m pip install .``            |
++---------------------------------+----------------------------------------+
+| ``python setup.py develop``     | ``python -m pip install --editable .`` |
++---------------------------------+----------------------------------------+
+| ``python setup.py sdist``       | ``python -m build`` [#needs-build]_    |
++---------------------------------+                                        |
+| ``python setup.py bdist_wheel`` |                                        |
++---------------------------------+----------------------------------------+
+
+
+.. [#needs-build] This requires the :ref:`build` dependency.
+    It is recommended to always build and publish both the source distribution
+    and wheel of a project, which is what ``python -m build`` does.
+    If necessary the ``--sdist`` and ``--wheel`` options can be used
+    to generate only one or the other.
+
+
+In order to install a setuptools based project,
+it was common to run :file:`setup.py`'s ``install`` command such as:
+``python setup.py install``.
+Nowadays, the recommended method is to use :ref:`pip` directly
+with a command like this one: ``python -m pip install .``.
+Where the dot ``.`` is actually a file system path,
+it is the path notation for the current directory.
+Indeed, *pip* accepts a path to
+a project's source tree directory on the local filesystem
+as argument to its ``install`` sub-command.
+So this would also be a valid command:
+``python -m pip install path/to/project``.
+
+As for the installation in *develop* mode aka *editable* mode,
+instead of ``python setup.py develop``
+one can use the ``--editable`` option of pip's *install* sub-command:
+``python -m pip install --editable .``.
+
+One recommended, simple, and straightforward method of building
+:term:`source distributions <Source Distribution (or "sdist")>`
+and :term:`wheels <Wheel>`
+is to use the :ref:`build` tool with a command like
+``python -m build``
+which triggers the generation of both distribution formats.
+If necessary the ``--sdist`` and ``--wheel`` options can be used
+to generate only one or the other.
+Note that the build tool needs to be installed separately.
+
+The command ``python setup.py install`` was deprecated
+in setuptools version *58.3.0*.
+
+
+What about custom commands?
+===========================
+
+Likewise, custom :file:`setup.py` commands are deprecated.
+The recommendation is to migrate those custom commands
+to a task runner tool or any other similar tool.
+Some examples of such tools are:
+chuy, make, nox or tox, pydoit, pyinvoke, taskipy, and thx.
+
+
+What about custom build steps?
+==============================
+
+Custom build steps that for example
+either overwrite existing steps such as ``build_py``, ``build_ext``, and ``bdist_wheel``
+or add new build steps are not deprecated.
+Those will be automatically called as expected.
+
+
+Should ``setup.py`` be deleted?
+===============================
+
+Although the usage of :file:`setup.py` as an executable script is deprecated,
+its usage as a configuration file for setuptools is absolutely fine.
+There is likely no modification needed in :file:`setup.py`.
+
+.. todo::
+
+    Link to a "How to modernize a setup.py based project?"
+
+
+Is ``pyproject.toml`` mandatory?
+================================
+
+While it is not technically necessary yet,
+it is **STRONGLY RECOMMENDED** for a project to have a :file:`pyproject.toml` file
+at the root of its source tree with a content like this:
+
+.. code:: toml
+
+    [build-system]
+    requires = ["setuptools"]
+    build-backend = "setuptools.build_meta"
+
+
+.. todo::
+
+    Link to "How to modernize a setup.py based project?"
+
+
+The standard fallback behavior for a :term:`build frontend <Build Frontend>`
+in the absence of a :file:`pyproject.toml` file and its ``[build-system]`` table
+is to assume that the :term:`build backend <Build Backend>` is setuptools.
+
+
+Why? What does it all mean?
+===========================
+
+One way to look at it is that the scope of setuptools
+has now been reduced to the role of a build backend.
+
+
+Where to read more about this?
+==============================
+
+* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
+
+* :doc:`setuptools:deprecated/commands`
--- a/source/discussions/index.rst
+++ b/source/discussions/index.rst
@@ -13,3 +13,4 @@ specific topic. If you're just trying to get stuff done, see
    install-requires-vs-requirements
    wheel-vs-egg
    src-layout-vs-flat-layout
+   setup-py-deprecated


source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,78 @@
...
+
+.. list-table::
+    :header-rows: 1
+
+    * - Deprecated
+      - Current recommendation
+    * - ``python setup.py install``
+      - ``python -m pip install .``
+    * - ``python setup.py develop``
+      - ``python -m pip install --editable .``
```
henryiii, line 32:
```suggestion
      - ``python -m pip install --editable .``
    * - ``python setup.py sdist``
      - ``python -m build --sdist`` (requires ``build``)
    * - ``python setup.py bdist_wheel``
      - ``python -m build --wheel`` (requires ``build``)
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,78 @@
...
+one can use the ``--editable`` option of pip's *install* sub-command:
+``python -m pip install --editable .``.
+
+
+Should the ``setup.py`` file be deleted from my project?
+========================================================
+
+Although it is recommended against running ``setup.py`` directly as a script,
+its usage as a configuration file for setuptools is absolutely fine.
+There is likely no modification whatsoever to do in ``setup.py``.
```
henryiii, line 60:
```suggestion
There is likely no modification whatsoever to do in ``setup.py``.
However, even setuptools doesn't require the file, so if you can configure
your project entirely in ``pyproject.toml`` or ``setup.cfg``, you don't
have to keep the ``setup.py`` file.
```

sinoroc:
This assumes https://github.com/pypa/packaging.python.org/pull/1329 gets merged first, which in turn depends on https://github.com/pypa/packaging.python.org/pull/1330.

I can do the rebases we get there.

I think we have stacked pull requests or something like this, but I do not know how to do it (yet). Sorry for the inconvenience.

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,78 @@
+..
+
+
+===========================
+Is ``setup.py`` deprecated?
+===========================
+
+Not exactly.
+What is deprecated is the usage of the ``setup.py`` file as a runnable script,
+in particular its ``install`` and ``develop`` commands.
```
abravalheri, line 10: Let's be more decisive and say "no"... people have trouble to understand that. I want to avoid to have even a slightly fuzzy response.

How about something along the lines:

```suggestion
No, ``setup.py`` is not deprecated, it is a valid configuration file for
``setuptools`` that happens to be written in Python
(a similar practice is used by other tools like ``nox``/``noxfile.py``,
``pytest``/``conftest.py``, etc).

It is, however, deprecated to run ``python setup.py`` as a CLI tool.
```

?

sinoroc: @abravalheri Thanks for the feedback :)

I tried to address your suggestions in the latest updates to this pull request.

source/discussions/setup-py-deprecated.rst
sinoroc: @henryiii Thanks :)

I tried to address this suggestion in the latest updates to this pull request.

source/discussions/setup-py-deprecated.rst
sinoroc: @henryiii Thanks :)

This is now incorporated in the latest updates to this pull request.

sinoroc:
I wrote this based on the questions I have seen on StackOverflow. It is not meant to be exhaustive, and I am sure it can be improved on over time. However imperfect this document is in this state, I think it is important to get such a document published as soon as possible and build from there.

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+as argument to its ``install`` sub-command.
+So the this would be also a valid command:
+``python -m pip install path/to/project``.
+
+As for the installation in *develop* mode aka *editable* mode,
+instead of ``python setup.py develop``
+one can use the ``--editable`` option of pip's *install* sub-command:
+``python -m pip install --editable .``.
+
+One recommended, simple, and strqightforward method of building
```
jamestwebber, line 67:
```suggestion
One recommended, simple, and straightforward method of building
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+
+
+Should ``setup.py`` be deleted?
+===============================
+
+Although the usage of ``setup.py`` as an executable script is deprecated,
+its usage as a configuration file for setuptools is absolutely fine.
+There is likely no modification whatsoever to do in ``setup.py``.
+
+But note that setuptools does not striclty require a ``setup.py`` file.
```
jamestwebber, line 83:
```suggestion
But note that setuptools does not strictly require a ``setup.py`` file.
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+===========================
+
+No, ``setup.py`` is not deprecated,
+it is a valid configuration file for ``setuptools``
+that happens to be written in Python, instead of in *TOML* for example
+(a similar practice is used by other tools
+like *nox* and its ``nox.py`` configuration file,
+or *pytest* and ``conftest.py``).
+
+It is however deprecated to run ``python setup.py`` as a command line tool.
```
pfmoore, line 15: One question that comes up a lot is "I have custom commands defined in my `setup.py` - how should I invoke these if running `setup.py` is deprecated?" I think this warrants its own section here, although it will probably need input from the setuptools maintainers on what the answer actually is. /cc @abravalheri 

Another question in a similar vein is "I've added custom command line options to the sdist/build/... command - how do I expose them now?"

source/discussions/setup-py-deprecated.rst
oscarbenjamin: Also worth mentioning that even if everything is moved out of setup.py you can still provide a mostly setup.py as a backwards compatibility shim for anyone downstream who might still expect to do `python setup.py install`:
```python
from setuptools import setup

setup()  # reads config from pypproject.toml in setuptools >= 0.61.0
```
(at least I think this is true)

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Should ``setup.py`` be deleted?
+===============================
+
+Although the usage of ``setup.py`` as an executable script is deprecated,
+its usage as a configuration file for setuptools is absolutely fine.
+There is likely no modification whatsoever to do in ``setup.py``.
+
+But note that setuptools does not striclty require a ``setup.py`` file.
+This file can be replaced by a ``setup.cfg`` file or a ``pyproject.toml`` file.
```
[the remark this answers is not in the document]
pfmoore: IMO we should keep this discussion simple, *without* multiple options or choices. To that extent, I'd be inclined to not even mention `setup.cfg` here at all.

source/discussions/setup-py-deprecated.rst
oscarbenjamin, line 15: I think that it would be useful to attach some versions to qualify this.

I think that the minimum version of setuptools that can read `pyproject.toml` is 0.61.0 and that the minimum version of Python supported by setuptools 0.61.0 is Python 3.7.

So setup.py is still basically mandatory for projects supporting Python < 3.7.

For Python >= 3.7 it should be possible to use setuptools >= 0.61.0 in which case it should be possible to begin moving content from setup.py to pyproject.toml.

I think that pip has supported PEP 517 further back than this meaning that someone on an older system that has an older version of setuptools should be still be able to `pip install` etc with a `pyproject.toml` based project but it would be good to be able to say something about how far back that pip support goes. The relevance of pip is that for Python >= 3.7 pip would be able to install `setuptools >= 0.61.0` in the build environment regardless of how old the version of `setuptools` is outside of that environment so if someone is using e.g. distro-provided `pip` then what matters is the `pip` version rather than `setuptools`.

source/discussions/setup-py-deprecated.rst
henryiii: You can, but that encourages `python setup.py <command>`, and I think it's somewhat implicit that if you optionally can delete a file, you can also optionally keep it. But showing the placeholder file probably is fine - the one thing I have seen people do is remove everything from the file, but keep the empty file, which every much does _not_ work!

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
oscarbenjamin: I think it would be worth clearly stating that setup.cfg is not needed because its contents can definitely be moved to pyproject.toml (at least as far as setuptools-related content is concerned).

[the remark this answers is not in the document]
henryiii: I think I'd mention static / dynamic here. Something like:

In most cases, `setup.py` does not need to be modified at all.

But note that setuptools supports static configuration in `pyproject.toml` or `setup.cfg`. If a project can rely solely on static configuration, you can move all
configuration to one of these files and even remove `setup.py` entirely.

source/discussions/setup-py-deprecated.rst
henryiii: Would that make more sense in a followup? (Yes, I've seen that a lot too).

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+which triggers the generation of both distribution formats.
+Note that the build tool needs to be installed separately, though.
+
+
+Should ``setup.py`` be deleted?
+===============================
+
+Although the usage of ``setup.py`` as an executable script is deprecated,
+its usage as a configuration file for setuptools is absolutely fine.
+There is likely no modification whatsoever to do in ``setup.py``.
```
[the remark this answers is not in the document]
henryiii: Ahh, good point! Somewhat related to https://github.com/pypa/packaging.python.org/pull/1331#discussion_r1380583864 - though I think a mention then later filling out details on the alternative (which is a bit messy still) would be fine.

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
henryiii: If we get into the "what do I give up" side of things that was mentioned in the original discussion, then that's important. You can't use pyproject.toml and support Python 3.6, for example, but you can use setup.cfg and support pretty much anything.

Maybe this could by `pyproject.toml` only (including showing >=61 or >=62 as 61 was super buggy for PEP 621 config IIRC), and a followup could add a little discussion about supporting older users while following best practices.

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+
+But note that setuptools does not striclty require a ``setup.py`` file.
+This file can be replaced by a ``setup.cfg`` file or a ``pyproject.toml`` file.
+
+
+Is the ``pyproject.toml`` file mandatory?
+=========================================
+
+No, it is not necessary for a project to have a ``pyproject.toml`` file.
```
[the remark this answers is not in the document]
henryiii: I'd very very much say yes, that's what will fix those warnings and cause the "no setuptools in X" (X being some Python 3.12 environment, like GitHub Actions) problems to go away.

You could put `requires = ["setuptools>=62"]` for example, then avoid mentioning `setup.cfg` directly.

I'd make sure you clarify here, though, that the pyproject.toml is _not_ required to have configuration in it. I think most people confused (or scared) by this think they need to _move_ from `setup.py` to `pyproject.toml` (and I've had people tell me they would not add a `pyproject.toml` file until it could replace `setup.py`), and it's important to clarify that this is _not_ asking them to move, just add a very small file.

source/discussions/setup-py-deprecated.rst
petsuter: Is there a reason [python -m pip wheel](https://pip.pypa.io/en/stable/cli/pip_wheel/) is not mentioned? I imagine most people prefer using fewer tools, and using pip for everything would seem much easier.

source/discussions/setup-py-deprecated.rst
henryiii: `setup.cfg` + `pyproject.toml` works fine for older versions (pretty far back I think as long as you aren't doing edible installs in a very old version of Python+pip). That's the reason I mentioned above for dropping setup.cfg but adding a section (later) on older version support.

source/discussions/setup-py-deprecated.rst
henryiii: You can't use pip to build SDists though, and I'm not sure how happy the pip team is with keeping that around indefinitely. It also doesn't just build that wheel, but any dependencies that don't already have wheels also will build and get their wheels added to the wheelhouse (I could be misremembering the exact details here). Whereas `build` only gives you the one you want.

CAM-Gerlach: Don't we generally want to just recommend running ``python -m build`` to build both artifacts (outside of special cases) instead of running the two separate commands, one for each artifact type, like in the previous workflow? Its simpler for users as it only requires one shorter command vs. two longer ones, ensures both get built, and is more robust and efficient as the wheel actually builds from the sdist rather than separately from the source tree.

henryiii: That doesn't fit very nicely in the table though, unless we can merge two cells vertically (can we? I think we can with normal RST...)

henryiii: It also has the slight downside that people with broken SDists then open issues on the pypa/build repo because they don't know why their wheel is missing stuff when `pip` installs it just fine. :P

source/discussions/setup-py-deprecated.rst
pfmoore: PEP 517 support was added to pip in version 19.0.

pradyunsg: It can definitely live in a follow up.

source/discussions/setup-py-deprecated.rst
sinoroc: If we want we can collapse rows and columns (and suggest `python -m build` for both distributions at once). I am fine with doing that.

Maybe things have changed since then, maybe I misremember things... but I kind of recall reading some time ago that `pip wheel` was kind of not something *pip* maintainers want to keep around, maybe something along the lines that it got implemented kind of quick to fill in some hole in the space and now of course we can't get rid of it.

Maybe @pfmoore can help with that. Should we suggest something `python -m pip wheel . --no-deps` instead of `python setup.py bdist_wheel`?

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
abravalheri: PEP 517 does mention that frontendd should treat the lack of `pyproject.toml` as equivalent to the minimal `requires = ["setuptools"]` and `build-backend = "setuptools.build_meta:__legacy__"]`, right?

`pip` has not *fully* adopted that yet just as part of the transition period (if I am not wrong, it was supposed to be the case for 23.1, but it got further postponed).

source/discussions/setup-py-deprecated.rst
pfmoore: Given that pip can't build sdists, I think the consistency of using `build` for both wheels and sdists is better. The "fewer tools" argument IMO doesn't hold water if you have to use `build` for sdists anyway.

I like the simple model of "pip is for installing, build is for building".

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: The target audience is people who a `setup.py` and wondering what to do with it. I guess we should try to push them towards `pyproject.toml`. For more in-depth questions regardingfor example `setup.cfg` and older Python versions, I guess we should push them towards *setuptools* documentation. My feeling is that we can not cover everything here and should not try to. If we have too many details, it risks being hard to read and confusing.

What do you htink?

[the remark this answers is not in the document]
oscarbenjamin: > I guess we should try to push them towards `pyproject.toml`

More important than pushing someone towards something is helping them to understand the implications of the changes that they make. I expect that most people reading this will already realise that the intended end-state is to use pyproject.toml but as a package maintainer I would worry about not understanding what all of the implications are of the different changes. Something like this that cannot really be tested until after you push out a release is especially nerve-wracking. I would like to make the information here as clear as possible about the kinds of questions that a maintainer might have.

As a maintainer I would want to see clear actionable statements like:

> *All* content from setup.cfg can be moved *directly* to pyproject.toml so there is no need to have both files except when supporting `Python < 3.7`.

Also (I don't know if this is true):

> Putting any particular data in either setup.py, setup.cfg or pyproject.toml is precisely equivalent so it does not matter which file is used for any metadata/configuration if both files are present.

I really don't know how setuptools combines metadata from setup.py, setup.cfg and pyproject.toml. This is the sort of thing that would make me nervous about adding a pyproject.toml to a widely used project or messing with setup.py: what effect does even *adding* pyproject.toml have (regardless of its content)? Could it be that just adding pyproject.toml would break someone's build (I have heard it said that doing so breaks conda packages for example).

I can see though that this is potentially just better in the setuptools docs rather than here and could be scope-creep. If the point here is just to say that you can keep your setup.py or other setuptools configuration but should add pyproject.toml and should not use `python setup.py bdist_wheel` then maybe that is fine. Maybe it is out of scope here but it would still be good to be able to link to something in the setuptools docs that gives more detail about other potential changes.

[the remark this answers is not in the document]
henryiii: > All content from setup.cfg can be moved directly to pyproject.toml  ... precisely equivalent ... Could it be that just adding pyproject.toml would break someone's build ... 

Ohh, that's a can of worms. Mostly fine, mostly the same, most projects should be fine, those sorts of things. But words like perfectly, no.

Adding a `pyproject.toml`, even with no content or unrelated content triggers the isolated builds. Pip will default to isolated builds, and it will call the PEP 517 hook instead of running setup.py commands. If someone doesn't have wheel installed, they would get `setup.py install`, but that won't ever trigger with the build without wheel (and wheel will be pulled in via the build hooks). But if you've customized install (with you shouldn't), you might notice that. It's hard to get an environment with pip and setuptools these days, but words like "perfectly" require all details. This also affects the build speed, which I've had some people refuse to add pyproject.toml over. This cause major fights over black requiring pyproject.toml for configuration, since adding one changes the build. It's also why we really want everyone to add one, since changing this in pip and removing the old setup.py command code will do this to everyone without their opting in.

If you don't set `build-system.build-backend`, you'll get a legacy backend from setuptools. This is similar (inherits from) the modern backend, but it has some minor differences. The main one I know of is the new one doesn't let you import from the current folder by default. This is considered bad practice, and you can add the current directory to the path in `setup.py`, but it's a difference.

Note that even if you do the above, you don't break setuptools < 42 or pip < 19, since the setup.py (and static setup.cfg) is just fine. In fact, Pip 10+ supported isolated builds and the requires table (PEP 518). However, if you have modern TOML syntax (like `a.b`), you can break a narrow band of old pip versions that were reading the toml but not using a modern TOML library yet. I've had ~1 complaint on that.

*Note: At this point the packaging is modernized and changes can stop here. Setup.py or setup.cfg still continue to work.*

Then, putting configuration in pyproject.toml changes several setuptools configuration options to have reasonable defaults and requires 61+; old version will no longer build. This is great for setting up a new package, and is closer to what most packages want, but it's a change. This was the one chance to update the defaults based on years of usage, and it was used. The metadata is similar, but there are some small differences, like Home-page will now be empty (actually setuptools might try to set this from one of the URLs), Author-email will probably be empty (everything is in the normal field now, supports multiple authors better). AFACIT, the only people affected are those that use `pip show` which still displays those fields (and hides some other useful ones by default).

Once you have a wheel, a wheel is a wheel. There's no setup.py or pyproject.toml in it. You can just compare the wheels (just zip files) and see what's changed. That's how I do conversions. For pure Python, users never see it. Third party packagers (like conda) might break, but they have their own packaging, and can adapt (lots of other things break them too!). Usually they need to make sure they are using `pip install` instead of `setup.py` commands.

All of this is probably well out of scope for most packages. If you are doing something "custom", then you should probably know how to fix it, and you'll be _much_ better off manually moving now and fixing it now than if/when this is changed for everyone.

But nice reassuring words like "perfectly" can't really be used. If they could, then I bet pip would have just switched over some time ago.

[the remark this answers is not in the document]
oscarbenjamin: Thanks @henryiii for clearly demonstrating why a package maintainer *should* be nervous: there is no way that any of them will understand all of the implications that you have just listed and some of them clearly do matter.

While many of these things are out of scope I think that at least this should be described somewhere:

> Adding a `pyproject.toml`, even with no content or unrelated content triggers the isolated builds.

source/discussions/setup-py-deprecated.rst
abravalheri: Custom build steps are supported in the form of a "build sub-commands", e.g. overwriting `build_py`, `build_ext`, `bdist_wheel` or even creating a new one (like in this [example](https://github.com/pypa/setuptools/discussions/3762#discussioncomment-4658230)). They will be automatically called as part of the build.

Custom commands for "project/file system management" or creating custom types of artefacts are deprecated. Possible replacements are general "task runners" and tools like [nox](https://nox.thea.codes/en/stable/)/[tox](https://tox.wiki/), [Invoke](https://www.pyinvoke.org), [pydoit](https://pydoit.org), etc...

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
pfmoore: > Adding a pyproject.toml, even with no content or unrelated content triggers the isolated builds.

Changes the *default* to isolated builds. You can get the old behaviour with an option. Yes it should be mentioned because it’s surprising and people often don’t understand what’s going on, but it’s easy to get the old behaviour back and that should be emphasised.

Of course, people then complain that the old behaviour is annoying because they have to manage their build dependencies manually, but that’s a different issue :-)

sinoroc:
Thanks a lot for the comments. Hopefully tomorrow, I will take some time to re-read all comments carefully. My intention is to rewrite in order to tighten the focus on the "deprecation of `setup.py` as a CLI tool". And if no one has done it yet, I intend to start a different page for the "How to modernize a `setup.py` based project?" as there seems to be some kind of consensus that it is what we should do.

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
+..
+
+
```
webknjaz, line 3: This isn't necessary, right?
```suggestion
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+===========================
+Is ``setup.py`` deprecated?
+===========================
+
+No, ``setup.py`` is not deprecated,
```
webknjaz, line 8: There's a term in glossary. Why not link it?
```suggestion
No, :term:`setup.py` is not deprecated,
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+it is a valid configuration file for ``setuptools``
+that happens to be written in Python, instead of in *TOML* for example
+(a similar practice is used by other tools
+like *nox* and its ``nox.py`` configuration file,
+or *pytest* and ``conftest.py``).
+
+It is however deprecated to run ``python setup.py`` as a command line tool.
+
+This means for example that the following commands should NOT be run anymore:
```
webknjaz, line 17: "must" is stronger:
```suggestion
This means for example that the following commands **MUST NOT** be run anymore:
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+This file can be replaced by a ``setup.cfg`` file or a ``pyproject.toml`` file.
+
+
+Is the ``pyproject.toml`` file mandatory?
+=========================================
+
+No, it is not necessary for a project to have a ``pyproject.toml`` file.
+The standard fallback behavior for a :term:`build frontend <Build Frontend>`
+in absence of a ``pyproject.toml`` file and its ``[build-system]`` table
+is to assume that the :term:`build backend <Build Backend>` is setuptools.
```
webknjaz, line 93: I believe, that's not the case. `pip` has always used the presence of `pyproject.toml` as an opt-in for using PEP 517 and falls back to doing legacy invocations, if the PEP 517 mode is not activated. The fallback to `setuptools` happens only in the PEP 517 mode, when there's `pyproject.toml`, but it doesn't specify any backend.

@pradyunsg is this still the case?

source/discussions/setup-py-deprecated.rst
sinoroc: I revert back to the original version. Plan is to write a different page where we will likely discuss how one can remove `setup.py` from their project and link to it from here.

source/discussions/setup-py-deprecated.rst
sinoroc: Fixed. Thanks :)

source/discussions/setup-py-deprecated.rst
sinoroc: Fixed, thanks :)

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: I believe this should be addressed in a separate page "How to modernize a `setup.py` based project?", that I plan to start but will most definitely need help with.

source/discussions/setup-py-deprecated.rst
sinoroc: I believe there should be a section about the "additional custom commands" in this document, and I plan to address the "custom build steps as build sub-commands" in a separate document "How to modernize a `setup.py` based project?".

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,115 @@
...
+
+This means for example that the following commands should NOT be run anymore:
+
+* ``python setup.py install``
+* ``python setup.py develop``
+* ``python setup.py sdist``
+* ``python setup.py bdist_wheel``
+
+
+What commands should be use instead?
```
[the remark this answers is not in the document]
sinoroc: Fixed, thanks :)

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: @jeanas Not sure I follow 100%, I guess I understand where you are going with this. Do you have an example to offer (real or made up, but simple)?

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: I will remove this section. Out of scope for my vision for this document. Can be addressed in a separate document.

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: Huh? Are we sure about this? Things should behave fine without `pyproject.toml`, no?

If yes, then I will likely remove this section entirely in order to keep the document focused on the "Is `setup.py` deprecated?" question and the message we want to give "only the `python setup.py` command is deprecated". Modernizing the packaging for a `setup.py` project should be addressed in different page.

source/discussions/setup-py-deprecated.rst
sinoroc: If we only consider the scope of the "Is setup.py deprecated?" question and the message that we want to spread "only the `python setup.py` command is deprecated", do we need to add info about specific versions or not?

source/discussions/setup-py-deprecated.rst
sinoroc: Fixed, thanks :)

source/discussions/setup-py-deprecated.rst
sinoroc: Fixed, thanks :)

source/discussions/setup-py-deprecated.rst
sinoroc: If I understood correctly from other earlier comments above, currently the presence of `pyproject.toml` changes *pip*'s default behavior to use build isolation: https://github.com/pypa/packaging.python.org/pull/1331#discussion_r1381819095

source/discussions/setup-py-deprecated.rst
sinoroc: Fixed, thanks :)

source/discussions/setup-py-deprecated.rst
sinoroc: Thanks for comments.

I believe I found a good compromise. It now reads "_It is recommended to always build and publish both the source distribution and wheel of a project, which is what python -m build does. If necessary the --sdist and --wheel options can be used to generate only one or the other._"

source/discussions/setup-py-deprecated.rst
oscarbenjamin: Usually deprecations have some sort of version connected to them like "distutils was deprecated in Python 3.10 and removed in Python 3.12". It is a bit tricky here to say what versions should be mentioned though (Python version, pip version, setuptools version, ...).

sinoroc: @oscarbenjamin Right. I think @abravalheri might be in the best position to tell us starting from which version of *setuptools* the usage of `setup.py` as command line tool has been deprecated.

For the Python version, my impression is that it is irrelevant.

For the *pip* version, my understanding is that the behavior has stayed relatively consistent throughout pip's releases, am I wrong?

Maybe some milestones (dates and/or version strings) we could add:

* when was *build* declared ready for mainstream usage
* when did the behaviors of *setup.py install* and *pip install* started diverging? same for *develop* and *editable*
* when did *setuptools* declare its command line usage deprecated

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: https://python-packaging-user-guide--1341.org.readthedocs.build/en/1341/discussions/modernize-setup-py-project/#how-to-handle-packaging-metadata

oscarbenjamin:
Assuming that this goes with #1341 I think that what is here is fine. I would just remove the section about pyproject.toml from this page so that it is purely focused on what is deprecated and what the direct alternatives are supposed to be.

sinoroc:
So https://github.com/pypa/packaging.python.org/pull/1341 has been started as a companion for this PR. So some bits can be moved to that other page.

Here are some of the latest updates:

* added note that *setuptools* iself is not deprecated either (obviously)
* added section about custom setup.py commands being deprecated, and list of tools that can be used as replacement: nox/tox, make, pyinvoke, and more
* added section about custom build steps NOT being deprecated.
* removed some things that are better expressed in that other page, I can not put links yet, but once both are merged we can add the links between those 2 pages
* kept the note that ``pyproject.toml`` is not necessary but STRONGLY recommended with a minimal example of `build-system` table (later we can link to the other page)

Of course, everything can be rephrased, reworded, reordered, clarified, but I guess I might be running out of steam (already? :D ), so maybe it is better to get this merged so that others can take over. There is clearly a bunch of things that have been commented here, but I either do not understand or do not know how to phrase.

source/discussions/setup-py-deprecated.rst
sinoroc: Note is added in the other page: https://python-packaging-user-guide--1341.org.readthedocs.build/en/1341/discussions/modernize-setup-py-project/#what-is-the-build-isolation-feature

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,134 @@
...
+to a task runner tool or any other similar tool.
+Some examples of such tools are:
+chuy, make, nox/tox, pydoit, pyinvoke, taskipy, and thx.
+
+
+Should ``setup.py`` be deleted?
+===============================
+
+Although the usage of ``setup.py`` as an executable script is deprecated,
+its usage as a configuration file for setuptools is absolutely fine.
```
cmeyer, line 91: I don't think this answers the question - but maybe a better (or additional) question is essentially "Is 'setup.py' still required if all configuration can be specified in the 'pyproject.toml' file?"

We have been using `setup.cfg` but still have minimal `setup.py` that look like this:

```
from setuptools import setup

if __name__ == "__main__":
    setup()
```

I'm guessing these were required sometime in the past for running `setup.py` directly; but I'm also guessing they're (a) not required with `setup.cfg` and (b) also not required with `pyproject.toml`, where those are used with `install` or `build` and where either of those do not run any custom Python steps.

source/discussions/setup-py-deprecated.rst
sinoroc: Minimal `setup.py` stub is in the other page in the other PR: https://python-packaging-user-guide--1341.org.readthedocs.build/en/1341/discussions/modernize-setup-py-project/#what-if-something-that-can-not-be-changed-expects-a-setup-py-file

source/discussions/setup-py-deprecated.rst
sinoroc: @cmeyer thanks for the comment :)

The target audience for this document is users who saw the deprecation notice for `python setup.py install` and are now wondering what it is all about and if they should:

* stop using *setuptools*
* stop using `setup.py` and even possibly delete it completely

The main focus is really on things that users should act on now. For example because `python setup.py install` definitely is broken right now and no one should use it anymore.

We of course want to also nudge people towards using the new standard and modern ways, but this is less a pressing matter, a different level of urgency, so we want to handle this in a separate document.

> maybe a better (or additional) question is essentially "Is 'setup.py' still required if all configuration can be specified in the 'pyproject.toml' file?"

I believe this is better addressed in https://github.com/pypa/packaging.python.org/pull/1341 where we talk about how to migrate a project based on legacy setup.py towards standards and more modern practices.

sinoroc:
> (Perhaps you could click "resolve" on the threads you addressed so it's easier to see what remains?)

Right, I never know what the etiquette is around this. Should I "resolve" the thread or should the person who opened the thread resolve it when they are satisfied with the outcome?

I did resolve some where it seemed obvious to me that is was resolved (like for typos for example).

oscarbenjamin:
> Right, I never know what the etiquette is around this. Should I "resolve" the thread or should the person who opened the thread resolve it when they are satisfied with the outcome?

I would say that if you are confident that the other person would consider it resolved then mark it as resolved. For example if the other person says "it should be X not Y" and you say "Oh, good point, I've changed it to X" then go ahead and mark it resolved. On the other hand if the your response is "no, you are wrong it should definitely be Y and not X" then you should leave it unresolved and let the other person respond.

source/discussions/setup-py-deprecated.rst
pfmoore: Works for me.

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
pfmoore: I'd go with "it's not technically necessary yet, but it's strongly recommended to have at least a minimal `pyproject.toml`". But I'm also fine with leaving this as it stands and debating the nuances in a follow-up PR.

source/discussions/setup-py-deprecated.rst
pfmoore: IMO, the statement here is sufficient for the target audience. Let's not get bogged down in details.

pfmoore (approved):

pfmoore:
> I would say that if you are confident that the other person would consider it resolved then mark it as resolved.

I think people have different opinions on this. Unless the interaction is of the form "I think you should do X"/"OK, I did that", I tend to prefer to leave it to the person who made the original comment to confrm that they are happy by being the one to mark the conversation as resolved. And I prefer it if others do that when I'm the commenter. But ultimately, I think there's no standard etiquette here - do what you feel is best.

> Of course, everything can be rephrased, reworded, reordered, clarified, but I guess I might be running out of steam (already? :D ), so maybe it is better to get this merged so that others can take over.

+1 - people can make inmprovements in follow-up PRs. Let's not burn people out by making it a huge task to make an incremental improvement.

source/discussions/setup-py-deprecated.rst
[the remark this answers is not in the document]
sinoroc: > I'd go with "it's not technically necessary yet, but it's strongly recommended to have at least a minimal `pyproject.toml`".

Sounds good, done

source/discussions/setup-py-deprecated.rst
cmeyer: The larger point may be better addressed elsewhere, but I still don't think the question "Should ``setup.py`` be deleted?" is answered here, even to the particular audience you indicate above.

The answer "it is fine to be used as configuration" doesn't answer the question.

It still leaves me with questions. Is it necessary to have setup.py? _Can_ it be deleted? Under what conditions _should_ it be deleted?

The answer only says "if you want to use it as a configuration file, it doesn't NEED to be deleted."

oscarbenjamin: The section here has a note:
```
.. todo::

    Link to a "How to modernize a setup.py based project?"
```
In other words that means to link to the page that would be added in #1341.  That other page would be a better place to address what to do with the setup.py file rather than this page that is focused on what is (or is not) deprecated.

pfmoore: > Is it necessary to have setup.py? Can it be deleted? Under what conditions should it be deleted?

No, yes, and "it depends". As @oscarbenjamin says, "it depends" is the subject of #1341, not here.

The question here is really more "people are telling me I should delete `setup.py` - are they right?" To which the answer is "no". You *might* want to delete it, depending on what you change, but the people telling you you "should" do so are over-simplifying. But if we keep trying to address all of the nuances here, we'll end up with a document that's so complex and confusing, that it *adds* to the problem rather than helping.

I appreciate your point that the text here still leaves you with questions, but IMO that's fine. The object was never to answer *all* your questions, just to leave you better informed than you were when you got here. And (again, IMO) the text we have now does that job just fine.

If, once this PR is merged, you still feel there's improvements to be made, please create a new PR with your suggested wording. It'll be much easier to have a focused discussion on that PR rather than trying to cover everything here.

oscarbenjamin (approved):

webknjaz:
> > (Perhaps you could click "resolve" on the threads you addressed so it's easier to see what remains?)
> 
> Right, I never know what the etiquette is around this. Should I "resolve" the thread or should the person who opened the thread resolve it when they are satisfied with the outcome?

@sinoroc so the guidance I follow and recommend other people do is:
1) Generally, the thread author should be the one to resolve it. This is because other people may not interpret the request, and ultimately only the author knows what they meant. Also, the author would be able to accurately judge whether the reaction (change in code or response comment) is adequate and solves their concern.
2) If a reviewer suggested to change something and the PR author did that, it's a good tone to add a comment to the thread stating this and linking the commit. Note that this might not work well if the PR submitter exercises the force-push strategy, which is one I've seen some project ask that only new commits are added, making the incremental updates discoverable more easily.
3) Clicking `Resolve` does not generate notifications (like emails). When this is done, the thread author would easily lose the reference to their thread, and it's often hard to find, once it's collapsed. For this reason, I typically recommend posting a comment saying “Resolving this thread.” and maybe some reasoning, _and only then_, actually clicking `Resolve`. This increases transparency (it's not obvious when things got “resolved” otherwise) and allows for people to find the way back to those threads, using email notifications which contain direct links…

webknjaz:
> > I would say that if you are confident that the other person would consider it resolved then mark it as resolved.
> 
> I think people have different opinions on this. Unless the interaction is of the form "I think you should do X"/"OK, I did that", I tend to prefer to leave it to the person who made the original comment to confrm that they are happy by being the one to mark the conversation as resolved. And I prefer it if others do that when I'm the commenter. But ultimately, I think there's no standard etiquette here - do what you feel is best.

+1

The other day, somebody introduced me to https://conventionalcomments.org in another PyPA repo, I think. It seems like this is something that could improve the communication even more.

> > Of course, everything can be rephrased, reworded, reordered, clarified, but I guess I might be running out of steam (already? :D ), so maybe it is better to get this merged so that others can take over.
> 
> +1 - people can make improvements in follow-up PRs. Let's not burn people out by making it a huge task to make an incremental improvement.

Yeah, a few days ago I merged a PR (https://github.com/jazzband/pip-tools/pull/1681) that took 13 months to complete, with hundreds of review comments (my fault partially). I'm pretty sure it was exhausting to all the parties involved.

I'll try to take a prompt look at this PR since I haven't had a chance to review properly and will try to get it merged unless there's a noticeable problem.

webknjaz:
Aha… So the CI is failing with:
```console
/home/runner/work/packaging.python.org/packaging.python.org/source/discussions/setup-py-deprecated.rst:129: WARNING: term not in glossary: 'Build Frontend'
/home/runner/work/packaging.python.org/packaging.python.org/source/discussions/setup-py-deprecated.rst:129: WARNING: term not in glossary: 'Build Backend'
```
_(https://github.com/pypa/packaging.python.org/actions/runs/6764239357/job/18382296998?pr=1331#step:5:62)_

webknjaz:
So this should be merged together with #1329 and it'll be fine...

webknjaz (approved):
@sinoroc I'm going to merge this, but I saw some punctuation and grammar mistakes in the text. Would you mind feeding it to LanguageTool and/or Grammarly and sending the fixes as a follow-up?

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,146 @@
...
+One way to look at it is that setuptools' scope
+has now been reduced to the role of a build backend.
+
+
+Where to read more about this?
+==============================
+
+* https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
+
+* https://setuptools.pypa.io/en/latest/deprecated/commands.html
```
webknjaz, line 146: FYI, this is linkable through intersphinx with
```rst
:doc:`deprecated/commands`
```
See: https://webknjaz.github.io/intersphinx-untangled/setuptools.rtfd.io/.

Could be another thing to update in a follow-up.

EpicWink:
Perhaps you also want to change list-item 2 of [the doc for setup.py](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#setup-py)

> [`setup.py` is] the command line interface for running various commands that relate to packaging tasks. To get a listing of available commands, run python3 setup.py --help-commands.

webknjaz:
@EpicWink that sounds like another follow-up. This PR is about a "discussion" document, while your suggestion concerns a "guide".

source/discussions/setup-py-deprecated.rst
webknjaz, line 1: Another suggestion for the future: this page could benefit from a Local TOC.

sinoroc: Not now that we have switched to furo, right?

source/discussions/setup-py-deprecated.rst
sinoroc: @webknjaz Thanks, done

sinoroc:
@webknjaz 

> I saw some punctuation and grammar mistakes in the text. Would you mind feeding it to LanguageTool and/or Grammarly and sending the fixes as a follow-up?

Thanks for the suggestion, done

Also rebased on main so that we have working links to the build backend and frontend glossary definitions.

And fixed the link to setuptools page.

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,149 @@
...
+that happens to be written in Python, instead of in *TOML* for example
+(a similar practice is used by other tools
+like *nox* and its :file:`nox.py` configuration file,
+or *pytest* and :file:`conftest.py`).
+
+And of course *setuptools* itself is not deprecated either.
+
+It is however deprecated to run ``python setup.py`` as a command line tool.
+
+This means for example that the following commands **MUST NOT** be run anymore:
```
pradyunsg, line 19:
```suggestion
This means for example that the following commands **MUST NOT** be run directly anymore:
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -0,0 +1,149 @@
...
+
+* ``python setup.py install``
+* ``python setup.py develop``
+* ``python setup.py sdist``
+* ``python setup.py bdist_wheel``
+
+
```
pradyunsg, line 26:
```suggestion

These commands may be executed by `setuptools` and `pip` internally,
typically to ensure backwards compatibility in certain configurations.

```

Let's mention that the default tooling can (and does) use these "under the hood" because... well... we don't really hide that fact. :)

pradyunsg (commented):
A couple more minor things.

source/discussions/setup-py-deprecated.rst
oscarbenjamin: I'm not sure why these need to be mentioned. For completeness it makes sense but this document is hardly complete in terms of describing all possibilities. The fact that the commands might be used is potentially relevant if considering removing the setup.py altogether but that is not the topic of this guide,

oscarbenjamin: It looks like @pradyunsg intended to merge this via auto-merge but the merge failed because of these unresolved conversations...

I haven't seen a setup like this that blocks merge unless conversations are resolved. I'm not sure how I feel about that: sometimes something should be merged even though some discussions are not entirely resolved. Future PRs might resolve them but in the meantime it is helpful to be able to see what those unresolved things are.

sinoroc: @oscarbenjamin Indeed, resolving the last 2 remaining unresolved threads (including this one and one closely related) seems to have triggered the "merge queue". I should read up on this merge queue.

oscarbenjamin:
Nice work @sinoroc !

PR #1335, opened 2023-11-03
Rework the landing page to reflect diataxis principles

willingc:
This PR is focused on the end user of the documentation. It edits the landing page to more closely reflect the diataxis structure of the documentation. The intent of the PR is to improve the experience of readers and help guide them to information that will be most valuable for their use case.

### DIFF
--- a/source/index.rst
+++ b/source/index.rst
@@ -32,34 +32,48 @@ happily accept any :doc:`contributions and feedback <contribute>`. 😊
 .. _GitHub: https://github.com/pypa/packaging.python.org
 
 
-Get started
-===========
+Overview
+========
+
+* To get an overview of packaging options for Python libraries and
+  applications, see the :doc:`Overview of Python Packaging <overview>`
+
+Tutorials
+=========
 
 Essential tools and concepts for working within the Python
-development ecosystem are covered in our :doc:`tutorials/index` section:
+development ecosystem are covered in our :doc:`tutorials/index` section.
 
-* To get an overview of the flow used to publish your code, see the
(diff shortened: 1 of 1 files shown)


henryiii:
I like the idea, though I should add the warning (here, not in the landing page / PR) that some of the pages in guides are extremely (5 or 10+ years) out of date. I think seriously rewriting them, cutting as needed, would really improve the user experience on the site. (Compare to the guides on https://learn.scientific-python.org/development/guides for an example of something more up to date.)

Once that is done, then linking a few of the most useful guide pages, similar to the tutorial section, might be in order.

BrenBarn:
@henryiii What is the relative utility of having a guide that is 10 years out of date vs. just removing it entirely?  Does it vary from one guide to another?  Or perhaps there could be some kind of "legacy" subsection where out-of-date guides are clearly separated from the rest?  I understand that actually updating the guides would require a lot of work, but if we at least know which ones are the most egregiously outdated, then maybe we can segregate them under a sub-index of some sort, so that a casual user visiting the page will know they're different from the more up-to-date parts.

willingc:
@henryiii This PR is only to make an incremental improvement as a stopgap to help new people landing on the page. There definitely needs to be a longer sustained effort to address out of date pages. I'm happy to help iteratively on some of those in separate PRs.

willingc:
We may wish to create an epilog or prolog in a separate PR to notify readers that sections may be out of date.

pradyunsg (approved):

pradyunsg:
(did a release via the web UI, to ensure that the merge is clean)

PR #1337, opened 2023-11-03
Import PEP 518 specification

jeanas:
This relates to #955 and #1093.

I am aware of the previous attempt in #1111. This PR has smaller commits and does less. (The integration could be much better, but let's please not be perfectionistic and get the text there first; we can improve in subsequent PRs.)

@pradyunsg @pfmoore @brettcannon @CAM-Gerlach


source/specifications/declaring-build-dependencies.rst
```diff
@@ -5,6 +5,97 @@
...
+specification <version-specifiers>`.
+
+An example ``build-system`` table for a project built with
+``setuptools`` is:
+
+.. code-block:: toml
+
+   [build-system]
+   # Minimum requirements for the build system to execute.
+   requires = ["setuptools", "wheel"]
```
sinoroc, line 35: If I am not mistaken `wheel` is never(?) necessary here.

sinoroc: Ah, right, understood.

pfmoore: You're making a number of other minor changes to the wording of the spec, I think updating this example is just as reasonable.

source/specifications/declaring-build-dependencies.rst
```diff
@@ -5,6 +5,97 @@
...
+
+Tables not defined by PyPA specifications are reserved for future use.
+
+
+build-system table
+------------------
+
+The ``[build-system]`` table is used to store build-related data.
+Currently,  only one key of the table is be valid and is mandatory
+for the table: ``requires``. This key must have a value of a list
```
pfmoore, line 23: This ignores the `build-system` keys added by PEP 517. Maybe go back to the word "Initially", used in the PEP, rather than "Currently"?

source/specifications/declaring-build-dependencies.rst
```diff
@@ -5,6 +5,97 @@
...
 
-``pyproject.toml`` is a build system independent file format defined in :pep:`518`
-that projects may provide in order to declare any Python level dependencies that
-must be installed in order to run the project's build system successfully.
+The ``pyproject.toml`` file is written in `TOML <https://toml.io>`_.
+Among other metadata (such as :ref:`project metadata <declaring-project-metadata>`),
+it declares any Python level dependencies that must be installed in order to
+run the project's build system successfully.
+
+.. TODO: move this sentence elsewhere
```
pfmoore, line 13: Rather than trying to find other places for the two items marked with TODO, maybe rename this specification as something like "Structure of `pyproject.toml`"?

And if you're concerned about PEPs 517 and 621 meaning that this isn't actually the "structure" in all its glory, maybe just add a note at the top saying something like "This is the initial definition of `pyproject.toml`, and is extended in later specifications" (maybe adding a list).

Someone motivated to do so can then reorganise the document structure for the various `pyproject.toml` related specs later.

source/specifications/declaring-build-dependencies.rst
pfmoore: Aargh. You're reacting quicker than I can type 🙂

Having just checked, pip uses `requires = ["setuptools>=40.8.0", "wheel"]` as its default. And while there's some version of setuptools that allows `requires = ["setuptools>=x.y.z"]` (because setuptools adds the wheel dependency itself) I don't know what that version is. @abravalheri would probably know. I'd be reluctant to use *just* setuptools with no version limit, even though I can't imagine a practical situation where a really old version would be the only thing that could be picked up by a bare `setuptools`.

"Build tools are expected to use the example configuration file above" is, IMO, worded as a suggestion rather than as a requirement, so I don't think we need to stress about whether such a change is backward compatible, what matters is that it's a good *example*.

OK, I think I've said my piece. You can react now 😉

source/specifications/declaring-build-dependencies.rst
pfmoore: I'd keep the two separate. You're right that smaller changes are easier to review. Ignore my previous comment.

pfmoore:
> By the way, I have a question: do we want to keep the JSON-schema structure?

I have no personal use for it. But some people might. I'd keep it, but don't worry about updating it. If it bothers you, add a note saying it reflectes the initial definition from PEP 518 and has not been updated to incorporate later specifications, and leave it at that.

A direct copy of PEP 518 with flaws that can be fixed later by interested parties is better than a PR that tries to describe the current situation accurately but doesn't get merged 🙂

source/specifications/declaring-build-dependencies.rst
pfmoore: Works for me.

webknjaz: @jeanas `wheel` is never necessary, indeed. `setuptools` depends on it dynamically, only for building wheels, through the `get_requires_for_build_wheel` PEP 517 hook since it was first implemented: https://github.com/pypa/setuptools/blob/2384d91/setuptools/build_meta.py#L355. It's been the case for the past 5 years: https://github.com/pypa/setuptools/commit/2fabfd3. I believe I contributed to the confusion by documenting `requires = ["setuptools", "wheel"]` in setuptools' own docs at some point, and many people copied that over into their `pyproject.toml` files.
Later, when I realized the mistake, I fixed it in said document almost 2 years ago, but people still have this dependency specified — there's a number of backreferences to that docs PR (https://github.com/pypa/setuptools/pull/3056) as folks are dropping `wheel`.

Specifying `requires = ["setuptools", "wheel"]` means that `wheel` will be present in the PEP 517 build env even when unnecessary, like when building sdist but not wheel.

So I think it's beneficial to delete `wheel` to avoid contributing to the confusion.

brettcannon (approved):

pradyunsg (approved):

PR #1338, opened 2023-11-04
Update how to guide to install using pip and venv

willingc:
This PR makes the following updates:
- remove all references to Python 2 and virtualenv and replace with a note at the top of the document
- change `env` environment name to `.venv` which is more commonly used and less likely to be confused with environment variables
- create 4 high level sections (pip, venv, install packages with pip, requirements files)

I view this PR as one of iteration of future iterations to improve individual documents in the guide.

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -338,7 +338,7 @@ installed package without needing to re-install:
 
 
 Install from version control systems
-------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 pip can install packages directly from their version control system. For
 example, you can install directly from a git repository:
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -2,7 +2,7 @@ Installing packages using pip and venv
 ======================================
 
 This guide discusses how to install packages using :ref:`pip` and
-the standard library's virtual environment manager :ref:`venv`. The guide
+the standard library's virtual environment :ref:`venv`. The guide
 covers how to:
 
 * Install and update pip
@@ -83,8 +83,8 @@ since packages can be installed confidently and will not interfere with the
 other project environments.
 
 .. tip::
-   It is always recommended to use a virtual environment while developing Python
-   applications.
+   It is recommended to use a virtual environment when working with third
+   party packages.
 
 To create a virtual environment, go to your project's directory and run
 ``venv``.
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -1,12 +1,12 @@
-Installing packages using pip and venv
-======================================
+Install packages in a virtual environment using pip and venv
+============================================================
 
-This guide discusses how to install packages using :ref:`pip` and
-the standard library's virtual environment :ref:`venv`. The guide
-covers how to:
+This guide discusses how to create and activate a virtual environment using
+the standard library's virtual environment :ref:`venv` and install packages.
+The guide covers how to:
 
-* Install and update pip
-* Create and use a virtual environment
+* Create and activate a virtual environment
+* Prepare pip
 * Install packages into a virtual environment using the ``pip`` command
 * Use and create a requirements file
 
@@ -21,52 +21,11 @@ covers how to:
     import modules in your Python source code.
 
 
-Install and update pip
-----------------------
-
-:ref:`pip` is the reference Python package manager.
-It's used to install and update packages.
-Make sure you have the latest version of pip installed.
-
-
-.. tab:: Unix/macOS
-
-    Debian and most other distributions include a `python-pip`_ package; if you
-    want to use the Linux distribution-provided versions of pip, see
-    :doc:`/guides/installing-using-linux-tools`.
-
-    You can also install pip yourself to ensure you have the latest version. It's
-    recommended to use the system pip to bootstrap a user installation of pip:
-
-    .. code-block:: bash
-
-        python3 -m pip install --user --upgrade pip
-        python3 -m pip --version
-
-    Afterwards, you should have the latest version of pip installed in your
-    user site:
-
-    .. code-block:: text
-
-        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
-
-    .. _python-pip: https://packages.debian.org/stable/python/python3-pip
-
-.. tab:: Windows
-
-    The Python installers for Windows include pip. You can make sure that pip is
-    up-to-date by running:
-
-    .. code-block:: bat
-
-        py -m pip install --upgrade pip
-        py -m pip --version
-
-    Afterwards, you should have the latest version of pip:
-
-    .. code-block:: text
-
-        pip 21.1.3 from c:\python39\lib\site-packages (Python 3.9.4)
+.. important::
+    This guide has the prerequisite that an official Python version from
+    <https://www.python.org/downloads/>. If you are using your operating
+    system's package manager to install Python, please ensure that Python is
+    installed before proceeding with these steps.
 
 
 Create and Use Virtual Environments
@@ -80,7 +39,7 @@ different projects. It creates a "virtual" isolated Python installation. When
 you switch projects, you can create a new virtual environment which is isolated
 from other virtual environments. You benefit from the virtual environment
 since packages can be installed confidently and will not interfere with the
-other project environments.
+another project's environment.
 
 .. tip::
    It is recommended to use a virtual environment when working with third
@@ -183,6 +142,72 @@ Reactivate a virtual environment
 If you want to reactivate an existing virtual environment, follow the same
 instructions about activating a virtual environment. There's no need to create
 a new virtual environment.
+
+
+Prepare pip
+-----------
+
+:ref:`pip` is the reference Python package manager.
+It's used to install and update packages into a virtual environment.
+
+
+.. tab:: Unix
+
+    Debian and most other distributions include a `python-pip`_ package; if you
+    want to use the Linux distribution-provided versions of pip, see
+    :doc:`/guides/installing-using-linux-tools`.
+
+    You can also install pip yourself to ensure you have the latest version. It's
+    recommended to use the system pip to bootstrap a user installation of pip:
+
+    .. code-block:: bash
+
+        python3 -m pip install --user --upgrade pip
+        python3 -m pip --version
+
+    Afterwards, you should have the latest version of pip installed in your
+    user site:
+
+    .. code-block:: text
+
+        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
+
+    .. _python-pip: https://packages.debian.org/stable/python/python3-pip
+
+.. tab:: macOS
+
+    The Python installers for macOS include pip. You can make sure that pip is
+    up-to-date by running:
+
+    .. code-block:: bash
+
+        python3 -m pip install --upgrade pip
+        python3 -m pip --version
+
+    Afterwards, you should have the latest version of pip installed in your
+    user site:
+
+    .. code-block:: text
+
+        pip 21.1.3 from $HOME/.local/lib/python3.9/site-packages (python 3.9)
+
+    .. _python-pip: https://packages.debian.org/stable/python/python3-pip
+
+.. tab:: Windows
+
+    The Python installers for Windows include pip. You can make sure that pip is
+    up-to-date by running:
+
+    .. code-block:: bat
+
+        py -m pip install --upgrade pip
+        py -m pip --version
+
+    Afterwards, you should have the latest version of pip:
+
+    .. code-block:: text
+
+        pip 21.1.3 from c:\python39\lib\site-packages (Python 3.9.4)
 
 
 Install packages using pip
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -2,7 +2,7 @@ Install packages in a virtual environment using pip and venv
 ============================================================
 
 This guide discusses how to create and activate a virtual environment using
-the standard library's virtual environment :ref:`venv` and install packages.
+the standard library's virtual environment tool :ref:`venv` and install packages.
 The guide covers how to:
 
 * Create and activate a virtual environment
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -1,22 +1,32 @@
-Installing packages using pip and virtual environments
-======================================================
+Installing packages using pip and venv
+======================================
 
 This guide discusses how to install packages using :ref:`pip` and
-a virtual environment manager: either :ref:`venv` for Python 3 or :ref:`virtualenv`
-for Python 2. These are the lowest-level tools for managing Python
-packages and are recommended if higher-level tools do not suit your needs.
+the standard library's virtual environment manager :ref:`venv`. The guide
+covers how to:
 
-.. note:: This doc uses the term **package** to refer to a
-    :term:`Distribution Package`  which is different from an :term:`Import
-    Package` that which is used to import modules in your Python source code.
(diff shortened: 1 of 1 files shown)


source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,22 +1,32 @@
-Installing packages using pip and virtual environments
-======================================================
+Installing packages using pip and venv
+======================================
 
 This guide discusses how to install packages using :ref:`pip` and
-a virtual environment manager: either :ref:`venv` for Python 3 or :ref:`virtualenv`
-for Python 2. These are the lowest-level tools for managing Python
-packages and are recommended if higher-level tools do not suit your needs.
+the standard library's virtual environment manager :ref:`venv`. The guide
```
sinoroc, line 5: Should we consider avoiding the term "manager" here? The way I see it `venv` can only "create" environments (not much else, there is some kind of update feature that I believe is rarely used and barely documented). On the other hand there are tools that keep track of virtual environments (store them all in one "managed" location), can delete them, can rename them or move them, freeze the list of dependencies it contains and so on.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -59,94 +69,68 @@ installed.
...
-you to manage separate package installations for
-different projects. They essentially allow you to create a "virtual" isolated
-Python installation and install packages into that virtual installation. When
-you switch projects, you can simply create a new virtual environment and not
-have to worry about breaking the packages installed in the other environments.
-It is always recommended to use a virtual environment while developing Python
-applications.
+.. tip::
+   It is always recommended to use a virtual environment while developing Python
+   applications.
```
sinoroc, line 87: I would say, virtual environments are only necessary when installing 3rd party packages with *pip* or other, from _PyPI_ or other). Doing Python development with standard library only is perfectly fine without virtual environment.

sinoroc (commented):
Great update!

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: Thanks for the review. I've updated bassed on your suggestions.

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: Updated. Thx.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -59,94 +69,68 @@ installed.
...
-
-
-:ref:`virtualenv` is used to manage Python packages for different projects.
-Using virtualenv allows you to avoid installing Python packages globally
-which could break system tools or other projects. You can install virtualenv
-using pip.
+Create a new virtual environment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+:ref:`venv` (for Python 3) allows you to manage separate package installations for
```
Kwpolska, line 78: Debian does not have working `venv` in the default install, it might be worth mentioning that you need to install either `python3-venv` or `python3-full` (which also includes idle and tkinter) on Debian and Ubuntu. (Virtually all non-Debian-related Linux distros do it correctly and include venv in the default install.)

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,22 +1,32 @@
...
+.. note:: This guide uses the term **package** to refer to a
+    :term:`Distribution Package`, which commonly is installed from an external
+    host. This differs from the term :term:`Import Package` which refers to
+    import modules in your Python source code.
+
+
+Install and update pip
+----------------------
+
+:ref:`pip` is the reference Python package manager.
```
Kwpolska, line 27: In the previous version of the guide, having pip installed system-wide was used to install venv. Right now, I can’t really see any need for it. Perhaps this guide could focus only on installing things in venvs instead of using the system pip (which may be `EXTERNALLY-MANAGED`)?

source/guides/installing-using-pip-and-virtual-environments.rst
sinoroc: Maybe anecdotal and complex to bring in this page, but personally I make sure to only ever install directly `python3-venv` (not `python3-full` or anything like that) so that I do not have any gloabl *pip* at all. The only way to have *pip* is from inside a virtual environment.

_[So yes, for me at least this split of Python into multiple parts on Debian and Ubuntu has this one advantage.]_

petsuter:
> Before you can start installing or using packages in your virtual environment you’ll need to activate it. 

> While a virtual environment is activated, pip will install packages into that specific environment.

> Activating a virtual environment will put the virtual environment-specific ``python`` and ``pip`` executables into your
shell's ``PATH``.

Could these parts be updated as well? They are incorrect and actively misleading, leading to a lot of misunderstandings. A virtual environment can be used without activation. It depends on how / which pip is used. It would be fantastic if this guide could explain that in slightly more detail: Explicitly explain that activating is only (one way of) putting the the virtual environment-specific ``python`` and ``pip`` executables into the current shell's ``PATH``. Specifying that path explicitly is also possible and often easier, preferred, clearer, more robust. The leaky abstraction "activation" and the lack of such an explanation is mystifying things unnecessarily, and leads to many problems and much confusion.

* ["Activation sucks as a method of switching environments"](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/422) -- pfmoore
* ["I use venvs extensively, but I never “activate” them."](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/181) -- fungi
* ["Why do we need to activate envs? One can always just env/bin/pip install whatever"](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/5) -- bernatgabor
* ["I have never once activated a virtual environment. I always use /full/path/to/venv/bin/python -m pip install …. And that’s what I teach to groups I mentor."](https://discuss.python.org/t/pip-plans-to-introduce-an-alternative-zipapp-deployment-method/17431/47) -- ericvsmith
* ...

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: @Kwpolska I think that is a good suggestion and one that I was thinking about when I edited this. I'm going to make this change and focus on installing packages in a `venv`

willingc:
> It seems that this page:
> 
> https://packaging.python.org/en/latest/tutorials/installing-packages/#requirements-for-installing-packages
> 
> and this page:
> 
> https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/
> 
> are mostly duplicates. Right?

That is true. One is a tutorial and this is the guide. According to Diataxis doc process, the guide is focused on one thing. I'm not updating the tutorial now but it can incorporate or refer to this after being merged.

willingc:
> > Before you can start installing or using packages in your virtual environment you’ll need to activate it.
> 
> > While a virtual environment is activated, pip will install packages into that specific environment.
> 
> > Activating a virtual environment will put the virtual environment-specific `python` and `pip` executables into your
> > shell's `PATH`.
> 
> Could these parts be updated as well? They are incorrect and actively misleading, leading to a lot of misunderstandings. A virtual environment can be used without activation. It depends on how / which pip is used. It would be fantastic if this guide could explain that in slightly more detail: Explicitly explain that activating is only (one way of) putting the the virtual environment-specific `python` and `pip` executables into the current shell's `PATH`. Specifying that path explicitly is also possible and often easier, preferred, clearer, more robust. The leaky abstraction "activation" and the lack of such an explanation is mystifying things unnecessarily, and leads to many problems and much confusion.
> 
> * ["Activation sucks as a method of switching environments"](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/422) -- pfmoore
> * ["I use venvs extensively, but I never “activate” them."](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/181) -- fungi
> * ["Why do we need to activate envs? One can always just env/bin/pip install whatever"](https://discuss.python.org/t/pep-582-python-local-packages-directory/963/5) -- bernatgabor
> * ["I have never once activated a virtual environment. I always use /full/path/to/venv/bin/python -m pip install …. And that’s what I teach to groups I mentor."](https://discuss.python.org/t/pip-plans-to-introduce-an-alternative-zipapp-deployment-method/17431/47) -- ericvsmith
> * ...

Let me think about this a bit more. I think what you raise is interesting but perhaps belongs in a new Discussion/Explanation document instead of a "How To" guide. https://diataxis.fr/

I think that taking @Kwpolska's suggestion to rename the guide to "Install packages into a virtual environment using pip and venv" will help.

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: PR updated.

petsuter:
I think it's even worse than I first realized:

> #### Install packages using pip
> When your virtual environment is activated, you can install packages.
> [...]
> .. tab:: Windows
>        `py -m pip install requests`

This unfortunately is completely wrong. That will *never* install in the virtual environment. It is exactly showcasing a case of confusion caused by the "activating" concept. `py` is not affected by the "activated" PATH manipulation. It always uses the global python.

Kwpolska:
> This unfortunately is completely wrong. That will _never_ install in the virtual environment. It is exactly showcasing a case of confusion caused by the "activating" concept. `py` is not affected by the "activated" PATH manipulation. It always uses the global python.

It seems to work for me, in both cmd and PowerShell:

```
C:\tmp>py -m venv testvenv

C:\tmp>testvenv\Scripts\activate

(testvenv) C:\tmp>py -m pip --version
pip 22.3.1 from C:\tmp\testvenv\Lib\site-packages\pip (python 3.11)
```

petsuter:
Thanks for the correction. I should have confirmed it before posting. Apparently I'm still confused, and there's more magic than I realized. Sorry for the distraction.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,12 +1,12 @@
-Installing packages using pip and venv
-======================================
+Install packages in a virtual environment using pip and venv
+============================================================
 
-This guide discusses how to install packages using :ref:`pip` and
-the standard library's virtual environment :ref:`venv`. The guide
-covers how to:
+This guide discusses how to create and activate a virtual environment using
+the standard library's virtual environment :ref:`venv` and install packages.
```
sinoroc, line 5:
```suggestion
the standard library's virtual environment tool :ref:`venv` and install packages.
```

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -21,52 +21,11 @@ covers how to:
...
-        py -m pip --version
-
-    Afterwards, you should have the latest version of pip:
-
-    .. code-block:: text
-
-        pip 21.1.3 from c:\python39\lib\site-packages (Python 3.9.4)
+.. important::
+    This guide has the prerequisite that an official Python version from
+    <https://www.python.org/downloads/>. If you are using your operating
```
sinoroc:
```suggestion
    <https://www.python.org/downloads/> is installed and used. If you are using your operating
```

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -21,52 +21,11 @@ covers how to:
...
-        py -m pip install --upgrade pip
-        py -m pip --version
-
-    Afterwards, you should have the latest version of pip:
-
-    .. code-block:: text
-
-        pip 21.1.3 from c:\python39\lib\site-packages (Python 3.9.4)
+.. important::
+    This guide has the prerequisite that an official Python version from
```
sinoroc, line 25:
```suggestion
    This guide has the prerequisite that you are using an official Python version obtained from
```

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,203 +1,225 @@
...
-You can confirm you're in the virtual environment by checking the location of your
-Python interpreter:
+:ref:`pip` is the reference Python package manager.
+It's used to install and update packages into a virtual environment.
 
-.. tab:: Unix/macOS
+
+.. tab:: Unix
+
+    Debian and most other distributions include a `python-pip`_ package; if you
```
Kwpolska, line 156: This entire paragraph is no longer relevant if this guide only supports working in a venv.

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: For now, I'm working under the prerequisite that a version of python from https://www.python.org/downloads is already installed.

source/guides/installing-using-pip-and-virtual-environments.rst
willingc: Removed. Thx.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,203 +1,220 @@
...
+* Create and activate a virtual environment
+* Prepare pip
+* Install packages into a virtual environment using the ``pip`` command
+* Use and create a requirements file
 
 
-Installing pip
---------------
+.. note:: This guide applies to Python 3.3 and higher. If using a
+    legacy version of Python 2.x, consider using :ref:`virtualenv`.
```
ofek, line 15: Please come up with different words for this note as it makes it seem like that project is for legacy use cases whereas in fact it is a better version of the standard library module.

willingc: Good point. I will mention virtualenv as an alternative to venv that may be used for Python 3 or 2.

Kwpolska: Is a mention of virtualenv really useful for an average reader of this guide? Are there any virtualenv features that outweigh the cost of having to `pip install` it somewhere? The virtualenv docs list a few advantages (speed, extensibility, upgradeability, specifying interpreter via `-p`) — not something an end-user trying to install packages needs (they can always `.venv\bin\python -m pip install -U pip` or wait the extra few seconds).

pradyunsg: I think it's OK for the tutorial to drop the reference to `virtualenv` entirely, and keeping a line in the guide is reasonable.

willingc: I'm taking the approach with this guide to reframe the note to include "supported versions of Python". In a separate PR, I will create a stub guide for "Installing packages using pip and virtualenv". I don't use virtualenv so someone else will need to create that "how to" guide beyond the stub.

willingc:
I believe that I have addressed all feedback. Thanks for all of the reviews.

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -1,203 +1,220 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-    Afterwards, you should have the latest version of pip installed in your
-    user site:
+:ref:`venv` (for Python 3) allows you to manage separate package installations for
+different projects. It creates a "virtual" isolated Python installation. When
+you switch projects, you can create a new virtual environment which is isolated
+from other virtual environments. You benefit from the virtual environment
+since packages can be installed confidently and will not interfere with the
+another project's environment.
```
chrysle, line 42:
```suggestion
since packages can be installed confidently and will not interfere with
another project's environment.
```

sinoroc:
I think it be merged as is...

I wonder though if it would not be better to have 3 tabs all the way. Sometimes unix and mac are merged into 1 tab sometimes not. So the auto switch of all tabs throughout the page does not work as expected.

chrysle:
Maybe link to the `virtualenv` stub now?

> I wonder though if it would not be better to have 3 tabs all the way. Sometimes unix and mac are merged into 1 tab sometimes not.

I think this has been fixed recently? Add least I couldn't find examples any more...

pradyunsg (approved):

sinoroc:
> I think this has been fixed recently? Add least I couldn't find examples any more...

https://python-packaging-user-guide--1338.org.readthedocs.build/en/1338/guides/installing-using-pip-and-virtual-environments/#prepare-pip

The "Prepare pip" section has 3 tabs.

willingc:
I'm going to go ahead and merge this. I appreciate all of the reviews and suggestions. Let's continue to iterate in separate PRs. ☀️

Issue #1166, opened 2022-11-19
Explore integrating `sphinx-lint`


webknjaz:
Ref: https://github.com/python/python-docs-uk/issues/3

nwray25498wg471637320483646:
git-2.37.0.tar.sign

sanvi0sh:
hi, is this issue still open? if so can you elaborate it a little bit?

rffontenelle:
Please see the above PR for adding sphinx-lint session to the noxfile

rffontenelle:
There are #1786 and #1800, not sure which or any would land in the repo.

PR #1343, opened 2023-11-04
Use correct reST syntax for emphasis

jeanas:
xref: https://github.com/sphinx-contrib/sphinx-lint/issues/98


sinoroc:
Related: https://github.com/pypa/packaging.python.org/issues/1166

pradyunsg (approved):

PR #1345, opened 2023-11-05
Remove the Creating Documentation tutorial

willingc:
This tutorial is dated. It's also not directly related to packaging. It would be better to create a new docs tutorial or refer to existing documentation in other places.

I could see a future "How to" guide about making documentation be an extra installation.


sinoroc (approved):

pradyunsg:
I do agree that we should drop this page, and I also share the concern of dropping the page without redirects breaking existing links.

However, I don't like the idea of silently redirecting to another domain though -- it's a hard-coded link in source code that can't be checked for whether it is out of date automatically via Sphinx's linkcheck.

@willingc How do you feel about keeping the toctree removal and replacing page with a orphan stub page? Something like...

```
:orphan:

Creating documentation
======================

This tutorial has been removed since it is not related to packaging and was unmaintained. Please see the `Sphinx tutorial <https://www.sphinx-doc.org/en/master/tutorial>`_ instead.
```

willingc:
Great suggestion about the breaking links. I like the @pradyunsg's suggestion for a stub.

sinoroc (approved):
Even better now.

pradyunsg (approved):

pradyunsg:
Hah, nice @jeanas! ^.^

PR #1348, opened 2023-11-05
Remove "Dropping support for older Python versions" guide

jeanas:
This is a bunch of outdated or duplicated info, and does not bring much today. Make it a redirect to the specification for `requires-python` in `pyproject.toml`.


sinoroc (approved):

pradyunsg (changes requested):
I don't think we should remove this page, and it would instead be better to update it.

It should describe what requires-python looks like, how it behaves and provide guidance on how to bump the version across releases.

requirements.txt
```diff
@@ -5,3 +5,4 @@ python-docs-theme==2023.9
 sphinx-copybutton==0.5.0
 pypa-docs-theme @ git+https://github.com/pypa/pypa-docs-theme.git
 sphinx-toolbox==3.5.0
+sphinx-reredirects==0.1.3
```
webknjaz, line 8: RTD already has its own redirects that work on HTTP level. We don't need to handle the same thing in Sphinx.

webknjaz:
I agree with Pradyun in general — wiping such pages is not the best solution, unless the important information is preserved elsewhere. One thing I noticed is the `universal` setting explanation, for example.

PR #1350, opened 2023-11-05
Split a pyproject.toml guide out of the "Declaring project metadata" specification

### LINKED ISSUE #1349: Create a pyproject.toml guide (vs specification)
The best page I've found to learn about `pyproject.toml` configuration (the `[project]` table, i.e., PEP 621) is https://packaging.python.org/en/latest/specifications/declaring-project-metadata/. Unfortunately, that page is also full of technical language like

“For optional-dependencies, it is a table where each key specifies an extra and whose value is an array of strings. The strings of the arrays must be valid [PEP 508](https://peps.python.org/pep-0508/) strings. The keys MUST be valid values for [Provides-Extra](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-provides-extra). Each value in the array thus becomes a corresponding [Requires-Dist](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-requires-dist) entry for the matching [Provides-Extra](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-provides-extra) metadata.”

which is normal because it's a specification, but not ideal for Joe User.

I propose to add a `pyproject.toml` configuration guide, separate from the specification, to describe the fields in accessible, non-technical language without the details of how they're translated to core metadata.

jeanas:
Fixes #1349


sinoroc:
Thanks @jeanas 

Shortcut to preview: https://python-packaging-user-guide--1350.org.readthedocs.build/en/1350/guides/writing-pyproject-toml/

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,383 @@
...
+   description = "Lovely Spam! Wonderful Spam!"
+   readme = "README.rst"
+   license = {file = "LICENSE.txt"}
+   keywords = ["egg", "bacon", "sausage", "tomatoes", "Lobster Thermidor"]
+   classifiers = [
+     "Development Status :: 4 - Beta",
+     "Programming Language :: Python"
+   ]
+
+   # dynamic = ["version", "description"]
```
chrysle, line 341: I'm in favour of just taking this out.

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,381 @@
+.. _writing-pyproject-toml:
+
+===============================
+Writing your ``pyproject.toml``
+===============================
+
+.. TODO: link :term:`build backends` when that's merged
```
chrysle, line 7: This should be done now.

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,381 @@
...
+
+.. note::
+
+   You may have heard of ``setup.py`` and ``setup.cfg`` for the setuptools_
+   build backend. For new projects, it is recommended to use ``pyproject.toml``
+   for basic metadata, and keep ``setup.py`` only if some programmatic configuration
+   is needed (especially building C extensions). However, putting basic project
+   metadata in ``setup.py`` or ``setup.cfg`` is still valid.
+
+.. TODO: link to "is setup.py deprecated?" page when merged
```
chrysle, line 22: Same as above.

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,381 @@
...
+
+.. code-block:: toml
+
+   [project]
+   readme = {file = "README.txt", content-type = "text/markdown"}
+   # or
+   readme = {file = "README.txt", content-type = "text/x-rst"}
+
+
+``license``
```
chrysle, line 228: I think we should mention https://peps.python.org/pep-0639/#spec-source-metadata here?

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,381 @@
...
+sidebar of your PyPI project page. If the key contains spaces, don't
+forget to quote it.
+
+.. code-block:: toml
+
+    [project.urls]
+    Homepage = "https://example.com"
+    Documentation = "https://readthedocs.org"
+    Repository = "https://github.com/me/spam.git"
+    "Bug Tracker" = "https://github.com/me/spam/issues"
```
chrysle, line 288: Probably it's better to avoid the quotes to comply with the TOML spec.

```suggestion
    Issues = "https://github.com/me/spam/issues"
```

source/specifications/declaring-project-metadata.rst
```diff
@@ -9,6 +9,12 @@ Declaring project metadata
 packaging-related tools to consume. It defines the following
 specification as the canonical source for the format used.
 
+.. warning::
```
willingc, line 12: I like this warning block. Why remove the content from the specification? I hesitate to change specifications. Specifications probably should rarely change unless there is formal agreement to do so.

willingc (commented):
Overall, I like this PR. I'm concerned about removing content from a specification. Do you think the removal is necessary?

source/guides/writing-pyproject-toml.rst
chrysle: Yes, I read the issue. I don't think that the fact many people use the old example should stop us from amending it.

source/specifications/declaring-project-metadata.rst
willingc: Here are some of my thoughts. In general, I don't like changing specs without versioning the spec (I'm making an assumption that this is the only place that the spec is captured in PyPA.)

I do think that the examples here provide clarity to users. As a user, I would view the spec as the source of truth. I agree with you that duplication does provide the likely potential for divergence.

I'm wondering if we could keep the examples in the spec and mention in the guide that the spec is the source of truth.

If others agree that we can remove the examples from the spec, then I'm cool with that. I'm just leaning toward being conservative with minimal changes, if any, to the specs.

willingc (commented):
I like having the new doc which will be helpful. I'm just not sure that it adds value to remove the examples from the spec.

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,383 @@
...
+
+    [project]
+    authors = [
+      {name = "Pradyun Gedam", email = "pradyun@example.com"},
+      {name = "Tzu-Ping Chung", email = "tzu-ping@example.com"},
+      {name = "Another person"},
+      {email = "different.person@example.com"},
+    ]
+    maintainers = [
+      {name = "Brett Cannon", email = "brett@python.org"}
```
willingc, line 183: I suggest changing this email to `brett@example.org` both here and on line 332. Bots that scrape docs and use as AI source for code gen will serve this up for users who will copy-paste into their `pyproject.toml`. Seems more prudent to have a non-active email.

willingc (approved):
@jeanas I'm approving this PR since I like the new doc. I will leave to you and others when to merge re: spec changes. We can always add the removals back as needed at a later date.

Very happy to see the activity the past few weeks to make improvements. ✨

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,383 @@
...
+``name``
+--------
+
+Put the name of your project on PyPI. This field is required and is the
+only field that cannot be marked as dynamic.
+
+.. code-block:: toml
+
+   [project]
+   name = "spam"
```
sinoroc, line 57: Add link to [Package name normalization](https://packaging.python.org/en/latest/specifications/name-normalization/) `` :ref:`name-normalization` ``?

source/guides/writing-pyproject-toml.rst
```diff
@@ -0,0 +1,383 @@
...
+
+    [project]
+    version = "2020.0.0"
+
+This field is required, although it is often marked as dynamic using
+
+.. code-block:: toml
+
+   [project]
+   dynamic = ["version"]
```
sinoroc, line 75: Add link to [version specifiers spec](https://packaging.python.org/en/latest/specifications/version-specifiers/) `` :ref:`version-specifiers` ``?

sinoroc (commented):
Great!

More links maybe?

source/guides/writing-pyproject-toml.rst
sinoroc: I guess I was trying to find some documentation or specification about what are valid names for Python projects (distribution packages). And I guess this is the closest I could find at the time. There is also the [Name Core Metadata](https://packaging.python.org/en/latest/specifications/core-metadata/#name). I don't know, feel free to add one or both or none.

willingc:
I've opened #1390 to document guidance on when to change specifications. This PR is better than the current documentation, and I will merge when CI completes. Thanks for the patience @jeanas.

source/guides/writing-pyproject-toml.rst
sinoroc: fine by me

PR #1351, opened 2023-11-05
Bump requirements

jeanas:
- Upgrade to latest version of Sphinx (and other dependencies),
- Use "language = 'en'" in configuration since recent Sphinx versions do not accept "language = None" anymore,
- Add now required "%s" in extlinks config,
- Refer to setuptools explicitly in :std:doc: because Sphinx no longer implicitly uses intersphinx (by default) for this role. See https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_disabled_reftypes

The generated HTML can be compared using

$ diff -u build-old/ build-new | colordiff | diff-highlight

The only visible change is "Permalink to this headling" becoming "Link to this heading".

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-sphinx==4.5.0
+sphinx==7.2.6
 sphinx-autobuild==2021.3.14
-sphinx-inline-tabs==2021.4.11b9
+sphinx-inline-tabs==2023.4.21
 python-docs-theme==2023.9
-sphinx-copybutton==0.5.0
+sphinx-copybutton==0.5.2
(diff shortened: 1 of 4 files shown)


willingc:
@jeanas Would you mind splitting the requirements changes out from the content changes? Thanks!

pradyunsg:
IIUC, the content changes are limited to those required as a result of moving to the newer Sphinx versions.

pradyunsg:
> diff -u build-old/ build-new | colordiff | diff-highlight

Could you post the output of this on a GitHub Gist?

pradyunsg (commented):
Let's move forward with this.

pradyunsg (approved):
Let's move forward with this.

PR #1353, opened 2023-11-06
Add local contents to Projects doc and edits to landing page

willingc:
This PR adds a local contents to the Projects Page for easier scanning and updates prose on the main landing page.


webknjaz:
This conflicts with #1354 since Furo don't allow in-document local toc.

willingc:
@webknjaz Happy to close this as unneeded if we swap to Furo (which we should as the page sidebar would be helpful for users).

Issue #1354, opened 2023-11-06
Update Sphinx theme


willingc:
The current Sphinx theme lacks a right-hand sidebar for page navigation. It would be lovely to move to something similar to the  devguide.python.org theme, Furo, or ???

cc/ @pradyunsg

webknjaz:
https://github.com/pypa/packaging.python.org/pull/791

pradyunsg:
I think I'm no longer opposed to Furo being used here, albeit we might have to spend some time after to clean up some of the rough edges that migration will pop up.

webknjaz:
I'll rebase the PR, then.

pradyunsg:
Done in #791.

PR #1356, opened 2023-11-06
[DO-NOT-MERGE] Use pydata-sphinx-theme

jeanas:
This is just so that we have a PR preview for easier discussion of #1354.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ sphinx-autobuild==2021.3.14
 sphinx-inline-tabs==2023.4.21
 python-docs-theme==2023.9
 sphinx-copybutton==0.5.2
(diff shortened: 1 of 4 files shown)


willingc:
Thanks @jeanas for the rendered version. I think the key takeaway is that a modern theme (pydata-sphinx, furo, or other) would be helpful for readers.

pradyunsg:
I looked at the preview for this on my mobile and it looks like that overflows to the right instead of wrapping content. 😅

pradyunsg:
So... I ended up triggering automerge of #791 before I saw this. Since that's landed, let's close this out.

Thanks @jeanas for opening this, even though we didn't end up going down this direction.

PR #1357, opened 2023-11-06
Clarify that packages don't need `__init__.py`

### LINKED ISSUE #1292: __init__ should not be empty
In the webpage: https://packaging.python.org/en/latest/tutorials/packaging-projects/  was stated the following:
"__init__.py is required to import the directory as a package, and should be empty."

It did not work for me. Instead, when I wrote my methods in the __init__ file as before then it worked.

abravalheri:
This is a follow up on #1342: the tutorial implies that a Python package requires an `__init__.py` file.

I think that is the best to avoid this implication, otherwise we might "mis-educate" "packaging beginners" from the start.

The idea of this change is to introduce minimal rewording and a footnote.
The terms "regular package" and "namespace package" were introduced in [PEP 420](https://peps.python.org/pep-0420/), which is also very intentional when it comes to equating regular and namespace packages:

> A namespace package is not fundamentally different from a regular package. It is just a different way of creating packages. Once a namespace package is created, there is no functional difference between it and a regular package.

The change still recommends people that are getting started with Python packaging to add `__init__.py` as a result of the discussion in #1292: the general understanding in the community is that *some tools* might not fully support namespace packages yet (or be a bit more fussy about it) - although I cannot point out examples of these tools from the top of my mind...

Closes #1292.

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -54,8 +54,8 @@ Create the following file structure locally:
 The directory containing the Python files should match the project name. This
 simplifies the configuration and is more obvious to users who install the package.
 
-:file:`__init__.py` is required to import the directory as a package,
-even if as is our case for this tutorial that file is empty.
+:file:`__init__.py` is recommended to import the directory as a regular package,
+even if as is our case for this tutorial that file is empty [#namespace-packages]_.
 
 :file:`example.py` is an example of a module within the package that could
 contain the logic (functions, classes, constants, etc.) of your package.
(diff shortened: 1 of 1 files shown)


willingc (approved):
Nice improvement. Thanks @abravalheri.

henryiii (approved):

webknjaz:
Love the change, thanks!

sinoroc:
> Love the change

I don't :D

It is not a big deal, probably can be considered nitpicking.

Honestly, I do not follow the logic of this change, and you can feel free to ignore this.

---

> `__init__.py` is recommended to import the directory as a regular package

It is not recommended, it is the only way to have the package be a regular package. If the file is not there then it is imported as a namespace package, isn't it? Python still considers it as a namespace package, doesn't it? Maybe there are no practical differences in most of the cases, but the sentence is still incorrect. Maybe the PEP says that the initializer can be omitted but it seems to me very much against the spirit of the PEP (what its author stated later on [[1]](https://discuss.python.org/t/-/7536/6) [[2]](https://discuss.python.org/t/-/7536/10)).

Is the intention of mentioning that `__init__.py` files can be omitted to make it easier for beginners?

If I recall correctly I have more than once helped fix people's projects by telling them to add missing `__init__.py`.




> the general understanding in the community is that _some tools_ might not fully support namespace packages yet (or be a bit more fussy about it) - although I cannot point out examples of these tools from the top of my mind...

Unless the behavior of those has changes since last time I checked (which was a long time ago, for sure):

* `setuptools.find_packages()`
* [`importlib.resources` ](https://stackoverflow.com/a/62931707)

abravalheri:
Hi my @sinoroc the idea was a bit different, the recommendation is for people to add `__init__.py` ***so that*** packages are considered "regular packages". I agree that this is the only way a package is considered regular.


We can further clarify that if you prefer.

I don't know about `importlib`, but `setuptools` supports namespace packages (the distinction is explicit via 2 separated API though).

PR #1358, opened 2023-11-06
Remove contents directives

jeanas:
They're not useful with Furo, which has a right sidebar with the local TOC, and they make it generate warnings in the rendered documentation.

@pradyunsg


sinoroc:
Preview: https://python-packaging-user-guide--1358.org.readthedocs.build/en/1358/

pradyunsg (approved):

PR #1361, opened 2023-11-06
Bump linters and add ruff for consistent formatting across the codebase

pradyunsg:
I'd like to avoid having any discussions about code style/formatting in this codebase, so let's adopt a tool that'll handle it for us.


webknjaz (commented):
Consistency doesn't necessarily mean changes. Can we leave the quotes alone?

pradyunsg:
> Can we leave the quotes alone?

It's the defaults of the tool, which I'd prefer to leave as-is. Do you feel strongly about this?

webknjaz:
Kinda

pradyunsg:
Given the non-committal answer, I'm inclined to suggest that we move forward with this as-is. This is not a Python codebase with thousands of lines of complex logic -- there's 2 configuration files written in Python with at most if statements.

webknjaz (approved):

pradyunsg:
(I should add the reformatting commit to blame ignore)

PR #1362, opened 2023-11-06
Modernise the conf.py file

pradyunsg:
For anyone curious about what keys have changed in this PR, you can use `python -i source/conf.py` followed by `locals()` to capture the assignments made.

Based on that, below is the diff between `{str(k): str(v) for k,v in locals()}` with the old conf.py (i.e. main) and new conf.py (i.e. this PR).

```py
>>> set(old_conf_py.items()) - set(new_conf_py.items())
{('master_doc', 'index'), ('templates_path', "['_templates']"), ('release', ''), ('version', ''), ('source_suffix', '.rst')}
>>> set(new_conf_py.items()) - set(old_conf_py.items())
{('root_doc', 'index')}
```

The keys removed were set to values that influenced ~no relevant part of the build output.


pradyunsg:
This builds upon #1361 (it's what motivated me filing that PR) so I'll keep this as a draft until that one is resolved.

source/conf.py
```diff
@@ -1,446 +1,174 @@
...
-locale_dirs = ['../locales']
-
-gettext_auto_build = True  # Build MO files from PO on each build
-
-# making all documents use single text domain
-gettext_compact = "messages"
+project = "Python Packaging User Guide"
 
-gettext_location = True  # Include location info in the translation files
+copyright = "2013–2020, PyPA"
```
sinoroc, line 6: I am not sure dates are really needed in copyright notices. If yes, then maybe this needs an update, maybe better would be to not hard-code it and get the real current year instead?

webknjaz: Many people misunderstand the meaning of copyright marks. Only the year of the content creation, not website redeployment, should be mentioned, normally. Though, when the content is updated, that year can also be mentioned as it's when new changes were first created, again.
Note that the presence of the copyright notice or the years is optional and has no legal power. The copyright owners gain their rights at the moment of creation regardless of whether they added this line or not.

pradyunsg: We should update this, but that is likely a separate concern.

source/conf.py
```diff
@@ -1,446 +1,174 @@
...
-#
-html_theme = 'furo'
+html_title = "Python Packaging User Guide"
 
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
+html_theme = "furo"
 html_theme_options = {
     "sidebar_hide_name": True,
```
chrysle, line 56: Could we turn this on? I currently don't see a way to return to the front page.

pradyunsg: Yea, let's do that in a follow up to this one.

pradyunsg: Or... I could just do that here. :)

pradyunsg:
This is good to go, and only need a review from another editor. :)

source/conf.py
```diff
@@ -366,41 +96,38 @@
...
-    "issue": (f"{github_repo_issues_url}/%s", "#%s"),  # noqa: WPS323
-    "pr": (f"{github_repo_url}/pull/%s", "PR #%s"),  # noqa: WPS323
-    "commit": (f"{github_repo_url}/commit/%s", "%s"),  # noqa: WPS323
-    "gh": (f"{github_url}/%s", "GitHub: %s"),  # noqa: WPS323
-    "user": (f"{github_sponsors_url}/%s", "@%s"),  # noqa: WPS323
+    "issue": (f"{github_repo_issues_url}/%s", "#%s"),
+    "pr": (f"{github_repo_url}/pull/%s", "PR #%s"),
+    "commit": (f"{github_repo_url}/commit/%s", "%s"),
+    "gh": (f"{github_url}/%s", "GitHub: %s"),
+    "user": (f"{github_sponsors_url}/%s", "@%s"),
```
chrysle, line 125:
```suggestion
    "user": (f"{github_sponsors_url}/%s", "@%s"),
    "pypi": ("https://pypi.org/project/%s", "%s"),
```

I think this could be real helpful.

chrysle:
BTW, could we somehow allow the footer to hold the donations link and license (via template)?

source/conf.py
sinoroc: Can be handled separately, and with actual usage. Just adding the alias without using it seems a bit strange.

Issue created: https://github.com/pypa/packaging.python.org/issues/1370

pradyunsg:
> BTW, could we somehow allow the footer to hold the donations link and license (via template)?

I don't know. 😅

chrysle (approved):

chrysle:
> Let's leave that for a separate PR?

Yes, I'll have a look now.

chrysle:
@pradyunsg Umm, unsure if you could arrange yourself with the idea of.... two footers?

image

chrysle:
@pradyunsg ?

pradyunsg:
I'm not sure what you're suggesting here. 😅

PR #1364, opened 2023-11-07
Allow Sphinx to do incremental builds

### LINKED ISSUE #1363: Stop deleting documentation build directory on rebuilds
`noxfile.py` currently contains this in the `build` session:

```python
    shutil.rmtree(target_build_dir, ignore_errors=True)
```

This nukes the build folder before proceeding.

I currently have to work on a relatively slow Internet connection. Each build is re-downloading the inventory files, which takes noticeable time.

Is there any particular reason not to use Sphinx's incremental build capabilities? After all, it's easy enough to remove the build folder by hand if you want to check that some oddity is not caused by incremental builds.

jeanas:
Fixes #1363; CC @pradyunsg

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -33,10 +33,6 @@ def build(session, autobuild=False):
     """
     session.install("-r", "requirements.txt")
 
-    target_build_dir = "build"
-
-    shutil.rmtree(target_build_dir, ignore_errors=True)
-
     if autobuild:
         command = "sphinx-autobuild"
         extra_args = "--host", "0.0.0.0"
@@ -56,7 +52,7 @@ def build(session, autobuild=False):
         "-W",  # Treat warnings as errors.
         *session.posargs,
         "source",  # where the rst files are located
-        target_build_dir,  # where to put the html output
+        "build",  # where to put the html output
     )
 
 


webknjaz:
@jeanas could you make sure that all Sphinx commands also have the same `-d` CLI arg?

webknjaz:
That specifies a directory for storing the doctree cache, which is Sphinx's intermediate representation of the documents before they are converted into the resulting output markup.
This is what actually allows for incremental builds.

pradyunsg:
> This is what actually allows for incremental builds.

For the specific thing that @jeanas is facing, it's the environment which does live in the build directory.

webknjaz:
I've never seen env loving there, unless set specifically. Will have to double-check when I get to the laptop..

webknjaz:
Okay.. I verified that doctrees are created under `./build/.doctrees/`. And `./build/objects.inv` with local targets is created. But I was unable to locate any downloaded `*.inv` files, which matches my experience..
Though, I know how to make *that* cache work. Kinda.

pradyunsg:
> I've never seen env living there, unless set specifically. 

Ah, you're correct.

> But I was unable to locate any downloaded *.inv files, which matches my experience..

It ends up in the `environment.pickle` file, which does end up in the .doctrees directory.

---

I just confirmed locally that the following does use the cache the inventories on the first build and reuses them for the subsequent runs.

```
sphinx-build . ../build/ -d ../doctrees.ignore
rm -r ../build/
sphinx-build . ../build/ -d ../doctrees.ignore
touch index.rst
rm -r ../build/
sphinx-build . ../build/ -d ../doctrees.ignore
```

pradyunsg:
I think #1365 does what @jeanas is hoping for -- could you confirm that it works correctly @jeanas?

Also, thanks @webknjaz for flagging the `-d` thingie -- that turned out to be the missing piece here. :)

webknjaz (approved):

webknjaz:
FWIW, this PR is also useful.

webknjaz:
> It ends up in the environment.pickle file, which does end up in the .doctrees directory.

Oh, interesting.. I guess that's why I was under an impression that the inventory cache was broken. Good to know!

As for fallbacks, intersphinx allows pointing to local inv files additionally. So that's what I was considering as an alternative.

webknjaz:
😂 I added both PRs to the merge queue and one closed the other! ROFL

PR #1365, opened 2023-11-07
Use a shared cache directory for Sphinx builds across nox sessions

### LINKED ISSUE #1363: Stop deleting documentation build directory on rebuilds

pradyunsg:
This enables incremental build with Sphinx.

Closes #1364 
Resolves #1363 by resolving the underlying concern.

https://gist.github.com/pradyunsg/27d99a7b08ea022fb28385b76c8d8e2b has the output, demonstrating the behaviour of this.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -4,8 +4,8 @@
 #   http://creativecommons.org/licenses/by-sa/3.0.
 
 import shutil
-import nox
 
+import nox
 
 nox.options.sessions = []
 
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -4,8 +4,8 @@
 #   http://creativecommons.org/licenses/by-sa/3.0.
 
 import shutil
-import nox
 
+import nox
```
pradyunsg, line 8: _sigh_

I have isort installed and set up to run on save, because all my projects sort imports. I'd appreciate if we let this specific change go through so that I don't have to fumble around to undo this one-line move. :)

webknjaz: _Sigh_

webknjaz (approved):

PR #1202, opened 2023-02-03
Document more package index host/cache/mirror options

EpicWink:
Update "Package index mirrors and caches" guide to include more options, and better describe classes of options.

I think the "Hosting your own simple repository" guide is very related, with a bit of overlap with this PR. I'm not sure what the path forward is here.

Also updates bandersnatch docs link #1083.

Possibly also related: #601

PR #1366, opened 2023-11-08
Add more package index projects

EpicWink:
Add many more package index projects to the "Project Summaries" page, add a table comparing these projects in the "Hosting your own simple repository" page, add another table comparing these projects in the "Package index mirrors and caches" page, and update the intro to "Package index mirrors and caches" to reflect the multitude of options

An alternative to #1202

Previews:
* https://python-packaging-user-guide--1366.org.readthedocs.build/en/1366/guides/index-mirrors-and-caches/
* https://python-packaging-user-guide--1366.org.readthedocs.build/en/1366/guides/hosting-your-own-index/


source/key_projects.rst
```diff
@@ -727,6 +845,20 @@ A package in the Python Standard Library that provides support for bootstrapping
 cases, end users won't use this module, but rather it will be used during the
 build of the Python distribution.
 
+.. _httpserver:
+
+http.server
+===========
+
+`Docs <https://docs.python.org/3/library/http.server.html>`__ |
```
chrysle, line 853:
```suggestion
`Docs <python:library/http.server.html>`__ |
```

source/key_projects.rst
```diff
@@ -675,6 +780,19 @@ Python zipapps as outlined in :pep:`441`, but with all their
...
+
+simpleindex
+===========
+
+:gh:`GitHub <uranusjr/simpleindex>` |
+`PyPI <https://pypi.org/project/simpleindex/>`__
+
+simpleindex is a :term:`package index <Package Index>` which routes URLs to
+multiple package indexes (including PyPI), serves local (or cloud-hosted,
+for example `AWS S3`_, with a custom plugin) directory of packages, and
```
chrysle, line 793:
```suggestion
for example `AWS S3`_, with a custom plugin) directories of packages, and
```

source/guides/hosting-your-own-index.rst
```diff
@@ -52,10 +52,81 @@ directory with autoindex enabled. For an example using the built in Web server
...
+     -
+     -
+     - also mirroring; manual synchronisation
+
+   * - :ref:`httpserver`
+     -
+     -
+     - standard-library
+
+   * - `Apache <https://httpd.apache.org/>`_
```
chrysle, line 113: Don't know if we should include Apache?

source/key_projects.rst
```diff
@@ -727,6 +845,20 @@ A package in the Python Standard Library that provides support for bootstrapping
 cases, end users won't use this module, but rather it will be used during the
 build of the Python distribution.
 
+.. _httpserver:
+
+http.server
+===========
+
+`Docs <https://docs.python.org/3/library/http.server.html>`__ |
+`Issues <https://github.com/python/cpython/issues>`__
```
chrysle, line 854:
```suggestion
:gh:`Issues <python/cpython/issues>`
```

sinoroc:
Previews:
* https://python-packaging-user-guide--1366.org.readthedocs.build/en/1366/guides/index-mirrors-and-caches/
* https://python-packaging-user-guide--1366.org.readthedocs.build/en/1366/guides/hosting-your-own-index/

source/guides/hosting-your-own-index.rst
EpicWink: I don't see the harm in including it

There's also an nginx plugin earlier

jeanas (approved):
Let's merge this?

webknjaz:
@EpicWink let's maybe mention `dumb-pypi` which is essentially a static site generator that makes a simple index for a bunch of pre-downloaded/created wheels.
As for the hosting, there's GitHub Pages that I use for having extra indexes with "nightly" bleeding-edge dists. So mentioning GitHub/GitLab Pages (or any static hosting for that matter) as possible solutions for limited use-cases would be useful too.
Here's an example: https://ansible.github.io/pylibssh/.

Finally, you mention Twisted, but omit the built-in web server that ships with Python — `python -Im http.server`. It's more of a local testing tool since it's not really designed to serve HTTP under the high load, but I believe does still deserve an honorary mention. After all, Twisted would be a production-side solution (although, this does omit all the other web servers capable of the same).

Could you bake these tools into the contributed documents, please?

EpicWink:
> let's maybe mention dumb-pypi which is essentially a static site generator that makes a simple index for a bunch of pre-downloaded/created wheels.

@webknjaz I've included `dumb-pypi` in the latest change-set.

---

> As for the hosting, there's GitHub Pages

GitHub pages and GitLab pages aren't really "projects", but services. Per [this comment](https://github.com/pypa/packaging.python.org/pull/1367#issuecomment-1806700551), we don't want to list commercial services.

I guess there's an opportunity to explicitly say package indexes can be static file servers.

---

> Finally, you mention Twisted, but omit the built-in web server that ships with Python — `python -Im http.server`

I have already included it: see [source/guides/hosting-your-own-index.rst:108](https://github.com/pypa/packaging.python.org/pull/1366/files#diff-f1d7ac8277e255bfae4ac8f0ee743d27c879f16bbb0439f630a0c1ad0c8066eeR108)

edgarrmondragon:
Would it be worth including https://github.com/gorilla-co/s3pypi?

EpicWink:
> Would it be worth including https://github.com/gorilla-co/s3pypi?

@edgarrmondragon I don't think so, as it relies on a commercial solution, and is not a self-contained project.

webknjaz:
@EpicWink re:gitlab — it can be a self-hosted platform with a free version, since it's an Open Core project. But yeah, maybe a more generic "use any web server to serve static if you want your index accessible over web" would be better (mentioning that it's also possible to point installers to a simple index / wheelhouse on the local file system). I was also thinking that we could mention such examples, but not link them, maybe? Anyway, these two platforms already host the majority of open source and do so for free. So while they're commercial, they also do a lot for the sustainability of the open source. I think it might be fine to mention them in a restricted context with disclaimers — it wouldn't be an advertisement but an acknowledgement that these things exists.

webknjaz:
> > Finally, you mention Twisted, but omit the built-in web server that ships with Python — `python -Im http.server`
> 
> I have already included it: see [source/guides/hosting-your-own-index.rst:113](https://github.com/pypa/packaging.python.org/blob/2241c5039af75a4199b2645c8dc9f61d1d0b9017/source/guides/hosting-your-own-index.rst?plain=1#L113)

Ah, maybe I missed it because it doesn't have a dot in RST. Though, I was thinking more in the context of having it instead of twisted in the paragraph since it wouldn't be something dependent on the presence of an external project.

EpicWink:
> maybe a more generic "use any web server to serve static if you want your index accessible over web" would be better

@webknjaz re-reading the page, the current `main` version already discusses static file serving (with Twisted as an example), and this pull-request doesn't make changes to that discussion. I think making it more generic, and discussing other services, would better be served in another pull-request.

webknjaz:
Okay, fair.

webknjaz:
> Per this comment, we don't want to list commercial services.

Looks like S3 is still linked?

EpicWink:
> Looks like S3 is still linked?

@webknjaz my understanding is that PyPA doesn't want to promote or recommend commercial services. I don't think listing a project which has optional functionality to utilise a commercial service (in this case S3) can be considered recommending.

A previous comment suggests `s3pypi`, which *requires* S3, which I think means that if `s3pypi` is listed, S3 is de-facto being recommended.

We could remove the explicit reference to S3, and just say "cloud object storage services", but honestly I left in S3 because I think it helps with Google (and other search engines!) searches, which helps novice package distributors who are familiar with cloud services.

webknjaz:
Alright, that's fair. Just wanted to clarify.

source/guides/hosting-your-own-index.rst
webknjaz: Looks like this is addressed. Marking the thread as resolved to unblock merging.

source/guides/hosting-your-own-index.rst
```diff
@@ -52,10 +52,86 @@ directory with autoindex enabled. For an example using the built in Web server
...
+       `mod_cache_disk
+       <https://httpd.apache.org/docs/current/mod/mod_cache_disk.html>`_,
+       you can cache requests to package indexes through an Apache server
+
 ----
 
 .. [1] For complete documentation of the simple repository protocol, see
        :ref:`simple repository API <simple-repository-api>`.
 
+.. [2] Can be configured to fall-back to PyPI (or another package index)
```
webknjaz, line 134:
```suggestion
.. [2] Can be configured to fall back to PyPI (or another package index)
```
(verb variants are whitespace-delimited)

webknjaz (approved):

webknjaz:
@EpicWink thanks!

PR #1367, opened 2023-11-08
Document hosted/managed package indexes

EpicWink:
Add a list of hosted and managed package index services to indicate to new users they don't have to build their own solution (as currently that's all the guide suggests at the moment, apart from devpi). Also include documentation links

### DIFF
--- a/source/guides/hosting-your-own-index.rst
+++ b/source/guides/hosting-your-own-index.rst
@@ -52,6 +52,87 @@ directory with autoindex enabled. For an example using the built in Web server
 in `Twisted`_, you would simply run ``twistd -n web --path .`` and then
 instruct users to add the URL to their installer's configuration.
 
+
+Hosted and managed solutions
+============================
+
+These options are provided by third-party organisations, and are generally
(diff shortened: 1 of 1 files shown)


chrysle:
Generally unsure if we should include non-free services in the guide.. Maybe someone else can say something about that?

sinoroc:
Preview: https://python-packaging-user-guide--1367.org.readthedocs.build/en/1367/guides/hosting-your-own-index/#hosted-and-managed-solutions

willingc:
Hmm...I don't think that we have ever recommended a commercial solution in other docs for Python. I think the decision is a governance one and what the PSF and PyPA policy is.

That said I think folks would find the info valuable hosted as a personal doc on RTD about commercial hosting options.

EpicWink:
Something like https://epicwink.github.io/python-index-services/?

jeanas:
It sounds like there is consensus on not listing proprietary and paid services, so I'm going to close this PR, but please feel free to reopen if I've misunderstood.

PR #1369, opened 2023-11-08
Add a GitHub Action to inject preview links in PRs

pradyunsg:
This makes easier to review pull requests by providing the preview links in the PR description.

### DIFF
--- a/.github/workflows/pr-preview-links.yml
+++ b/.github/workflows/pr-preview-links.yml
@@ -1,0 +1,22 @@
+name: Read the Docs PR preview
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


pradyunsg:
This is also used on https://github.com/python/peps/, where you can see how this injects the name (eg: https://github.com/python/peps/pull/3522).

sinoroc:
Thanks! I knew I had seen something like this before.

.github/workflows/pr-preview-links.yml
```diff
@@ -0,0 +1,22 @@
+name: Read the Docs PR preview
+
+on:
+  pull_request_target:
+    types:
+      - opened
```
chrysle, line 6: Should we add this?

```suggestion
    types:
      - opened
      paths:
        - "source/**"
```

pradyunsg: I considered this, but we have .readthedocs.yml and requirements.txt outside of this directory which do affect the RTD build. After adding those, I figured that the RTD builds are going to happen again anyway and it's easy-enough to see things. 😅

chrysle: Sounds fair, resolving!

chrysle (approved):
LGTM, but let's wait for the expert's (@webknjaz's) review.

willingc:
@webknjaz Sorry if I merged prematurely. I missed @chrysle's note.

PR #1372, opened 2023-11-08
Import PEP 668 specification

jeanas:
This is similar to #1337 and #1340.

The first commit imports the PEP text as-is. All changes are done in the second commit so that reviewers can see what changed exactly. It is mostly trimming rationale/motivation sections and fixing markup for the PUG.

### DIFF
--- a/source/specifications/externally-managed-environments.rst
+++ b/source/specifications/externally-managed-environments.rst
@@ -21,6 +21,1102 @@ neither install nor remove packages into the interpreter’s default installatio
 environment, and should instead guide the end user towards using
 :ref:`virtual-environments`.
 
+PEP: 668
+Title: Marking Python base environments as “externally managed”
+Author: Geoffrey Thomas <geofft@ldpreload.com>,
+        Matthias Klose <doko@ubuntu.com>,
+        Filipe Laíns <lains@riseup.net>,
+        Donald Stufft <donald@stufft.io>,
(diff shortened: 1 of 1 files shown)


willingc (approved):
Thanks @jeanas.

source/specifications/externally-managed-environments.rst
```diff
@@ -15,12 +15,460 @@ environments can be confusing at best and outright break the entire underlying
...
+
+which provides useful and distro-relevant information
+to a user trying to install a package. Optionally,
+translations can be provided in the same file:
+
+.. code-block:: ini
+
+    Error-de_DE=Wenn ist das Nunstück git und Slotermeyer?
+
+     Ja! Beiherhund das Oder die Virtualenvironment gersput!
```
chrysle, line 238: This German example is outright horrible.... No wonder `codespell` complains.

We're in a bad spot; `codespell` doesn't support inline ignores. I don't think passing `--ignore-regex` to the hook is the best idea.

chrysle: We could take a look at https://github.com/python/peps/blob/main/.codespell/exclude-file.txt.

chrysle: > [On purpose](https://en.m.wikipedia.org/wiki/The_Funniest_Joke_in_the_World) 😄

They should add a warning next time ;-)

PR #1373, opened 2023-11-09
Version specifiers: replace 'PEP' with 'specification'

hugovk:
Follow on from https://github.com/pypa/packaging.python.org/pull/1340 which imported [PEP 440](https://peps.python.org/pep-0440/).

I left this "PEP" intact:

```rst
* Note: this comparison is to ``pkg_resourses.parse_version`` as it existed at
  the time :pep:`440` was written. After the PEP was accepted, setuptools 6.0 and
  later versions adopted the behaviour described here.
```

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1373.org.readthedocs.build/en/1373/


chrysle (approved):
Thanks!

PR #1377, opened 2023-11-11
Remove "Using MANIFEST.in" guide, moved to setuptools docs

jeanas:
This page: https://packaging.python.org/en/latest/guides/using-manifest-in/ is very setuptools-specific, describing the syntax of setuptools' `MANIFEST.in` file. I think the setuptools documentation is a better home for this. I've opened a [PR](https://github.com/pypa/setuptools/pull/4108) on the setuptools repo to add it. Once it's merged, this PR can be undrafted and merged. It removes the document, which should be replaced by a redirect in the ReadTheDocs configuration.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1377.org.readthedocs.build/en/1377/


abravalheri:
I was wondering if it would be possible to add a HTTP redirect instead of removing[^1]... This way any existing links in places other than the setuptools docs themselves (thanks Jean) would still "work".

In setuptools, I use [sphinx-reredirects](https://documatt.com/sphinx-reredirects/usage.html) in a couple of places to achieve something similar.


[^1]: I mean just for direct access to the webpage address itself. Removing from any ToC is completely fine.

willingc (commented):
I believe that we have been creating orphan stubs instead of completely removing docs.

sinoroc:
> The PUG uses redirects configured from the RTD Web UI instead of sphinx-reredirects.

Why? Seems like we are locking ourselves to RTD for no reason.

I asked on some other thread but it might be a bit hidden, forgive me for asking again...

willingc (approved):
Thanks for the update @jeanas.

willingc:
@sinoroc I think @pradyunsg would be better to respond to your question.

pradyunsg:
Filed https://github.com/pypa/packaging.python.org/issues/1408 which, I think makes it easier to track the RtD redirects if we're going to be adding them for _anything_ or already have a bunch.

PR #1378, opened 2023-11-11
Consistent tabs in installing-using-pip-and-virtual-environments.rst

### LINKED ISSUE #1375: Issue with tabs in guide "Install packages in a virtual environment using pip and venv"
I wonder if it would not be better to have 3 tabs all the way. Unix and Mac are merged into 1 tab except in the "[_Prepare pip_](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#prepare-pip)" section. So the auto switch of tabs throughout the page does not work as expected.

_Originally posted by @sinoroc in https://github.com/pypa/packaging.python.org/issues/1338#issuecomment-1796293328_

jeanas:
Fixes #1375

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1378.org.readthedocs.build/en/1378/

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -155,25 +155,10 @@ Prepare pip
 It's used to install and update packages into a virtual environment.
 
 
-.. tab:: Unix
+.. tab:: Unix/macOS
 
-    You can make sure that pip is up-to-date by running:
-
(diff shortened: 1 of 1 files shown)


sinoroc (approved):
Thanks!

willingc (approved):
Thanks @jeanas for cleaning this up :sunny:

PR #1379, opened 2023-11-12
Improve namespace packaging guide

### LINKED ISSUE #689: Confusing example linked in packaging-namespace-packages
The section on the [packaging-namespace-packages page for "Native namespace packages"](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages) includes a link to the [native namespace package example project](https://github.com/pypa/sample-namespace-packages/tree/master/native)

This is a bit confusing because the packaing.python.org page uses `find_namespace_packages` in setup.py but the sample package [explicitly lists the packages to include](https://github.com/pypa/sample-namespace-packages/blob/df7530eeb8fa0cb7dbb8ecb28363e8e36bfa2f45/native/pkg_b/setup.py#L31): `packages=['example_pkg.b']`

I'm not sure if there's a fix I can think of here. Maybe update the sample project?

### LINKED ISSUE #1222: stop recommending deprecated pkg_resources namespace packages
The recommendations here are now outdated because plg_resources is deprecated: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#creating-a-namespace-package

> This method is recommended if you need compatibility with packages already using this method or if your package needs to be zip-safe.

https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkg-resources-style-namespace-packages

> If you are creating a new distribution within an existing namespace package that uses this method then it’s recommended to continue using this as the different methods are not cross-compatible and it’s not advisable to try to migrate an existing package.

### LINKED ISSUE #1305: Clarify `__init__.py` syntax in namespace packaging guide
The [docs](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) say:

> The __init__.py file for the namespace package needs to contain only the following:
> Every distribution that uses the namespace package must include an identical __init__.py

I wonder:

1. does "only" mean "a semantically equivalent set of instructions that appear first in the file"?
2. does "identical" mean "semantically equivalent" or "character-by-character identical"?

To give an example, consider the two following `__init__.py` modules. Are they valid paired-modules for a non-native namespace package?

```python
# arbitrary comments like:
# pylint: disable=missing-module-docstring
__path__ = __import__("pkgutil").extend_path(__path__, __name__) # type: ignore
# note the type ignore comment as well
```

```python
from pkgutil import extend_path
__path__ = extend_path(__path__, __name__)
```

Note how neither the textual contents nor the AST is the same, but how they actually do the same thing.

Does it still work if there are other statements after extending `__path__`?

```python
from pkgutil import extend_path
__path__ = extend_path(__path__, __name__)
print("done extending __path__")
```

Don't hesitate to redirect me to the code responsible for this :slightly_smiling_face:

chrysle:
This addresses the linked issues, and adds a `pyproject.toml` example for how to find native or implicit namespace packages with `setuptools`.

- Closes #1222
- Closes #689
- Closes #1305

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1379.org.readthedocs.build/en/1379/

### DIFF
--- a/source/guides/packaging-namespace-packages.rst
+++ b/source/guides/packaging-namespace-packages.rst
@@ -19,7 +19,7 @@ have the following package structure:
             __init__.py
             ...
         module_b.py
-    setup.py
+    pyproject.toml
 
 And you use this package in your code like so::
 
@@ -31,17 +31,19 @@ Then you can break these sub-packages into two separate distributions:
 .. code-block:: text
 
     mynamespace-subpackage-a/
-        setup.py
-        mynamespace/
-            subpackage_a/
-                __init__.py
+        pyproject.toml
+        src/
+            mynamespace/
+                subpackage_a/
+                    __init__.py
 
     mynamespace-subpackage-b/
-        setup.py
-        mynamespace/
-            subpackage_b/
-                __init__.py
-            module_b.py
+        pyproject.toml
+        src/
+            mynamespace/
+                subpackage_b/
+                    __init__.py
+                module_b.py
 
 Each sub-package can now be separately installed, used, and versioned.
 
@@ -73,44 +75,29 @@ Native namespace packages
 Python 3.3 added **implicit** namespace packages from :pep:`420`. All that is
 required to create a native namespace package is that you just omit
 :file:`__init__.py` from the namespace package directory. An example file
-structure:
+structure (following :ref:`src-layout <setuptools:src-layout>`):
 
 .. code-block:: text
 
     mynamespace-subpackage-a/
-        setup.py # AND/OR pyproject.toml, setup.cfg
-        mynamespace/ # namespace package
-        # No __init__.py here.
-            subpackage_a/
-                # Sub-packages have an __init__.py.
-                __init__.py
-                module.py
+        pyproject.toml # AND/OR setup.py, setup.cfg
+        src/
+            mynamespace/ # namespace package
+            # No __init__.py here.
+                subpackage_a/
+                    # Sub-packages have an __init__.py.
+                    __init__.py
+                    module.py
 
 It is extremely important that every distribution that uses the namespace
 package omits the :file:`__init__.py` or uses a pkgutil-style
 :file:`__init__.py`. If any distribution does not, it will cause the namespace
 logic to fail and the other sub-packages will not be importable.
 
-Because ``mynamespace`` doesn't contain an :file:`__init__.py`,
-:func:`setuptools.find_packages` won't find the sub-package. You must
-use :func:`setuptools.find_namespace_packages` instead or explicitly
-list all packages in your :file:`setup.py`. For example:
-
-.. code-block:: python
-
-    from setuptools import setup, find_namespace_packages
-
-    setup(
-        name='mynamespace-subpackage-a',
-        ...
-        packages=find_namespace_packages(include=['mynamespace.*'])
-        # or list a single package explicitly:
-        # packages=['mynamespace.subpackage_a'],
-    )
-
-The same can be accomplished by replacing the :file:`setup.py` in the
-namespace packages' parent directory with a :file:`pyproject.toml`,
-with the following contents:
+The ``src-layout`` directory structure allows automatic discovery of packages
+by most :term:`build backends <Build Backend>`. If however you want to manage
+exclusions or inclusions of packages yourself, this is possible to be configured
+in the top-level :file:`pyproject.toml`:
 
 .. code-block:: toml
 
@@ -119,12 +106,37 @@ with the following contents:
     build-backend = "setuptools.build_meta"
 
     [tool.setuptools.packages.find]
-    where = ["."]
-    include = ["mynamespace.*"]
+    where = ["src/"]
+    include = ["mynamespace.subpackage_a"]
 
     [project]
     name = "mynamespace-subpackage-a"
     ...
+
+The same can be accomplished with a :file:`setup.cfg`:
+
+.. code-block:: ini
+
+   [options]
+    package_dir =
+        =src
+    packages = find_namespace:
+
+    [options.packages.find]
+    where = src
+
+Or :file:`setup.py`:
+
+.. code-block:: python
+
+    from setuptools import setup, find_namespace_packages
+
+    setup(
+        name='mynamespace-subpackage-a',
+        ...
+        packages=find_namespace_packages(where='src/', include=['mynamespace.subpackage_a']),
+        package_dir={"": "src"},
+    )
 
 :ref:`setuptools` will search the directory structure for implicit namespace
 packages by default.
@@ -170,12 +182,14 @@ To create a pkgutil-style namespace package, you need to provide an
 
 .. code-block:: text
 
-    setup.py # AND/OR pyproject.toml, setup.cfg
-    mynamespace/
-        __init__.py  # Namespace package __init__.py
-        subpackage_a/
-            __init__.py  # Sub-package __init__.py
-            module.py
+    mynamespace-subpackage-a/
+        src/
+            pyproject.toml # AND/OR setup.cfg, setup.py
+            mynamespace/
+                __init__.py  # Namespace package __init__.py
+                subpackage_a/
+                    __init__.py  # Sub-package __init__.py
+                    module.py
 
 The :file:`__init__.py` file for the namespace package needs to contain
 the following:
@@ -215,22 +229,24 @@ To create a pkg_resources-style namespace package, you need to provide an
 
 .. code-block:: text
 
-    setup.py # AND/OR pyproject.toml, setup.cfg
-    mynamespace/
-        __init__.py  # Namespace package __init__.py
-        subpackage_a/
-            __init__.py  # Sub-package __init__.py
-            module.py
+    mynamespace-subpackage-a/
+        src/
+            pyproject.toml # AND/OR setup.cfg, setup.py
+            mynamespace/
+                __init__.py  # Namespace package __init__.py
+                subpackage_a/
+                    __init__.py  # Sub-package __init__.py
+                    module.py
 
 The :file:`__init__.py` file for the namespace package needs to contain
-**only** the following:
+the following:
 
 .. code-block:: python
 
     __import__('pkg_resources').declare_namespace(__name__)
 
-**Every** distribution that uses the namespace package must include an
-identical :file:`__init__.py`. If any distribution does not, it will cause the
+**Every** distribution that uses the namespace package must include such an
+:file:`__init__.py`. If any distribution does not, it will cause the
 namespace logic to fail and the other sub-packages will not be importable.  Any
 additional code in :file:`__init__.py` will be inaccessible.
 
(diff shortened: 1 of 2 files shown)


source/guides/packaging-namespace-packages.rst
```diff
@@ -86,13 +77,14 @@ structure:
 
 .. code-block:: text
 
-    setup.py
-    mynamespace/
+    mynamespace-subpackage-a/
+        setup.py # AND/OR pyproject.toml, setup.cfg
```
sinoroc, line 81: I would for sure not put `setup.py` in first place here. My personal preference is to only put standardized things in `pyproject.toml` so I would put `setup.cfg` in first place, but I guess the preferred way in the community is to put as much as possible in `pyproject.toml` so that is probably what should be done here as well.

source/guides/packaging-namespace-packages.rst
```diff
@@ -112,8 +104,31 @@ list all packages in your :file:`setup.py`. For example:
         name='mynamespace-subpackage-a',
         ...
         packages=find_namespace_packages(include=['mynamespace.*'])
+        # or list a single package explicitly:
+        # packages=['mynamespace.subpackage_a'],
     )
 
+The same can be accomplished by replacing the :file:`setup.py` in the
+namespace packages' parent directory with a :file:`pyproject.toml`,
```
sinoroc, line 112:
```suggestion
project's source tree root directory with a :file:`pyproject.toml`,
```

source/guides/packaging-namespace-packages.rst
```diff
@@ -125,8 +140,25 @@ the `native namespace package example project`_.
     only support Python 3 and pkgutil-style namespace packages in the
     distributions that need to support Python 2 and 3.
 
+
+Legacy namespace packages
+-------------------------
+
+These to methods, that were used to create namespace packages prior to :pep:`420`,
```
sinoroc, line 147:
```suggestion
These two methods, that were used to create namespace packages prior to :pep:`420`,
```

sinoroc (commented):
Maybe it is a bigger change... but have you considered migrating the examples in this document to `src` layout?

As far as I know, with the `src` layout the build backends are able to automatically find the packages (namespace and regular), no configuration needed.

chrysle:
That #@**-%$!!! pre-commit blames me for typos my PR doesn't include!

sinoroc:
> That #@**-%$!!! pre-commit blames me for typos my PR doesn't include!

That is surprising. I have no idea why _codespell_ did not trigger on those issues before...

chrysle:
> I have no idea why _codespell_ did not trigger on those issues before...

Well it did, but only **after** #1372 was merged. I guess that useless pre-commit-ci wasn't configured at the time that PR was opened.

chrysle:
Thanks for reviewing, I addressed your comments! 

> Maybe it is a bigger change... but have you considered migrating the examples in this document to `src` layout?

I was unsure if I shouldn't just leave the examples under the _Legacy namespace packages_ section alone, considering they're "archived"?

source/guides/packaging-namespace-packages.rst
```diff
@@ -57,52 +59,73 @@ import object short).
...
-            # Sub-packages have __init__.py.
-            __init__.py
-            module.py
+    mynamespace-subpackage-a/
+        pyproject.toml # AND/OR setup.py, setup.cfg
+        src/
+            mynamespace/ # namespace package
+            # No __init__.py here.
+                subpackage_a/
+                    # Sub-packages have an __init__.py.
```
sinoroc, line 88: I would align indentation like this:

```suggestion
        src/
            mynamespace/ # namespace package
                # No __init__.py here.
                subpackage_a/
                    # Sub-packages have an __init__.py.
```

source/guides/packaging-namespace-packages.rst
```diff
@@ -57,52 +59,73 @@ import object short).
...
 package omits the :file:`__init__.py` or uses a pkgutil-style
 :file:`__init__.py`. If any distribution does not, it will cause the namespace
 logic to fail and the other sub-packages will not be importable.
 
-Because ``mynamespace`` doesn't contain an :file:`__init__.py`,
-:func:`setuptools.find_packages` won't find the sub-package. You must
-use :func:`setuptools.find_namespace_packages` instead or explicitly
-list all packages in your :file:`setup.py`. For example:
+The ``src-layout`` directory structure allows automatic discovery of packages
+by most :term:`build backends <Build Backend>`. If however you want to manage
```
sinoroc, line 98: Add link to "[_src layout vs flat layout_](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)"? It does not seem to have a target yet though, so it might need to be added.

source/guides/packaging-namespace-packages.rst
```diff
@@ -57,52 +59,73 @@ import object short).
...
+
+    [project]
+    name = "mynamespace-subpackage-a"
+    ...
+
+The same can be accomplished with a :file:`setup.cfg`:
+
+.. code-block:: ini
+
+   [options]
```
sinoroc, line 120:
```suggestion
    [options]
```

source/guides/packaging-namespace-packages.rst
```diff
@@ -111,9 +134,13 @@ list all packages in your :file:`setup.py`. For example:
     setup(
         name='mynamespace-subpackage-a',
         ...
-        packages=find_namespace_packages(include=['mynamespace.*'])
+        packages=find_namespace_packages(where='src/', include=['mynamespace.subpackage_a']),
+        package_dir={"": "src"},
```
sinoroc, line 138: Fixing quotes for consistency of this clock of code.
Maybe we should have some formatting rules for the whole guide?

```suggestion
        package_dir={'': 'src'},
```

source/guides/packaging-namespace-packages.rst
chrysle: Done!

source/guides/packaging-namespace-packages.rst
chrysle: Yes, we might dig out a hook somewhere.

sinoroc: Something that formats Python code embedded inside reStructuredText? No idea if that exists, but maybe it does exist, I have no idea.

source/guides/packaging-namespace-packages.rst
```diff
@@ -57,52 +59,73 @@ import object short).
...
-            # Sub-packages have __init__.py.
-            __init__.py
-            module.py
+    mynamespace-subpackage-a/
+        pyproject.toml # AND/OR setup.py, setup.cfg
+        src/
+            mynamespace/ # namespace package
+                # No __init__.py here.
+                subpackage_a/
+                    # Sub-packages have an __init__.py.
```
sinoroc, line 88:
```suggestion
                    # Regular import packages have an __init__.py.
```

source/guides/packaging-namespace-packages.rst
```diff
@@ -57,52 +59,73 @@ import object short).
...
+by most :term:`build backends <Build Backend>`. See :ref:`src-layout-vs-flat-layout`
+for more information. If however you want to manage exclusions or inclusions of packages
+yourself, this is possible to be configured in the top-level :file:`pyproject.toml`:
+
+.. code-block:: toml
+
+    [build-system]
+    requires = ["setuptools", "setuptools-scm"]
+    build-backend = "setuptools.build_meta"
+
```
sinoroc, line 107: If the goal of this code block is to show how to configure setuptools to handle namespace packages, then maybe we can skip the `build-system` table. What do you think?

source/guides/packaging-namespace-packages.rst
```diff
@@ -138,22 +182,24 @@ To create a pkgutil-style namespace package, you need to provide an
...
-        subpackage_a/
-            __init__.py  # Sub-package __init__.py
-            module.py
+    mynamespace-subpackage-a/
+        src/
+            pyproject.toml # AND/OR setup.cfg, setup.py
+            mynamespace/
+                __init__.py  # Namespace package __init__.py
+                subpackage_a/
+                    __init__.py  # Sub-package __init__.py
```
sinoroc, line 191:
```suggestion
                    __init__.py  # Regular package __init__.py
```

source/guides/packaging-namespace-packages.rst
```diff
@@ -183,22 +229,24 @@ To create a pkg_resources-style namespace package, you need to provide an
...
```
sinoroc, line 238:
```suggestion
                    __init__.py  # Regular package __init__.py
```

sinoroc (commented):
Some more suggestions. Sorry that I am bringing yet another round of those instead of all at once... :/

source/guides/packaging-namespace-packages.rst
chrysle: How about the following?

```suggestion
    [build-system]
    ...

```

source/guides/packaging-namespace-packages.rst
chrysle: I find this a bit confusing, since the namespace package _is_ a sub-package and the root directory is named accordingly?

Probably 
```suggestion
                    __init__.py  # Regular sub-package __init__.py
```

is more like it ;-)

sinoroc: My reasoning for this is that from the point of view of the Python interpreter there are two kinds of importable packages: the regular ones without qualifier and the "namespace" packages.

Whether they are sub-packages or not does not matter.

For example:

```none
.
├── namespace
│   └── regular
│       └── __init__.py
└── regular
    ├── __init__.py
    └── namespace
```

```pycon
>>> import namespace
>>> namespace
<module 'namespace' (<_frozen_importlib_external.NamespaceLoader object at 0x7fb3d531f090>)>
>>> import namespace.regular
>>> namespace.regular
<module 'namespace.regular' from '/tmp/tmp.wRR0QJjvej/namespace/regular/__init__.py'>
>>> import regular
>>> regular
<module 'regular' from '/tmp/tmp.wRR0QJjvej/regular/__init__.py'>
>>> import regular.namespace
>>> regular.namespace
<module 'regular.namespace' (<_frozen_importlib_external.NamespaceLoader object at 0x7fb3d4b6af90>)>
```

I know that this is valid for PEP420 native namespace packages, and this example is not PEP420, but from my point of view we should try to stick to this terminology.

I guess I am fine with "regular sub-package".

source/guides/packaging-namespace-packages.rst
sinoroc: I tend do write `# ...` or `# [...]` in such cases. But I am fine with anything, even the whole build system table, I don't mind much, it is just nit-picking.

sinoroc (approved):
I approve, no reason to block this.

source/guides/packaging-namespace-packages.rst
chrysle: Let's do it like you suggest, then. I think I'm the one who's nit-picking.

source/guides/packaging-namespace-packages.rst
chrysle: Using my suggestion since that comes to you use in several other places.

chrysle:
@sinoroc Thank you for you thorough review!

willingc (approved):
This looks good to me. Thanks!

PR #1380, opened 2023-11-12
Warn of exceeding PyPI project size limit in GHA guide

### LINKED ISSUE #777: GitHub Actions tutorial: Warn about filling up test pypi package limit if you publish every commit as suggested

chrysle:
Closes #777 

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1380.org.readthedocs.build/en/1380/

### DIFF
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -219,6 +219,14 @@ And it'll publish any push to TestPyPI which is useful for
 providing test builds to your alpha users as well as making
 sure that your release pipeline remains healthy!
 
+.. attention::
+
+  If your repository has frequent commit activity and every push is uploaded
+  to TestPyPI as described, the project might exceed the
+  `PyPI project size limit <https://pypi.org/help/#project-size-limit>`_.
+  The limit could be increased, but a better solution may constitute to
+  use a PyPI-compatible server like :ref:`pypiserver` in the CI for testing purposes.
+
 .. note::
 
    It is recommended to keep the integrated GitHub Actions at their latest


willingc (approved):
lgtm thanks.

willingc (approved):
Thanks @chrysle. Merging :tada:

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -219,6 +219,14 @@ And it'll publish any push to TestPyPI which is useful for
 providing test builds to your alpha users as well as making
 sure that your release pipeline remains healthy!
 
+.. attention::
+
+  If your repository has frequent commit activity and every push is uploaded
+  to TestPyPI as described, the project might exceed the
+  `PyPI project size limit <https://pypi.org/help/#project-size-limit>`_.
+  The limit could be increased, but a better solution may constitute to
+  use a PyPI-compatible server like :ref:`pypiserver` in the CI for testing purposes.
```
webknjaz, line 228: This suggests a different use case. The original one I had in mind was keeping the release pipeline healthy, which included going through a (Test)PyPI auth.
This isn't bad, just something to consider...

willingc: @webknjaz Do you want to create a PR with your intended use case? :sunny:

Issue #1384, opened 2023-11-13
Replace News document with an automated monthly changelog


willingc:
Help give maintainers and readers information on what has been changed/added/removed in the Packaging Guide by creating a monthly sectioned changelog. This would be similar to the existing News Page. Ideally, we would automate this.

pradyunsg:
I think the news section was meant to cover "what happened in Python packaging" but, evidently, we haven't been keeping up with that and reorienting this to focus on how has this website is evolving seems like a good idea.

The question is: How do you want to do this? Do we want curated entries, AKA https://github.com/twisted/towncrier style where someone explicitly writes the human-facing summary, or do we want to just use the commit history to provide this context as is common with the NodeJS community / GitHub's automated release notes thingie?

chrysle:
>  or do we want to just use the commit history to provide this context as is common with the NodeJS community / GitHub's automated release notes thingie?

I personally am more comfortable with the authors describing their work "human-readable" / in detail themselves.

willingc:
Thanks for the context @pradyunsg. Perhaps 2 changes in 2 separate PRs:
- [ ] Add the original News purpose as a note at the top of the news file
- [ ] Automate a changelog (given the time/effort to make human readable, I would just pull the actual commits)

Alternatively, we could just clarify the News purpose with a note for now.

chrysle:
> I'm not sure what exactly you mean by this, but I'd be wary of adding work to the contribution process.

I mean [newsfragments](https://towncrier.readthedocs.io/en/stable/tutorial.html#creating-news-fragments).

> (given the time/effort to make human readable, I would just pull the actual commits)

I'd argue we don't always have a PR frequency like now.

PR #1387, opened 2023-11-14
Add note that News doc is not currently being updated

willingc:
Partially addresses #1384 

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1387.org.readthedocs.build/en/1387/


pradyunsg (approved):
LGTM, although the linters aren't happy.

PR #1388, opened 2023-11-14
Import PEP 723 specification

jeanas:
This resembles https://github.com/pypa/packaging.python.org/pull/1372, https://github.com/pypa/packaging.python.org/pull/1340 and https://github.com/pypa/packaging.python.org/pull/1337.

Like those other PRs, it has one first commit importing the PEP text verbatim, and a second commit changing it. This is so that reviewers can see exactly what changed (content explaining the motivation and alternatives is trimmed, and some references are adjusted).

I'm fully aware that PEP 723 was only provisionally/conditionally accepted. I'd still like to have it in the PUG so that if/when it becomes accepted, all the announcements can read "Tool X now supports inline script metadata" instead of "Tool X now supports PEP 723". I guess it's clear by now that I don't like concepts referred to by PEP number :wink:

The provisional status is clearly reflected in the text.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1388.org.readthedocs.build/en/1388/

### DIFF
--- a/source/specifications/declaring-build-dependencies.rst
+++ b/source/specifications/declaring-build-dependencies.rst
@@ -50,6 +50,8 @@ should consider it an error.
 
 .. TODO: move elsewhere
 
+.. _pyproject-tool-table:
+
 tool table
 ----------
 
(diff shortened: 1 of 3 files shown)


ofek (approved):

willingc (approved):
Thanks @jeanas

PR #1396, opened 2023-11-14
A unified pyproject.toml specification

jeanas:
There are currently two specifications describing `pyproject.toml`, [Declaring build system dependencies](https://packaging.python.org/en/latest/specifications/declaring-build-dependencies) and [Declaring project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata). This split corresponds to history: first PEP 518 came, then PEP 621.

The two specifications are closely related (they define tables of the same file), and it's also slightly confusing to have the `[tool]` table defined in the "Declaring build system dependencies" spec. This PR proposes a unified spec.

Each commit is atomic so you can check that I didn't sneak in spec changes :)

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1396.org.readthedocs.build/en/1396/


source/specifications/pyproject-toml.rst
```diff
@@ -1,23 +1,105 @@
...
+``pyproject.toml`` specification
+================================
 
 .. warning::
 
    This is a **technical, formal specification**. For a gentle,
    user-friendly guide to ``pyproject.toml``, see
    :ref:`writing-pyproject-toml`.
 
+The ``pyproject.toml`` file acts as a configuration file for packaging-related
```
willingc, line 13: Perhaps add a note at the top here that the specification history include the 2 peps.

source/specifications/pyproject-toml.rst
```diff
@@ -1,23 +1,105 @@
...
+=================================================================
+
+The ``[build-system]`` table declares any Python level dependencies that
+must be installed in order to run the project's build system
+successfully.
+
+.. TODO: merge with PEP 517
+
+The ``[build-system]`` table is used to store build-related data.
+Initially,  only one key of the table is valid and is mandatory
```
willingc, line 35:
```suggestion
Initially, only one key of the table is valid and is mandatory
```

willingc (approved):
Nice merge of the 2 specs. I would suggest putting the note at the top to identify the history of the 2 PEPs.

source/specifications/pyproject-toml.rst
sinoroc: In general I believe it might be helpful to have the PEP numbers in a relatively prominent place, because a lot of people still know these things by their PEP number. And hopefully soon (I guess in a couple of years) it will be that useful anymore.

Maybe a note at the top would be nice.

> Note: This is the specification that was introduced in PEP xxx and PEP xxx.

If not, it is not a deal breaker.

sinoroc (approved):

sinoroc (approved):

pradyunsg:
IIUC, this is approved already and reviewed by multiple people -- trusting the group here and marking this into the queue without doing a review myself.

hugovk:
See also https://github.com/pypa/packaging.python.org/pull/1434, would it be okay to keep the `declaring-project-metadata` reference?

We're relying on it to build PEP 621 in the PEPs repo via intersphinx. And/or we can fix PEP 621 to use the new ref.

But it might be worth keeping the old refs to avoid breaking intersphinx for other people/avoid linkrot. (We usually do this in https://github.com/python/devguide.)

PR #1397, opened 2023-11-15
Add discussion of package formats, expanding "Wheel vs Egg" discussion

jeanas:
This a discussion that I [would have found useful](https://discuss.python.org/t/sdists-for-pure-python-projects/25191) not long ago. I seem [not to be alone](https://discuss.python.org/t/why-build-wheels-for-pure-python-projects/33262).

After merging this, a redirect should be added in RTD from https://packaging.python.org/en/latest/discussions/wheel-vs-egg/ to https://packaging.python.org/en/latest/discussions/package-formats/ .

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1397.org.readthedocs.build/en/1397/


source/discussions/wheel-vs-egg.rst
```diff
@@ -1,55 +0,0 @@
-.. _`Wheel vs Egg`:
```
pradyunsg, line 1: We should keep this page, as an orphan pointing to the new package-formats page.

merwok: I don’t think you can redirect fragment identifiers

pradyunsg: Hmm... Actually, the suggested Web UI redirect is a good idea.

pradyunsg: The web UI redirect forwards fragments.

sinoroc: Naive question:

Aren't we locking ourselves to readthedocs unnecessarily? Wouldn't it be possible to implement the redirects at the Sphinx level with something like https://setuptools-scm.readthedocs.io/en/latest/ ?

merwok: did you paste the wrong link?

sinoroc: > did you paste the wrong link?

Absolutely. Fixed. Thanks :)

pradyunsg: > Aren't we locking ourselves to readthedocs unnecessarily?

Yes, but the vendor lock-in isn't particularly strong in these cases and I'm not as skeptical of them as a more general corporation personally.

jeanas:
Fixed the merge conflict.

Review would still be appreciated 🙂

source/glossary.rst
```diff
@@ -73,9 +73,8 @@ Glossary
     Egg
 
         A :term:`Built Distribution` format introduced by :ref:`setuptools`,
-        which is being replaced by :term:`Wheel`.  For details, see
-        :doc:`The Internal Structure of Python Eggs <setuptools:deprecated/python_eggs>` and
-        `Python Eggs <http://peak.telecommunity.com/DevCenter/PythonEggs>`_
+        which has been replaced by :term:`Wheel`.  For details, see
+        :ref:`egg-format`.
```
sinoroc, line 77: What about something like this?

```suggestion
        which has been deprecated in favor of :term:`Wheel`.  For details, see
        :ref:`egg-format`.
```

source/discussions/package-formats.rst
```diff
@@ -0,0 +1,193 @@
...
+Linux and macOS), or :ref:`the command line interface of Python's tarfile module
+<python:tarfile-commandline>` on any platform.
+
+Sdists serve several purposes in the packaging ecosystem. When :ref:`pip`, the
+standard Python package installer, cannot find a wheel to install, it will fall
+back on downloading a source distribution, compiling a wheel from it, and
+installing the wheel. Furthermore, sdists are often used as the package source
+by downstream packagers (such as Linux distributions, Conda, Homebrew and
+MacPorts on macOS, ...), who, for various reasons, may prefer them over, e.g.,
+pulling from a Git repository.
```
sinoroc, line 42: I guess it does not matter much, but...

Is it true? I was under the impression that re-packagers prefer using the source code repositories rather than sdists nowadays. I am sorry I can not find any quote/source right now, I guess I must have read this on <https://discuss.python.org> at some point.

source/discussions/package-formats.rst
```diff
@@ -0,0 +1,193 @@
...
+
+A source distribution is recognized by its file name, which has the form
+:samp:`{package_name}-{version}.tar.gz`, e.g., ``pip-23.3.1.tar.gz``.
+
+.. TODO: provide clear guidance on whether sdists should contain docs and tests.
+   Discussion: https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578
```
sinoroc, line 48: Yes that would be useful info to present, but that should probably be a different document, maybe a guide. It might be wise to create a ticket (if there isn't one yet) and remove the TODO.

source/discussions/package-formats.rst
```diff
@@ -0,0 +1,193 @@
...
+
+If you are wondering about ``.pyc`` bytecode files: they are not included in
+wheels, since they are cheap to generate, and including them would unnecessarily
+force a huge number of packages to distribute one wheel per Python version
+instead of one single wheel. Instead, installers like :ref:`pip` generate them
+while installing the package.
+
+With that being said, there are still important differences between sdists and
+wheels, even for pure Python projects. Wheels are meant to contain exactly what
+is to be installed, and nothing more. In particular, wheels should never include
```
sinoroc, line 85: > exactly what is to be installed, and nothing more

Not sure we need to be so strict here. As said a few lines below, there are files such as `RECORD`, `METADATA`, and I guess depending on the cases, a (small) number of other files like the license.

This is nitpick, it can stay like this, I do not have better phrasing to offer

sinoroc (approved):
Looks good to me.

Some comments/nitpicks, but they should not block getting this merged

source/glossary.rst
jeanas: I personally prefer the stronger wording. PyPI does not accept eggs anymore, pip has dropped support for installing them and is soon even going to remove all support for detecting/listing/showing/uninstalling them (pypa/pip#12330). I'd like to avoid giving the impression that there is still some time to move away from eggs, because their last users really need to migrate now.

source/discussions/package-formats.rst
jeanas: I think it depends on the packagers (hence “may”). I don't have much experience with Linux distro packaging but I believe both practices exist in the wild. I seem to remember someone on d.p.o. complaining about the stability of hashes of GitHub-generated archives w.r.t. changes in the compression algorithm or something like that.

source/discussions/package-formats.rst
jeanas: Ok, I'll create an issue. However, IMHO, if it is a separate guide it should at least be linked here.

source/discussions/package-formats.rst
jeanas: Well, these special files are also technically installed…

source/discussions/package-formats.rst
merwok: I have seen strong opinions on both sides, even in the same distro.

source/discussions/package-formats.rst
jeanas: https://github.com/pypa/packaging.python.org/issues/1494

source/glossary.rst
sinoroc: Indeed... works for me :)

(Consider this solved. I can not solve it. Has something really changed about this? I think I have seen others wondering why they can not resolve a thread when they expected to be able to.)

source/discussions/package-formats.rst
sinoroc: OK. Consider this solved from my side :)

source/discussions/package-formats.rst
sinoroc: Right. Solved :)

source/discussions/package-formats.rst
jeanas: Assuming that this thread can be resolved.

sinoroc: @jeanas From my point view, yes it can be resolved. But there is no "resolution" button that I can click.

chrysle (approved):

chrysle:
Thank you!

zahlman:
I want to dispute a point made in this discussion:

> Wheel archives do not include `.pyc` files. Therefore, when the distribution only contains Python files (i.e. no compiled extensions), and is compatible with Python 2 and 3, it’s possible for a wheel to be “universal”, similar to an sdist.

This differs from the explanation offered [for wheels specifically](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#binary-distribution-format):

> Wheel, being an installation format that is intended to work across multiple versions of Python, does not *generally* include .pyc files.

(Emphasis mine.)

There is nothing in the description there (nor in the [original PEP](https://peps.python.org/pep-0427/)), as far as I can tell, which *forbids* `.pyc` files in a wheel. In fact, by my understanding, it's perfectly valid to publish a wheel (as long as the wheel tags specify the Python version, since the bytecode will be specific to that version) which contains *only* `.pyc` representations of the Python code; or, indeed, where any Python code included is only a minimal wrapper for a native-code compiled library. (As far as I'm aware, there's no requirement for packages on PyPI to be open-source, either.)

If `.egg` files were *required* (I can't recall) to contain `.pyc` files (or maybe it's just that Setuptools was the only game in town, and it only made eggs with `.pyc` files?), and were thus *necessarily* specific to a Python version, this should be clarified (e.g.: "Wheel archives do not *have to* include `.pyc` files"). Otherwise the point should be removed entirely, since the distinction doesn't really exist.

jeanas:
This is a discussion intended to be a helpful explanation for confused user, not a standard. Virtually nobody includes `.pyc`s in wheels in practice (and I don't believe anybody would recommend that, I haven't tested if they actually work in `pip`), so this wording is fine IMHO.

zahlman:
I wish to dispute/clarify a point made here:

> Wheel archives do not include `.pyc` files. Therefore, when the distribution only contains Python files (i.e. no compiled extensions), and is compatible with Python 2 and 3, it’s possible for a wheel to be “universal”, similar to an sdist.

I'll take it as given that Eggs *required* `.pyc` files - since the format wasn't well specified, and as far as I can tell it was just Setuptools making them. However, as far as I can tell, wheels absolutely *may* contain `.pyc` files - and may even use them exclusively. (Or use native code almost-exclusively, with only a minimal Python wrapper.) Elsewhere in the guide, [where wheels are described specifically](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#binary-distribution-format), we see (emphasis mine):

> Wheel, being an installation format that is intended to work across multiple versions of Python, does not *generally* include .pyc files.

[The original specification](https://peps.python.org/pep-0427/) uses the same wording, and even explicitly contemplates wheels containing `.pyc` files:

>  All possible entries are hashed, including any generated files such as .pyc files, but not RECORD which cannot contain its own hash.

The description in the PEP of the difference might be better:

> Wheel is an installation format; egg is importable. Wheel archives do not need to include .pyc and are less tied to a specific Python version or implementation. Wheel can install (pure Python) packages built with previous versions of Python so you don’t always have to wait for the packager to catch up.

zahlman:
(Sorry if you saw the double-post - I had some kind of glitch with the GitHub submission form.)

I'm pretty sure that wheels are intended to be allowed to include `.pyc` files based on other wording in PEP 427, even if it isn't common practice (all emphasis mine):

>Wheel, being an installation format that is intended to work across multiple versions of Python, does not *generally* include .pyc files.

(same as in the other PyPA documentation)

> All possible entries are hashed, including any generated files *such as .pyc files*, but not RECORD which cannot contain its own hash.

> Wheel is an installation format; egg is importable. Wheel archives *do not need to* include .pyc and are less tied to a specific Python version or implementation.

If installers like Pip are meant to reject or filter such files then that really ought to be made explicit *somewhere*. But I would be surprised. Wheel is a "binary" distribution and open-source is not supposed to be a requirement to distribution Python code, to the best of my knowledge. (In my testing, Pip will happily unpack the `.pyc` files, just like any other files in the archive.)

PR #1400, opened 2023-11-16
Exclude precommit codespell check for translation po files

### LINKED ISSUE #1389: Relax codespell requirements for CI of translated docs
#1359 indicates failure due to dictionary items missing for different translations. We should relax codespell so that it is not a required check for translations. Alternatively, set it to report results but not block a PR.

willingc:
Closes #1389 

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1400.org.readthedocs.build/en/1400/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
   rev: v2.2.6
   hooks:
   - id: codespell
-    args: ["-L", "ned,ist,oder","--skip","*.po"]
+    args: ["-L", "ned,ist,oder", "--skip", "*.po"]
 
 - repo: local
   hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
   rev: v2.2.6
   hooks:
   - id: codespell
-    args: ["-L", "ned,ist,oder"]
+    args: ["-L", "ned,ist,oder","--skip","*.po"]
```
hugovk, line 18: The littlest of nits, otherwise looks good 👍 

```suggestion
    args: ["-L", "ned,ist,oder", "--skip", "*.po"]
```

Oh, and did the ned, ist and/or oder come from a po file? If so, we can remove those from this config too.

hugovk (approved):

pradyunsg (approved):

PR #1407, opened 2023-11-18
Fix virtualenv description

zahlman:
Okay, I managed to figure out how to do this properly (I think). Maybe I'm the only one who feels this way, but sometimes GitHub is harder for me to understand than Git itself.

Anyway, here are a few quick changes to update the description of virtualenv and make it read more cleanly.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1407.org.readthedocs.build/en/1407/

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -341,11 +341,9 @@ virtualenv
 virtualenv is a tool for creating isolated Python :term:`Virtual Environments
 <Virtual Environment>`, like :ref:`venv`. Unlike :ref:`venv`, virtualenv can
 create virtual environments for other versions of Python, which it locates
-using the PATH environment variable. It also provides
-additional functionality, compared to :ref:`venv`, by supporting Python
-2.7 and by providing convenient features for configuring, maintaining,
-duplicating, and troubleshooting the virtual environments. For more
-information, see the section on :ref:`Creating and using Virtual
+using the PATH environment variable. It also provides convenient features for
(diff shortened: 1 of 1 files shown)


source/key_projects.rst
```diff
@@ -333,18 +333,17 @@ it's fast and secure, it's maintained, and it reliably works.
 virtualenv
 ==========
 
-:doc:`Docs <virtualenv:index>` |
+`Docs <https://virtualenv.pypa.io/en/stable/index.html>`__ |
```
willingc, line 336: I would keep the original link since it will provide a better search experience across documentation.

willingc (commented):
I like the updated prose. I would like the link to remain as written. If you update the link back to the original content, I will be happy to rereview.

source/key_projects.rst
zahlman: I used a separate commit for that change exactly because I anticipated that there would be a reason this part was as it was. I have no objection to cherry-picking (cherry-unpicking?) this, but I want to understand why the hover text for the current link says

> (in virtualenv v0.1)

and whether that is actually as intended. What is supposed to be significant about that version? The hover text for e.g. the Conda doc link does not list a version number.

willingc: Interesting re: hover text. It does seem that anything of the form `:doc:` does have a hover text which I agree is confusing since the version seems off. I'm going to merge this as is. Thanks @zahlman and good eye for detail.

@

willingc (approved):

PR #1411, opened 2023-11-19
Proofread 'How to modernize a setup.py based project?'

hugovk:
Thank you for this guide, it's very useful!

Here's a few small proofreading suggestions.

* No need to italicise "pip"

* If we "**MUST NOT**" use the old commands, then upgrade "can be used" to "should be used" for the replacement commands

* Simplify "Current recommendation" -> "Recommendation". When it's no longer current, let's update it. (Same for the whole website.)

* Fix typo: `build-sytem` -> `build-system`

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1411.org.readthedocs.build/en/1411/


sinoroc (approved):

willingc (approved):
Looks good. Thanks for editing @hugovk.

PR #1415, opened 2023-11-24
Add "python setup.py --version" replacement command

jeanas:
https://discuss.python.org/t/replacement-for-setup-py-version

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1415.org.readthedocs.build/en/1415/

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -38,7 +38,11 @@ What commands should be used instead?
 +---------------------------------+                                        |
 | ``python setup.py bdist_wheel`` |                                        |
 +---------------------------------+----------------------------------------+
+| | ``python setup.py --version`` |  ``python -m setuptools_scm``          |
+| | (when using setuptools-scm_)  |                                        |
++---------------------------------+----------------------------------------+
 
+.. _setuptools-scm: https://setuptools-scm.readthedocs.io
 
 .. [#needs-build] This requires the :ref:`build` dependency.
     It is recommended to always build and publish both the source distribution
--- a/source/guides/modernize-setup-py-project.rst
+++ b/source/guides/modernize-setup-py-project.rst
@@ -37,7 +37,11 @@ and their recommended replacement commands should be used instead:
 +---------------------------------+                                        |
 | ``python setup.py bdist_wheel`` |                                        |
 +---------------------------------+----------------------------------------+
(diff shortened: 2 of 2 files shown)


sinoroc:
I am not convinced we need to add this.

I believe it was essential to write about `install`, `develop`, `sdist`, and `bdist_wheel` because using those can actually be damaging (they do produce out of date results), and also there is a clear obvious replacement. The replacement commands are even build backend agnostic and based on standards, so this was a no-brainer (from my point of view).

With things like `--name`, and `--version` (and also `test` and so on) this is not as straightforward in my opinion. As far as I know there is no real "damage" that can occur when using those. And also there is no "standard" replacement.

Maybe we could list those (`--name` `--version` `test` and so on) in a separate table further down below in a less prominent place where we can expand a bit more on the use cases.

sinoroc:
I thought merely having setuptools-scm installed had side effects, but looking at the documentation it seems like it doesn't. It has to be listed in build dependencies and there has to be a setuptools-scm table in `pyproject.toml` for it to be enabled. Is that right? If yes, then I might be more likely to agree with this suggestion.

sinoroc:
I am thinking this should be a "suggestion" rather than a recommendation.

I am thinking maybe we could add a whole new subsection of suggestions (soft recommendations), something along the lines of:

> ## What commands should be used instead?
> [...]
>
> ## What about other commands?
> 
> `setup.py --version` => a possible replacement (among others) is to rely on *setuptools-scm*: `python -m setuptools-scm`
> `setup.py test` => we recommend using a test runner, for example *pytest*
> `setup.py check` and `setup.py upload` => we suggest using *twine*: `python -m twine check` and `python -m twine upload`
>
> ## What about custom commands?
> [...]

What do you think? Would that work for you?

I do not want the guide to be seen as recommending specific tools unless they are the absolute obvious choice (such as *build*). Maybe *twine* could be one of those, I guess it is recommended in some other places in this guide. But for *setuptools-scm* (and *pytest*) I am a bit more reluctant, first there are other alternatives, and second the whole question about version strings is its own (hotly debated) topic.

sinoroc:
> this doesn't imply any sort of recommendation

Yes but this table is for "hard" recommendations, as the header of the right-hand column says: "_Current recommendation_". That is why I suggest moving this to another location outside of this recommendation table.

sinoroc:
Sorry for being picky about this. But this motivated me to open my own pull request on this. So effort is not lost. :)

PR #1418, opened 2023-11-26
Improve discussion "Is `setup.py` deprecated?"

sinoroc:
Add alternative solutions for more commands.

GitHub: relates to https://github.com/pypa/packaging.python.org/pull/1415

---

Direct link to added section in preview: https://python-packaging-user-guide--1418.org.readthedocs.build/en/1418/discussions/setup-py-deprecated/#what-about-other-commands

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1418.org.readthedocs.build/en/1418/

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -5,18 +5,20 @@
 Is ``setup.py`` deprecated?
 ===========================
 
-No, :term:`setup.py` is not deprecated,
-it is a valid configuration file for :ref:`setuptools`
+No, :term:`setup.py` and :ref:`setuptools` are not deprecated.
+
+Setuptools is perfectly usable as a :term:`build backend`
+for packaging Python projects.
+And :file:`setup.py` is a valid configuration file for :ref:`setuptools`
 that happens to be written in Python, instead of in *TOML* for example
 (a similar practice is used by other tools
 like *nox* and its :file:`nox.py` configuration file,
 or *pytest* and :file:`conftest.py`).
 
-And of course *setuptools* itself is not deprecated either.
+However, ``python setup.py`` and the use of :file:`setup.py`
+as a command line tool are deprecated.
 
-It is however deprecated to run ``python setup.py`` as a command line tool.
-
-This means for example that the following commands **MUST NOT** be run anymore:
+This means that commands such as the following **MUST NOT** be run anymore:
 
 * ``python setup.py install``
 * ``python setup.py develop``
@@ -28,7 +30,7 @@ What commands should be used instead?
 =====================================
 
 +---------------------------------+----------------------------------------+
-| Deprecated                      | Current recommendation                 |
+| Deprecated                      | Recommendation                         |
 +=================================+========================================+
 | ``python setup.py install``     | ``python -m pip install .``            |
 +---------------------------------+----------------------------------------+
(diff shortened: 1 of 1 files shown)


source/discussions/setup-py-deprecated.rst
```diff
@@ -79,6 +79,33 @@ The command ``python setup.py install`` was deprecated
...
+
+.. _pytest: https://docs.pytest.org/
+
+
+.. rubric:: ``python setup.py check`` and ``python setup.py upload``
+
+A trusted replacement is :ref:`twine`:
+
+* ``python -m twine check``
+* ``python -m twine upload``
```
chrysle, line 97: Feel free to work it in as you like, but I think this should be mentioned.

```suggestion
.. rubric:: ``python setup.py register``, ``python setup.py check`` and ``python setup.py upload``

A trusted replacement is :ref:`twine`:

* ``python -m twine register``
* ``python -m twine check``
* ``python -m twine upload``
```

source/discussions/setup-py-deprecated.rst
```diff
@@ -28,7 +28,7 @@ What commands should be used instead?
 =====================================
 
 +---------------------------------+----------------------------------------+
-| Deprecated                      | Current recommendation                 |
+| Deprecated                      | Recommendation                         |
```
chrysle, line 31: Should we mention deprecated commands like `bdist_rpm` here or elsewhere that don't have a replacement?

source/discussions/setup-py-deprecated.rst
chrysle: `devpi` for example does require it. Should we limit the guide to PyPI?

source/discussions/setup-py-deprecated.rst
sinoroc: I guess we could.

Is there a full list somewhere of all the commands `setup.py` used to have?

Then I do not know what we should say about those? Should we say something like this: "_This guide does not make suggestions of replacement solutions for those deprecated commands._"?

source/discussions/setup-py-deprecated.rst
sinoroc: Added `register` as well. Thanks.

source/discussions/setup-py-deprecated.rst
chrysle: > Is there a full list somewhere of all the commands `setup.py` used to have?

Just call `python setup.py --help`, using any setup.py you might have flying around.
 
> Then I do not know what we should say about those? Should we say something like this: "_This guide does not make suggestions of replacement solutions for those deprecated commands._"?

There's for example [PEP 527](https://peps.python.org/pep-0527/) explaining why building dists for several file types was removed from PyPI, and packaging tools adopted it (see pypa/setuptools#1988).

source/discussions/setup-py-deprecated.rst
```diff
@@ -16,7 +16,7 @@ And of course *setuptools* itself is not deprecated either.
 
 It is however deprecated to run ``python setup.py`` as a command line tool.
 
-This means for example that the following commands **MUST NOT** be run anymore:
+This means that commands such as the following **MUST NOT** be run anymore:
```
willingc, line 19: I believe we can improve this response beginning at Line 8 by being more direct in the question's response.

```
No, :term:`setup.py` and :ref:`setuptools` are not deprecated.
:file:`setup.py` is valid as a Python configuration file for :ref:`setuptools` (similar to the * TOML*:file:`pyproject.toml`).

However, ``python setup.py`` and the use of :term:`setup.py` as a command line tool are deprecated.
The following commands **must not** be run anymore:

File: source/discussions/setup-py-deprecated.rst [hunk 4]
```diff
@@ -79,6 +79,34 @@ The command ``python setup.py install`` was deprecated
 in setuptools version *58.3.0*.
 
 
+What about other commands?
```
willingc, line 82: What are some additional replacements for deprecated ``python setup.py`` commands?

willingc (commented):
Thanks for the PR.

source/discussions/setup-py-deprecated.rst
sinoroc: Added here: https://python-packaging-user-guide--1418.org.readthedocs.build/en/1418/discussions/setup-py-deprecated/#remaining-commands

source/discussions/setup-py-deprecated.rst
sinoroc: Thanks

willingc (approved):
Thanks @sinoroc. Merging.

PR #1423, opened 2023-11-30
DOC: Document RECORD modification

larsoner:
Distilled based on the conversation in https://github.com/pypa/pip/issues/11835#issuecomment-1833546937 and related comments (cc @pfmoore). Happy to modify language or whatever, or close if this recommendation is incorrect!

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1423.org.readthedocs.build/en/1423/

### DIFF
--- a/source/specifications/recording-installed-packages.rst
+++ b/source/specifications/recording-installed-packages.rst
@@ -175,14 +175,15 @@ attempt to uninstall or upgrade the package.
 (This restriction does not apply to tools that rely on other sources of information,
 such as system package managers in Linux distros.)
 
-.. warning::
+.. note::
 
-   If your package dynamically generates files at runtime, you will need to update
-   the ``RECORD`` file to reflect the new files. If you do not, the files will be
-   retained during uninstallation (e.g., by ``pip uninstall my_package``) likely leading
-   to issues such as the package remaining importable but errantly appearing as a
-   namespace package. See :gh:`pypa/pip/issues/11835` for discussion.
-
+   It is *strongly discouraged* for an installed package to modify itself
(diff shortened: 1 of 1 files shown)


pfmoore (changes requested):
I agree that this is a subtle "gotcha" that probably warrants a warning, but we need to be careful not to make too much of this particular issue. After all, it's basically just saying "if you don't maintain the metadata the way you should, then things will break". Which is fairly self-evident when you put it like that.

I don't expect the people who are writing files into site-packages will even think they *need* to read this spec - so this probably isn't going to signficantly reduce how often this issue comes up in practice. It's more about raising general awareness, and should be written from that perspective.

pfmoore:
> However, I think that this text could lead to the impression that dynamically changing installed packages is officially supported (as long as you do update RECORD). Is there consensus on that?

I think the opposite is true, the presumption is that only installers should be writing to package directories. It's not a *rule* (how would you even define it without having to pin down what counts as an installer?) but it's definitely baked into how people think about things.

I'd be a strong -1 on anything that suggested modifying installed packages was supported. Do it if you have to, but when everything breaks, you have nobody to blame but yourself :-) And in particular, don't expect us to tell you what you need to do in order to do it safely.

larsoner:
👍 works for me! @jeanas do you want to open a new PR, comment inline as a `suggestion` so you can get credit, or want me to just copy-paste here? Would be nice for you to get credit but if you don't care then I can easily copy-paste

larsoner:
Put in @jeanas wording, @pfmoore do you want to look?

willingc (approved):

Issue #1316, opened 2023-10-18
The tutorial https://packaging.python.org/en/latest/tutorials/packaging-projects/ does not work with hatch


LLyaudet:
Hello,

I hope this is the right place :) after https://github.com/python/cpython/issues/111009
This tutorial
https://packaging.python.org/en/latest/tutorials/packaging-projects/
says:
"this tutorial uses [Hatchling](https://packaging.python.org/en/latest/key_projects/#hatch) by default, but it will work identically with [setuptools](https://packaging.python.org/en/latest/key_projects/#setuptools), [Flit](https://packaging.python.org/en/latest/key_projects/#flit), [PDM](https://packaging.python.org/en/latest/key_projects/#pdm), and others that support the [project] table for [metadata](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata)."

But it seems it was converted from a tutorial with setuptools.
Hatchling does not seem to work well when the source of the package is in a src/ folder.
My repository here
https://github.com/LLyaudet/django-monkey-patches
demonstrates the problem.
I had to switch from hatch to setuptools to have a wheel that is correct.
Otherwise code was installed in lib/python3.11/site-packages/src/django-monkey-patches instead of lib/python3.11/site-packages/django_monkey_patches.

Best regards,
   Laurent Lyaudet

sinoroc:
@LLyaudet Where is the version of your example that reproduces the issue? In other words where is the version of your example project that uses *hatchling* as build back-end? The link in your post points to a project with *setuptools* as build back-end.

LLyaudet:
@sinoroc  Version 1.0.0 was build with hatchling, versions 1.0.1 and 1.1.0 were build with setuptools. Look at the commits history ; there is only 8 commits right now.
First dist is here : https://github.com/LLyaudet/django-monkey-patches/commit/b721dab517741ce2980f768b73cd4cc531d048ab
The previous commit creates file pyproject.toml with hatchling : https://github.com/LLyaudet/django-monkey-patches/commit/a720461cba0aa6e4796b1a3cc9711b267a6ac6a3

LLyaudet:
git clone and git checkout to this previous commit and you can reproduce the issue :)

LLyaudet:
or if you only want to check the result of install : `pip install django-monkey-patches==1.0.0`.

ofek:
The tutorial works perfectly if you follow it as-is (as recommended). The following line may be misleading:

[...]

PR #1424, opened 2023-11-30
 #1316 Update packaging-projects.rst with a tip on good module naming 

LLyaudet:
----

📚 Documentation preview 📚: https://python-packaging-user-guide--1424.org.readthedocs.build/en/1424/discussions/choosing-a-module-name/


LLyaudet:
https://github.com/pypa/packaging.python.org/issues/1316

LLyaudet:
Hello @jeanas :)

I don't know what kind of pedagogy study supports this:
"""
Offer only minimum, necessary, explanation

If the learner doesn’t need an explanation in order to complete the tutorial, don’t explain.
"""
When I learn, I always ask myself much more questions than what is required to just follow instructions in a tutorial.
And usually I try to "poke the system" by trying things that are slightly distinct from the tutorial.
And when I teach I always give more context information than the strict minimum.
I understand it may not always be welcome, but that's why I put this tip at the end as I said.
And that I suggested it may be in an "advanced tutorial".

> Also see: https://devguide.python.org/documentation/style-guide/#audience
Please be specific. The tone of this link is contemptuous on some part.
Please write here what part of this link you think should apply in this situation.

Best regards,
    Laurent Lyaudet

LLyaudet:
> who is looking for vindication
Moreover, there was no vindication when I opened https://github.com/pypa/packaging.python.org/issues/1316.
Third paragraph makes a free trial of intent to anyone who makes a mistake.

source/tutorials/packaging-projects.rst
```diff
@@ -538,7 +538,38 @@ some things you can do:
...
+
+   - The following works:
+
+     .. code-block:: python
+
+        >>> import abcd_1234
+        >>> from abcd_1234 import something
+
+   - Hatch will recognize that the module corresponding to the package is abcd_1234 instead of defaulting to src and building a not working wheel.
+   More information about `Python imports`_ and its grammar_.
```
webknjaz, line 566: Looks like this might be misindented. Or, maybe, an empty line is missing after this one.

webknjaz (changes requested):
Thanks for the attempt to improve this tutorial! We can't accept this change _as is_, though.

This change is made to a tutorial document, but this document type is expected to include simple instructions exclusively, exactly as is, without any deviations. It probably fits guides or discussions better. Frankly, #1425 is probably what the guide needs instead.
I'm also concerned that this mostly duplicates the explanation of how Python itself works, which isn't directly related to packaging.

Like many other projects over at least half a decade, we're doing our best to follow the Diátaxis framework for writing technical documentation. I witnessed it gaining a rising popularity since at least 2018, since it's formalized and preached by a respected member of the Python community. Some projects/organizations have their own variants of techwriting guidelines that mostly resemble Diátaxis or are direct derivatives of it.
In PyPUG, we also aspire to follow it. Please, refer to https://diataxis.fr which has a lot of information about the different aspects of using it.
Moreover, you can find the author on the Write The Docs Slack instance where there's a corresponding channel: https://www.writethedocs.org/slack/. Daniele is quite responsive there, and this place has a very high concentration of experienced techwriters, exchanging ideas and discussing practical applications of this framework and other things (in separate channels).

I'm afraid, this issue is not a place for debating what Diátaxis got right and wrong. What's important here is that the maintainers and frequent contributors agree that this is the way to go and everybody is expected to respect the convention.

source/tutorials/packaging-projects.rst
LLyaudet: Yes, there was a missing empty line after the item in the list. But preview on GitHub did not complain. I added the empty line.

source/tutorials/packaging-projects.rst
```diff
@@ -538,7 +538,38 @@ some things you can do:
...
 
+.. _Python identifier: https://docs.python.org/3/reference/lexical_analysis.html#identifiers
+.. _importlib: https://docs.python.org/3/library/importlib.html
+.. _question: https://stackoverflow.com/questions/8350853/how-to-import-module-when-module-name-has-a-dash-or-hyphen-in-it
+.. _Python imports: https://docs.python.org/3/reference/import.html
+.. _grammar: https://docs.python.org/3/reference/grammar.html
```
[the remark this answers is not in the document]
LLyaudet: Done, sorry. I did not know. I corrected directly on GitHub but the preview on GitHub does not work for intersphinx links. I reread twice. Hopefuly, I did not mess up the syntax.

LLyaudet:
Hello,
I corrected the points given in the review.
Best regards,
   Laurent Lyaudet

LLyaudet:
Hello,
> > I corrected the points given in the review.
> 
> I wouldn't say so. The references are invalid (they must be identifiers Sphinx understands, not filenames). The location of the changes hasn't changed either. Both are hard blockers.

Sorry, I corrected the references. It should be ok now.
I switched the content in a new discussion Choosing a Module Name.
And linked between it and Packaging Python Projects.
Let me know if you see anything else to improve :).

Best regards,
  Laurent Lyaudet

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,32 @@
+Choosing a Module Name
+=========================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
+
+Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your module name.
+The PyPI project/dist name and the Python module may differ slightly.
```
webknjaz, line 7: Moreover, one PyPI project/dist may ship more than one module or importable package — it only possible that one matches the name, others can't.

webknjaz:
FTR, I'm not entirely convinced this deserves such attention, as others mentioned earlier. So I'll be waiting for other PyPA folks to share more opinions before deciding to merge.

source/discussions/choosing-a-module-name.rst
webknjaz, line 1: It seems like it's best to have this absorbed in #1426, it's much more comprehensive and there's a section for this.

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,32 @@
...
+For example, your package in pyproject.toml and on PyPI may have the name abcd-1234.
+But a module named abcd-1234 would be cumbersome to import in Python,
+since it isn't a valid identifier.
+(There is a way to import it anyway, see :doc:`importlib <python:library/importlib>` and this question_.)
+
+.. code-block:: python
```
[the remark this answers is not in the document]
LLyaudet: I copy pasted from the tutorial. I will correct the tutorial also then.

source/discussions/choosing-a-module-name.rst
LLyaudet: I added it. It somehow complicates the succession of sentences, but I did not found a clever way to explain the progression from the tutorial with: same name > slightly varying name > many distinct names. Feel free to restructure, if you find a more elegant phrasing.

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,32 @@
+Choosing a Module Name
+=========================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
+
+Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your module name.
+The PyPI project/dist name and the Python module may differ slightly.
+For example, your package in pyproject.toml and on PyPI may have the name abcd-1234.
```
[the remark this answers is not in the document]
LLyaudet: done :)

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,33 @@
+Choosing a Module Name
```
sinoroc, line 1: Is there a specific reason why we talk about "module" here? I think I would rather write "import package" instead. With mention that the rule also applies to "import modules" (obviously?).

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,33 @@
...
+=========================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
+
+Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your module name.
+The PyPI project/dist name and the Python module may differ slightly.
+Moreover, one PyPI project/dist may ship more than one module or importable package — it is only possible that one matches the name, others can't.
+For example, your package in :file:`pyproject.toml` and on PyPI may have the name ``abcd-1234``.
+But a module named ``abcd-1234`` would be cumbersome to import in Python,
```
sinoroc, line 10: Not only cumbersome, but completely unnatural and against the long-established conventions in the Python community.

source/discussions/choosing-a-module-name.rst
LLyaudet: You're right talking about import packages would be more in line with the tutorial.

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,33 @@
+Choosing a Module Name
+=========================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
```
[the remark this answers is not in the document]
LLyaudet: I think it is because it adds a link to the main tutorial and I think doing/reading the tutorial is a recommended prerequisite to this discussion.

source/discussions/choosing-a-module-name.rst
LLyaudet: Probably, I do not have knowledge about the long-established conventions. I will reread PEP 8 to see if something applies here. If you have any reference, I would be happy to add it.

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,33 @@
...
+since it isn't a valid identifier.
+(There is a way to import it anyway, see :doc:`importlib <python:library/importlib>` and this question_.)
+
+.. code-block:: pycon
+
+   >>> import abcd-1234
+   >>> from abcd-1234 import something
+
+would not work.
+But having a directory structure with ``src/abcd_1234/`` instead of ``src/abcd-1234/`` has 2 consequences:
```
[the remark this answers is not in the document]
LLyaudet: That was in part the goal of the first sentence :).

source/discussions/choosing-a-module-name.rst
```diff
@@ -0,0 +1,33 @@
+Choosing a Module Name
+=========================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
+
+Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your module name.
+The PyPI project/dist name and the Python module may differ slightly.
```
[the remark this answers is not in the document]
LLyaudet: I used the terminology in messages of @webknjaz  : https://github.com/pypa/packaging.python.org/issues/1316#issuecomment-1793318908

> Obviously, the rules for the PyPI project name, dist name and the installable content are different.

I am not an expert, but his comment helped me realize that there are 3 names : one for the project in pyproject.toml and on PyPI, one for the tar file distribution name, and one/many for the modules that can then be imported in Python code.
I do not know if the project and the dist name may differ. I have seen in some comments they were linked.

source/discussions/choosing-a-module-name.rst
LLyaudet: I changed the discussion name and the writing to talk about "import package/module".

source/discussions/choosing-a-module-name.rst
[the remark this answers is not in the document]
LLyaudet: I have read the glossary. In it, the three terms are defined "Project", "Distribution package", "Import package". I will clarify this sentence with the terms from the glossary.

sinoroc:
@LLyaudet Thanks for working on this.

Clearly we are not sure under what form this topic should appear in the guide and where. I believe @henryiii wanted only to extract it out of the "packaging tutorial", which is now done, so it is good. And I believe @webknjaz would prefer not to have two "discussion" documents on the same topic (#1426). So my suggestion is to make it into some kind of short "how-to guide" document (I suggested "tutorial" in an earlier comment, but I meant "guide"). This way we would have a [tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/), a [specification](https://packaging.python.org/en/latest/specifications/name-normalization/), a [discussion](https://github.com/pypa/packaging.python.org/pull/1426), and finally a guide (which is what [Diataxis](https://diataxis.fr/) is all about).

If we can agree on this, then we still need to find a nice way to structure and phrase all of this. Maybe @willingc can help with this part.

LLyaudet:
Hello :),

Thanks for the suggestions @sinoroc :).
It is a little late. I will look at it in detail, tomorrow I hope.
This evening, I did this : https://github.com/LLyaudet/testpypa
There is 8 subfolders: (Flit, Hacthling,PDM, setuptools) times (kebab-case, snake_case)-for import packages.
Each subfolder corresponds to a Project in the glossary.
The structure of each project is the same and roughly that of the tutorial.
There is only 5 commits : the last commit is the result of the build of the 8 projects.
Only one project failed to build : Flit crash when the import package has kebab-case.
For the 7 other projects, a distribution package is successfully created,
but it may be not importable after installation as was the case with Hatchling in my original mistake.
It is too late for me to look in detail at the contents of the distribution packages.
Here also, I hope to look in detail tomorrow.
Feel free to look at it before me :)

Best regards,
    Laurent Lyaudet

LLyaudet:
Hello :),

I finished analyzing the build results in [testpypa](https://github.com/LLyaudet/testpypa).

The following applies to my test but may differ in other contexts.
The archive `tar.gz` suffix is always `-1.0.0.tar.gz`.
The wheel `whl` suffix is always `-1.0.0-py3-none-any.whl`.
The `dist-info` suffix is always `-1.0.0.dist-info`.
The `egg-info` suffix is always `-1.0.0.egg-info` for setup-tools.
All suffixes are kebab-case.

When the correct import package is found, it is always leaved as is (no normalisation): no lowercase, no underscore instead of dash.
`whl`, `dist-info`, and `egg-info` *prefixes* are always normalized.
Thus, a mismatch between import package name and `dist-info` after install is the norm if import package name is not normalized.
 `tar.gz` prefix is normalized by Flit, Hatchling, PDM, but not by setup-tools.

- with kebab-case:
  - Flit: Fail;
  - Hatchling: no error but the wheel is broken with `src` as importable;
  - PDM: success;
  - setuptools: , success;
- with snake_case: success for Flit, Hatchling, PDM, and setuptools

I tried to be synthetic and group similar results.
Not sure what would be interesting for the documentation,
maybe just two sentences:
"If your import package name is not a valid identifier, Flit will crash at build, Hatchling will create a not working wheel with ``src`` as import package, both PDM and setuptools will build correctly.
If your import package name is not a normalized valid identifier or if you have many import packages in your project, there will also be a mismatch between ``.dist-info`` file and import package(s) name(s)."

It can replace "Hatch will recognize that the module corresponding to the package is ``abcd_1234`` instead of defaulting to ``src`` and building a not working wheel."

I'll try to improve the PR tomorrow :).

Best regards,
     Laurent Lyaudet

source/discussions/choosing-a-package-module-name.rst
```diff
@@ -0,0 +1,36 @@
+Choosing a Package/Module Name
+==============================
+
+This discussion is a complement to :doc:`/tutorials/packaging-projects`.
+
+Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your import package/module name.
```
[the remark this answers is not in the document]
LLyaudet: Sorry for the late answer. I validated your commit suggestion of the 4th. The 3rd december  I modified like this:
```
Make sure to have a valid :ref:`Python identifier <python:identifiers>` for your import package/module name.
The (PyPI) :term:`project<Project>`/:term:`distribution package <Distribution Package>` name
and the Python :term:`import package <Import Package>`/module may differ slightly.
```
So the terms are already referenced.
I found it nice to have all 3 terms referenced in the same sentence when I modified.
But I understand the need to reference one of the 3 terms in the preceding sentence since it is asked here:
<https://github.com/pypa/packaging.python.org/blob/e8a97163a1e0d7f1b882da0a41387ce1386c9d67/source/contribute.rst?plain=1#L255>
I have a question: Should I reference the glossary on each appearance of a term in the same document?
I know that when a document is long, it is nice when you can hover over any acronym and check again its meaning if it does not stuck the first time.
I think on Wikipedia, it is frequent that most occurrences are cross-links.
As of now, I kept the second reference in the following sentence.
Let me know if I should remove it.
Moreover, if there is a consensus on this question, I think it should be in contribute.rst :)

source/discussions/choosing-a-package-module-name.rst
```diff
@@ -0,0 +1,36 @@
...
+The (PyPI) :term:`project<Project>`/:term:`distribution package <Distribution Package>` name
+and the Python :term:`import package <Import Package>`/module may differ slightly.
+Moreover, one PyPI project/dist may ship more than one module or importable package — it is only possible that one matches the name, others can't.
```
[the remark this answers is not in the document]
LLyaudet: I made it a paragraph with a second sentence recommending to keep only one importable package with a name as similar as possible to the dist-info file.

LLyaudet:
Hello :),

Sorry for the delay.
I modified the contents with most of the suggestions.
I merged your 2 last suggestions @sinoroc  but I switched order and wrote it slightly differently.

I have reopened a conversation where I applied your commit suggestion, since I have a question for best practices regarding term references (only one per document or for each occurrence).
It would be nice if that best practice could be explicit in contribute.rst.

Apart from that, I included the results of my testing of the 4 build backends covered in the tutorial.
I am quite happy that it supports and illustrates further the importance of good naming.
I hope you will see this interest also :).

I build the doc with nox and everything is ok.

If you do not mind, I would like to know if it is ok to move again the document as is from discussions/ to guides/.
Or if it is not in a shape that can be considered a guide by now.
I anticipate that you will probably make more corrections and improvements.
I would feel more comfortable to put the document at the right place,
and then let other authors directly commit on the PR or after the PR is merged.
The document as is conveys the informations I wanted to see in it.
I fear that at the end the document will not look like what I wrote and it is time consuming and painful to just validate others modifications.
Sometimes I see them as improvements and sometimes I just see them as rewording which is not pleasant.
I prefer that each person authors its modifications.

Best regards,
    Laurent Lyaudet

jeanas:
Sorry, this got auto-closed because I wrote "Fixes #1424" in #1426 while I meant "Supersedes #1424". Reopening.

I do think that the info added in #1426 (now [here](https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/#how-do-distribution-package-names-and-import-package-names-compare)) is sufficient, but I will leave it to committers to choose whether to continue this, or to re-close it and also close the associated issue #1316.

LLyaudet:
Hello :),
Thanks for reopening it.
I do not think it should be closed, since it provides information that are not covered in https://github.com/pypa/packaging.python.org/pull/1426.
I'm still opened to suggestions to the correct way of giving this information in the final documentation.
At last iteration, we wanted to put it in a guide.
But I had no feedback on my last comment.
Best regards,
    Laurent Lyaudet

Issue #1408, opened 2023-11-18
Use a version-controlled file as source of truth for Read the Docs redirects


pradyunsg:
The idea is basically the same as that of https://github.com/pypa/pip/pull/11652.

The proposal is that we adopt the same here. This serves to move the ability to add redirects to everyone with the ability to commit to this repository, and makes it possible to ensure that the actually-configured redirects are more generally visibke/accessible.

The details of how it'd work are in the commit messages in the aforementioned PR.

pradyunsg:
If we decide that we want this, I think it makes sense to make this into a chunk of reusable code -- and, I'd like to have co-maintainer(s) on that code as well to avoid having a bus factor of 1 as well as avoiding the need to fix multiple places if any of the relevant platforms (GitHub Actions, RtD) evolve. 😅

pradyunsg:
X-ref https://github.com/pypa/packaging.python.org/pull/1397 where some of this was discussed.

X-ref https://github.com/pypa/packaging.python.org/issues/1401 which is likely where most of this discussion will take place.

pradyunsg:
> Is it important to keep using RTD's HTTP redirects, as opposed to HTML redirects (that we can create with sphinx-reredirects or similar)?

Let's discuss this in https://github.com/pypa/packaging.python.org/issues/1401, to avoid splitting the discussion.

jeanas:
I'm working on this. @webknjaz @di @ewdurbin Could one of you please post a screenshot of the current redirects so that I can initialize the new file with them? They should be on

https://readthedocs.org/dashboard/python-packaging-user-guide/redirects/

jeanas:
(Or you could add @pradyunsg as maintainer on RTD, since he seems interested in this...)

ewdurbin:
```
{
  "count": 55,
  "next": "https://readthedocs.org/api/v3/projects/python-packaging-user-guide/redirects/?limit=10&offset=10",
  "previous": null,
  "results": [
    {
      "_links": {
        "_self": "https://readthedocs.org/api/v3/projects/python-packaging-user-guide/redirects/21919/",
        "project": "https://readthedocs.org/api/v3/projects/python-packaging-user-guide/"
      },
      "created": "2023-12-06T20:17:57.062443Z",
      "from_url": "/specifications/declaring-build-dependencies/",
      "modified": "2023-12-06T20:20:47.137238Z",
      "pk": 21919,
      "project": "python-packaging-user-guide",
[...]

PR #1427, opened 2023-12-01
Update "install_requires vs. requirements files" discussion

jeanas:
Use pyproject.toml metadata instead of referring to setuptools' install_requires setup() keyword.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1427.org.readthedocs.build/en/1427/


source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
...
 (i.e. dependencies of your dependencies).  This is overly-restrictive, and
 prevents the user from gaining the benefit of dependency upgrades.
 
-Lastly, it's important to understand that ``install_requires`` is a listing of
-"Abstract" requirements, i.e just names and version restrictions that don't
-determine where the dependencies will be fulfilled from (i.e. from what
-index or source).  The where (i.e. how they are to be made "Concrete") is to
-be determined at install time using :ref:`pip` options. [1]_
+Lastly, it's important to understand that ``dependencies`` is a listing of
+"abstract" requirements, i.e, it just names and version restrictions, but doesn't
```
sinoroc, line 56:
```suggestion
"abstract" requirements, i.e., just names and version restrictions, but doesn't
```

sinoroc (approved):

webknjaz:
> Would you mind making a PR yourself after this one is merged? The scope of this PR is basically updating the metadata format and `install_requires` terminology; the rest is just a couple nitpicking changes like `i.e.` → `i.e.,` that I did “in passing”.

Yeah, I'm not sure about myself, just felt related so I shared. Trying to post the info in places that seem relevant first, so it's not overlooked when somebody gets to describing this.

webknjaz:
Also, kinda dislike replacing one implementation detail (`install_requires` key name in a config file) with another (`dependencies` key name in a config file). That's also why I shared that post.
The problem is that the semantics of these configuration options tends to be ambiguous and I feel like we need to start talking about "a package runtime dependencies in metadata" in place of "a config setting name". And in that context, it's not just `install_requires` but also `extras` that is a public API for declaring what a project needs in runtime.

webknjaz:
@dstufft since this document what heavily inspired by your blog post originally, would you mind sharing an opinion?

webknjaz:
Although, I missed that you already implemented some of it. @jeanas do you think we could reduce repeat use of `dependencies` as an implementation detail towards saying "abstract/loose runtime dependencies"?

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
-.. _`install_requires vs requirements files`:
```
webknjaz, line 1: Perhaps, we should keep the old reference or at least make a new one.

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
...
+==========================================
+Package dependencies vs requirements files
```
webknjaz, line 2: How about "Dependencies in package metadata vs environment requirements files"?

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
...
+====================
 
-install_requires
-----------------
+Packages can declare dependencies, i.e., other packages that they need
+to function. The standard method to do so is to set the
+:ref:`dependencies key <writing-pyproject-toml-dependencies>` in the
+``[project]`` section of a ``pyproject.toml`` file. When installing a
+package, installers like :ref:`pip` will automatically install the
+dependencies.  This should be used for packages that the project
```
webknjaz, line 14: It appears to establish a false direct link between a config for build backends and what installers do. The reality is that the installers work with package metadata directly, not with whatever configures backend.

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
...
+dependencies to specific versions, or to specify transitive dependencies
 (i.e. dependencies of your dependencies).  This is overly-restrictive, and
 prevents the user from gaining the benefit of dependency upgrades.
 
-Lastly, it's important to understand that ``install_requires`` is a listing of
-"Abstract" requirements, i.e just names and version restrictions that don't
-determine where the dependencies will be fulfilled from (i.e. from what
-index or source).  The where (i.e. how they are to be made "Concrete") is to
-be determined at install time using :ref:`pip` options. [1]_
+Lastly, it's important to understand that ``dependencies`` is a listing of
```
webknjaz, line 55: and extras?

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,90 @@
...
 just a list of :ref:`pip:pip install` arguments placed into a file.
 
-Whereas ``install_requires`` defines the dependencies for a single project,
+Whereas ``dependencies`` defines the dependencies for a single project,
 :ref:`Requirements Files <pip:Requirements Files>` are often used to define
 the requirements for a complete Python environment.
 
-Whereas ``install_requires`` requirements are minimal, requirements files
+Whereas ``dependencies`` requirements are minimal, requirements files
 often contain an exhaustive listing of pinned versions for the purpose of
```
webknjaz, line 73: This is debatable — I'd argue the constraint files exist for this instead.

source/discussions/install-requires-vs-requirements.rst
webknjaz, line 1: I just realized that the filename still has the reference to "install requires".

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+Whereas metadata dependencies are "abstract", i.e., not associated with any
+particular index, requirements files often contain pip options like
+``--index-url`` or ``--find-links`` to make requirements "concrete", i.e.,
+associated with a particular index or directory of packages. [1]_
 
-Whereas ``install_requires`` metadata is automatically analyzed by pip during an
+Whereas metadata dependencies are automatically analyzed by pip during an
 install, requirements files are not, and only are used when a user specifically
-installs them using ``python -m pip install -r``.
+installs them using :samp:`python -m pip install -r {requirement_file.txt}`.
```
sinoroc, line 97:
```suggestion
installs them using :samp:`python -m pip install --requirement requirements.txt`.
```

I prefer having the long forms of option flags, because they are more explicit, less cryptic, which is important in documentation. Users will be able to figure out the short forms on their own.

Also I do not see any reason to write some invalid syntax here (with the curly braces).

sinoroc (approved):
One nitpick, otherwise looks good to me

source/discussions/install-requires-vs-requirements.rst
sinoroc: Oh, I did not know about the curly braces, sorry :/

jeanas: Should we resolve this thread?

sinoroc: Sure

jeanas:
@webknjaz Is anything still needed here? I'd like to defer the file renaming until I get around to implementing #1408. From my point of view this is ready to be merged.

source/discussions/install-requires-vs-requirements.rst
webknjaz: I'm going to mark this as resolved for now.

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+(i.e., dependencies of your dependencies).  This is overly restrictive, and
 prevents the user from gaining the benefit of dependency upgrades.
 
-Lastly, it's important to understand that ``install_requires`` is a listing of
-"Abstract" requirements, i.e just names and version restrictions that don't
-determine where the dependencies will be fulfilled from (i.e. from what
-index or source).  The where (i.e. how they are to be made "Concrete") is to
-be determined at install time using :ref:`pip` options. [1]_
+Lastly, it's important to understand that metadata dependencies are "abstract"
+requirements, i.e., just names and version restrictions, but don't determine
```
webknjaz, line 69: I don't think having a comma there is correct.
```suggestion
requirements, i.e. just names and version restrictions, but don't determine
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+     "A >= 1, < 2",
+     "B >= 2"
+    ]
 
-It is not considered best practice to use ``install_requires`` to pin
-dependencies to specific versions, or to specify sub-dependencies
-(i.e. dependencies of your dependencies).  This is overly-restrictive, and
+It is not considered best practice to use metadata dependencies to pin
+dependencies to specific versions, or to specify transitive dependencies
+(i.e., dependencies of your dependencies).  This is overly restrictive, and
```
webknjaz, line 65: I don't think I ever saw a comma in such structures. Besides, it's inconsistent with other similar places.
```suggestion
(i.e. dependencies of your dependencies).  This is overly restrictive, and
```

source/discussions/install-requires-vs-requirements.rst
jeanas: Apparently, it's usually done in American English but not in British English?

https://jakubmarian.com/comma-after-i-e-and-e-g/

jeanas: I don't mind though, applying.

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+may use different methods. There can also be groups of optional dependencies,
+also called "extras", which are typically specified in the
+``optional-dependencies`` key of the ``[project]`` table. Both dependencies and
+extras are ultimately written by the build backend to the package's distribution
+metadata. On this page, we'll refer to these as "metadata dependencies".
 
-For example, if the project requires A and B, your ``install_requires`` would be
-like so:
+When installing a package, installers like :ref:`pip` will automatically install
+the metadata dependencies. They should be used for packages that the project
```
webknjaz, line 27: I think this is confusing. Pip installs concrete deps, but it derives them from the abstract ones. The first sentence in this paragraph makes it sounds like it skips the resolution and installs the abstract deps which is not something installable.

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
-install_requires
-----------------
+Metadata dependencies
+=====================
 
-``install_requires`` is a :ref:`setuptools` :file:`setup.py` keyword that
-should be used to specify what a project **minimally** needs to run correctly.
-When the project is installed by :ref:`pip`, this is the specification that is
-used to install its dependencies.
+Packages can declare dependencies, i.e., other packages that they need to
```
jeanas, line 16:
```suggestion
Packages can declare dependencies, i.e. other packages that they need to
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
 often contain an exhaustive listing of pinned versions for the purpose of
 achieving :ref:`repeatable installations <pip:Repeatability>` of a complete
 environment.
 
-Whereas ``install_requires`` requirements are "Abstract", i.e. not associated
-with any particular index, requirements files often contain pip
-options like ``--index-url`` or ``--find-links`` to make requirements
-"Concrete", i.e. associated with a particular index or directory of
-packages. [1]_
+Whereas metadata dependencies are "abstract", i.e., not associated with any
```
jeanas, line 90:
```suggestion
Whereas metadata dependencies are "abstract", i.e. not associated with any
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+particular index, requirements files often contain pip options like
+``--index-url`` or ``--find-links`` to make requirements "concrete", i.e.,
```
jeanas, line 92:
```suggestion
``--index-url`` or ``--find-links`` to make requirements "concrete", i.e.
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
-install_requires vs requirements files
-======================================
+============================================
+Metadata dependencies vs. requirements files
+============================================
 
+There are two main places where you will find list of "needed packages to
+install", perhaps with version constraints, like ``requests`` or
+``requests==2.31.0``. These are metadata dependencies, typically in a
+``pyproject.toml`` (or ``setup.py``) file, and requirements files, often called
```
webknjaz, line 10:
```suggestion
:file:`pyproject.toml` (or :file:`setup.cfg` / :file:`setup.py`) file, and requirements files, often called
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+``requirements.txt``. This page breaks down the differences.
```
webknjaz, line 11:
```suggestion
:file:`requirements.txt`. This page breaks down the differences.
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+=====================
 
-``install_requires`` is a :ref:`setuptools` :file:`setup.py` keyword that
-should be used to specify what a project **minimally** needs to run correctly.
-When the project is installed by :ref:`pip`, this is the specification that is
-used to install its dependencies.
+Packages can declare dependencies, i.e. other packages that they need to
+function. The standard method to do so is to set the :ref:`dependencies key
+<writing-pyproject-toml-dependencies>` in the ``[project]`` section of a
+``pyproject.toml`` file -- although other :term:`build backends <build backend>`
```
webknjaz, line 19:
```suggestion
:file:`pyproject.toml` file -- although other :term:`build backends <build backend>`
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
 
-For example, if the project requires A and B, your ``install_requires`` would be
-like so:
+When installing a package, installers like :ref:`pip` will automatically install
+the metadata dependencies. They should be used for packages that the project
+**minimally** needs to run correctly.
 
-::
+For example, suppose the project requires A and B. When using the ``[project]``
+table to declare metadata, the ``pyproject.toml`` would be like so:
```
webknjaz, line 31:
```suggestion
table to declare metadata, the :file:`pyproject.toml` would be like so:
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
-Lastly, it's important to understand that ``install_requires`` is a listing of
-"Abstract" requirements, i.e just names and version restrictions that don't
-determine where the dependencies will be fulfilled from (i.e. from what
-index or source).  The where (i.e. how they are to be made "Concrete") is to
-be determined at install time using :ref:`pip` options. [1]_
+Lastly, it's important to understand that metadata dependencies are "abstract"
+requirements, i.e. just names and version restrictions, but don't determine
+where the dependencies will be fulfilled from (from what package index or
+source). The where (i.e. how they are to be made "concrete") is to be determined
+at install time, e.g., using :ref:`pip` options. [1]_
```
webknjaz, line 72:
```suggestion
at install time, e.g. using :ref:`pip` options. [1]_
```

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
+   [project]
+   dependencies = [
+     "A >= 1, < 2",
+     "B >= 2"
+    ]
 
-It is not considered best practice to use ``install_requires`` to pin
-dependencies to specific versions, or to specify sub-dependencies
-(i.e. dependencies of your dependencies).  This is overly-restrictive, and
+It is not considered best practice to use metadata dependencies to pin
```
webknjaz, line 63: Honestly, it seems weird to read of this as "metadata dependencies". Abstract dependencies can be in requirements files, concrete pins can be in constraint files.

webknjaz:
I'd like to hear from more PyPA folks. This patch doesn't feel entirely accurate, so I welcome more inputs.

source/discussions/install-requires-vs-requirements.rst
jeanas: I've changed “will automatically install the metadata dependencies” to “will automatically resolve the metadata dependencies and install them”. There's a later paragraph that goes into more detail (“Lastly, it's important to understand…”).

source/discussions/install-requires-vs-requirements.rst
jeanas: I'm not sure I understand your comment, would you mind expanding please?

source/discussions/install-requires-vs-requirements.rst
jeanas: (I've resolved this comment and others because I've simply applied the exact edit that you suggested [manually, but GitHub would auto-resolve it if I had used “commit suggestion”]. If this is not fine please let me know.)

source/discussions/install-requires-vs-requirements.rst
webknjaz: Well, if you introduce a concept of metadata requirements (which I feel weird about by itself for some reason), and later on the document suggests that the requirements are opposite because they can contain pip options. The reality is that the requirements aren't opposite. They can have the same properties and be abstract.

I argue that more places need to be updated to reduce such confusion. What you call "metadata" deps refers to the location of specification. Just like "requirements". But that's about it. "requirements" can have the both same property of being abstract and the same semantics, except for the tool-specific additions, while the package metadata is standardized.

Perhaps I should've started this thread in a different place in the diff, though.

pfmoore: I agree with @webknjaz - "Metadata dependencies" aren't a thing. "Dependencies" is the correct term, and remains the correct term regardless of where they are declared.

Requirements files hold *requirements* - because they are pip-specific, some of the subtleties of what a "requirement" is in this context are pip-dependent, but basically you can consider them as concrete requirements used to build an environment or application. There's a subtle but important difference between "dependencies" and concrete requirements, but this text isn't making it very clear.

Constraints are a whole different thing, very much pip specific. They *add* limits to specifiers (concrete requirements or abstract dependencies) when resolving them, but they aren't dependencies themselves. And if that last sentence seems hard to follow, that's because we don't have well-defined terms for everything we're discussing here 🙁

jeanas: > I agree with @webknjaz - "Metadata dependencies" aren't a thing. "Dependencies" is the correct term, and remains the correct term regardless of where they are declared.

In the abstract, I agree. But I don't want to use just “dependencies” to explain the distinction “dependencies vs. requirements files”, because you can often hear that “the dependencies to do XXX are in this requirements files”, so a more precise term is needed to help Joe User understand that this isn't the same meaning of “dependency”. “Metadata dependencies” is the best I could come up with (I also considered “package dependencies”). Do you have another suggestion?

source/discussions/install-requires-vs-requirements.rst
```diff
@@ -1,89 +1,102 @@
...
-:ref:`Requirements Files <pip:Requirements Files>` described most simply, are
+:ref:`Requirements Files <pip:Requirements Files>`, described most simply, are
 just a list of :ref:`pip:pip install` arguments placed into a file.
 
-Whereas ``install_requires`` defines the dependencies for a single project,
-:ref:`Requirements Files <pip:Requirements Files>` are often used to define
-the requirements for a complete Python environment.
+Whereas metadata dependencies define the dependencies for a single
+project, requirements files are often used to define the requirements
+for a complete Python environment.
```
pfmoore, line 83: I know this is just re-wording the original text, but frankly I don't like it.

> metadata dependencies define the dependencies for a single project

Well, obviously. It's project metadata. This doesn't seem to be saying anything useful. And it's not a useful contrast with requirement files (IMO).

> requirements files are often used to define the requirements for a complete Python environment

Requirement files can be used for many things. They are a *mechanism* (and a pip-specific one, at that) and don't have any particular "intended use". People use requirement files to define the concrete dependencies of a standalone application, pinning exact versions to offer some level of reproducibility. That's arguably "the dependencies for the (application) project", but in a very different sense than the dependencies in `pyproject.toml` (if the application even *has* a `pyproject.toml`).

What's really being discussed here is the "abstract requirements vs concrete requirements" distinction, which is about *concepts*, not about how those concepts are implemented. The implementation is (and should be!) irrelevant, as long as the concepts are used correctly. But the implementation is important, because it's tricky to map concrete/abstract requirements onto the features available - whether we restrict ourselves to solely standards-defined features, or allow for tool-specific ones (which may be pip's requirement files, or whatever features hatch/pdm/poetry provide for "adding requirements to a project").

Things are made even harder because of the mess that is projects which are in fact applications, but which are built as if they were libraries (with entry point metadata providing a command line executable). That method of application deployment mixes up abstract and concrete requirements in a way that cannot be disentangled by a general document like this is intended to be.

I don't have a good answer here. In a very real sense, this is way too complex to be appropriate for an introductory document like this is supposed to be. But conversely, new users *need* some guidance on how to set up their dependencies, if for no other reason than many of the examples available to them are terrible, and won't teach them good practices 🙁

pfmoore:
> I'd like to hear from more PyPA folks. This patch doesn't feel entirely accurate, so I welcome more inputs.

Well, I added a review comment which sort of turned into a comment on the whole page (to an extent, both before and after the PR). I agree that this patch doesn't feel accurate, although I can't be sure how much of that is because the original material is inaccurate and the changes simply fail to improve it.

While I don't want the perfect to be the enemy of the good here, I sort of feel that a better resolution would just be to delete this page altogether. Having just re-read the current version, I'm left with the feeling that I hope new users never actually find it - and that's more because of the advice it gives than because it refers to `install_requires` rather than `pyproject.toml`...

source/discussions/install-requires-vs-requirements.rst
jeanas: > What's really being discussed here is the "abstract requirements vs concrete requirements" distinction, which is about concepts, not about how those concepts are implemented.

Ok, I understand better where you're coming from now.

We have packages which can define metadata dependencies which are always abstract, and we have requirements files which define dependencies which can be either abstract or concrete. You're framing the document as “what are abstract vs concrete dependencies”. On the other hand, I am framing it as “what are metadata dependencies vs requirements files” (as in the title).

Yes, abstract vs concrete dependencies is important here, and we need to explain the link, but realistically, the question Joe Random Beginner is going to ask themselves is “I saw package names in pyproject.toml and also in requirements.txt, what's the difference?” and that's what they're going to Google for. There could be a separate document to explain abstract vs concrete dependencies or a new section, but it's not what I'm trying to explain here (and it's not what the page before this PR is explaining either).

The text you quoted said “are *often* used”, not “are used”, so it's correct, even though it doesn't give details. Those details IMHO don't belong here.

source/discussions/install-requires-vs-requirements.rst
pfmoore: Not really. The terms really are confused. We don't have well-defined terminology here, and getting consensus on terminology isn't something we should be doing in a discussion on a single PR.

We had this debate before, a long time ago. The term "package" is horribly overloaded, so we coined the terms "distribution package" and "install package". No-one uses them. They use "package", project", and probably other terms as well, and work out what is meant from context. It sucks, but that's human communication for you 😉

And to make things worse here, **requirement files aren't standardised**. So how can we have standard terminology for something that's fundamentally tool-specific? Someone using hatch or PDM to manage their "dependencies for running the script that deploys to the internal test server" may be using something that's not a requirement file, but performs the same logical function. But their tool may not (for example) support using local filenames as requirements, unlike requirement files.

The existing text seems to come from a background of requirement-files-as-lockfiles. Which is certainly *one* valid use case, but by no means the only one.

source/discussions/install-requires-vs-requirements.rst
pfmoore: > requirements files which define dependencies which can be either abstract or concrete

Nope. I don't agree. Requirement files define *requirements*. They may not be "dependencies" at all, in that there may be nothing (except in a very abstract sense) that actually *depends* on them. In pip, a "requirement" is a very real thing. Unfortunately, it's not something that's captured in any standard, so you can't talk about (pip's) requirements in a tool-agnostic context (which this is, at least in the sense that it's not the pip documentation).

> realistically, the question Joe Random Beginner is going to ask themselves is “I saw package names in pyproject.toml and also in requirements.txt, what's the difference?”

Understood. But the real answer is "well, it's complicated..."

Do we want to give an over-simplified, and often wrong, answer here, and risk reinforcing an incorrect understanding? Personally, I don't think we do, but the original authors of this page thought we did. Maybe you prefer to just paper over the cracks and do (in effect) a search and replace on the idea `install_requires`, replacing it with "dependencies as specified in `pyproject.toml`". I don't think that's a worthwhile thing to spend time on, and I don't have any insight into how to make it "correct"[^1].

[^1]: Because to my mind, it simply isn't correct no matter how you word it.

> The text you quoted said “are often used”, not “are used”, so it's correct, even though it doesn't give details. Those details IMHO don't belong here.

I have no stomach for nitpicking over this. I don't think[^2] it's right to tie requirement files to the idea of being "the definition of an environment" like this. Take that opinion or leave it. It's not as if your PR added that text anyway, so you can legitimately say you don't intend to change the sense of the text, just the specific fact that `install_requires` is out of date.

[^2]: As a pip maintainer, someone who is fundamentally responsible for *defining* what requirement files are for.

pfmoore:
I should probably drop out of this conversation now. It's bringing back all of my reservations and frustrations with the PUG that I expressed when these pages were initially written, and which led to me burning out and deciding to leave the decisions to the people who were writing the text, even if that meant I didn't consider the result to be something I would recommend. I think I'm going the same way again, so for everyone's sake, I'll quit sooner rather than later this time.

source/discussions/install-requires-vs-requirements.rst
jeanas: > It's not as if your PR added that text anyway, so you can legitimately say you don't intend to change the sense of the text, just the specific fact that install_requires is out of date.

:shrug: It was not my intent to do more than that, but reviewers seemed to be against merging without more improvements.

Honestly, I'm having trouble understanding everything you write, probably because I lack context on pip's implementation and the terms used in it. For example, I'm not sure I completely get the distinction you make between dependencies and requirements. (I was using them in the same sense.)

If your problem is with excessive framing of requirements files as “defining environments”, then I'm happy to change that. But fundamentally, yes, I'm really just trying to get rid of setuptools-specific language and trying to do whatever it takes to get this merged.

I'm not as pessimistic as you on the possibility of getting the page to a useful state, though — and, importantly, I don't think we can just delete it without adding a redirect from it to some explanation somewhere (so it might as well be here).

jeanas: Also, this is a discussion, not a specification — the level we're trying to take people from is “I am writing a package for PyPI, where should I write my dependencies, pyproject.toml or requirements.txt”. There's a tradeoff we have to make with “little white lies” that are needed to keep the text understandable for laymen even if it occasionally hurts the eyes of a pip maintainer.

pfmoore: > Honestly, I'm having trouble understanding everything you write, probably because I lack context on pip's implementation and the terms used in it.

To a large extent, that's my fault - because we don't have well-defined meanings for the various terms we're using, it's easy to misunderstand each other. I didn't have time to clearly define and explain the concepts I was using so I went with what I thought was "common usage". Clearly it's not (demonstrating nicely the problem here...).

Let me try to clarify. This is simply my view on what needs to be part of everyone's "shared understanding". I don't want to get sucked into the question of "how do we teach this", as long as whatever we *do* present doesn't lead to a different understanding, or reinforces misconceptions that are commonly held in the community.

For me, the distinction between "concrete requirements" and "abstract requirements" is crucial, and should be a core part of any vocabulary we use when talking about dependencies/requirements/etc. The caremad article linked from this page remains a good description of the concepts - I don't know if any better description has been published since.

Building on that, the next important concept I see is the idea of what pip calls "top level requirements" - the things that the user requests an installer actually installs into an environment. These can be concrete or abstract, the key point here is that the user is explicitly asking for them. Pip's requirement files are simply a way of bundling together a set of such top-level requirements for reusability. Whether you fully pin the contents of a requirement file, or whether you view what's in such a file as concrete or abstract requirements depends entirely on what you plan on using the requirement file *for*. Note that other tools can have different ideas of what counts as a "top level requirement", or how to enable reusability. One such idea is that of "adding a requirement to a project", which IMO isn't very well-defined, but is nevertheless a common thing for people to want to do[^1]. I can't comment on how the idea of "adding a requirement to a project" fits into this, as I don't work on a tool that supports such an idea, and as a user I've never really understood what tools that *do* offer such a concept actually mean by it...

At the next level, we have package *dependencies*, which are stored in the package metadata, and typically defined in the `pyproject.toml` file in the project source tree. These are *abstract* requirements, and say what else must be present for that package to function correctly. An installer will *add* these requirements to the top-level requirements requested by the user, as part of the process of dependency resolution, which turns an install request into an actionable list of packages and versions to install.

And that's it, IMO. With these concepts, a user stands a reasonable chance of making good decisions on how to express their own project's requirements and dependencies, and of understanding how other projects do so.

A note on terminology: In the above, I'm using the specific term "requirement" (which isn't standardised) in the way pip uses it - to mean "something that can be installed". Typically this is what is referred to [in the standards](https://packaging.python.org/en/latest/specifications/dependency-specifiers/) as a "dependency specifier", but it's not always a dependency, and individual tools may allow extra options (pip allows local paths, for example). Yes, this is confusing. But I didn't decide on the terminology, and it's IMO too late to expect people to change at this point. So we live with it (and from a teaching point of view, making new users aware that the world isn't entirely perfect is a good learning experience!)

I hope this is a little clearer. As I said, it's just my view on how these ideas hang together. You're free to do whatever you want with it - I don't want to get sucked into editorial discussions on "how do we express this in the packaging guide" because I'll only end up burning out if I do.

[^1]: Step one should be "clarify the purpose of the requirement" 🙂

jeanas:
I do not have the time to shepherd PRs at the moment; if anybody feels this is worth saving, please feel free to pick it up.

PR #1428, opened 2023-12-03
Correction of two code-blocks python -> pycon after suggestion from #1424

LLyaudet:
Hello :),
I just used:
```bash
grep -r 'code-block:: python' -A 3 .
```
in case there was anything else to correct and found another spot.
Best regards,
    Laurent Lyaudet

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1428.org.readthedocs.build/en/1428/


LLyaudet:
@sinoroc @jeanas I wanted to create an issue to add a warning in the rst parser. I think it would be very easy to find code-blocks with python starting with >>> and output a warning saying it should probably be pycon instead of python. What is the right package to submit this idea ? Is it https://pypi.org/project/docutils/ ?
I will try to improve https://github.com/pypa/packaging.python.org/pull/1424 first, before looking for that in detail. But if you can help, it would be nice. I think this error may be in many more python projects with rst files.

LLyaudet:
Thanks for the validation and the merge :)

PR #1429, opened 2023-12-03
Fix bits of "Is `setup.py` deprecated?" discussion

sinoroc:
GitHub: refs https://github.com/pypa/packaging.python.org/pull/1418

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1429.org.readthedocs.build/en/1429/

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -105,7 +105,7 @@ A trusted replacement is :ref:`twine`:
 * ``python -m twine upload dist/*``
 
 .. [#not-pypi] Not necessary, nor supported on :term:`PyPI <Python Package Index (PyPI)>`.
-    But might be necessary on other :term:`package indexes <package index>`.
+    But might be necessary on other :term:`package indexes <package index>` (for example :ref:`devpi`).
 
 
 ``python setup.py --version``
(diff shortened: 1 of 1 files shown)


source/discussions/setup-py-deprecated.rst
```diff
@@ -100,17 +100,20 @@ The recommendation is to use a test runner such as pytest_.
...
 
-* ``python -m twine check``
-* ``python -m twine register``
-* ``python -m twine upload``
+* ``python -m twine check --strict dist/*``
+* ``python -m twine register dist/*.whl`` [#not-pypi]_
+* ``python -m twine upload dist/*``
+
+.. [#not-pypi] Not necessary, nor supported on :term:`PyPI <Python Package Index (PyPI)>`.
+    But might be necessary on other :term:`package indexes <package index>`.
```
chrysle, line 108:
```suggestion
    But might be necessary on other :term:`package indexes <package index>` (for example :ref:`devpi`).
```

webknjaz (approved):
Thanks!

source/discussions/setup-py-deprecated.rst
webknjaz: @sinoroc could you apply this suggestion?

PR #1434, opened 2023-12-04
Keep old reference to fix PEP 621 build

hugovk:
The removal of the `.. _declaring-project-metadata:`reference in https://github.com/pypa/packaging.python.org/pull/1396 has broken the PEPs build because [PEP 621](https://peps.python.org/pep-0621/) used to link here via Intersphinx:

```
peps/pep-0621.rst:23: WARNING: undefined label: 'packaging:declaring-project-metadata'
```

https://github.com/python/peps/actions/runs/7092751804/job/19304712978

https://github.com/python/peps/blob/ff0702ebfce6e5196a0969e609d2a45e5bd41843/peps/pep-0621.rst?plain=1#L23

I think retaining the old reference should fix it.

Instead/also, we could update PEP 621 to point to the new reference. I don't really mind, but there could be other sites also linking via intersphinx so perhaps this is the more general fix.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1434.org.readthedocs.build/en/1434/


webknjaz:
@hugovk looks like maybe an empty line is needed?

hugovk:
No, they can be next to each other, see for example https://github.com/python/devguide/blob/main/documentation/start-documenting.rst?plain=1

The other warnings were also introduced by the re-org in https://github.com/pypa/packaging.python.org/pull/1396 plus another PR landing at the same time, see https://github.com/pypa/packaging.python.org/pull/1396#issuecomment-1839385271.

sinoroc (approved):

hugovk:
We've fixed this for the PEPs by updating the reference:

https://github.com/python/peps/pull/3563

However, we're still running into build failures when updating PRs created more than a week ago, for example:


* https://github.com/python/peps/pull/3562
* https://github.com/python/peps/issues/3583
*

We can fix them by updating them from `main`, however it would still be useful to merge this to preserve the old name, and avoid surprises for others. 

Thank you!

willingc (approved):

willingc:
Thanks @hugovk.

isyafiq (commented):
hugovk:patch-1

isyafiq (commented):
gh pr checkout 1434

PR #1435, opened 2023-12-05
Do not fail when there's nothing to commit

rffontenelle:
Instead of the translation CI workflow exit with error status when there is nothing to commit, run a check for staged files and exit normally if none is found.


webknjaz (approved):
Thanks!

PR #1436, opened 2023-12-05
Prevent workflow failures on forks

jeanas:
I'm getting notifications every day that workflow runs on my fork of this repository had failures.

- Test workflow: add condition to skip the daily run on forks in the `check` job, just like in the `build` job
- Translation workflow: likewise, skip on forks

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1436.org.readthedocs.build/en/1436/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
     # This job does nothing and is only used for the branch protection
     # or multi-stage CI jobs, like making sure that all tests pass before
     # a publishing job is started.
-    if: always()
+    if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
 
     needs:
     - build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,9 @@ jobs:
     # This job does nothing and is only used for the branch protection
     # or multi-stage CI jobs, like making sure that all tests pass before
     # a publishing job is started.
-    if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
+    if: >-
+      github.repository_owner == 'pypa'
+      || github.event_name != 'schedule'
 
     needs:
     - build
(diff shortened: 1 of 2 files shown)


.github/workflows/translation.yml
```diff
@@ -15,6 +15,7 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: $${ github.repository_owner == 'pypa' }}
```
webknjaz, line 18:
```suggestion
    if: github.repository_owner == 'pypa'
```

.github/workflows/test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     # This job does nothing and is only used for the branch protection
     # or multi-stage CI jobs, like making sure that all tests pass before
     # a publishing job is started.
-    if: always()
+    if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
```
webknjaz, line 50:
```suggestion
    if: >-
      github.repository_owner == 'pypa'
      || github.event_name != 'schedule'
```

webknjaz (commented):
No need for extra wrapping. Also, I'm not sure this is what we want, anyway.
Why can't you just disable the workflow in the fork? It's done in the UI.
And if this is still needed for some reason, I'd rearrange it through reusable workflows similar to pip-tools. As for the translations one, it would probably only guard specific steps.

webknjaz:
Well, maintaining such exclusions is additional burden to upstreams. Especially on the scale. So I'd like to minimize that.

That's why I'd like the reusable workflows solution instead. A separate workflow for cron runs would be easier to manage/disable.

pradyunsg (approved):

pradyunsg:
Let's move forward with this -- it's an improvement over status quo and we can iterate to move to reusable workflows in a separate follow up.

webknjaz:
Thanks, here's the follow-up: https://github.com/pypa/packaging.python.org/pull/1475

PR #1439, opened 2023-12-05
Remove the guide from its own Intersphinx mapping

### LINKED ISSUE #1438: Avoid the intersphinx mapping to ourselves?
After the merge of #1396, builds started to fail with errors about undefined references, which was normal because some labels were removed that indeed had references pointing to them. (This was fixed in #1433.) The problem is that this was not caught before the merge, even though the GitHub merge queue was used to merge the PR. The merge queue is supposed to check the result of merging the PR branch into master (as opposed to just checking the PR branch).

I did some digging into this mystery and it seems to me that this was caused by this oddity in our `conf.py`:

```python
intersphinx_mapping = {
    # ...
    "packaging.python.org": ("https://packaging.python.org/en/latest/", None),
    # ...
}
```

So... what happened is that while the PR was not merged, the references `` :ref:`declaring-build-dependencies` `` was resolved to to be an Intersphinx reference to the deployment https://packaging.python.org ! And then, after the PR got merged, the targets were removed by the deployment and the references logically became dangling.

You can see this on the PR preview of #1396:

https://python-packaging-user-guide--1396.org.readthedocs.build/en/1396/guides/modernize-setup-py-project/

image

Here I'm hovering on the first link, which points to inside the PR preview.

image

And here I hover on the second link, which points to https://packaging.python.org instead, with a tooltip indicating an external link.

I'm going to open a PR to remove that entry in the Intersphinx mapping.

jeanas:
Fixes #1438

Also replace a few hyperlinks inside packaging.python.org with normal references.

----
:books: Documentation preview :books:: https://python-packaging-user-guide--1439.org.readthedocs.build/en/1439/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -145,7 +145,6 @@ intersphinx_mapping = {
     "nox": ("https://nox.thea.codes/en/latest/", None),
     "openstack": ("https://docs.openstack.org/glance/latest/", None),
     "packaging": ("https://packaging.pypa.io/en/latest/", None),
-    "packaging.python.org": ("https://packaging.python.org/en/latest/", None),
     "pip": ("https://pip.pypa.io/en/latest/", None),
     "pipenv": ("https://pipenv.pypa.io/en/latest/", None),
     "piwheels": ("https://piwheels.readthedocs.io/en/latest/", None),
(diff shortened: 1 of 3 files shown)


pradyunsg (approved):

pradyunsg:
Ooof, thank you for figuring this out!

webknjaz:
> from `ouroboros`

I admire your sense of humor! :laughing:

jeanas:
Well, doesn't this remind you of anything?

image

:wink:

PR #1452, opened 2023-12-15
Update packaging-projects.rst

schlich:
These instructions seem outdated.  My username was not appended to my package, and appending the package broke subsequent installs.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1452.org.readthedocs.build/en/1452/


schlich:
Yep, I had mistakenly thought that the username would automatically be appended by testpypi.  It was an easy line to miss but a surprisingly hard one to backtrack from.

PR #1450, opened 2023-12-13
Start updating distributing-packages-using-setuptools guide

jeanas:
The [Packaging and Distributing Projects](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/) guide is quite outdated. It uses `setup.py` and no `pyproject.toml`. It's also out of sync with the sample project.

I'd like to update it, but this is going to be difficult to review if I do it in a single sweeping PR, so here's a start, to be continued later. This PR removes the descriptions of `setup()` arguments that have better equivalents in `[project]`. Where these descriptions had more details than the [Writing pyproject.toml](https://packaging.python.org/guides/writing-pyproject-toml/) guide, I have moved the details there.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1450.org.readthedocs.build/en/1450/

PR #1455, opened 2023-12-17
Create a discussion on versioning

jeanas:
Move content on versioning schemes from
distributing-packages-using-setuptools.rst to a new discussion on versioning, and update it.

This is the continuation of #1450.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1455.org.readthedocs.build/en/1455/discussions/versioning/


chrysle:
Looks good, thanks! Probably mention a few projects that comply with the different versioning systems?

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+of a :ref:`local version identifier <local-version-identifiers>`, which can be
+used to identify local development builds not intended for publication, or
+modified variants of a release maintained by a redistributor.
+
+A local version identifier takes the form of a public version identifier,
+followed by "+" and a local version label. For example:
+
+.. code-block:: text
+
+   1.2.0.dev1+hg.5.b11e5e6f0b0b  # 5th VCS commit since 1.2.0.dev1 release
```
webknjaz, line 132: Can we show versions that `setuptools-scm` makes too? I feel like this is something that people would see in the wild more frequently.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+Semantic versioning is not a suitable choice for all projects, such as those
+with a regular time based release cadence and a deprecation process that
+provides warnings for a number of releases prior to removal of a feature.
+
+A key advantage of date-based versioning, or `calendar versioning <calver_>`_,
+is that it is straightforward to tell how old the base feature set of a
+particular release is given just the version number.
+
+Calendar version numbers typically take the form *year.month* (for example,
+23.10 for December 2023).
```
webknjaz, line 84: Did you mean to use "October" or "12"?
```suggestion
23.10 for October 2023).
```

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+examples of version numbers:
+
+.. code-block:: text
+
+  1.2.0.dev1  # Development release
+  1.2.0a1     # Alpha Release
+  1.2.0b1     # Beta Release
+  1.2.0rc1    # Release Candidate
+  1.2.0       # Final Release
+  1.2.0.post1 # Post Release
```
webknjaz, line 28: How about showing off something more complex? Like a pre-release of a dev-release's post-release?

Also, can we include examples using the epoch version component? https://peps.python.org/pep-0440/#version-epochs

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+Different Python projects may use different versioning schemes based on the
+needs of that particular project, but in order to be compatible with tools like
+:ref:`pip`, all of them are required to comply with a flexible format for
+version identifiers, for which the authoritative reference is the
+:ref:`specification of version specifiers <version-specifiers>`. Here are some
+examples of version numbers:
+
+.. code-block:: text
+
+  1.2.0.dev1  # Development release
```
webknjaz, line 23: Nit: in RST, we usually match the block indentation with the start of the directive name. So it'd need to start under the `c` letter of `code-block`.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+
+Semantic versioning vs. calendar versioning
+===========================================
+
+A versioning scheme is a way to interpret version numbers of a package, and to
+decide which should be the next version number for a new release of a package.
+Two versioning schemes are commonly used for Python packages, semantic
+versioning and calendar versioning.
+
+Semantic versioning is recommended for most new projects.
```
webknjaz, line 41: By... whom? It's got a lot of problems (like the maintainer and end-user perception mismatches, especially in the context of locking the envs). I think it'd be more accurate to say "semantic versioning is what people tend to be most familiar with and therefore expect to see", right?

It might be a good idea to link https://hynek.me/articles/semver-will-not-save-you/ somewhere too.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+  1.2.0       # Final Release
+  1.2.0.post1 # Post Release
+  15.10       # Date based release
+  23          # Serial release
+
+
+Semantic versioning vs. calendar versioning
+===========================================
+
+A versioning scheme is a way to interpret version numbers of a package, and to
```
webknjaz, line 36: Maybe, "formalized way to interpret semantics of the version segments"?

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+particular release is given just the version number.
+
+Calendar version numbers typically take the form *year.month* (for example,
+23.10 for December 2023).
+
+
+Other schemes
+-------------
+
+Serial versioning refers to the simplest possible versioning scheme, which
```
webknjaz, line 90: Can we link some external resource?

source/discussions/versioning.rst
jeanas: I hesitated to include this. I only did so because it's in the current guide. I don't like this recommendation either, and it feels a bit silly given that pip itself, arguably the most high-profile Python packaging project, uses calver.

I'll change it to something softer.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,139 @@
...
+
+Regardless of the base versioning scheme, pre-releases for a given final release
+may be published as:
+
+* Zero or more dev releases, denoted with a ".devN" suffix,
+* Zero or more alpha releases, denoted with a ".aN" suffix,
+* Zero or more beta releases, denoted with a ".bN" suffix,
+* Zero or more release candidates, denoted with a ".rcN" suffix.
+
+Pip and other modern Python package installers ignore pre-releases by default
```
webknjaz, line 114: It's not very accurate — if a pre-release is met in the user-requests or in the dependency declarations, that flags that the resolver should consider pre-releases for those specific projects, regardless of the explicit opt-in.

source/guides/distributing-packages-using-setuptools.rst
```diff
@@ -288,125 +288,6 @@ points (see below).
 
 
 
-
```
webknjaz, line 291: Can we link the new document from in place of the removed content?

source/discussions/index.rst
```diff
@@ -8,6 +8,7 @@ specific topic. If you're just trying to get stuff done, see
 .. toctree::
    :maxdepth: 1
 
+   versioning
```
webknjaz, line 11: Should this be the first item in the ToC? Thoughts?

webknjaz:
@jeanas the first round of review complete.

source/discussions/versioning.rst
jeanas: I'm not sure we're going to find such a resource since it seems virtually nobody advocates this kind of scheme...

source/discussions/index.rst
jeanas: Not sure. It was a gut feeling, feel free to propose another sorting.

jeanas:
> Looks good, thanks! Probably mention a few projects that comply with the different versioning systems?

Good idea; done. (I've used Sphinx, NumPy and pip.)

jeanas:
Thanks for the detailed comments! I've revised much of the page, please take another look.

source/discussions/versioning.rst
jeanas: I changed the wording to be objective ("most projects do this and that").

> It might be a good idea to link https://hynek.me/articles/semver-will-not-save-you/ somewhere too.

I don't feel very comfortable linking to a blog post on an official guide; do you think it's okay?

chrysle: > I don't feel very comfortable linking to a blog post on an official guide; do you think it's okay?

That's also been done in the new _Is setup.py deprecated?_ guide. I personally don't have any reservations.

chrysle:
Could you add an own paragraph for [ZeroVer](https://0ver.org/)?

jeanas:
> Could you add an own paragraph for [ZeroVer](https://0ver.org/)?

It's a satire, not a serious thing :)

chrysle:
I know, it would be to warn users. Maybe that's stupid.

sinoroc:
The comments on this PR seem to show that we are venturing into territory where things are much less clear-cut and I wonder whether there is a risk people come to complain because we "bless" things that they don't like. Not sure you know what I mean and maybe it is unwarranted worry from my side, I just wanted to mention it. I do not have any suggestion to offer. In the end, since it is a "discussion", it might be more forgiving...

jeanas:
@sinoroc Sorry, I don't understand your comment at all. What “things” are less clear-cut in the new document?

sinoroc:
> I don't understand your comment at all.

Fair enough :D

I think my comment can be ignored. I will hide it as well as this one.

webknjaz:
> I wonder whether there is a risk people come to complain because we "bless" things that they don't like

That's literally the story of the Python Packaging in the nutshell 😂

webknjaz:
I think adding a 0ver remark might be fun, just don't pass judgement..

source/discussions/versioning.rst
jeanas: Done; I've also linked a couple other blog posts by packaging tool maintainers.

jeanas:
- Added references to blog posts on SemVer's weaknesses
- Added link to https://0ver.org

I think the only unresolved issue right now is https://github.com/pypa/packaging.python.org/pull/1455#discussion_r1429250729

source/discussions/versioning.rst
```diff
@@ -0,0 +1,176 @@
...
+- A simple version with only two components: 23.12
+- A simple version with just one component (possible, but discouraged): 42
+- A version with an epoch: 1!1.0
+
+Projects can use a cycle of pre-releases to support testing by their users
+before a final release. In order, the steps are: alpha releases, beta releases,
+release candidates, final release. Pip and other modern Python package
+installers ignore pre-releases by default when deciding which versions of
+dependencies to install, unless explicitly requested (e.g., with
+``pip install pkg==1.1a3``).
```
sinoroc, line 37: What about `pip install --pre pkg`? Honestly I never use it so I am not sure what it does, but I know it exists.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,176 @@
...
+- *maintenance*, when they make backwards-compatible bug fixes.
+
+A majority of Python projects use a scheme that resembles semantic
+versioning. However, most projects, especially larger ones, do not strictly
+adhere to semantic versioning, since many changes are technically breaking
+changes but affect only a small fraction of users. Such projects tend to
+increment the major number when the incompatibility is high, or to signal a
+shift in the project, rather than for any tiny incompatibility
+[#semver-strictness]_.
+
```
sinoroc, line 84: Somewhere in there I would like to read that incrementing the major version number does not imply incompatible change. Sometimes projects increment their major version number for vanity purposes although the library is still 100% backwards compatible. This detail becomes important when dependent projects put unnecessary upper caps on a dependency, assuming that the next major release will be incompatible, which there is have absolutely no way of knowing until the next major release does actually happen.

source/discussions/versioning.rst
chrysle: This flag [installs](https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions) release candidates and beta versions. Should be combined with `--upgrade`, I think.

source/discussions/versioning.rst
jeanas: Done.

sinoroc (approved):

source/discussions/index.rst
jeanas: @webknjaz Friendly ping?

webknjaz: Let's leave it be, then. Marking as resolved.

source/discussions/versioning.rst
webknjaz: @chrysle no, `--upgrade` is orthogonal. `--pre` essentially allows the dependency resolve to consider pre-releases instead of rejecting them right away. It's global and allows transitive pre-releases.

But using `pip install pkg==1.1a3` allows pre-releases implicitly just for `pkg`, because it's an explicit user request, but not other packages (transitive or otherwise) — it's not global.

There's also a corner case that PEP 440 demands but pip doesn't (yet) implement — when the dependency resolution can only be satisfied with pre-releases, the resolver is supposed to allow inevitable pre-releases in order to succeed. Only one sub-case is implemented — if `pkg` *only* has pre-releases published on the PyPI and zero stable releases, `pip` will allow pre-releases for that project too.

`--upgrade` is only significant when the installed version of a package in the tree satisfies the constraints the resolver encountered, but the user wants to force the latest version of all the projects in the tree anyway.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+- A simple version (final release): 1.2.0
+- A development release: 1.2.0.dev1
+- An alpha release: 1.2.0a1
+- A beta release: 1.2.0b1
+- A release candidate: 1.2.0rc1
+- A post-release: 1.2.0.post1
+- A post-release of an alpha release (possible, but discouraged): 1.2.0a1.post1
+- A simple version with only two components: 23.12
+- A simple version with just one component (possible, but discouraged): 42
+- A version with an epoch: 1!1.0
```
webknjaz, line 30: Can we mark all the versions with double backticks? It may be weird to read things like this as normal text.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+:ref:`specification of version specifiers <version-specifiers>`. Here are some
+examples of version numbers:
+
+- A simple version (final release): 1.2.0
+- A development release: 1.2.0.dev1
+- An alpha release: 1.2.0a1
+- A beta release: 1.2.0b1
+- A release candidate: 1.2.0rc1
+- A post-release: 1.2.0.post1
+- A post-release of an alpha release (possible, but discouraged): 1.2.0a1.post1
```
webknjaz, line 27: Is it? Based on what?

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+- A simple version with only two components: 23.12
+- A simple version with just one component (possible, but discouraged): 42
```
webknjaz, line 29: Same — what's the source of discouragement?

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+
+
+.. _zerover: https://0ver.org
+.. _calver: https://calver.org
+.. _semver: https://semver.org
+.. _semver-bernat-gabor: https://bernat.tech/posts/version-numbers/
+.. _semver-brett-cannon: https://snarky.ca/why-i-dont-like-semver/
+.. _semver-donald-stufft: https://caremad.io/posts/2016/02/versioning-software/
+.. _semver-hynek-schlawack: https://hynek.me/articles/semver-will-not-save-you/
+.. _setuptools-scm: https://setuptools-scm.readthedocs.io
```
webknjaz, line 178: use intersphinx, maybe?

source/discussions/versioning.rst
jeanas: This is from the spec:

https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases

webknjaz: Okay, I see in the spec that this is noted. Resolved.

source/discussions/versioning.rst
webknjaz: I don't see that the spec discourages this. It simply mentions that 2-3 segment version numbers are more common.
So I'd drop that statement, unless there's another information source to support this claim.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+number, and to decide which should be the next version number for a new release
+of a package. Two versioning schemes are commonly used for Python packages,
+semantic versioning and calendar versioning.
+
+
+Semantic versioning
+-------------------
+
+The idea of *semantic versioning* (or SemVer) is to use 3-part version numbers,
+*major.minor.maintenance*, where the project author increments:
```
webknjaz, line 71: Where is *maintenance* coming from? https://semver.org defines the third component as *patch*.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+versions like 1.0, the comparison between 1.0 and 23.12 will go the wrong way.
+To correct this, the new version numbers should have an explicit epoch, as in
+"1!1.0", in order to be treated as more recent than the old version numbers.
+
+
+
+Semantic versioning vs. calendar versioning
+===========================================
+
+A versioning scheme is a formalized way to interpret the segments of a version
```
webknjaz, line 61: Let's emphasize that the interpretation is a reflection of the maintainer's view which is not necessarily how the end-users would perceive their version bump choice.

And also, it's important to mention that there are exceptions, like security releases that break the versioning scheme assumptions. As in patch releases having security fixes that are inevitably breaking changes by nature.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,178 @@
...
+increment the major number when the incompatibility is high, or to signal a
+shift in the project, rather than for any tiny incompatibility
+[#semver-strictness]_. Conversely, a bump of the major version number
+is sometimes used to signal significant but backwards-compatible new
+features.
+
+For those projects that do use strict semantic versioning, this approach allows
+users to make use of :ref:`compatible release version specifiers
+<version-specifiers-compatible-release>`, with the ``~=`` operator.  For
+example, ``name ~= X.Y`` is roughly equivalent to ``name >= X.Y, == X.*``, i.e.,
```
webknjaz, line 90: Not roughly, exactly equivalent:
```suggestion
example, ``name ~= X.Y`` is equivalent to ``name >= X.Y, == X.*``, i.e.,
```

webknjaz:
@jeanas I stumbled upon an article by Seth. Maybe it could be linked from somewhere too: https://sethmlarson.dev/pep-440.

source/discussions/versioning.rst
jeanas: Done.

source/discussions/versioning.rst
jeanas: Ok, done.

source/discussions/versioning.rst
jeanas: We can't, because these docs are made with MkDocs.

source/discussions/versioning.rst
jeanas: Changed

source/discussions/versioning.rst
jeanas: I'm not sure how to phrase this and if you want to add something or to rephrase the paragraph below about "strict" vs "loose" semver. Could you perhaps propose a wording or give more details on what you envision?

source/discussions/versioning.rst
webknjaz: Ah, I forgot. Well, then. Marking as resolved.

source/discussions/versioning.rst
webknjaz: Marking as resolved.

source/discussions/versioning.rst
webknjaz: Marking as resolved.

source/discussions/versioning.rst
webknjaz: Marking as resolved.

source/discussions/versioning.rst
```diff
@@ -0,0 +1,181 @@
...
+
+
+Semantic versioning vs. calendar versioning
+===========================================
+
+A versioning scheme is a formalized way to interpret the segments of a version
+number, and to decide which should be the next version number for a new release
+of a package. Two versioning schemes are commonly used for Python packages,
+semantic versioning and calendar versioning.
+
```
webknjaz, line 65:
```suggestion

.. caution::

   The decision which version number to choose is up to a
   project's maintainer. This effectively means that version
   bumps reflect the maintainer's view. That view may differ
   from the end-users' perception of what said formalized
   versioning scheme promises them.

   There are known exceptions for selecting the next version
   number. The maintainers may consciously choose to break the
   assumption that the last version segment only contains
   backwards-compatible changes.
   One such case is when security vulnerability needs to be
   addressed. Security releases often come in patch versions
   but contain breaking changes inevitably.

```

source/discussions/versioning.rst
webknjaz: Here you go https://github.com/pypa/packaging.python.org/pull/1455/files#r1468739775

jeanas:
@webknjaz I've addressed the latest review.

webknjaz (approved):
Thanks 🙏

PR #1456, opened 2023-12-18
glossary.rst: Update link to requirement specifier syntax to PEP 508

noamraph:
The current link doesn't actually give the syntax of the requirement specifier. PEP 508 is much better.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1456.org.readthedocs.build/en/1456/

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -209,10 +209,8 @@ Glossary
     Requirement Specifier
 
        A format used by :ref:`pip` to install packages from a :term:`Package
-       Index`. For an EBNF diagram of the format, see the
-       `pkg_resources.Requirement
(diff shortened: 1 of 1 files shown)


jeanas (changes requested):
Thanks, but this should rather be `` :ref:`dependency-specifiers` `` to link to the canonical specification on https://packaging.python.org/en/latest/specifications/dependency-specifiers/. PEP 508 is a historical document, it's the change proposal, not the now authoritative reference. It should have an “Attention” banner like [PEP 440](https://peps.python.org/440) for example, but I was waiting to have imported a few more PEPs into packaging.python.org before submitting PR adding these in batch on the PEPs repo.

noamraph:
@jeanas thanks, I updated the PR.

chrysle (approved):

PR #1457, opened 2023-12-18
docs: update testpypi guide for api token

miketheman:
With the upcoming change to require 2FA for all users (already enforced on TestPyPI), update guidance to point the user in the right direction.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1457.org.readthedocs.build/en/1457/


webknjaz (approved):
Thanks!

PR #1460, opened 2023-12-19
Correct path to venv scripts on Windows

### LINKED ISSUE #1459: Windows Activate.{ps1/bat} is under Scripts/ not bin/
https://github.com/pypa/packaging.python.org/blob/49ceefc15b5707553c847682390d5a26ef15e4f2/source/guides/installing-using-pip-and-virtual-environments.rst?plain=1#L87-L93

https://docs.python.org/3/library/venv.html#how-venvs-work

image

jeanas:
Fixes #1459

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1460.org.readthedocs.build/en/1460/

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -88,7 +88,7 @@ shell's ``PATH``.
 
     .. code-block:: bat
 
-        .venv\Scripts\Activate.bat
+        .venv\Scripts\activate
 
 To confirm the virtual environment is activated, check the location of your
 Python interpreter:
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -88,7 +88,7 @@ shell's ``PATH``.
 
     .. code-block:: bat
 
-        .venv\bin\Activate.bat
+        .venv\Scripts\Activate.bat
 
 To confirm the virtual environment is activated, check the location of your
 Python interpreter:
@@ -118,7 +118,7 @@ filepath that includes the ``.venv`` directory, by ending with the following:
 
     .. code-block:: bat
 
-        .venv\bin\python.exe
+        .venv\Scripts\python.exe
 
 
 While a virtual environment is activated, pip will install packages into that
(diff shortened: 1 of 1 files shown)


source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -88,7 +88,7 @@ shell's ``PATH``.
 
     .. code-block:: bat
 
-        .venv\bin\Activate.bat
+        .venv\Scripts\Activate.bat
```
chrysle, line 91:
```suggestion
        .venv\Scripts\activate
```

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -118,7 +118,7 @@ filepath that includes the ``.venv`` directory, by ending with the following:
 
     .. code-block:: bat
 
-        .venv\bin\python.exe
+        .venv\Scripts\python.exe
```
chrysle, line 121:
```suggestion
        .venv\Scripts\python
```

chrysle (commented):
I believe simpler is better here?

source/guides/installing-using-pip-and-virtual-environments.rst
jeanas: Oh, does that also exist? (I don't have Windows, and the venv docs don't mention it.)

chrysle: Windows will automatically detect the executable. Therefore this is better because it will work on both PowerShell (_.ps1_) and cmd.exe (_.bat_). I also don't use Windows mainly.

NSQY: This seems like a good change to me, when using Powershell the .bat file will silently fail. You may also want to add the Powershell command ``gcm`` to find python; though it is technically redundant as Powershell will inform the user if they are inside a venv much like ``zsh`` or ``fish``

```
PS C:\tmp> .venv\Scripts\Activate.ps1^C
PS C:\tmp> python -m venv venv
PS C:\tmp> .\venv\Scripts\activate.bat
PS C:\tmp> where python
PS C:\tmp> .\venv\Scripts\activate
(venv) PS C:\tmp> where python
(venv) PS C:\tmp> gcm python

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     python.exe                                         3.11.5150… C:\tmp\venv\Scripts\python.exe

(venv) PS C:\tmp>
```

chrysle (approved):
Thanks!

PR #1463, opened 2023-12-22
Deduplicate specification of valid package names

jeanas:
- Rename the "Package name normalization" specification into "Package name grammar and normalization", to make it explicitly about the format of valid names too.

- Link to this spec in the core metadata specification instead of duplicating the part about valid names.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1463.org.readthedocs.build/en/1463/

### DIFF
--- a/source/specifications/name-normalization.rst
+++ b/source/specifications/name-normalization.rst
@@ -1,16 +1,16 @@
-======================================
-Package name grammar and normalization
-======================================
+=======================
+Names and normalization
+=======================
 
-Project names obey a restricted format, and are "normalized" for use in various
-contexts. This document describes what the valid project names are, and how
-project names should be normalized.
+This specification defines the format that names for packages and extras are
+required to follow. It also describes how to normalize them, which should be
+done before lookups and comparisons.
 
 
-.. _name-grammar:
+.. _name-format:
 
-Name grammar
-============
+Name format
+===========
 
 A valid name consists only of ASCII letters and numbers, period,
 underscore and hyphen. It must start and end with a letter or number.
--- a/source/specifications/name-normalization.rst
+++ b/source/specifications/name-normalization.rst
@@ -1,13 +1,16 @@
-.. _name-normalization:
+======================================
+Package name grammar and normalization
+======================================
 
-==========================
-Package name normalization
-==========================
+Project names obey a restricted format, and are "normalized" for use in various
+contexts. This document describes what the valid project names are, and how
+project names should be normalized.
 
-Project names are "normalized" for use in various contexts. This document describes how project names should be normalized.
 
-Valid non-normalized names
---------------------------
+.. _name-grammar:
+
+Name grammar
+============
 
 A valid name consists only of ASCII letters and numbers, period,
 underscore and hyphen. It must start and end with a letter or number.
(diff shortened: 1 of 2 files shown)


source/specifications/name-normalization.rst
```diff
@@ -16,10 +19,15 @@ following regex (run with ``re.IGNORECASE``)::
...
-Normalization
--------------
 
-The name should be lowercased with all runs of the characters ``.``, ``-``, or ``_`` replaced with a single ``-`` character. This can be implemented in Python with the re module:
+.. _name-normalization:
+
+Name normalization
+==================
+
+The name should be lowercased with all runs of the characters ``.``, ``-``, or
```
pradyunsg, line 28: nit/off-topic: Did we have agreement at any point on whether to wrap or not-wrap in this docset?

pradyunsg (changes requested):

source/specifications/name-normalization.rst
```diff
@@ -1,13 +1,16 @@
...
-==========================
-Package name normalization
-==========================
 
-Project names are "normalized" for use in various contexts. This document describes how project names should be normalized.
+.. _name-grammar:
 
-Valid non-normalized names
---------------------------
+Name grammar
```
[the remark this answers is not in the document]
jeanas: Well, doing a bit of introspection, the reason it was natural to me is that a language recognized by a classic regex can be viewed as a language recognized by a regular grammar. But let's assume not everybody knows that formal language theory stuff... I'd prefer to eschew "valid names" because it feels weird to link to the "valid name specification". Let me change it to "Name format", this is a good idea.

source/specifications/name-normalization.rst
jeanas: Do you mean whether paragraphs should be wrapped in general (to which I think the answer is yes based on the vast majority of the content), or whether it's fine to wrap them in PRs if they're not wrapped?

source/specifications/name-normalization.rst
```diff
@@ -1,13 +1,16 @@
-.. _name-normalization:
+======================================
+Package name grammar and normalization
+======================================
+
+Project names obey a restricted format, and are "normalized" for use in various
```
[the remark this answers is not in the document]
jeanas: Changed, let me know if the next text is OK.

source/specifications/name-normalization.rst
```diff
@@ -1,13 +1,16 @@
-.. _name-normalization:
+======================================
+Package name grammar and normalization
```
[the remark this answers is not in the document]
jeanas: Applied.

source/specifications/name-normalization.rst
pradyunsg: > Do you mean whether paragraphs should be wrapped in general 

This.

> or whether it's fine to wrap them in PRs if they're not wrapped?

... which informs this because I'd prefer that we enforce consistency in a one-shot. 😅

pradyunsg (approved):

source/specifications/name-normalization.rst
webknjaz: I think the problem with this is that on one hand wrapping is nice but reflowing together with content changes in the same PR has the same problem as reformatting normal source code while also changing behavior. They just have to be separate. And when they're smashed, it's impossible to review sanely.

OTOH, I know there's i18n-related tools that demand complete sentences on dedicated lines and use that for tracking sentence changes across versions when they're updated. That's a translation tooling aspect. I think these is used in "conventional markdown" or something like that.

webknjaz (approved):

PR #1465, opened 2023-12-22
Increase number of linkcheck retries

jeanas:
We're getting frequent flaky CI failures due to the huge number of links that have to be checked, making the check sensitive to random connection errors.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1465.org.readthedocs.build/en/1465/


webknjaz (approved):
Oh, I didn't realize that it defaults to just one attempt... Thanks!

PR #1469, opened 2023-12-23
Add problem matcher for test-translations.yml

rffontenelle:
Changes in here include:

- Make sure that the lint is ran even when build fails
- Add problem matcher for both build and lint
- Update version of the action/setup-python action

Link of tests done: rffontenelle/packaging.python.org#3

Screenshot from my personal tests showing errors in a language:

Annotations for sphinx-lint

Screenshot of the summary of a pull request:

sphinx-lint outputs shown as annotations

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1469.org.readthedocs.build/en/1469/

### DIFF
--- a/.github/workflows/test-translations.yml
+++ b/.github/workflows/test-translations.yml
@@ -67,7 +67,7 @@ jobs:
       run: python -m pip install --upgrade nox virtualenv sphinx-lint
 
     - name: Set Sphinx problem matcher
-      uses: sphinx-doc/github-problem-matcher@master
+      uses: sphinx-doc/github-problem-matcher@v1.0
 
     - name: Build translated docs in ${{ matrix.language }}
       run: nox -s build -- -q -D language=${{ matrix.language }}
--- a/.github/workflows/test-translations.yml
+++ b/.github/workflows/test-translations.yml
@@ -72,7 +72,10 @@ jobs:
     - name: Build translated docs in ${{ matrix.language }}
       run: nox -s build -- -q -D language=${{ matrix.language }}
 
+    - name: Set Sphinx Lint problem matcher
+      if: always()
+      run: echo '::add-matcher::.github/sphinx_lint_matcher.json'
+
     - name: Lint translation file
-      run: |
-        echo '::add-matcher::.github/sphinx_lint_matcher.json'
-        sphinx-lint locales/${{ matrix.language }}/LC_MESSAGES/messages.po
+      if: always()
+      run: sphinx-lint locales/${{ matrix.language }}/LC_MESSAGES/messages.po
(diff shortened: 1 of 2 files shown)


.github/workflows/test-translations.yml
```diff
@@ -58,16 +58,21 @@ jobs:
...
+      uses: actions/setup-python@v5
       with:
         python-version: >-
           3.10
 
     - name: Install Python tooling
       run: python -m pip install --upgrade nox virtualenv sphinx-lint
 
+    - name: Set Sphinx problem matcher
+      uses: sphinx-doc/github-problem-matcher@master
```
webknjaz, line 70: Maybe, pin to a commit hash?

Also, let's add this one to other workflows that run Sphinx, right in this PR.

.github/workflows/test-translations.yml
```diff
@@ -58,16 +58,21 @@ jobs:
...
+
     - name: Build translated docs in ${{ matrix.language }}
       run: nox -s build -- -q -D language=${{ matrix.language }}
 
     - name: Lint translation file
-      run: sphinx-lint locales/${{ matrix.language }}/LC_MESSAGES/messages.po
+      run: |
+        echo '::add-matcher::.github/sphinx_lint_matcher.json'
```
webknjaz, line 77: Could you move this to a separate step? Squashing multiple commands in the same script tends to be problematic and hard to distinguish.

rffontenelle: Looks like I forgot to include in the PR a line for `if: always()` which is required to run lint when build fails.

I'm saying this because the reason I didn't separate them was that I would have to duplicate the `if: always()` condition in order to make sure that the linter would be called.

Do you prefer me to duplicate `if: always()` or to keep them together (with the `if: always()` I will add) ?

.github/workflows/test-translations.yml
rffontenelle: prefer a commit or `v0` (for the current v0.1)?

webknjaz: I contributed the conversion to composite actions, which made it much faster and cross-compatible, and asked for a new tag. So it's now
```suggestion
      uses: sphinx-doc/github-problem-matcher@v1.0
```

.github/workflows/test-translations.yml
webknjaz: Yes, please. Or add all the matchers before running any tests so they run early and don't need conditionals.

rffontenelle: Done.

rffontenelle:
How it looks now:

image

jeanas (approved):

PR #1471, opened 2023-12-23
Update "Dropping older Python versions" guide

chrysle:
An attempt to improve the guide. Actually I find the title a bit inaccurate. What about _"Changing required Python versions in the package metadata"_?

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1471.org.readthedocs.build/en/1471/guides/dropping-older-python-versions/


source/guides/dropping-older-python-versions.rst
```diff
@@ -4,23 +4,17 @@
...
 
 Requirements
 ------------
 
-This workflow requires that:
-
-1. The publisher is using the latest version of :ref:`setuptools`,
-2. The latest version of :ref:`twine` is used to upload the package,
-3. The user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
+This workflow requires that the user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
```
jeanas, line 17: This is fine as it is, but I'd suggest to just drop this sentence, pip 9.0 is 7 years old at this point. (As written on another review, my guess is that many commands in the PUG will not work with pip versions that are that old anyway.)

source/guides/dropping-older-python-versions.rst
```diff
@@ -52,42 +46,50 @@ explicitly set ``universal`` to ``0``:
...
 
 .. code-block:: text
 
-    Requires-Python: ">=3"
-    Requires-Python: ">2.7,!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+    Requires-Python: ">= 3"
+    Requires-Python: ">= 2.7, != 3.0.*, != 3.1.*, != 3.2.*, != 3.3.*"
+
+
+The way to set those values is within your :file:`pyproject.toml`. The :ref:`requires-python` field of the
```
jeanas, line 79: I think “the way” is not really accurate because it's not in, e.g., Poetry. Maybe just `` :ref:`requires-python` `` to link to the pyproject.toml guide? (There's a note at the top of that guide, explaining that not all build backends use the `[project]` table.)

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
 
     setup(
         # Your setup arguments
-        python_requires='>=2.7',  # Your supported Python ranges
+        python_requires='>= 3.8',
     )
 
+It is warned against adding upper bounds to the version ranges, e. g. ``">= 3.8 < 3.10"``. This can cause different errors
+and version conflicts. See the `discourse discussion`_ for more information.
```
jeanas, line 105: Nit: reST labels normally use hyphens.

source/guides/dropping-older-python-versions.rst
chrysle: > This is fine as it is

Does that mean you suggest restoring the list? I took out the `setuptools` requirement since this guide now focuses on `pyproject.toml`, and the `twine` requirement is already set in the linked distributing guide (but I could as well reinstate that).

webknjaz: A lot of LTS distro users (e.g. RHEL) may still be constrained to quite old versions. Here's what RHEL 8.3 has:
```console
$ podman run --rm -it -v "$(pwd):/io" ubi8/ubi:8.3 /usr/libexec/platform-python -m ensurepip --version    
pip 9.0.3
```

RHEL 7.9 has one instance of pip 8:
```console
[root@92c71f44a6f9 /]# yum info python27-python-pip
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager

This system is not receiving updates. You can use subscription-manager on the host to register and assign subscriptions.

Installed Packages
Name        : python27-python-pip
Arch        : noarch
Version     : 8.1.2
Release     : 7.el7
Size        : 6.6 M
Repo        : installed
From repo   : ubi-server-rhscl-7-rpms
Summary     : A tool for installing and managing Python packages
URL         : http://www.pip-installer.org
License     : MIT and Python and ASL 2.0 and BSD and ISC and LGPLv2 and (ASL 2.0 or BSD)
Description : Pip is a replacement for `easy_install
            : <http://peak.telecommunity.com/DevCenter/EasyInstall>`_.  It uses mostly the
            : same techniques for finding packages, so packages that were made
            : easy_installable should be pip-installable as well.
```

Of course, it's quite old but a version hint is still relevant. `manylinux1` images that are still in use, are based on an even older CentOS 6, for example.

source/guides/dropping-older-python-versions.rst
webknjaz: I don't think that Poetry should be special-cased as we're building guides around the accepted and widely implemented standards and not focus on tools that don't want to play ball. Besides, whenever they get to implementing it, it'll work for them too.

OTOH, I'd not focus on `pyproject.toml` solely as I still prefer a tool-specific `setup.cfg` myself. So the argument of not presenting it as the only possible way of declaring the metadata makes sense to me. I'd probably say something like along the lines of "the way declared by the standard/PEP X is this, but tool specific alternatives exist".

source/guides/dropping-older-python-versions.rst
webknjaz: +1 to using dashes. Using snake case in contexts differing from Python code helps notices said difference visually. This is the reason why I use dashes in YAML identifiers and other places where it's allowed — it clearly communicates that this syntax is from some other context.

source/guides/dropping-older-python-versions.rst
```diff
@@ -4,23 +4,17 @@
 Dropping support for older Python versions
 ==========================================
 
-Dropping support for older Python versions is supported by the standard :ref:`core-metadata` 1.2 specification via a "Requires-Python" attribute.
+Dropping support for older Python versions is supported by the standard :ref:`core-metadata` 1.2 specification via a :ref:`"Requires-Python" <core-metadata-requires-python>` attribute.
```
webknjaz, line 7: > support ... is supported

Can we change one of these words to something else to improve the writing style?

webknjaz: Also, perhaps using `s/a/the/` would be more grammatically correct.

source/guides/dropping-older-python-versions.rst
jeanas: Yes, sorry for not being clear, I'm not suggesting to list how every tool does it, just not to lead the reader to believe that `[project]` is always used. This is already explained in a note near the beginning of the [pyproject.toml guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/), hence the proposal to link that document here.

source/guides/dropping-older-python-versions.rst
jeanas: > Does that mean you suggest restoring the list?

Sorry, I meant that your modified version is fine as it is.

Maybe the version hint should stay but in a footnote?

source/guides/dropping-older-python-versions.rst
```diff
@@ -52,42 +46,50 @@ explicitly set ``universal`` to ``0``:
...
-metadata values based on the argument you provide in ``python_requires``.
+   [build-system]
+   ...
+
+   [project]
+   requires-python = ">= 3.8" # At least Python 3.8
+
+
+For :ref:`setuptools` users, another way to achieve this is using the ``python_requires`` parameter
+in the call to ``setup`` within your :file:`setup.py` script.
```
webknjaz, line 92: Let's make this a bit more accurate:
```suggestion
For :ref:`setuptools` users, another way to achieve this is using the ``python_requires`` parameter in your :file:`setup.cfg` config or the :file:`setup.py` script.
``setuptools < 61`` does not support declaring the package metadata in :file:`pyproject.toml`.
```

source/guides/dropping-older-python-versions.rst
```diff
@@ -52,42 +46,50 @@ explicitly set ``universal`` to ``0``:
...
 
 .. code-block:: python
```
webknjaz, line 94: Can we demonstrate using a declarative config instead of a script here?
```suggestion
.. code-block:: ini

   # setup.cfg

   [options]
   python_requires = >= 3.8
```

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
 
     setup(
         # Your setup arguments
-        python_requires='>=2.7',  # Your supported Python ranges
+        python_requires='>= 3.8',
     )
 
+It is warned against adding upper bounds to the version ranges, e. g. ``">= 3.8 < 3.10"``. This can cause different errors
```
webknjaz, line 104: Could you add an admonition here? Like `.. caution::`? https://pradyunsg.me/furo/reference/admonitions/#caution

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
+and version conflicts. See the `discourse discussion`_ for more information.
+
 3. Validating the Metadata before publishing
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Within a Python source package (the zip or the tar-gz file you download) is a text file called PKG-INFO.
 
-This file is generated by :ref:`distutils` or :ref:`setuptools` when it generates the source package.
-The file contains a set of keys and values, the list of keys is part of the PyPa standard metadata format.
+This file is generated by the build backend when it generates the source package.
```
webknjaz, line 112: Could you link the build backend term?

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
+The file contains a set of keys and values, the list of keys is part of the PyPA standard metadata format.
 
 You can see the contents of the generated file like this:
 
 .. code-block:: bash
 
-    tar xfO dist/my-package-1.0.0.tar.gz my-package-1.0.0/PKG-INFO
+    tar xf dist/my-package-1.0.0.tar.gz my-package-1.0.0/PKG-INFO -O
```
webknjaz, line 119: Why?

source/guides/dropping-older-python-versions.rst
webknjaz: Also, `setuptools` now has comprehensive examples in the docs. See:
- https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#python-requirement
- https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
- https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

So perhaps linking them instead of having a tool-specific example spelled out, would be a good option.

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
 -------------------------
 
-Once you have published a package with the Requires-Python metadata, you can then make a further update removing that Python runtime from support.
+In principle, at least metadata support for Python versions should be kept as long as possible, because
+once that has been dropped, people still depending on a version will be forced to downgrade.
+If however supporting a specific version becomes a blocker for a new feature or other issues occur, the metadata
+``Requires-Python`` should be amended. Of course this also depends on whether the project needs to be stable and
+well-covered for a wider range of users.
+
+Each version compatibility change should have an own release.
```
webknjaz, line 140:
```suggestion
Each version compatibility change should have its own release.
```

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
+
+For example, you published version 1.0.0 of your package with ``Requires-Python: ">= 2.7"`` metadata.
+
+If you then update the version string to ``">= 3.5"``, and publish a new version 2.0.0 of your package, any users running Pip 9.0+ from version 2.7 will have version 1.0.0 of the package installed, and any ``>= 3.5`` users will receive version 2.0.0.
```
webknjaz, line 144: Not sure I agree with this at all. It's absolutely orthogonal to SemVer. The dependency resolvers will always select a runtime-compatible version for as long as this information is present in the package metadata.

Dropping a runtime support is not a feature change. And in SemVer, the major version release is for breaking APIs a project provides to the end-users. This metadata change may happen anytime and is a subject to different considerations that are dependent on the conventions of specific projects.

Let's not overload this guide with assumptions based on one specific schema for communicating API changes, that is known to be problematic anyway.

The maintainers can choose to synchronize dropping a runtime with a versioning scheme of their choice but that's not something that is mandatory for changing a metadata field value.

I think @jaraco mentioned this too somewhere.

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
 
-It must be done in this order for the automated fallback to work.
+It may be a good idea to create a minor release, stating that it is the last one compatible with the Python version to be removed, just before dropping that version.
```
webknjaz, line 146: I don't think this advice makes sense anymore. It could be a patch release too. The idea here was that before `Requires-Python` became a thing, many projects were released without this bit of metadata. And this would cause the dependency resolver to treat the old releases as supporting any Python runtime. So if people were to add a high minimum Python version right away, the resolver would backtrack to the one immediately preceding release.

So this is a good advice for old project that never had this metadata field set. Once they start using it, they can update it anytime, regardless of any other pieces of the metadata.

source/guides/dropping-older-python-versions.rst
```diff
@@ -4,23 +4,17 @@
...
 ------------
 
-This workflow requires that:
-
-1. The publisher is using the latest version of :ref:`setuptools`,
-2. The latest version of :ref:`twine` is used to upload the package,
-3. The user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
+This workflow requires that the user installing the package has at least Pip 9.0, or a client that supports the Metadata 1.2 specification.
 
 Dealing with the universal wheels
```
webknjaz, line 19: @chrysle this paragraph is specifically talking about `setuptools`, `bdist_wheel` as a `setuptools`' command that invokes the `wheel` project APIs internally and a CLI option (in the tip block below) that is supposed to be passed to the direct CLI invocation of `setup.py`, which is also discouraged.

So while you're on it, could to add clarifications that it's solely about the setuptools' mechanisms and maybe drop the suggestions to use the CLI flag?

source/guides/dropping-older-python-versions.rst
```diff
@@ -96,43 +98,54 @@ metadata values based on the argument you provide in ``python_requires``.
...
+
+For example, you published version 1.0.0 of your package with ``Requires-Python: ">= 2.7"`` metadata.
+
+If you then update the version string to ``">= 3.5"``, and publish a new version 2.0.0 of your package, any users running Pip 9.0+ from version 2.7 will have version 1.0.0 of the package installed, and any ``>= 3.5`` users will receive version 2.0.0.
 
-It must be done in this order for the automated fallback to work.
+It may be a good idea to create a minor release, stating that it is the last one compatible with the Python version to be removed, just before dropping that version.
 
-For example, you published the Requires-Python: ">=2.7" as version 1.0.0 of your package.
+When dropping a Python version, it might also be rewarding to upgrade the project's code syntax generally, apart from updating the versions used in visible places (like the testing environment). Tools like pyupgrade_ can simplify this task.
```
webknjaz, line 148: Let's put this into an admonition like `.. hint::` or `.. tip::`.

chrysle:
Thank you for your comments! I addressed them.

jeanas (approved):
Sorry for the lack of explicit approval here. (I'm also going to resolve my comments.)

chrysle:
> (I'm also going to resolve my comments. **Edit:** In fact I don't have the permissions to resolve them.)

I did that now, thank you for your review!

source/guides/dropping-older-python-versions.rst
```diff
@@ -43,69 +36,74 @@ explicitly set ``universal`` to ``0``:
...
-Examples:
+You can specify version ranges and exclusion rules (complying with the :ref:`version-specifiers` specification),
+such as at least Python 3. Or, Python 2.7, 3.4 and beyond:
 
 .. code-block:: text
 
-    Requires-Python: ">=3"
-    Requires-Python: ">2.7,!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+    Requires-Python: ">= 3"
+    Requires-Python: ">= 2.7, != 3.0.*, != 3.1.*, != 3.2.*, != 3.3.*"
```
edgarrmondragon, line 74: Should this example be updated to use newer Python versions? I don't think most people reading this guide would be interested in supported Python 2.7 or 3.4-3.7.

source/guides/dropping-older-python-versions.rst
jaraco: Agreed on not making assumptions about versions schemes.

It's been my instinct that dropping a platform _isn't_ a breaking change because of the reasons described (it doesn't break anyone who's likely to download it).

source/guides/dropping-older-python-versions.rst
chrysle: Done!

chrysle:
@jaraco Could you review this small hunk? I'm interested in moving it forward.

source/guides/dropping-older-python-versions.rst
```diff
@@ -43,69 +36,74 @@ explicitly set ``universal`` to ``0``:
...
+
+For :ref:`setuptools` users, another way to achieve this is using the ``python_requires`` parameter
+in your :file:`setup.cfg` config or the :file:`setup.py` script. ``setuptools < 61`` does not support
+declaring the package metadata in :file:`pyproject.toml`.
+
+Consult the ``setuptools`` `dependency-management`_ documentation for information about the appropriate
+way to configure each of these files.
+
+.. caution::
+        It is warned against adding upper bounds to the version ranges, e. g. ``">= 3.8 < 3.10"``. This can cause different errors
```
jaraco, line 97: Use imperative voice, fix the missing comma, and avoid ambiguous "this":

```suggestion
        Avoid adding upper bounds to the version ranges, e. g. ``">= 3.8, < 3.10"``. Doing so can cause different errors
```

source/guides/dropping-older-python-versions.rst
```diff
@@ -116,23 +114,30 @@ You can see the contents of the generated file like this:
...
+If however supporting a specific version becomes a blocker for a new feature or other issues occur, the metadata
+``Requires-Python`` should be amended. Of course this also depends on whether the project needs to be stable and
+well-covered for a wider range of users.
+
+Each version compatibility change should have its own release.
+
+.. tip::
 
-It must be done in this order for the automated fallback to work.
+        When dropping a Python version, it might also be rewarding to upgrade the project's code syntax generally, apart from updating the versions used in visible         places (like the testing environment). Tools like pyupgrade_ can simplify this task.
```
jaraco, line 137:
```suggestion
        When dropping a Python version, it might also be rewarding to upgrade the project's code syntax generally, apart from updating the versions used in visible places (like the testing environment). Tools like pyupgrade_ or `ruff <https://docs.astral.sh/ruff/linter/>`_ can automate some of this work.
```

jaraco (approved):
LGTM after addressing my comments.

chrysle:
Thanks for reviewing, good points! I think I managed to address them all.

chrysle:
Thank you all for reviewing!

PR #1474, opened 2023-12-30
Update tool recommendations

### LINKED ISSUE #1468: Discuss how to update the "Tool recommendations" page
Quoting from https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/:

> Outdated Tool Recommendations from the PyPA
>
> Finally I find [Tool recommendations](https://packaging.python.org/en/latest/guides/tool-recommendations/) in the PyPA User Guide. Under [Packaging tool recommendations](https://packaging.python.org/en/latest/guides/tool-recommendations/#packaging-tool-recommendations) it says:
>
> - Use setuptools to define projects.
> - Use build to create Source Distributions and wheels.
> - If you have binary extensions and want to distribute wheels for multiple platforms, use cibuildwheel as part of your CI setup to build distributable wheels.
> - Use twine for uploading distributions to PyPI.
>
> Finally, some canonical documentation from the PyPA that comes out and suggests what to use!
>
> But my relief immediately turns to questioning whether this tooling recommendations documentation is up to date:
>
> 1. If setuptools is recommended, why does the [Packaging Python Projects](https://packaging.python.org/en/latest/tutorials/packaging-projects/) tutorial use Hatch?
> 2. How exactly should I be using setuptools to define projects? Is this referring to setuptools as a [build-system] backend? The existence of define seemingly implies using setup.py or setup.cfg to define metadata. But I thought these distutils/setuptools specific mechanisms were deprecated in favor of the more generic pyproject.toml?
> 3. Why aren't other tools like Hatch, pip, poetry, flit, and pdm mentioned on this page? Where's the guidance on when to use these alternative tools?
> 4. There are footnotes referencing distutils as if it is still a modern practice. No mention that it was removed from the standard library in Python 3.12.
> 5. But the build tool is referenced and that tool is relatively new. So the docs have to be somewhat up-to-date, right?

These are all valid concerns.

I'm willing to do the work of updating that page, but I think this is a potentially very delicate/controversial topic, so I figured I'd open an issue to agree on a plan first, so as not to mix this discussion with PR review details.

Some thoughts by tool category:

- I think we can agree that pip should be recommended. But this could/should also mention pipx for the specific case of CLI tools. I don't think pipx has much competition right now (though it [may have](https://github.com/pypa/hatch/issues/1105) in the future).
- The recommendation of virtualenv/venv is fine. But we could mention that there many (a *great* many) tools out there to manage environments more automatically (virtualenvwrapper, Poetry/PDM/Hatch, etc.)
- The recommendation of pip-tools, Pipenv or poetry for lock files could be expanded to include PDM.
- I think we should drop buildout and Hashdist, both of these are dead judging from Git commit history, especially Hashdist.
- I do not think the PyPA should officially recommend setuptools as a build backend. After all, the explicit purpose of PEP 517 was to allow other backends. Here, I think we should simply list the different alternatives without making anything that would look like an official choice, since there is no consensus on such a choice. The backends could be listed in the same order as the tabs in the "Packaging projects" tutorial.
- The cibuildwheel recommendation is fine. I don't think there are any serious alternatives.
- I'm not sure about recommending `build` and `twine`. Certainly, they're better than `python setup.py bdist_wheel` and `python setup.py upload`, but the equivalent Hatch/Poetry/PDM commands are fine too.

By and large, I think we should make less recommendations on that page, be more objective, and especially avoid recommending a given tool when there is no consensus among PyPA members that this should be the blessed tool, since the officialness of the page makes users believe that a consensus exists. For better or worse, there is more good advice to be given on what *not* to use — e.g., this could be a good page to remind things like "don't use `python setup.py upload`", "don't use distutils", and "whichever backend you choose, declare it in the `[build-system]` table".

One other thing I'm not sure about is the status of Pipenv (how well-maintained is it and are many people using it after all the controversy that there was around it?).

jeanas:
This is my proposal to address #1468.

Fixes #1468

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1474.org.readthedocs.build/en/1474/guides/tool-recommendations/

### DIFF
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -4,111 +4,140 @@
 Tool recommendations
 ====================
 
-If you're familiar with Python packaging and installation, and just want to know
-what tools are currently recommended, then here it is.
+The Python packaging landscape consists of many different tools. For many tasks,
+the Python Packaging Authority (PyPA, the umbrella organization which
+encompasses many packaging tools and maintains this guide) purposefully does not
+make a blanket recommendation; for example, the reason there exist many build
+backends is that the landscape was opened in order to enable the development of
+new backends serving certain users' needs better than the previously unique
+backend, setuptools. This guide does point to some tools that are widely
+recognized, and also makes some recommendations of tools that you should *not*
+use because they are deprecated or insecure.
 
 
-Application dependency management
-=================================
+Virtual environments
+====================
 
-* Use :ref:`pip` in a `secure manner`_ to install a Python application and its
-  dependencies during deployment.
-
-* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
-  application-specific dependencies from a shared Python installation. [4]_
-
-* Use `pip-tools`_, :ref:`pipenv`, or `poetry`_ to generate the fully-specified
-  application-specific dependencies, when developing Python applications.
-
-.. _secure manner: https://pip.pypa.io/en/latest/topics/secure-installs/
-.. _pip-tools: https://github.com/jazzband/pip-tools
-.. _Poetry: https://python-poetry.org/
-
-Installation tool recommendations
-=================================
-
-* Use :ref:`pip` to install Python :term:`packages <Distribution Package>` from
-  :term:`PyPI <Python Package Index (PyPI)>`. [1]_ [2]_ Depending on how :ref:`pip`
-  is installed, you may need to also install :ref:`wheel` to get the benefit
-  of wheel caching. [3]_
-
-* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
-  project-specific dependencies from a shared Python installation. [4]_
-
-* If you're looking for management of fully integrated cross-platform software
-  stacks, consider:
-
-  * :ref:`buildout`: primarily focused on the web development community
-
-  * :ref:`spack`, :ref:`hashdist`, or :ref:`conda`: primarily focused
-    on the scientific community.
+The standard tools to create and use virtual environments manually are
+:ref:`virtualenv` (PyPA project) and :doc:`venv <python:library/venv>` (part of
+the Python standard library, though missing some features of virtualenv).
 
 
-Packaging tool recommendations
-==============================
+Installing packages
+===================
 
-* Use :ref:`setuptools` to define projects. [5]_ [6]_
+:ref:`Pip` is a standard tool to install packages from :term:`PyPI <Python
+Package Index (PyPI)>`. It can install into the global environment or into
+virtual environments. You may want to read pip's recommendations for
+:doc:`secure installs <pip:topics/secure-installs>`. Pip is available by default
+in most Python installations.
 
-* Use :ref:`build` to create :term:`Source Distributions
-  <Source Distribution (or "sdist")>` and :term:`wheels <Wheel>`.
+Alternatively, for installing Python command line applications specifically,
+consider :ref:`pipx`, which is a wrapper around pip that installs each
+application into a dedicated virtual environment in order to avoid conflicts
+with other applications and, on Linux, conflicts with the system.
 
-If you have binary extensions and want to distribute wheels for multiple
-platforms, use :ref:`cibuildwheel` as part of your CI setup to build
-distributable wheels.
+For scientific software specifically, consider :ref:`Conda` or :ref:`Spack`.
 
-* Use `twine <https://pypi.org/project/twine>`_ for uploading distributions
-  to :term:`PyPI <Python Package Index (PyPI)>`.
+.. todo:: Write a "pip vs. Conda" comparison, here or in a new discussion.
+
+Do **not** use ``easy_install`` (part of :ref:`setuptools`), which is deprecated
+in favor of pip (see :ref:`pip vs easy_install` for details). Likewise, do
+**not** use ``python setup.py install`` or ``python setup.py develop``, which
+are also deprecated (see :ref:`setup-py-deprecated` for background and
+:ref:`modernize-setup-py-project` for migration advice).
 
 
-Publishing platform migration
-=============================
+Lock files
+==========
 
-The original Python Package Index implementation (previously hosted at
-`pypi.python.org <https://pypi.python.org>`_) has been phased out in favour
-of an updated implementation hosted at `pypi.org <https://pypi.org>`_.
+:ref:`pip-tools` and :ref:`Pipenv` are two recognized tools to create lock
+files, which contain the exact versions of all packages installed into an
+environment, for reproducibility purposes.
 
-See :ref:`Migrating to PyPI.org` for more information on the status of the
-migration, and what settings to change in your clients.
 
-----
+Build backends
+==============
 
-.. [1] There are some cases where you might choose to use ``easy_install`` (from
-       :ref:`setuptools`), e.g. if you need to install from :term:`Eggs <Egg>`
-       (which pip doesn't support).  For a detailed breakdown, see :ref:`pip vs
-       easy_install`.
+Popular :term:`build backends <build backend>` for pure-Python packages include:
 
-.. [2] The acceptance of :pep:`453` means that :ref:`pip`
-       will be available by default in most installations of Python 3.4 or
-       later.  See the :pep:`rationale section <453#rationale>` from :pep:`453`
-       as for why pip was chosen.
+- Hatchling, which is part of :ref:`Hatch` (but can be used without
+  Hatch as well). Hatchling is extensible through a plugin system.
 
-.. [3] `get-pip.py <https://github.com/pypa/get-pip/#readme>`_ and
-       :ref:`virtualenv` install
-       :ref:`wheel`, whereas :ref:`ensurepip` and :ref:`venv <venv>` do not
-       currently.  Also, the common "python-pip" package that's found in various
-       linux distros, does not depend on "python-wheel" currently.
+- :ref:`setuptools`, the historical build backend. It can be configured
+  programmatically through the :file:`setup.py` file.
 
-.. [4] Beginning with Python 3.4, ``venv`` will create virtualenv environments
-       with ``pip`` installed, thereby making it an equal alternative to
-       :ref:`virtualenv`. However, using :ref:`virtualenv` will still be
-       recommended for users that need cross-version consistency.
+  If you use setuptools, please be aware that it contains many deprecated
+  features which are currently kept for compatibility, but should not be used.
+  For example, do not use ``python setup.py`` invocations
+  (cf. :ref:`setup-py-deprecated`), the ``python_requires`` argument to
+  ``setup()`` (use the :ref:`[build-system] table
+  <pyproject-guide-build-system-table>` of :file:`pyproject.toml` instead), or
+  the ``easy_install`` command (cf. :ref:`pip vs easy_install`).
 
-.. [5] Although you can use pure :ref:`distutils` for many projects, it does not
-       support defining dependencies on other projects and is missing several
-       convenience utilities for automatically populating distribution metadata
-       correctly that are provided by ``setuptools``. Being outside the
-       standard library, ``setuptools`` also offers a more consistent feature
-       set across different versions of Python, and (unlike ``distutils``),
-       recent versions of ``setuptools`` support all of the modern metadata
-       fields described in :ref:`core-metadata`.
+- Flit-core, part of :ref:`Flit` (but usable standalone). It is meant to be a
+  minimal and opinionated build backend. It is not extensible.
 
-       Even for projects that do choose to use ``distutils``, when :ref:`pip`
-       installs such projects directly from source (rather than installing
-       from a prebuilt :term:`wheel <Wheel>` file), it will actually build
-       your project using :ref:`setuptools` instead.
+- PDM-backend_, part of :ref:`PDM` (but usable standalone). It provides build
+  hooks for extensibility.
 
-.. [6] `distribute`_ (a fork of setuptools) was merged back into
-       :ref:`setuptools` in June 2013, thereby making setuptools the default
-       choice for packaging.
+- Poetry-core, part of :ref:`Poetry` (but usable standalone). It is extensible
+  through plugins.
 
-.. _distribute: https://pypi.org/project/distribute
+Do **not** use distutils, which is deprecated, and has been removed from the
+standard library in Python 3.12, although it still remains available from
+setuptools.
+
+For packages with :term:`extension modules <extension module>`, you may use
+setuptools, but consider using a build system dedicated to the language the
+extension is written in, such as Meson or CMake for C/C++, or Cargo for Rust,
+and bridging this build system to Python using a dedicated build backend:
+
+- :ref:`meson-python` for Meson,
+
+- :ref:`scikit-build-core` for CMake,
+
+- :ref:`maturin` for Cargo.
+
+
+Building distributions
+======================
+
+The standard tool to build :term:`source distributions <source distribution (or
+"sdist")>` and :term:`wheels <wheel>` for uploading to PyPI is :ref:`build`.  It
+will invoke whichever build backend you :ref:`declared
+<pyproject-guide-build-system-table>` in :file:`pyproject.toml`.
+
+Do **not** use ``python setup.py sdist`` and ``python setup.py bdist_wheel`` for
+this task. All direct invocations of :file:`setup.py` are :ref:`deprecated
+<setup-py-deprecated>`.
+
+If you have :term:`extension modules <extension module>` and want to distribute
+wheels for multiple platforms, use :ref:`cibuildwheel` as part of your CI setup
+to build distributable wheels.
+
+
+Uploading to PyPI
+=================
+
+The standard tool for this task is :ref:`twine`.
+
+**Never** use ``python setup.py upload`` for this task. In addition to being
+:ref:`deprecated <setup-py-deprecated>`, it is insecure.
+
+
+Integrated workflow tools
+=========================
+
+These are tools that combine many features in one command line application, such
+as automatically managing virtual environments for a project, building
+distributions, uploading to PyPI, or creating and using lock files.
+
+- :ref:`Hatch`,
+- :ref:`Flit`,
+- :ref:`PDM`,
+- :ref:`Poetry`.
+
+
+
+.. _pdm-backend: https://backend.pdm-project.org
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -78,11 +78,10 @@ in alphabetical order:
 - Poetry-core_ (developed with but separate from :ref:`Poetry`). It is extensible
   through plugins.
 
-- :ref:`setuptools` (which used to be the only build backend). It can be configured
-  using modern standards like `pyproject.toml`, but can also be extended
-  and supports customisation via `setup.py`.
-  programmatically through the :file:`setup.py` file (but for basic metadata,
-  :file:`pyproject.toml` is preferred).
+- :ref:`setuptools` (which used to be the only build backend). It can be
+  configured using modern standards like the :ref:`[project] table in
+  pyproject.toml <writing-pyproject-toml>`, but can also be extended and
+  customized via :file:`setup.py`.
 
   If you use setuptools, please be aware that it contains many deprecated
   features which are currently kept for compatibility, but should not be used.
@@ -100,16 +99,10 @@ For packages with :term:`extension modules <extension module>`, you may use
 a build system with dedicated support for the language the extension is written in,
 for example:
 
-- :ref:`setuptools` - natively supports C/C++ (with 3rd-party plugins for Go and Rust);
-- :ref:`meson-python` - C, C++, Fortran, and Rust and other languages supported by Meson;
-- :ref:`scikit-build-core` - C, C++, Fortran and other languages supported by CMake;
-- :ref:`maturin` - Rust, via Cargo.
-
-- :ref:`meson-python` for Meson,
-
-- :ref:`scikit-build-core` for CMake,
-
-- :ref:`maturin` for Cargo.
+- :ref:`setuptools` -- natively supports C and C++ (with third-party plugins for Go and Rust),
+- :ref:`meson-python` -- C, C++, Fortran, Rust, and other languages supported by Meson,
+- :ref:`scikit-build-core` -- C, C++, Fortran, and other languages supported by CMake,
+- :ref:`maturin` -- Rust, via Cargo.
 
 
 Building distributions
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -35,11 +35,11 @@ in most Python installations through the standard library package
 :doc:`ensurepip <python:library/ensurepip>`.
 
 Alternatively, consider :ref:`pipx` for the specific use case of installing Python
-command line applications that are distributed through PyPI.
+applications that are distributed through PyPI and run from the command line.
 Pipx is a wrapper around pip and venv that installs each
 application into a dedicated virtual environment. This avoids conflicts between
-the dependencies of different applications, and also with the system
-(especially on Linux).
+the dependencies of different applications, and also with system-wide applications
+making use of the same Python interpreter (especially on Linux).
 
 For scientific software specifically, consider :ref:`Conda` or :ref:`Spack`.
 
@@ -66,8 +66,8 @@ Build backends
 Popular :term:`build backends <build backend>` for pure-Python packages include,
 in alphabetical order:
 
-- Flit-core_ -- developed with but separate from :ref:`Flit`. A minimal and
-  opinionated build backend. It does not support plugins.
+- :doc:`Flit-core <flit:pyproject_toml>` -- developed with but separate from :ref:`Flit`.
+  A minimal and opinionated build backend. It does not support plugins.
 
 - Hatchling_ -- developed with but separate from :ref:`Hatch`. Supports plugins.
 
@@ -82,21 +82,23 @@ in alphabetical order:
 
 - :ref:`setuptools`, which used to be the only build backend. Supports plugins.
 
-  If you use setuptools, please be aware that some features that predate
-  standardisation efforts are now deprecated and only *temporarily kept*
-  for compatibility.
+  .. caution::
 
-  In particular, do **not** use direct ``python setup.py`` invocations. On the
-  other hand, configuring setuptools with a :file:`setup.py` file is still fully
-  supported, although it is recommended to use the modern :ref:`[project] table
-  in pyproject.toml <writing-pyproject-toml>` whenever possible and keep
-  :file:`setup.py` only if programmatic configuration is needed. See
-  :ref:`setup-py-deprecated`.
+     If you use setuptools, please be aware that some features that predate
+     standardisation efforts are now deprecated and only *temporarily kept*
+     for compatibility.
 
-  Other examples of deprecated features you should **not** use include the
-  ``setup_requires`` argument to ``setup()`` (use the :ref:`[build-system] table
-  <pyproject-guide-build-system-table>` in :file:`pyproject.toml` instead), and
-  the ``easy_install`` command (cf. :ref:`pip vs easy_install`).
+     In particular, do **not** use direct ``python setup.py`` invocations. On the
+     other hand, configuring setuptools with a :file:`setup.py` file is still fully
+     supported, although it is recommended to use the modern :ref:`[project] table
+     in pyproject.toml <writing-pyproject-toml>` whenever possible and keep
+     :file:`setup.py` only if programmatic configuration is needed. See
+     :ref:`setup-py-deprecated`.
+
+     Other examples of deprecated features you should **not** use include the
+     ``setup_requires`` argument to ``setup()`` (use the :ref:`[build-system] table
+     <pyproject-guide-build-system-table>` in :file:`pyproject.toml` instead), and
+     the ``easy_install`` command (cf. :ref:`pip vs easy_install`).
 
 Do **not** use :ref:`distutils`, which is deprecated, and has been removed from
 the standard library in Python 3.12, although it still remains available from
@@ -144,7 +146,7 @@ Task runners
 These tools allow you to define and invoke "tasks", such as running tests,
 compiling documentation, regenerating some files, etc. Each task can be executed
 in a dedicated virtual environment with the dependencies it requires. The most
-common tools for this are tox_ and nox_.
+common tools for this are :doc:`tox <tox:index>` and :doc:`nox <nox:index>`.
 
 
 Integrated workflow tools
@@ -163,9 +165,6 @@ order:
 - :ref:`Poetry`.
 
 
-.. _flit-core: https://pypi.org/project/flit-core/
 .. _hatchling: https://pypi.org/project/hatchling/
-.. _nox: https://nox.thea.codes
 .. _pdm-backend: https://backend.pdm-project.org
 .. _poetry-core: https://pypi.org/project/poetry-core/
-.. _tox: https://tox.wiki
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -137,7 +137,7 @@ Task runners
 These tools allow you to define and invoke "tasks", such as running tests,
 compiling documentation, regenerating some files, etc. Each task can be executed
 in a dedicated virtual environment with the dependencies it requires. The most
-common tools for this are :ref:`tox` and :ref:`nox`.
+common tools for this are tox_ and nox_.
 
 
 Integrated workflow tools
@@ -158,5 +158,7 @@ order:
 
 .. _flit-core: https://pypi.org/project/flit-core/
 .. _hatchling: https://pypi.org/project/hatchling/
+.. _nox: https://nox.thea.codes
 .. _pdm-backend: https://backend.pdm-project.org
 .. _poetry-core: https://pypi.org/project/poetry-core/
+.. _tox: https://tox.wiki
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -131,6 +131,15 @@ The standard tool for this task is :ref:`twine`.
 :ref:`deprecated <setup-py-deprecated>`, it is insecure.
 
 
+Task runners
+============
+
+These tools allow you to define and invoke "tasks", such as running tests,
+compiling documentation, regenerating some files, etc. Each task can be executed
+in a dedicated virtual environment with the dependencies it requires. The most
+common tools for this are :ref:`tox` and :ref:`nox`.
+
+
 Integrated workflow tools
 =========================
 
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -134,7 +134,7 @@ to build distributable wheels.
 Uploading to PyPI
 =================
 
-For projects hosted on GitHub, PyPI recommends using :ref:`trusted publishing
+For projects hosted on GitHub, it is recommended to use the :ref:`trusted publishing
 <trusted-publishing>`, which allows the package to be securely uploaded to PyPI
 from a GitHub Actions job. (This is not yet supported on software forges other
 than GitHub.)
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -134,7 +134,12 @@ to build distributable wheels.
 Uploading to PyPI
 =================
 
-The standard tool for this task is :ref:`twine`.
+For projects hosted on GitHub, PyPI recommends using :ref:`trusted publishing
+<trusted-publishing>`, which allows the package to be securely uploaded to PyPI
+from a GitHub Actions job. (This is not yet supported on software forges other
+than GitHub.)
+
+The other available method is to upload the package manually using :ref:`twine`.
 
 **Never** use ``python setup.py upload`` for this task. In addition to being
 :ref:`deprecated <setup-py-deprecated>`, it is insecure.
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -62,6 +62,12 @@ environment, for reproducibility purposes.
 Build backends
 ==============
 
+.. important::
+
+   Please, remember: this document does not seek to steer the reader towards
+   a particular tool, only to enumerate common tools. Different use cases often
+   need specialized workflows.
+
 Popular :term:`build backends <build backend>` for pure-Python packages include,
 in alphabetical order:
 
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -79,6 +79,8 @@ in alphabetical order:
   through plugins.
 
 - :ref:`setuptools` (which used to be the only build backend). It can be configured
+  using modern standards like `pyproject.toml`, but can also be extended
+  and supports customisation via `setup.py`.
   programmatically through the :file:`setup.py` file (but for basic metadata,
   :file:`pyproject.toml` is preferred).
 
@@ -95,11 +97,13 @@ the standard library in Python 3.12, although it still remains available from
 setuptools.
 
 For packages with :term:`extension modules <extension module>`, you may use
-setuptools. However, you can also use a build system with dedicated support for
-the language the extension is written in. For example, you could choose Meson or
-CMake for C, C++, Fortran and many other compiled languages; or Cargo for Rust
-specifically. You can then bridge this build system to Python using a dedicated
-build backend:
+a build system with dedicated support for the language the extension is written in,
+for example:
+
+- :ref:`setuptools` - natively supports C/C++ (with 3rd-party plugins for Go and Rust);
+- :ref:`meson-python` - C, C++, Fortran, and Rust and other languages supported by Meson;
+- :ref:`scikit-build-core` - C, C++, Fortran and other languages supported by CMake;
+- :ref:`maturin` - Rust, via Cargo.
 
 - :ref:`meson-python` for Meson,
 
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -5,14 +5,15 @@ Tool recommendations
 ====================
 
 The Python packaging landscape consists of many different tools. For many tasks,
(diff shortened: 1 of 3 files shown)


source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,140 @@
 Tool recommendations
 ====================
 
-If you're familiar with Python packaging and installation, and just want to know
-what tools are currently recommended, then here it is.
+The Python packaging landscape consists of many different tools. For many tasks,
+the Python Packaging Authority (PyPA, the umbrella organization which
```
chrysle, line 8:
```suggestion
the :term:`Python Packaging Authority <Python Packaging Authority (PyPA)>` (PyPA, the umbrella organization which
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,140 @@
...
-.. _pip-tools: https://github.com/jazzband/pip-tools
-.. _Poetry: https://python-poetry.org/
+- :ref:`setuptools`, the historical build backend. It can be configured
+  programmatically through the :file:`setup.py` file.
 
-Installation tool recommendations
-=================================
+  If you use setuptools, please be aware that it contains many deprecated
+  features which are currently kept for compatibility, but should not be used.
+  For example, do not use ``python setup.py`` invocations
```
chrysle, line 72:
```suggestion
  For example, do **not** use ``python setup.py`` invocations
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,140 @@
...
+.. todo:: Write a "pip vs. Conda" comparison, here or in a new discussion.
+
+Do **not** use ``easy_install`` (part of :ref:`setuptools`), which is deprecated
+in favor of pip (see :ref:`pip vs easy_install` for details). Likewise, do
+**not** use ``python setup.py install`` or ``python setup.py develop``, which
+are also deprecated (see :ref:`setup-py-deprecated` for background and
+:ref:`modernize-setup-py-project` for migration advice).
+
+
+Lock files
```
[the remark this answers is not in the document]
jeanas: Yes, but follow-up PR?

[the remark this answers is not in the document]
chrysle: Yup.

jeanas:
@chrysle Thanks for reviewing! I've addressed your comments, sometimes with slightly different wordings.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+- :ref:`setuptools`, the historical build backend. It can be configured
+  programmatically through the :file:`setup.py` file (but for basic metadata,
+  :file:`pyproject.toml` is preferred).
 
-Installation tool recommendations
-=================================
+  If you use setuptools, please be aware that it contains many deprecated
+  features which are currently kept for compatibility, but should not be used.
+  For example, do **not** use ``python setup.py`` invocations
+  (cf. :ref:`setup-py-deprecated`), the ``python_requires`` argument to
```
jeanas, line 77:
```suggestion
  (cf. :ref:`setup-py-deprecated`), the ``setup_requires`` argument to
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
 Tool recommendations
 ====================
 
-If you're familiar with Python packaging and installation, and just want to know
-what tools are currently recommended, then here it is.
+The Python packaging landscape consists of many different tools. For many tasks,
+the :term:`Python Packaging Authority <Python Packaging Authority (PyPA)>`
+(PyPA, the umbrella organization which encompasses many packaging tools and
```
webknjaz, line 9: I wouldn't say "organization", it's more like a bunch of people and projects that are sometimes connected loosely, and sometimes not.
https://www.pypa.io/ presents the PyPA as a working group, which is probably a better term to use here.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+maintains this guide) purposefully does not make a blanket recommendation; for
+example, the reason there exist many build backends is that the landscape was
```
webknjaz, line 11:
```suggestion
example, the reason there are many build backends is that the landscape was
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+opened in order to enable the development of new backends serving certain users'
```
webknjaz, line 12:
```suggestion
opened up in order to enable the development of new backends serving certain users'
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+
+
+Installing packages
+===================
+
+:ref:`Pip` is the standard tool to install packages from :term:`PyPI <Python
+Package Index (PyPI)>`. It can install into the global environment or into
+virtual environments. You may want to read pip's recommendations for
+:doc:`secure installs <pip:topics/secure-installs>`. Pip is available by default
+in most Python installations.
```
webknjaz, line 34: ".. through the stdlib package `ensurepip`".

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+
+Alternatively, for installing Python command line applications specifically,
+consider :ref:`pipx`, which is a wrapper around pip that installs each
```
webknjaz, line 37: "pip and virtualenv"?

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+application into a dedicated virtual environment. This avoids conflicts between
+the dependencies of different applications. On Linux, pipx is especially
```
webknjaz, line 39: "applications that are distributed as Python package distributions through the PyPI"?

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+important because it also avoids conflicts with the system (which are the reason
```
webknjaz, line 40: Not sure I'd separate Linux from others here — it's bad to sure globally shared envs in so many contexts. In GNU/Linux you can break some system software, including package managers, but a similar influence exists in other OS-independent situations.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+Build backends
+==============
 
-* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
-  application-specific dependencies from a shared Python installation. [4]_
+Popular :term:`build backends <build backend>` for pure-Python packages include:
 
-* Use `pip-tools`_, :ref:`pipenv`, or `poetry`_ to generate the fully-specified
-  application-specific dependencies, when developing Python applications.
+- Hatchling, which is part of :ref:`Hatch` (but can be used without
```
webknjaz, line 67: It's a part of the same Git repository but is published a fully separate distribution on PyPI. This should either clarify that it's a part of the Hatch ecosystem or just drop the implication.

Hatchling is a build backend, but Hatch is a frontend, installer, resolver, virtualenv manager, command runner, workflow tool. Let's not smash them together.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+  Hatch as well). Hatchling is extensible through a plugin system.
 
-.. _secure manner: https://pip.pypa.io/en/latest/topics/secure-installs/
-.. _pip-tools: https://github.com/jazzband/pip-tools
-.. _Poetry: https://python-poetry.org/
+- :ref:`setuptools`, the historical build backend. It can be configured
```
webknjaz, line 70: It's not historical. It's old, but it's also modern. It has a lot of legacy in its own codebase but that's an implementation detail.

cc @jaraco @abravalheri

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+  programmatically through the :file:`setup.py` file (but for basic metadata,
```
webknjaz, line 71: Or `setup.cfg`, if we're listing the configuration details here. But then, why don't we do this for the other list items? If we're listing the config files, why not mention `hatch.toml`, `.pip-tools.toml`, `Pipfile` above too?

I'd drop these from here too.

source/guides/tool-recommendations.rst
jeanas: I didn't mean that it shouldn't be used anymore, only that it was the only choice for a long time. I think this is useful to put on this page because many people will be wondering why the heck they didn't have to make any choice of backend when they created a package years ago. Maybe `` :ref:`setuptools`, which used to be the only build backend `` would be clearer?

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+  (cf. :ref:`setup-py-deprecated`), the ``setup_requires`` argument to
+  ``setup()`` (use the :ref:`[build-system] table
+  <pyproject-guide-build-system-table>` of :file:`pyproject.toml` instead), or
+  the ``easy_install`` command (cf. :ref:`pip vs easy_install`).
 
-* Use :ref:`pip` to install Python :term:`packages <Distribution Package>` from
-  :term:`PyPI <Python Package Index (PyPI)>`. [1]_ [2]_ Depending on how :ref:`pip`
-  is installed, you may need to also install :ref:`wheel` to get the benefit
-  of wheel caching. [3]_
+- Flit-core, part of :ref:`Flit` (but usable standalone). It is meant to be a
```
webknjaz, line 82: Same as with hatch — let's not mix self-sufficient front ends into a section about backends.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
 
+Do **not** use ``python setup.py sdist`` and ``python setup.py bdist_wheel`` for
+this task. All direct invocations of :file:`setup.py` are :ref:`deprecated
+<setup-py-deprecated>`.
 
-Publishing platform migration
-=============================
+If you have :term:`extension modules <extension module>` and want to distribute
+wheels for multiple platforms, use :ref:`cibuildwheel` as part of your CI setup
+to build distributable wheels.
```
webknjaz, line 121: Let's mention `pip wheel` that can be used as a low-level piece of automation for building platform-specific wheels in places where `cibuildwheel` may be unavailable. With a note that pypa/build can be used for the same.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
-`pypi.python.org <https://pypi.python.org>`_) has been phased out in favour
-of an updated implementation hosted at `pypi.org <https://pypi.org>`_.
 
-See :ref:`Migrating to PyPI.org` for more information on the status of the
-migration, and what settings to change in your clients.
+Uploading to PyPI
+=================
 
-----
+The standard tool for this task is :ref:`twine`.
```
webknjaz, line 127: Let's mention that some of the workflow tools (hatch, poetry, flit) invoke it under the hood.

Also, let's link the secretless publishing guide which wraps it but should probably be the first choice for any GitHub-hosted projects using GHA: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
+Integrated workflow tools
+=========================
 
-.. [4] Beginning with Python 3.4, ``venv`` will create virtualenv environments
-       with ``pip`` installed, thereby making it an equal alternative to
-       :ref:`virtualenv`. However, using :ref:`virtualenv` will still be
-       recommended for users that need cross-version consistency.
+These are tools that combine many features in one command line application, such
+as automatically managing virtual environments for a project, building
+distributions, uploading to PyPI, or creating and using lock files.
```
webknjaz, line 138:
```suggestion
distributions,
uploading to PyPI, or creating and using tool-specific lock files.
They often call the tools mentioned above under the hood.
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
-       convenience utilities for automatically populating distribution metadata
-       correctly that are provided by ``setuptools``. Being outside the
-       standard library, ``setuptools`` also offers a more consistent feature
-       set across different versions of Python, and (unlike ``distutils``),
-       recent versions of ``setuptools`` support all of the modern metadata
-       fields described in :ref:`core-metadata`.
+- :ref:`Hatch`,
+- :ref:`Flit`,
+- :ref:`PDM`,
+- :ref:`Poetry`.
```
webknjaz, line 143: I suppose `Pipenv` can be mentioned in this section too.

source/guides/tool-recommendations.rst
chrysle: > "pip and virtualenv"?

Pip and Venv, actually. Virtualenv might be supported in future versions.

source/guides/tool-recommendations.rst
webknjaz: Yep, that would be better. Thanks!

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
-  * :ref:`buildout`: primarily focused on the web development community
+Do **not** use :ref:`distutils`, which is deprecated, and has been removed from
+the standard library in Python 3.12, although it still remains available from
+setuptools.
 
-  * :ref:`spack`, :ref:`hashdist`, or :ref:`conda`: primarily focused
-    on the scientific community.
+For packages with :term:`extension modules <extension module>`, you may use
+setuptools, but consider using a build system dedicated to the language the
+extension is written in, such as Meson or CMake for C/C++, or Cargo for Rust,
```
eli-schwartz, line 97: Just want to point out that meson and cmake aren't specific to C/C++

They also support C#, Cuda, Swift, Asm, ObjC, ObjC++, Fortran.
Meson supports D, rust, vala, Java, and cython. CMake supports Hip and ISPC.

They are particularly valuable in cases where you have multiple languages, e.g. a mixture of rust and C++ and Fortran in a single project.

Meson also explicitly allows you to select `*.py files for installation`, do code generation of .py files (and other files), compile data such as gettext .po -> .mo catalogs, produce bundled `*.exe`s, and in general any form of custom build step. In theory CMake allows you to do precisely the same thing, although the scikit-build(-core) docs don't clearly indicate whether they support packaging that into a wheel.

tl;dr saying to use it "to support a dedicated extension language" is vastly underselling them. Even the "C/C++" part is simply *factually* wrong, if you need C/C++ support then setuptools works okay. The killer feature in the data science community is support for *Fortran*.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,144 @@
...
-=================================
 
-* Use :ref:`pip` in a `secure manner`_ to install a Python application and its
-  dependencies during deployment.
+Build backends
+==============
 
-* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
-  application-specific dependencies from a shared Python installation. [4]_
+Popular :term:`build backends <build backend>` for pure-Python packages include:
```
eli-schwartz, line 65:
```suggestion
Popular :term:`build backends <build backend>` for pure-Python packages include (in alphabetical order):
```

... and sort them.

eli-schwartz (changes requested):

source/guides/tool-recommendations.rst
jeanas: I'm just trying to list which backends support plugins / programmatic config. (I think this is important, because those that do have an ecosystem of plugins around them.) `pyproject.toml` was originally not mentioned here, but added per https://github.com/pypa/packaging.python.org/pull/1474#discussion_r1438610971

source/guides/tool-recommendations.rst
jeanas: I thought the pip developers were unhappy with `pip wheel` and considering to deprecate it (but I may be misremembering, and I can't find the pip issue right now). @pfmoore @pradyunsg What's your opinion?

source/guides/tool-recommendations.rst
jeanas: > Let's mention that some of the workflow tools (hatch, poetry, flit) invoke it under the hood.

I don't think they do? I didn't find any trace of this in the Hatch, Flit and Poetry repos.

source/guides/tool-recommendations.rst
```diff
@@ -79,23 +90,16 @@ Popular :term:`build backends <build backend>` for pure-Python packages include:
...
 
 For packages with :term:`extension modules <extension module>`, you may use
-setuptools, but consider using a build system dedicated to the language the
-extension is written in, such as Meson or CMake for C/C++, or Cargo for Rust,
-and bridging this build system to Python using a dedicated build backend:
+setuptools. However, you can also use a build system with dedicated support for
+the language the extension is written in. For example, you could choose Meson or
+CMake for C, C++, Fortran and many other compiled languages; or Cargo for Rust
+specifically. You can then bridge this build system to Python using a dedicated
+build backend:
```
abravalheri, line 102: This is probably more direct:

```suggestion
a build system with dedicated support for the language the extension is written in:

- :ref:`setuptools` - natively supports C/C++ (with 3rd-party plugins for Go and Rust);

- :ref:`meson-python` - C, C++, Fortran, and Rust and other languages supported by Meson;

- :ref:`scikit-build-core` - C, C++, Fortran and other languages supported by CMake;

- :ref:`maturin` - Rust, via Cargo.
```

source/guides/tool-recommendations.rst
```diff
@@ -62,12 +63,22 @@ environment, for reproducibility purposes.
...
+  be used without Hatch. Hatchling is extensible through a plugin system.
+
+- PDM-backend_ (developed with but separate from :ref:`PDM`). It provides build
+  hooks for extensibility.
 
-- :ref:`setuptools`, the historical build backend. It can be configured
+- Poetry-core_ (developed with but separate from :ref:`Poetry`). It is extensible
+  through plugins.
+
+- :ref:`setuptools` (which used to be the only build backend). It can be configured
```
abravalheri, line 81:
```suggestion
- :ref:`setuptools` (which used to be the only build backend). It can be configured
  using modern standards like `pyproject.toml`, but can also be extended
  and supports customisation via `setup.py`.
```

source/guides/tool-recommendations.rst
pfmoore: I'm not aware of plans to deprecate `pip wheel`, but it's worth noting that:

1. *Building* wheels isn't its sole/primary purpose - it downloads wheels, or downloads sdists and builds them into wheels. `pip wheel .` builds a wheel of the current project, but somewhat as a side issue (we could potentially use a cached wheel if we had a way to be sure it matched the source code).
2. There's no pip command to build a sdist, and there may well never be one.

If you want to build a local project, `build` is a much more appropriate tool.

jeanas: Ok, then I'd rather not mention `pip wheel` here.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,161 @@
...
-  project-specific dependencies from a shared Python installation. [4]_
+- :ref:`setuptools` (which used to be the only build backend). It can be
+  configured using modern standards like the :ref:`[project] table in
+  pyproject.toml <writing-pyproject-toml>`, but can also be extended and
+  customized via :file:`setup.py`.
 
-* If you're looking for management of fully integrated cross-platform software
-  stacks, consider:
+  If you use setuptools, please be aware that it contains many deprecated
+  features which are currently kept for compatibility, but should not be used.
```
abravalheri, line 87:
```suggestion
  If you use setuptools, please be aware that some features that predate
  standardisation efforts are now deprecated and only *temporarily kept* for compatibility.
```

jeanas:
I believe I've addressed all review comments. There were many reviews and it's not clear to me whether reviewers are now OK with the PR or just didn't have the time to take another look.

webknjaz:
I didn't have time, but I hope to find some.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
-       :ref:`setuptools` in June 2013, thereby making setuptools the default
-       choice for packaging.
 
-.. _distribute: https://pypi.org/project/distribute
+.. _flit-core: https://pypi.org/project/flit-core/
+.. _hatchling: https://pypi.org/project/hatchling/
+.. _nox: https://nox.thea.codes
+.. _pdm-backend: https://backend.pdm-project.org
+.. _poetry-core: https://pypi.org/project/poetry-core/
+.. _tox: https://tox.wiki
```
webknjaz, line 171: Tox is wired through intersphinx.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+- :ref:`Poetry`.
 
-.. [6] `distribute`_ (a fork of setuptools) was merged back into
-       :ref:`setuptools` in June 2013, thereby making setuptools the default
-       choice for packaging.
 
-.. _distribute: https://pypi.org/project/distribute
+.. _flit-core: https://pypi.org/project/flit-core/
+.. _hatchling: https://pypi.org/project/hatchling/
+.. _nox: https://nox.thea.codes
```
webknjaz, line 168: Nox is wired through intersphinx.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+- :ref:`PDM`,
+- :ref:`Pipenv`,
+- :ref:`Poetry`.
 
-.. [6] `distribute`_ (a fork of setuptools) was merged back into
-       :ref:`setuptools` in June 2013, thereby making setuptools the default
-       choice for packaging.
 
-.. _distribute: https://pypi.org/project/distribute
+.. _flit-core: https://pypi.org/project/flit-core/
```
webknjaz, line 166: Flit is wired through intersphinx.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+============
 
-.. [2] The acceptance of :pep:`453` means that :ref:`pip`
-       will be available by default in most installations of Python 3.4 or
-       later.  See the :pep:`rationale section <453#rationale>` from :pep:`453`
-       as for why pip was chosen.
+These tools allow you to define and invoke "tasks", such as running tests,
+compiling documentation, regenerating some files, etc. Each task can be executed
+in a dedicated virtual environment with the dependencies it requires. The most
+common tools for this are tox_ and nox_.
```
webknjaz, line 147:
```suggestion
common tools for this are :doc:`tox <tox:index>` and :doc:`nox <nox:index>`.
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+
+Build backends
+==============
 
+Popular :term:`build backends <build backend>` for pure-Python packages include,
+in alphabetical order:
 
-Application dependency management
-=================================
+- Flit-core_ -- developed with but separate from :ref:`Flit`. A minimal and
```
webknjaz, line 69:
```suggestion
- :doc:`Flit-core <flit:pyproject_toml>` -- developed with but separate from :ref:`Flit`. A minimal and
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+The standard tools to create and use virtual environments manually are
+:ref:`virtualenv` (PyPA project) and :doc:`venv <python:library/venv>` (part of
+the Python standard library, though missing some features of virtualenv).
+
+
+Installing packages
+===================
+
+:ref:`Pip` is the standard tool to install packages from :term:`PyPI <Python
+Package Index (PyPI)>`. It can install into the global environment or into
```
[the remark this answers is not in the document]
webknjaz: Also, "virtual" (as a concept/word) is not an opposite of "global". I think this may be confusing to some.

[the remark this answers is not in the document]
webknjaz: How about "isolated" vs. "shared"?

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+virtual environments. You may want to read pip's recommendations for
+:doc:`secure installs <pip:topics/secure-installs>`. Pip is available by default
+in most Python installations through the standard library package
+:doc:`ensurepip <python:library/ensurepip>`.
+
+Alternatively, consider :ref:`pipx` for the specific use case of installing Python
+command line applications that are distributed through PyPI.
```
webknjaz, line 38: I think it's wrong to emphasize CLI apps, while realistically, it is useful in the context of any packages that are considered distributing apps. GUI included, maybe even web(ish) apps with strict pins could be included. So I'd just drop this clarification as not factual.
```suggestion
applications that are distributed through PyPI.
```

webknjaz: Alternatively, clarify that there's more to it:
```suggestion
command line (and other types of) applications that are distributed through PyPI.
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+Pipx is a wrapper around pip and venv that installs each
+application into a dedicated virtual environment. This avoids conflicts between
+the dependencies of different applications, and also with the system
+(especially on Linux).
```
webknjaz, line 42:
```suggestion
the dependencies of different applications, and also with the
ones installed system-wide making use of the same Python
interpreter (especially on GNU/Linux).
```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+Lock files
+==========
+
+:ref:`pip-tools` and :ref:`Pipenv` are two recognized tools to create lock
+files, which contain the exact versions of all packages installed into an
+environment, for reproducibility purposes.
+
+
+Build backends
+==============
```
webknjaz, line 65: I feel like this might be the place to remind the readers not to treat this list as sorted by popularity.

```suggestion

.. important::

   Please, remember, this document does not seek to steer the reader towards
   a one-stop show solution. It merely enumerates tools observed within the
   packaging community. Different use-cases often need specialized workflows.

```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+the :term:`Python Packaging Authority <Python Packaging Authority (PyPA)>`
+(PyPA, the working group which encompasses many packaging tools and
+maintains this guide) purposefully does not make a blanket recommendation; for
+example, the reason there are many build backends is that the landscape was
+opened up in order to enable the development of new backends serving certain users'
+needs better than the previously unique backend, setuptools. This guide does
+point to some tools that are widely recognized, and also makes some
+recommendations of tools that you should *not* use because they are deprecated
+or insecure.
+
```
webknjaz, line 17: Should we mention that some tools focus on many things?
```suggestion

.. tip::

   Note that while most tools are is only mentioned under one
   section in this document, some of them cover several use
   cases.

```

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+  in pyproject.toml <writing-pyproject-toml>` whenever possible and keep
+  :file:`setup.py` only if programmatic configuration is needed. See
+  :ref:`setup-py-deprecated`.
 
-* If you're looking for management of fully integrated cross-platform software
-  stacks, consider:
+  Other examples of deprecated features you should **not** use include the
+  ``setup_requires`` argument to ``setup()`` (use the :ref:`[build-system] table
+  <pyproject-guide-build-system-table>` in :file:`pyproject.toml` instead), and
+  the ``easy_install`` command (cf. :ref:`pip vs easy_install`).
```
webknjaz, line 99: How about wrapping this with `.. caution::`?

source/guides/tool-recommendations.rst
webknjaz: Ah, interesting. I thought they did.

@jeanas still worth mention the GHA, I think.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,168 @@
...
+:ref:`deprecated <setup-py-deprecated>`, it is insecure.
 
-----
 
-.. [1] There are some cases where you might choose to use ``easy_install`` (from
-       :ref:`setuptools`), e.g. if you need to install from :term:`Eggs <Egg>`
-       (which pip doesn't support).  For a detailed breakdown, see :ref:`pip vs
-       easy_install`.
+Task runners
+============
```
webknjaz, line 142: I think it's more accurate to say
```suggestion
Workflow tools
==============
```

Saying "task runners" reduce them to a GNU/make replacement. But really, there's things like venv management that this skips.

webknjaz:
@jeanas another round of review complete.

source/guides/tool-recommendations.rst
jeanas: Yes, but I'd like to point to Flit-core here, and preferably not to the main Flit docs, to emphasize the distinction. (The text is `` Flit-core_ -- developed with but separate from :ref:`Flit`. ``.)

source/guides/tool-recommendations.rst
jeanas: Ah, I only see this now. Well, ok, why not.

source/guides/tool-recommendations.rst
jeanas: I'm not sure I understand your meaning of “one-stop shop”. I'm not a native speaker, but I'd read it as “a tool which does everything”, not “a build backend that you should always use”. Also I feel there is already a lot of warning at the beginning of the page.

source/guides/tool-recommendations.rst
jeanas: Well, the intent is that all tools presented are single-purpose, except those which are presented under “Integrated workflow tools” (but maybe that title should be more explicit?).

source/guides/tool-recommendations.rst
jeanas: I agree that this could be more precise, but if we make this change then I don't know how to call the section about Flit, Hatch, PDM, Pipenv and Poetry, which is currently called “Integrated workflow tools”. Thoughts?

pfmoore: Fundamentally, my thought here is that we're trying too hard to classify the tools available in the ecosystem. We might be better leaving it to the tools themselves to choose how they want to present themselves, and not have any headings at all...

jeanas: I'm not enthusiastic. The problem that this page solves is “it's too hard to understand how the tools fit together”. If it doesn't classify them, then it doesn't serve its purpose well, IMO.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Yeah, I wasn't completely sure about this one. Though, I see that you still applied the suggested change where I opted to link the doc that talks about `[build-system]` which is closer to the middle ground, it seems.

Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: I guess that's debatable. But fine. I don't have a better suggestion anyway. Do that.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: True. Let's wait for more opinions, then. And you're right, I did mean "a tool that does everything". Merely because a lot of people demand one without understanding that it entails. It seems to me that build backends are treated similarly even though, it's a different context. So I kinda wanted to assert the position once again in this section.

source/guides/tool-recommendations.rst
webknjaz: Perhaps yes. I also don't fully understand why tox/nox are ripped off of that section. I'd say they belong in it just as well.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: I suppose, I'd be in favor of combining these two sections into one. I feel like there's not enough differences in general to justify outing tox/nox vs. poetry/hatch et al.

source/guides/tool-recommendations.rst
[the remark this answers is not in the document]
webknjaz: @jeanas @pfmoore ideas?

source/guides/tool-recommendations.rst
webknjaz: @eli-schwartz are you okay with how it's been updated after your comment?

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,167 @@
...
+     If you use setuptools, please be aware that some features that predate
+     standardisation efforts are now deprecated and only *temporarily kept*
+     for compatibility.
 
-* Use :ref:`virtualenv` or :doc:`venv <python:library/venv>` to isolate
-  project-specific dependencies from a shared Python installation. [4]_
+     In particular, do **not** use direct ``python setup.py`` invocations. On the
+     other hand, configuring setuptools with a :file:`setup.py` file is still fully
+     supported, although it is recommended to use the modern :ref:`[project] table
+     in pyproject.toml <writing-pyproject-toml>` whenever possible and keep
```
webknjaz, line 94: I'd like this to avoid ignoring the existence of `setup.cfg`:
```suggestion
     in pyproject.toml <writing-pyproject-toml>` (or :file:`setup.cfg`) whenever possible and keep
```

source/guides/tool-recommendations.rst
webknjaz: Okay, but I still want https://github.com/pypa/packaging.python.org/pull/1474/files#r1466878549.

I'll mark this as resolved in favor of that thread.

source/guides/tool-recommendations.rst
jeanas: Sorry, I forgot about this one, done.

webknjaz: Marking this as resolved.

source/guides/tool-recommendations.rst
```diff
@@ -4,111 +4,172 @@
...
 
-* Use `twine <https://pypi.org/project/twine>`_ for uploading distributions
-  to :term:`PyPI <Python Package Index (PyPI)>`.
 
+Uploading to PyPI
+=================
 
-Publishing platform migration
-=============================
+For projects hosted on GitHub, PyPI recommends using :ref:`trusted publishing
```
webknjaz, line 137: Maybe, mention it's the PyPA?
```suggestion
For projects hosted on GitHub, the PyPA and the PyPI recommend using :ref:`trusted publishing
```

webknjaz: Alternatively,
```suggestion
For projects hosted on GitHub, it is recommended to use the :ref:`trusted publishing
```

jeanas: I'll go for the second, it's simpler.

source/guides/tool-recommendations.rst
jeanas: It looks like not everyone has the same mental model as me on the nox/tox vs. Flit/Hatch/PDM/Pipenv/Poetry distinction.

For background, the two I've used the most are tox and Hatch, and I've sometimes used nox (in this repo). Tox is an environment manager that lets me define virtual environments and run tasks in them. Meanwhile, Hatch also does environment management and task execution, but also: building, uploading to PyPI, and managing Python interpreters (plus reformatting the code). So I view tox as a single-purpose tool (environment manager) along with other single-purpose tools (build, twine, pyenv) and Hatch as a tool that combines the features of lots of single-purpose tools (build, twine, tox, pyenv).

Merging these two sections is possible but I'm not sure how I'd fill the description.

Making the list task-oriented is also possible, but then Flit, Hatch, Poetry and PDM all are going to get listed in three categories (building, publishing, managing environments, except Flit doesn't manage environments).

This also makes me think that by my categorization, Pipenv should be with tox and nox. (For my defense, I know very little about it except that it sparked controversies years ago.)

source/guides/tool-recommendations.rst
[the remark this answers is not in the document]
jeanas: Maybe we should just remove this sentence? Or replace it with “It can install into arbitrary environments.”?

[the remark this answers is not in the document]
pfmoore: I think remove the sentence. It isn't in the version of the text that's currently live, and I don't think it adds anything.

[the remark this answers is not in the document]
jeanas: Done.

source/guides/tool-recommendations.rst
webknjaz: Technically, tox is used in the same way. People automate calling the underlying tools (like build or twine) and have tox as a workflow tool with a unified interfaces. It doesn't have some things built into the codebase but it also has plugins and stuff. So the use case for it is exactly the same, it just allows more low-level configuration. And it does implement some things that others don't (like parallel execution of several things), but that's just one of the differences. Other tools also differ between each other.

All of these are workflow tools. They just allow different levels of configuration and different have things implemented by default. But from the project contributor (end-user of those tools / the caller), they provide a CLI to call those processes as a part of their workflow.

@gaborbernat correct me if I'm missing something...

source/guides/tool-recommendations.rst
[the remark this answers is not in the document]
webknjaz: Marking as resolved.

source/guides/tool-recommendations.rst
gaborbernat: You're correct.

jeanas: Ok, I've merged the two sections.

source/guides/tool-recommendations.rst
jeanas: I've merged them.

source/guides/tool-recommendations.rst
webknjaz: Marking as resolved, then.

source/guides/tool-recommendations.rst
webknjaz: Marking as resolved.

source/guides/tool-recommendations.rst
webknjaz: Marking as resolved.

source/guides/tool-recommendations.rst
webknjaz: @pfmoore do you have any thoughts on this?

pfmoore: Looks ok to me. I’d say “a particular solution” rather than “a one-stop shop solution”. But otherwise it’s good.

jeanas: Applied, with a slightly more concise wording.

webknjaz (approved):
I marked all the threads as resolved as there doesn't seem to be anything more actionable or blocking.

webknjaz:
The CI seems to be failing for an unrelated reason:
```console
(tutorials/installing-packages: line  444) broken    https://stackoverflow.com/a/14638025 - 403 Client Error: Forbidden for url: https://stackoverflow.com/a/14638025
```

The deps don't seem to differ compared to `main` so it's probably SO that's blocking HTTP requests from GH somehow. Will need to figure it out first.

willingc:
@webknjaz Maybe for CI, we change the SO link in the source to: https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix/14638025#14638025 It's where the CI link actually redirects if entered in the browser.

webknjaz:
@willingc that might work but needs checking. It should be done in a separate PR, though. An alternative would be adding a pattern to `linkcheck_ignore` if it's found that this problem exists with any StackOverflow links.

webknjaz:
@jeanas while that's being sorted out, what do you think about possibly linking https://pip.pypa.io/en/latest/topics/workflow/ / https://github.com/pypa/pip/pull/12503 in follow-ups? Not sure if it'd go in this specific doc, though.

PR #1479, opened 2024-01-04
Remove the explicit list of PaaS platforms

brettcannon:
The list can be misconstrued as an endorsement and we can't keep up w/ all PaaS platforms which support Python to make the list comprehensive/impartial.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1479.org.readthedocs.build/en/1479/


sinoroc (approved):

chrysle (approved):
Thanks!

PR #1488, opened 2024-01-25
Use intersphinx to link to `hashlib` docs

edgarrmondragon:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1488.org.readthedocs.build/en/1488/

### DIFF
--- a/source/specifications/direct-url-data-structure.rst
+++ b/source/specifications/direct-url-data-structure.rst
@@ -79,10 +79,10 @@ MUST be present as a dictionary with the following keys:
 
   These hash names SHOULD always be normalized to be lowercase.
 
-  Any hash algorithm available via ``hashlib`` (specifically any that can be passed to
-  ``hashlib.new()`` and do not require additional parameters) can be used as a key for
+  Any hash algorithm available via :py:mod:`hashlib` (specifically any that can be passed to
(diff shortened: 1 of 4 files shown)


chrysle (approved):
Thanks! Actually, this should be done consistently in all specifications.

source/specifications/direct-url-data-structure.rst
```diff
@@ -79,10 +79,10 @@ MUST be present as a dictionary with the following keys:
 
   These hash names SHOULD always be normalized to be lowercase.
 
-  Any hash algorithm available via ``hashlib`` (specifically any that can be passed to
-  ``hashlib.new()`` and do not require additional parameters) can be used as a key for
+  Any hash algorithm available via :py:mod:`hashlib` (specifically any that can be passed to
+  :py:func:`hashlib.new()` and do not require additional parameters) can be used as a key for
```
webknjaz, line 83: Does this actually work with `()`? I always thought they should be omitted in this syntax..

webknjaz: So it does work: https://python-packaging-user-guide--1488.org.readthedocs.build/en/1488/specifications/direct-url-data-structure/#archive-urls.

TIL!

PR #1491, opened 2024-01-26
Replace regex which allows dubble hyphens by one that does not.

Darsstar:
The current version disallows tripple hyphens but allows dubble hyphens. (It also seems overcomplicated)

Alternatively we can make it a negative lookbehind instead of a negative lookahead, but the complexity is much higher.

See https://regex101.com/r/Xyu0bt/1 for proof the current version matches double hyphens

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1491.org.readthedocs.build/en/1491/

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -631,7 +631,7 @@ of lowercase ASCII letters, ASCII numbers, and hyphen. It must start and end
 with a letter or number. Hyphens cannot be followed by another hyphen. Names are
 limited to those which match the following regex (which guarantees unambiguity)::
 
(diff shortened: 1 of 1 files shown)


chrysle:
LGTM. @brettcannon you made the last change to this regx in #1076. Do you agree with this?

jeanas (approved):

brettcannon (approved):

brettcannon:
Yep, the change seems to work!

chrysle (approved):

chrysle:
Thanks!

PR #1493, opened 2024-01-27
Re-ordered build backends in packaging tutorial.

venthur:
The new order reflects more the popularity of build backends used in pyproject.toml files:

* Setuptools: ~50%
* Poetry: ~33%
* Hatch: ~10%
* Flit: ~3%
* Other: ~4%

Setuptools as a build backend used in pyproject.toml, has the biggest "market share" since 2019, followed by poetry (which is not mentioned here). Hatch only started in 2022 and never gained more than 10%.

Setuptools is also the [default](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour) build backend when you don't specify one in your pyproject.toml.

I think it makes sense to swap the tabs and put setuptools before hatch in this tutorial. No information is lost, we just point the readers to the most popular choice by default.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1493.org.readthedocs.build/en/1493/

### DIFF
--- a/source/tutorials/packaging-projects.rst
+++ b/source/tutorials/packaging-projects.rst
@@ -117,8 +117,8 @@ and input files. Build backends have different levels of functionality, such as
 whether they support building :term:`extension modules <Extension Module>`, and
 you should choose one that suits your needs and preferences.
 
-You can choose from a number of backends; this tutorial uses :ref:`Hatchling
-<hatch>` by default, but it will work identically with :ref:`setuptools`,
+You can choose from a number of backends; this tutorial uses :ref:`setuptools`,
+by default, but it will work identically with :ref:`Hatchling <hatch>`,
 :ref:`Flit <flit>`, :ref:`PDM <pdm>`, and others that support the ``[project]``
 table for :ref:`metadata <configuring metadata>`.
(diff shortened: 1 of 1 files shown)


sinoroc:
This has been debated many times. Not sure this approach specifically has been suggested and debated, but I guess it has.

In general, there are reasons to avoid pushing/nudging (intentionally or not) users (especially beginners) towards setuptools, because it has a lot of pitfalls: a whole bunch of setuptools features are deprecated, the generally available documentation (blogs, tutorials, StackOverflow answers, and so on) is often out of date or plainly wrong.

chrysle:
Let's leave this how it is? The tool recommendations page will serve the purpose of build backend guidance better, I think.

jeanas:
With 4 contributors including 2 maintainers (@webknjaz, @chrysle, @sinoroc, me) in agreement with @sinoroc's comment above, I'm going to close this PR. For a recent summary of the reasons behind the choice of order, please see https://github.com/pypa/packaging.python.org/issues/1468#issuecomment-1875703126. Thank for your understanding and for trying to improve the guide.

PR #1504, opened 2024-02-21
Update writing-pyproject-toml.rst: Meson

ngrigri:
Add "Meson" in "Declaring the build backend" section.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1504.org.readthedocs.build/en/1504/

### DIFF
--- a/source/guides/writing-pyproject-toml.rst
+++ b/source/guides/writing-pyproject-toml.rst
@@ -86,7 +86,13 @@ Here are the values for some common build backends:
         requires = ["pdm-backend"]
         build-backend = "pdm.backend"
 
+.. tab:: Meson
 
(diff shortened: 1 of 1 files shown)


chrysle:
The `meson-python` build backend is a tool specific to Meson projects. This change seems to suggest it can be / is used generally. I think we should limit the build backends showcased in this guide to the most common ones.

webknjaz:
Looks like everyone's in agreement that this document isn't mean to enumerate every build backend that ever existed for any purpose, just give a few common/generic examples to start with. An attentive reader would be able to figure out they need to read the docs of the respective backend anyway, which would show them what to configure in a more coherent manner, anyway. So I'm going to reject this PR, unfortunately.

PR #1507, opened 2024-02-22
Update link that was failing on CI

willingc:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1507.org.readthedocs.build/en/1507/

### DIFF
--- a/source/tutorials/installing-packages.rst
+++ b/source/tutorials/installing-packages.rst
@@ -445,7 +445,7 @@ directory to your :envvar:`PATH`:
   ``python -m site --user-base`` and adding ``bin`` to the end. For example,
   this will typically print ``~/.local`` (with ``~`` expanded to the absolute
   path to your home directory) so you'll need to add ``~/.local/bin`` to your
-  ``PATH``.  You can set your ``PATH`` permanently by `modifying ~/.profile`_.
+  ``PATH``.  You can set your ``PATH`` permanently by modifying ``~/.profile``.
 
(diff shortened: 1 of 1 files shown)


source/tutorials/installing-packages.rst
```diff
@@ -457,7 +457,7 @@ directory to your :envvar:`PATH`:
   ``PATH`` changes to take effect.
 
 .. _user base: https://docs.python.org/3/library/site.html#site.USER_BASE
-.. _modifying ~/.profile: https://stackoverflow.com/a/14638025
+.. _modifying ~/.profile: https://stackoverflow.com/questions/14637979/how-to-permanently-set-path-on-linux-unix/14638025#14638025
```
willingc, line 460:
```suggestion
```

webknjaz:
Why do you think it doesn't?

webknjaz:
@dukecat0 @jwodder @theacodes since you were the ones to touch/add that link, do you have any opinion on keeping or dropping it?

dukecat0:
I personally prefer keeping it as a quick reference for beginners, who are the main audience of this tutorial.

webknjaz:
Yeah, that's my concern as well. I prefer keeping the reference and ignoring the link instead.

willingc:
Sounds good. I'll add back the link.

willingc:
Close and replace with #1510 to ignore stack overflow for linkchecks

PR #1510, opened 2024-02-26
Add stack overflow to linkcheck ignore list

willingc:
Related to failing queries to Stack overflow https://github.com/pypa/packaging.python.org/pull/1474

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1510.org.readthedocs.build/en/1510/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -107,7 +107,8 @@ linkcheck_ignore = [
     # Temporarily ignored. Ref:
     # https://github.com/pypa/packaging.python.org/pull/1308#issuecomment-1775347690
     "https://www.breezy-vcs.org/*",
-    # Blocked due to https://github.com/pypa/packaging.python.org/pull/1474
+    # Ignore while StackOverflow is blocking GitHub CI. Ref:
+    # https://github.com/pypa/packaging.python.org/pull/1474
     "https://stackoverflow.com/*",
 ]
 linkcheck_retries = 5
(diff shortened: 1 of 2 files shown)


willingc:
Thanks for the reviews and guidance @webknjaz. This should ignore the problematic link.

source/conf.py
```diff
@@ -107,6 +107,8 @@
     # Temporarily ignored. Ref:
     # https://github.com/pypa/packaging.python.org/pull/1308#issuecomment-1775347690
     "https://www.breezy-vcs.org/*",
+    # Blocked due to https://github.com/pypa/packaging.python.org/pull/1474
```
webknjaz, line 110: I don't think this comment is correct. The check started failing due to reasons that are not related to the PyPUG project. It's because Stackoverflow.com now blocks queries coming from GitHub Actions CI/CD or likely the entire GitHub platform, probably by IP range.
That PR is merely where we first noticed the effect of that.

willingc: That was poorly worded on my part. Will update. Thanks!

webknjaz: Thank you! Marking the thread as resolved.

webknjaz (approved):

webknjaz:
:roll_eyes: This is now failing due to https://github.com/pypa/pip/pull/12537 :laughing:

willingc:
Fixed the unrelated failing link @webknjaz.

webknjaz:
Oh, I already made a branch with a different fix that doesn't involve removing the link, just got distracted and didn't open a PR :smiley: Will probably do that anyway.

webknjaz:
Here I just make a link use an immutable URL to the last commit that had this file https://github.com/pypa/packaging.python.org/pull/1514. I think it might still be useful in the context it's presented in that document. Removing it should probably be done together with changing that context there.

PR #1514, opened 2024-02-29
Point pip's `setup.py` link to its latest version

webknjaz:
This file no longer exists on the `main` branch as of https://github.com/pypa/pip/pull/12537 /
https://github.com/pypa/pip/commit/0ad4c94b.

It's an alternative to removing the link performed in https://github.com/pypa/packaging.python.org/commit/3e65739d8078603eec17bc2a38a1bab1a74ccb37.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1514.org.readthedocs.build/en/1514/

### DIFF
--- a/source/guides/single-sourcing-package-version.rst
+++ b/source/guides/single-sourcing-package-version.rst
@@ -9,7 +9,8 @@ Single-sourcing the package version
 There are many techniques to maintain a single source of truth for the version
 number of your project:
 
-#.  Read the file in :file:`setup.py` and get the version. Example::
+#.  Read the file in :file:`setup.py` and get the version. Example (from `pip setup.py
+    <https://github.com/pypa/pip/blob/003c7ac/setup.py>`_)::
 
         import codecs
         import os.path


willingc (approved):
Thanks @webknjaz

source/guides/single-sourcing-package-version.rst
```diff
@@ -9,7 +9,8 @@ Single-sourcing the package version
 There are many techniques to maintain a single source of truth for the version
 number of your project:
 
-#.  Read the file in :file:`setup.py` and get the version. Example::
+#.  Read the file in :file:`setup.py` and get the version. Example (from `pip setup.py
+    <https://github.com/pypa/pip/blob/f4a54391/setup.py>`_)::
```
sinoroc, line 13: But this link does not point to a `setup.py` file that shows what we want to show, does it?

Shouldn't we link to a version of `setup.py` that does "version single sourcing"?

webknjaz: You're right! Thanks for pointing this out! An older commit would be better.

webknjaz: Updated to https://github.com/pypa/pip/blob/003c7ac/setup.py. Marking this thread as resolved.

PR #1530, opened 2024-04-11
Update packages installed in venv

cbrnr:
As of Python 3.12, `python -m venv` does not install `setuptools` in the venv anymore (https://docs.python.org/3/library/venv.html#creating-virtual-environments).

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1530.org.readthedocs.build/en/1530/


chrysle (approved):

chrysle:
Thanks!

PR #1537, opened 2024-04-30
Update glossary.rst

Sam1320:
fix seeming typo and add link to the correct term

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1537.org.readthedocs.build/en/1537/

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -15,7 +15,7 @@ Glossary
     Build Backend
 
         A library that takes a source tree
-        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or 
+        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or
         :term:`built distribution <Built Distribution>` from it.
         The build is delegated to the backend by a
         :term:`frontend <Build Frontend>`.
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -15,7 +15,8 @@ Glossary
     Build Backend
 
         A library that takes a source tree
-        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or :term:`built distribution <Built Distribution>` from it.
+        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or 
+        :term:`built distribution <Built Distribution>` from it.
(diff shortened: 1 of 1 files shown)


webknjaz (changes requested):
The phrasing is sub-optimal but it's correct: a build backend does indeed build an sdist or a wheel.

Sam1320:
> The phrasing is sub-optimal but it's correct: a build backend does indeed build an sdist or a wheel.

Good point. What about the phrasing in the last commit? I feel is less ambiguous this way. I also removed the reference to "wheel" for clarity and also because its already inside the "Built Distribution" description.

source/glossary.rst
```diff
@@ -14,9 +14,8 @@ Glossary
 
     Build Backend
 
-        A library that takes a source tree or
-        :term:`source distribution <Source Distribution (or "sdist")>`
-        and builds a source distribution or :term:`wheel <Wheel>` from it.
+        A library that takes a source tree
+        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or :term:`built distribution <Built Distribution>` from it.
```
chrysle, line 18: Let's not make the line too long.

```suggestion
        and builds a :term:`source distribution <Source Distribution (or "sdist")>` or 
        :term:`built distribution <Built Distribution>` from it.
```

sinoroc (approved):
I wrote the initial glossary entry. I recall struggling to find a phrasing I would be entirely happy with, but gave up and settled on this awkward phrasing.

I guess, technically the step from sdist to wheel goes through unpacking the sdist into a source tree before this source tree to the build backend. I am not sure the build backends really do accept an sdist as input. So this new phrasing might be correct enough.

webknjaz:
> I am not sure the build backends really do accept an sdist as input.

They don't. The frontends do the unpacking as a part of the ephemeral build environment provisioning and call the backends against that dir on disk. cibuildwheel does that, and build recently got this feature. Pip was the first frontend to implement this logic, I think.

webknjaz (approved):

Mohdhafez86 (approved):

PR #1538, opened 2024-05-03
Update packaging-projects.rst: rephrase first explanation of __init__.py

wyattscarpenter:
I thought the explanation of __init__.py was kind of confusing, so I have improved it somewhat.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1538.org.readthedocs.build/en/1538/


abravalheri:
To be honest, that is still a weak reasoning.

Users can import a directory even if it does not have an `__init__.py`. Sure, it will not be a "regular" package, it will be a "namespace" package, but the text does not explain what is so great about regular packages and why they are recommended over namespaces.

chrysle (approved):

PR #1539, opened 2024-05-05
publish-to-test-pypi: bump action versions

woodruffw:
This bumps two referenced action versions:

1. The latest stable `setup-python` is `v5`, which is compatible with the previous use.
2. The example workflow previously referred to a very old version of `gh-action-sigstore-python`, which was no longer supported. This bumps it to a newer version that is similarly compatible with the previous use.

For (2), see https://github.com/sigstore/gh-action-sigstore-python/issues/122 for an example of a user who was snarled by the old version :slightly_smiling_face: 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1539.org.readthedocs.build/en/1539/


chrysle (approved):

webknjaz (approved):
Thank you!

PR #1542, opened 2024-05-07
Update installing-using-pip-and-virtual-environments.rst

dharmatech:
Use `Get-Command` if on PowerShell

`where` in PowerShell is an alias for `Where-Object`.

If you use `where` in Powershell, no output is returned. Use `Get-Command` instead:

image

### DIFF
--- a/source/guides/installing-using-pip-and-virtual-environments.rst
+++ b/source/guides/installing-using-pip-and-virtual-environments.rst
@@ -103,7 +103,8 @@ Python interpreter:
 
     .. code-block:: bat
 
-        where python
+        Get-Command python   # If using PowerShell
+        where python         # If using Command Prompt
 
 While the virtual environment is active, the above command will output a
 filepath that includes the ``.venv`` directory, by ending with the following:


chrysle:
Thanks! Let's put the command into another code block with PowerShell code lexer?

webknjaz:
Hey @dharmatech, would you address the review comment?

source/guides/installing-using-pip-and-virtual-environments.rst
```diff
@@ -103,7 +103,8 @@ Python interpreter:
 
     .. code-block:: bat
 
-        where python
+        Get-Command python   # If using PowerShell
+        where python         # If using Command Prompt
```
merwok, line 107:
```suggestion
        where python

..tab:: Windows (PowerShell)

    .. code-block:: bat
        Get-Command python
```

merwok: and edit line 102 to be `.. tab:: Windows (Command Prompt)`

PR #1543, opened 2024-05-07
Update section-install.rst

dharmatech:
installing-using-virtualenv has no instructions.

Removing the reference to it.

Here's what the page looks like:

image


dwcoen1234:
[Import transactions (3).pdf](https://github.com/pypa/packaging.python.org/files/15245911/Import.transactions.3.pdf)

PR #1551, opened 2024-05-22
binary-distribution-format: Update a mention of distutils

geofft:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1551.org.readthedocs.build/en/1551/

PR #1564, opened 2024-06-19
Add HTML snippet for Plausible metrics

### LINKED ISSUE #1561: Enable Plausible analytics for the hosted packaging user guide
Activity tracking was recently enabled on the docs.python.org site, and it shows that one of the most frequently visited pages is the one for `venv`: https://plausible.io/docs.python.org/

The detailed page for the Google referals provides more detail on the search terms used: https://plausible.io/docs.python.org/?filters=((is,page,(/3/library/venv.html)),(is,source,(Google)))

With the high 86% bounce rate, it would be very interesting to know whether folks are leaving the site entirely because their question has already been answered, or if they're following the cross-reference link to https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#create-and-use-virtual-environments

Since `packaging.python.org` is a peer subdomain of `docs.python.org`, enabling the data collection should just be a matter of adding in `<script defer data-domain="docs.python.org" src="https://plausible.io/js/script.js"></script>` as part of the Sphinx template when deploying to the main packaging.python.org environment.

The CPython docs use an [environment variable](https://github.com/python/cpython/pull/119977/files) to control whether or not to insert the analytics snippet (so it gets omitted in local builds of the packaging docs), and I think it would make sense to do the same thing for the packaging user guide (we could even use the same environment variable name: `PYTHON_DOCS_ENABLE_ANALYTICS`)

ncoghlan:
Metrics are only collected for the main packaging.python.org documentation deployments.

Closes #1561

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1564.org.readthedocs.build/en/1564/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -1,6 +1,14 @@
 # -- Project information ---------------------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
+import os
+
+# Some options are only enabled for the main packaging.python.org deployment builds
+RTD_BUILD = bool(os.getenv("READTHEDOCS"))
+RTD_PR_BUILD = RTD_BUILD and os.getenv("READTHEDOCS_VERSION_TYPE") == "external"
+RTD_URL = os.getenv("READTHEDOCS_CANONICAL_URL")
+RTD_CANONICAL_BUILD = RTD_BUILD and not RTD_PR_BUILD and "packaging.python.org" in RTD_URL
+
 project = "Python Packaging User Guide"
 
 copyright = "2013–2020, PyPA"
@@ -54,6 +62,15 @@ html_theme = "furo"
 
 html_favicon = "assets/py.png"
 html_last_updated_fmt = ""
+
+_metrics_js_files = [
+    ("https://plausible.io/js/script.js", {"data-domain": "packaging.python.org"})
+]
+html_js_files = []
+if RTD_CANONICAL_BUILD:
+    # Enable collection of the visitor metrics reported at
+    # https://plausible.io/packaging.python.org
+    html_js_files.extend(_metrics_js_files)
 
 # -- Options for HTML help output ------------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-help-output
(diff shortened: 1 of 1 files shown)


source/conf.py
```diff
@@ -55,6 +65,15 @@
 html_favicon = "assets/py.png"
 html_last_updated_fmt = ""
 
+_metrics_js_files = [
+    ("https://plausible.io/js/script.js", {"data-domain": "packaging.python.org"})
```
hugovk, line 69: Do we need the `defer` attribute from https://github.com/pypa/packaging.python.org/issues/1561#issuecomment-2178330434?

source/conf.py
```diff
@@ -55,6 +65,15 @@
...
+]
+html_js_files = []
+if RTD_CANONICAL_BUILD:
+    # Enable collection of the visitor metrics reported at
+    # https://plausible.io/packaging.python.org
+    html_js_files.extend(_metrics_js_files)
```
ncoghlan, line 75: Reading https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file, I'm wondering if this should use `sphinx.application.Sphinx.add_js_file` so we can set `loading_method="defer"`

source/conf.py
ncoghlan: Although, if `"defer": None` in the script tag options does the right thing, that would be much easier

ncoghlan: The [Sphinx test suite](https://github.com/sphinx-doc/sphinx/blob/37ff33dff7265908c22ef82ee7a85f37be4dbd07/tests/roots/test-html_assets/conf.py#L9) uses `async="async"` in a test case, so presumably, `defer="defer"` will also work.

ncoghlan: TIL something new about HTML script tags (i.e. setting the boolean options to a string to cope with interfaces that expect all attributes to have values, even boolean ones)

hugovk: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer says it's a Boolean attribute, so I think `defer="defer"` should be fine.

It's used a lot: https://github.com/search?q=%2Fdefer%3D%22defer%22%2F&type=code

hugovk (approved):
Looks good, shall we merge and check if we get any hits?

ncoghlan:
Alas, I can't give my own PR merge approval, so we'll need a tick in the box from @pradyunsg to land it. (I reviewed and merged #1551 to confirm that was the problem, rather than there being any issues with my write access)

pradyunsg (approved):

PR #1572, opened 2024-07-12
Fix conflicting classification of install

covracer:
Lines 21 to 23 state that the `install` command must not be run anymore.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1572.org.readthedocs.build/en/1572/

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -141,7 +141,6 @@ This guide does not make suggestions of replacement solutions for those commands
     * ``easy_install``
     * ``editable_wheel``
     * ``egg_info``
-    * ``install``
     * ``install_data``
     * ``install_egg_info``
     * ``install_headers``


jeanas (approved):

sinoroc (approved):

webknjaz (changes requested):
Thanks for the change! None of the commands should be run anymore. I think that this just states for the record that things still exist, but the document does not provide suggestions on how to deal with them right now. Removing one but leaving the others doesn't feel right. Not sure we'd want it in this form.

sinoroc:
@webknjaz for all the commands in this one particular list we do not make any recommendation for an alternative solution, but the issue pointed here is that we actually do have a recommendation for `install` (`python -m pip install .`). So `install` does not need to be listed here.

So while the text in the body of the commit message is incorrect, the change itself is correct.

sinoroc:
@webknjaz Can we get this merged? This is a straightforward fix. I was the one to write this page and introduce this mistake. I am sure this fix is correct.

covracer:
> Thanks for the change! None of the commands should be run anymore. I think that this just states for the record that things still exist, but the document does not provide suggestions on how to deal with them right now. Removing one but leaving the others doesn't feel right. Not sure we'd want it in this form.

I don't understand. Lines 21 to 26 list 4 commands as "must not be run": install, develop, sdist, bdist_wheel.

Lines 121 through 153 lists 25 commands as remaining / no recommended replacement.

I'm trying to fix the one overlapping element between the two lists.

sinoroc:
@webknjaz @chrysle Can we get this merged? It is a simple, uncontroversial fix. I am the original author of this page, I know that this fix is correct.

webknjaz (approved):

sinoroc:
Thanks : )

PR #1575, opened 2024-07-15
Refer to platform tag as such in use example

scop:
There's no arch tag.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1575.org.readthedocs.build/en/1575/


chrysle (approved):

chrysle:
Kiitos!

PR #1578, opened 2024-07-25
WIP: updated single_source_version with a much simpler page.

ChrisBarker-NOAA:
Initial comment edited as of Jul 30 2024

This was quick draft I whipped out -- turns out there's interest, 

As per discussion on Discourse, and #1276, here is a single source page that is far simpler, and essentially refers users to their build systems for instructions.

NOTE: There is disagreement about what *should* be done re versioning and a `__version__` attribute. This PR is NOT the place to resolve those disagreements.

Rather, I'm hoping this page can capture the state of the practice without specification of every detail of what should be done.

I am hoping to at least capture what I think is the consensus:

The very short version:
- for a given distribution, the version string(s) should be fully consistent, and be specified in only one place.

Some detailed points:

1) When a distribution is built, it should have proper PEP-conforming version info, which is consistent within that entire distribution.

2) The user, when setting up their distribution build, should specify the version string in *only one place* (or only one way) - and the build system should be responsible for making sure it is consistent everywhere else (e.g. the filenames, the metadata, or attributes.

3) **If** a distribution maintainer wants to put an attribute in the installed importable module it should be consistent with the metadata of the installed package.

4) It's best for the community that an importable version attribute uses a consistent name: `__version__` is a defacto standard for that name.

If you disagree with these points, please make that argument specifically, rather than suggesting edits that don't conform to these points. That is, let's get a consensus for what this doc should say first, and then figure out how to say it.

I think (4) is the perhaps the most controversial -- though I don't understand why. I can't see how it benefits ANYONE for different distributions to use a different name for the version string. Frankly, the only explanation I can think of for why people advocate for other names, is that that way it's harder for anyone to *think* that having a version string is a standard -- and they really don't want having a version string to be a standard.
But I ask that if you feel that way, please make sure that the text is clear that providing a version string is OPTIONAL, rather than arguing against us using the most common name in the text.

Anyway -- my goal with this doc is to capture those points, and not be prescriptive about  what should be done about having or not having a version string in the installed package.

NOTE: the preview doesn't seem to be working -- I get a 404 *maybe* because of the merge conflict? (which I thought I'd wait for something closer to final before worrying about)

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1578.org.readthedocs.build/en/1578/

### DIFF
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -9,15 +9,13 @@ Single-sourcing the Project Version
 
 One of the challenges in building packages is that the version string can be required in multiple places.
 
-* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
+* It needs to be specified when building the package (e.g. in :file:``pyproject.toml``)
    - That will assure that it is properly assigned in the distribution file name, and in the installed package.
 
-* Some projects prefer that there be a version string available as an attribute in the importable module, e.g::
+* A package may, if the maintainers think it useful, to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level ``__version__`` attribute.
+The value of ``__version__`` attribute and that passed to the build system can (and should) be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
-    import a_package
-    print(a_package.__version__)
-
-* In the metadata of the artifacts for each of the packaging ecosystems    
+Should a package not set a top level ``__version__`` attribute, the version can still be accessed using ``importlib.metadata.version("distribution_name")``.
 
 While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
 
@@ -27,17 +25,15 @@ In general, the options are:
 
 2) The version can be hard-coded into the `pyproject.toml` file -- and the build system can copy it into other locations it may be required.
 
-3) The version string can be hard-coded into the source code -- either in a special purpose file, such as `_version.txt`, or as a attribute in the `__init__.py`, and the build system can extract it at build time.
-
-If the version string is not in the source, it can be extracted at runtime with code in `__init__.py`, such as::
-
-    import importlib.metadata
-    __version__ = importlib.metadata.version('the_distribution_name')
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as ``_version.txt``, or as a attribute in the ``__init__.py``, and the build system can extract it at build time.
 
 
 Consult your build system documentation for how to implement your preferred method.
 
-Here are the common ones:
+.. _how_to_set_version_links:
+
+Build System Version Handling
+----------------------------
 
 * `Hatch <https://hatch.pypa.io/1.9/version/>`_
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -32,7 +32,7 @@ Consult your build system's documentation for their recommended method.
 .. _how_to_set_version_links:
 
 Build System Version Handling
-----------------------------
+-----------------------------
 
 * `Flit <https://flit.pypa.io/en/stable/>`_
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -16,6 +16,7 @@ One of the challenges in building packages is that the version string can be req
 
     import a_package
     print(a_package.__version__)
+* In the metadata of the artifacts for each of the packaging ecosystems    
 
 While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -17,7 +17,7 @@ The value of ``__version__`` attribute and that passed to the build system can (
 
 In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
 
-While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
+To ensure that version numbers do not get out of sync, it is recommended that there is a single source of truth for the version number.
 
 In general, the options are:
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -12,7 +12,7 @@ One of the challenges in building packages is that the version string can be req
 * It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
    - That will assure that it is properly assigned in the distribution file name, and in the installed package metadata.
 
-* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the installed package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
 In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
 :Page Status: Complete
 :Last Reviewed: 2015-09-08
 
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in pyproject.toml)
+   - That will assure that it is properly assigned in the distribution file name, and in teh installed package.
 
-#.  Read the file in ``setup.py`` and parse the version with a regex. Example (
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* Some projects require that there be a version string available as an attribute in the importable module, e.g::
 
-        def read(*names, **kwargs):
-            with io.open(
-                os.path.join(os.path.dirname(__file__), *names),
-                encoding=kwargs.get("encoding", "utf8")
-            ) as fp:
-                return fp.read()
+    import a_package
+    print(a_package.__version__)
 
-        def find_version(*file_paths):
-            version_file = read(*file_paths)
-            version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]",
-                                      version_file, re.M)
-            if version_match:
-                return version_match.group(1)
-            raise RuntimeError("Unable to find version string.")
+While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
 
-        setup(
-           ...
-           version=find_version("package", "__init__.py")
-           ...
-        )
+In general, the options are:
 
-    .. note::
+1) If the code is in a version control system (VCS), e.g. git, then the version can be extracted from the VCS.
 
-        This technique has the disadvantage of having to deal with complexities of regular expressions.
+2) The version can be hard-coded into the `pyproject.toml` file -- and the build system can copy it into other locations it may be required.
 
-#.  Use an external build tool that either manages updating both locations, or
-    offers an API that both locations can use.
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as `_version.txt`, or as a attribute in the `__init__.py`, and the build system can extract it at build time.
 
-    Few tools you could use, in no particular order, and not necessarily complete:
-    `bumpversion <https://pypi.python.org/pypi/bumpversion>`_,
-    `changes <https://pypi.python.org/pypi/changes>`_, `zest.releaser <https://pypi.python.org/pypi/zest.releaser>`_.
+If the version string is not in the source, it can be extracted at runtime with code in `__init__.py`, such as::
+
+    import importlib.metadata
+    __version__ = importlib.metadata.version('the_distribution_name')
 
 
-#.  Set the value to a ``__version__`` global variable in a dedicated module in
-    your project (e.g. ``version.py``), then have ``setup.py`` read and ``exec`` the
-    value into a variable.
+Consult your build system documentation for how to implement your preferred method.
 
-    Using ``execfile``:
-
-    ::
-
-        execfile('...sample/version.py')
-        # now we have a `__version__` variable
-        # later on we use: __version__
-
-    Using ``exec``:
-
-    ::
-
-        version = {}
-        with open("...sample/version.py") as fp:
-            exec(fp.read(), version)
-        # later on we use: version['__version__']
-
-    Example using this technique: `warehouse <https://github.com/pypa/warehouse/blob/master/warehouse/__about__.py>`_.
-
-#.  Place the value in a simple ``VERSION`` text file and have both ``setup.py``
-    and the project code read it.
-
-    ::
-
-        with open(os.path.join(mypackage_root_dir, 'VERSION')) as version_file:
-            version = version_file.read().strip()
-
-    An advantage with this technique is that it's not specific to Python.  Any
-    tool can read the version.
-
-    .. warning::
-
-        With this approach you must make sure that the ``VERSION`` file is included in
-        all your source and binary distributions.
-
-#.  Set the value in ``setup.py``, and have the project code use the
-    ``pkg_resources`` API.
-
-    ::
-
-        import pkg_resources
-        assert pkg_resources.get_distribution('pip').version == '1.2.0'
-
-    Be aware that the ``pkg_resources`` API only knows about what's in the
-    installation metadata, which is not necessarily the code that's currently
-    imported.
+Put links in to build system docs?
+-- I have no idea which are currently robust and maintained -- do we want to get into seeming endorsing particular tools in this doc?
 
 
-#.  Set the value to ``__version__`` in ``sample/__init__.py`` and import
-    ``sample`` in ``setup.py``.
+* setuptools:
 
-    ::
+* hatch:
 
-        import sample
-        setup(
-            ...
-            version=sample.__version__
-            ...
-        )
+* poetry:
 
-    Although this technique is common, beware that it will fail if
-    ``sample/__init__.py`` imports packages from ``install_requires``
-    dependencies, which will very likely not be installed yet when ``setup.py``
-    is run.
+* PyBuilder:
+
+* Others?
+
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -12,9 +12,9 @@ One of the challenges in building packages is that the version string can be req
 * It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
    - That will assure that it is properly assigned in the distribution file name, and in the installed package metadata.
 
-* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.
 
-In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
+In any case, The installed distribution version should be accessible using ``importlib.metadata.version("distribution_name")``.
 
 To ensure that version numbers do not get out of sync, it is recommended that there is a single source of truth for the version number.
 
@@ -29,10 +29,12 @@ In general, the options are:
 
 Consult your build system's documentation for their recommended method.
 
-.. _how_to_set_version_links:
+.. _Build system version handling:
 
 Build System Version Handling
 -----------------------------
+
+The following are links to some build system's documentation for handling version strings.
 
 * `Flit <https://flit.pypa.io/en/stable/>`_
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -28,7 +28,7 @@ In general, the options are:
 3) The version string can be hard-coded into the source code -- either in a special purpose file, such as ``_version.txt``, or as a attribute in the ``__init__.py``, and the build system can extract it at build time.
 
 
-Consult your build system documentation for how to implement your preferred method.
+Consult your build system's documentation for their recommended method.
 
 .. _how_to_set_version_links:
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -12,8 +12,7 @@ One of the challenges in building packages is that the version string can be req
 * It needs to be specified when building the package (e.g. in :file:``pyproject.toml``)
    - That will assure that it is properly assigned in the distribution file name, and in the installed package.
 
-* A package may, if the maintainers think it useful, to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level ``__version__`` attribute.
-The value of ``__version__`` attribute and that passed to the build system can (and should) be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the installed package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
 In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -10,12 +10,13 @@ Single-sourcing the Project Version
 One of the challenges in building packages is that the version string can be required in multiple places.
 
 * It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
-   - That will assure that it is properly assigned in the distribution file name, and in teh installed package.
+   - That will assure that it is properly assigned in the distribution file name, and in the installed package.
 
 * Some projects require that there be a version string available as an attribute in the importable module, e.g::
 
     import a_package
     print(a_package.__version__)
+
 * In the metadata of the artifacts for each of the packaging ecosystems    
 
 While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
@@ -36,17 +37,15 @@ If the version string is not in the source, it can be extracted at runtime with 
 
 Consult your build system documentation for how to implement your preferred method.
 
-Put links in to build system docs?
--- I have no idea which are currently robust and maintained -- do we want to get into seeming endorsing particular tools in this doc?
+Here are the common ones:
 
+* `Hatch <https://hatch.pypa.io/1.9/version/>`_
 
-* setuptools:
+* `Setuptools <https://setuptools.pypa.io/en/latest/userguide/distribution.html#specifying-your-project-s-version>`_
 
-* hatch:
+  -  `setuptools_scm <https://setuptools-scm.readthedocs.io/en/latest/>`_
 
-* poetry:
+* `Flit <https://flit.pypa.io/en/stable/>`_
 
-* PyBuilder:
-
-* Others?
+* `PDM <https://pdm-project.org/en/latest/reference/pep621/#__tabbed_1_2>`_
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -15,7 +15,7 @@ One of the challenges in building packages is that the version string can be req
 * A package may, if the maintainers think it useful, to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level ``__version__`` attribute.
 The value of ``__version__`` attribute and that passed to the build system can (and should) be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
-Should a package not set a top level ``__version__`` attribute, the version can still be accessed using ``importlib.metadata.version("distribution_name")``.
+In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
 
 While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
 
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -9,7 +9,7 @@ Single-sourcing the Project Version
 
 One of the challenges in building packages is that the version string can be required in multiple places.
 
-* It needs to be specified when building the package (e.g. in pyproject.toml)
+* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
    - That will assure that it is properly assigned in the distribution file name, and in teh installed package.
 
 * Some projects require that there be a version string available as an attribute in the importable module, e.g::
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -12,7 +12,7 @@ One of the challenges in building packages is that the version string can be req
 * It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
    - That will assure that it is properly assigned in the distribution file name, and in the installed package.
 
-* Some projects require that there be a version string available as an attribute in the importable module, e.g::
+* Some projects prefer that there be a version string available as an attribute in the importable module, e.g::
 
     import a_package
     print(a_package.__version__)
--- a/source/single_source_version.rst
+++ b/source/single_source_version.rst
@@ -5,7 +5,7 @@ Single-sourcing the Project Version
 ===================================
 
 :Page Status: Complete
-:Last Reviewed: 2015-09-08
+:Last Reviewed: 2024-??
 
 One of the challenges in building packages is that the version string can be required in multiple places.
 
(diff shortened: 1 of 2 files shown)


source/single_source_version.rst
```diff
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
...
+* Some projects require that there be a version string available as an attribute in the importable module, e.g::
 
-        def read(*names, **kwargs):
-            with io.open(
-                os.path.join(os.path.dirname(__file__), *names),
-                encoding=kwargs.get("encoding", "utf8")
-            ) as fp:
-                return fp.read()
+    import a_package
+    print(a_package.__version__)
```
tacaswell, line 18:
```suggestion
    print(a_package.__version__)
* In the metadata of the artifacts for each of the packaging ecosystems    
```

source/single_source_version.rst
```diff
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
 :Page Status: Complete
 :Last Reviewed: 2015-09-08
 
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in pyproject.toml)
```
webknjaz, line 12:
```suggestion
* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
```

webknjaz: Also… Perhaps, not “specified” but “sourced”? There are cases when it's generated dynamically, based on the Git repo state, for example. In which case, it's not put into the metadata config file directly. I feel like “specified” here implies “hardcoded” which would be misleading. WDYT?

source/single_source_version.rst
```diff
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
...
-    offers an API that both locations can use.
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as `_version.txt`, or as a attribute in the `__init__.py`, and the build system can extract it at build time.
 
-    Few tools you could use, in no particular order, and not necessarily complete:
-    `bumpversion <https://pypi.python.org/pypi/bumpversion>`_,
-    `changes <https://pypi.python.org/pypi/changes>`_, `zest.releaser <https://pypi.python.org/pypi/zest.releaser>`_.
+If the version string is not in the source, it can be extracted at runtime with code in `__init__.py`, such as::
 
+    import importlib.metadata
+    __version__ = importlib.metadata.version('the_distribution_name')
```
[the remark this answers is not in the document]
ChrisBarker-NOAA: Actually, I think that whether having an attribute is a good idea or not is debatable, it's not debatable that **if** you have an attribute, everyone should use the same name -- and `__version__` is by far the most common convention.

So yes, it *does* have some importance and [should be] be followed :-)

If you really don't want to make that clear, then we can simply say that you can access the distribution version at run time with `importlib.metadata.version()`

source/single_source_version.rst
ChrisBarker-NOAA: I'm accepting this, but I'm not sure how clear to newbies "In the metadata of the artifacts" is. 

Anyone have a idea for a better phrasing?

source/single_source_version.rst
```diff
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
...
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in pyproject.toml)
+   - That will assure that it is properly assigned in the distribution file name, and in teh installed package.
 
-#.  Read the file in ``setup.py`` and parse the version with a regex. Example (
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* Some projects require that there be a version string available as an attribute in the importable module, e.g::
```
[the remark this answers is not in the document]
ChrisBarker-NOAA: And we're back to the debate.

There is no consensus on whether `__version__` is a good idea or not.

I think this guide should be documenting the state of the practice -- not documenting what some people think are best practices.

Should a best practice be determined? I suppose so, but this PR is not the place to have that discussion.

"Why is this so far up?" Because it is a common practice. 

> . Ideally __version__ is mentioned 0 times in the main part of this guide 

That is not the ideal we all agree on -- that's EXACTLY why the previous PRs were derailed.

source/single_source_version.rst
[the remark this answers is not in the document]
bwoodsend: This is implying that `__version__` should be set by `importlib.metadata.version()` which is something that seemingly almost everyone opposes (it adds to startup time whilst solving none of the issues that the pro-`__version__` people have with using metadata). I think the two concepts need to be separated out:
* A package *may*, if the maintainers think it useful to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level `__version__` attribute. The value of `__version__` and that passed to the build system can (and should) be kept in sync in ``:ref:`the build systems's recommended way <link to per build system breakout section>` ``.
* Should a package not set a top level `__version__` attribute, the version can still be accessed using `importlib.metadata.version("distribution_name")`. Since loading `importlib.metadata` has a non negligible startup cost, this lookup should ideally be deferred until when it's required (e.g. a `--version` flag is provided or an *About* dialog is opened)

source/single_source_version.rst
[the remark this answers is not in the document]
bwoodsend: How about just replacing `require` with `prefer`? That hopefully keeps both sides of the fence happy.

source/single_source_version.rst
```diff
@@ -7,111 +7,45 @@ Single-sourcing the Project Version
...
-#.  Set the value to a ``__version__`` global variable in a dedicated module in
-    your project (e.g. ``version.py``), then have ``setup.py`` read and ``exec`` the
-    value into a variable.
 
-    Using ``execfile``:
+Consult your build system documentation for how to implement your preferred method.
 
-    ::
+Put links in to build system docs?
+-- I have no idea which are currently robust and maintained -- do we want to get into seeming endorsing particular tools in this doc?
```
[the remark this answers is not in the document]
ChrisBarker-NOAA: great, thanks! -- I'll add those

ChrisBarker-NOAA:
By the way, when putting in links to the build system's docs, I noticed that they all support `__version__`, and at least flit does it by default.

So it is absolutely a common practice.

source/single_source_version.rst
tacaswell: I was trying to refer to `.deb`, `.rpm`, ... for each of the downstream packaging ecosystems, not just the variety of Python side build tools.

source/single_source_version.rst
```diff
@@ -7,111 +7,41 @@ Single-sourcing the Project Version
...
+* A package may, if the maintainers think it useful, to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level ``__version__`` attribute.
+The value of ``__version__`` attribute and that passed to the build system can (and should) be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
-        def read(*names, **kwargs):
-            with io.open(
-                os.path.join(os.path.dirname(__file__), *names),
-                encoding=kwargs.get("encoding", "utf8")
-            ) as fp:
-                return fp.read()
+Should a package not set a top level ``__version__`` attribute, the version can still be accessed using ``importlib.metadata.version("distribution_name")``.
```
tacaswell, line 18:
```suggestion
In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,41 @@ Single-sourcing the Project Version
...
 
-        def find_version(*file_paths):
-            version_file = read(*file_paths)
-            version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]",
-                                      version_file, re.M)
-            if version_match:
-                return version_match.group(1)
-            raise RuntimeError("Unable to find version string.")
+While different projects have different needs, it's important to make sure that there is a single source of truth for the version number.
```
tacaswell, line 20:
```suggestion
To ensure that version numbers do not get out of sync, it is recommended that there is a single source of truth for the version number.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,41 @@ Single-sourcing the Project Version
...
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in :file:``pyproject.toml``)
+   - That will assure that it is properly assigned in the distribution file name, and in the installed package.
 
-#.  Read the file in ``setup.py`` and parse the version with a regex. Example (
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* A package may, if the maintainers think it useful, to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level ``__version__`` attribute.
```
tacaswell, line 15:
```suggestion
* A package may set a top level ``__version__`` attribute to provide trivial runtime access to the version without needing to rely on any package manager's metadata.
```

source/single_source_version.rst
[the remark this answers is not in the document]
flying-sheep: I’m not trying to be complicated, I just really don’t know what is meant by “projects” here.
- individual packages that opted to use `__version__`? Then you could say something like “you might encounter a `__version__` variable set in `__init__.py` of a top level package, which is a convention that existed before standard methods to access package metadata existed”
- projects that access that variable automatically? I only know of one, and I wouldn’t recommend using it because of various issues that this is not the place for expanding.

source/single_source_version.rst
```diff
@@ -7,111 +7,41 @@ Single-sourcing the Project Version
...
 
-#.  Use an external build tool that either manages updating both locations, or
-    offers an API that both locations can use.
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as ``_version.txt``, or as a attribute in the ``__init__.py``, and the build system can extract it at build time.
 
-    Few tools you could use, in no particular order, and not necessarily complete:
-    `bumpversion <https://pypi.python.org/pypi/bumpversion>`_,
-    `changes <https://pypi.python.org/pypi/changes>`_, `zest.releaser <https://pypi.python.org/pypi/zest.releaser>`_.
 
+Consult your build system documentation for how to implement your preferred method.
```
bwoodsend, line 31: Can we loose the word *implement* – it's too close to *roll your own DIY method*?
```suggestion
Consult your build system's documentation for their recommended method.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,41 @@ Single-sourcing the Project Version
...
 
-    .. note::
+1) If the code is in a version control system (VCS), e.g. git, then the version can be extracted from the VCS.
 
-        This technique has the disadvantage of having to deal with complexities of regular expressions.
+2) The version can be hard-coded into the `pyproject.toml` file -- and the build system can copy it into other locations it may be required.
 
-#.  Use an external build tool that either manages updating both locations, or
-    offers an API that both locations can use.
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as ``_version.txt``, or as a attribute in the ``__init__.py``, and the build system can extract it at build time.
```
bwoodsend, line 28: What exactly is this ``_version.txt``? It sounds uncomfortably like the approach where both the `setup.py` and the `__init__.py` both read from a shared text file and at least one release ends up defunct because the text file didn't get packaged properly. I'd be more comfortable if it wasn't mentioned.

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
 :Page Status: Complete
 :Last Reviewed: 2015-09-08
 
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in :file:``pyproject.toml``)
```
webknjaz, line 12:
```suggestion
* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
```

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
-    Few tools you could use, in no particular order, and not necessarily complete:
-    `bumpversion <https://pypi.python.org/pypi/bumpversion>`_,
-    `changes <https://pypi.python.org/pypi/changes>`_, `zest.releaser <https://pypi.python.org/pypi/zest.releaser>`_.
 
+Consult your build system's documentation for their recommended method.
 
-#.  Set the value to a ``__version__`` global variable in a dedicated module in
-    your project (e.g. ``version.py``), then have ``setup.py`` read and ``exec`` the
-    value into a variable.
+.. _how_to_set_version_links:
```
webknjaz, line 32: It looks like other references use spaces. Perhaps, follow the same style?

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
-           version=find_version("package", "__init__.py")
-           ...
-        )
+In general, the options are:
 
-    .. note::
+1) If the code is in a version control system (VCS), e.g. git, then the version can be extracted from the VCS.
 
-        This technique has the disadvantage of having to deal with complexities of regular expressions.
+2) The version can be hard-coded into the `pyproject.toml` file -- and the build system can copy it into other locations it may be required.
```
webknjaz, line 25:
```suggestion
2) The version can be hard-coded into the :file:`pyproject.toml` file -- and the build system can copy it into other locations it may be required.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
 
-#.  Use an external build tool that either manages updating both locations, or
-    offers an API that both locations can use.
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as ``_version.txt``, or as a attribute in the ``__init__.py``, and the build system can extract it at build time.
```
webknjaz, line 27:
```suggestion
3) The version string can be hard-coded into the source code -- either in a special purpose file, such as :file:`_version.txt`, or as a attribute in the :file:`__init__.py`, and the build system can extract it at build time.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
 :Page Status: Complete
 :Last Reviewed: 2015-09-08
```
webknjaz, line 8: Update this before merging?

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
 
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in :file:``pyproject.toml``)
+   - That will assure that it is properly assigned in the distribution file name, and in the installed package.
```
webknjaz, line 13:
```suggestion
   - That will assure that it is properly assigned in the distribution file name, and in the installed package metadata.
```

source/single_source_version.rst
tacaswell: Every version of this is going to have failure modes.  One of the reasons I'm so in favor of the "git-tag-is-truth" and letting the build system sort it out is that in the past I have done broken releases using the "manually update hard code static strings" approach you are advocating @bwoodsend .

If the existence of a broken release eliminates a method, we would be unable to make any more releases 🤣 .

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
 
-#.  Read the file in ``setup.py`` and parse the version with a regex. Example (
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the installed package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
```
webknjaz, line 15: I feel like this introduced a lot of controversy in the past. The way it's written here, it reads as though the “`__version__`” name specifically must be used and nothing else. The reality is that the variable name can be arbitrary. And it doesn't have to be top-level to work, either. Additionally, I see that here and in the previous point, the word “package” is used, but it means different things. Here, you talk about an importable, while there it's about a distribution. Can we do anything about this?

Could you try rephrasing this to say that an in-project importable module may have a hard-coded variable that can be wired into packaging metadata by name? We can mention that many projects use `__version__` for the variable name, but there is no standard mandating it to be that.

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
 
-        def read(*names, **kwargs):
-            with io.open(
-                os.path.join(os.path.dirname(__file__), *names),
-                encoding=kwargs.get("encoding", "utf8")
-            ) as fp:
-                return fp.read()
+In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
```
webknjaz, line 17: Additionally, we should maybe mention that `__version__` may be populated from `importlib.metadata.version(__name__)` to provide the version string from installed packaging metadata as an importable.

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
 
-        setup(
-           ...
-           version=find_version("package", "__init__.py")
-           ...
-        )
+In general, the options are:
 
-    .. note::
+1) If the code is in a version control system (VCS), e.g. git, then the version can be extracted from the VCS.
```
webknjaz, line 23:
```suggestion
1) If the code is in a version control system (VCS), e.g. Git, then the version can be extracted from the VCS.
```

source/single_source_version.rst
```diff
@@ -7,111 +7,40 @@ Single-sourcing the Project Version
...
-    your project (e.g. ``version.py``), then have ``setup.py`` read and ``exec`` the
-    value into a variable.
+.. _how_to_set_version_links:
 
-    Using ``execfile``:
+Build System Version Handling
+----------------------------
 
-    ::
+* `Hatch <https://hatch.pypa.io/1.9/version/>`_
```
webknjaz, line 37: Let's keep the order alphabetical here, as it is done in other parts of PyPUG. Otherwise, there's going to be debate regarding how to sort these. If we'd want to sort by adoption, we'd put `setuptools` first, of course.

webknjaz: Additionally, Hatch is a build front-end. Their related build backend is called Hatchling.

source/single_source_version.rst
webknjaz: *(ref https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709137303)*

webknjaz:
@ChrisBarker-NOAA could you try working my diagram from https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176 into this?

P.S. Looks like there's a merge conflict to be solved.

source/single_source_version.rst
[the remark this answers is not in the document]
webknjaz: > This is implying that `__version__` should be set by `importlib.metadata.version()` which is something that seemingly almost everyone opposes (it adds to startup time whilst solving none of the issues that the pro-`__version__` people have with using metadata). I think the two concepts need to be separated out:
>

I would say that `__version__` should be set from `importlib.metadata.version()` as opposed to any other methods (it just can be called something else if somebody wants that).

>     * A package _may_, if the maintainers think it useful to be able to determine the version of the package from Python code without relying on a package manager's metadata, set a top level `__version__` attribute. The value of `__version__` and that passed to the build system can (and should) be kept in sync in `` :ref:`the build systems's recommended way <link to per build system breakout section>`  ``.

Agreed.

>     * Should a package not set a top level `__version__` attribute, the version can still be accessed using `importlib.metadata.version("distribution_name")`. Since loading `importlib.metadata` has a non negligible startup cost, this lookup should ideally be deferred until when it's required (e.g. a `--version` flag is provided or an _About_ dialog is opened)

The performance bit is a good point that I didn't consider. In that case, I'd recommend perhaps wrapping that into a module-level `__getattr__()` and computing it dynamically on first access (plus caching). This would still expose the same API/interface while pushing the lookup cost to the moment when it's needed.

source/single_source_version.rst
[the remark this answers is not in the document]
webknjaz: > individual packages that opted to use `__version__`? Then you could say something like “you might encounter a `__version__` variable set in `__init__.py` of a top level package, which is a convention that existed before standard methods to access package metadata existed”

I'd say “which is a convention some people follow that is not backed by standards, so feel free to opt for using any other name, depending on your use case”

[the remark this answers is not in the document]
webknjaz: > What I’m saying here is that it should be documented as the convention it is: I’m pretty sure the right way to structure a guide is to first talk about standards and what to do achieve the challenge at hand (single sourcing the version) in conjunction with them, _then_ go for how to integrate conventions with the solution that has just been outlined.

I like this explanation.

source/single_source_version.rst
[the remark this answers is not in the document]
webknjaz: @flying-sheep that's a tutorial. In guides, we specifically avoided any bias-based sorting to avoid the perception of one tool being more blessed than the other: https://packaging.python.org/en/latest/guides/tool-recommendations/#workflow-tools.

source/single_source_version.rst
tacaswell, line 15:
```suggestion
* An importable module (formally packaged for distribution or not) may set a top level attribute to provide runtime access to the version of the installed package.  The defacto standard name for this attribute, dating back to at least 2011 and with wide community usage, is ``__version__``.   If the runtime version string is set, the distribution version should be kept in sync with the runtime version using :ref:`the build systems's recommended way <how_to_set_version_links>`.
```

tacaswell: I suggested an edit in response to your comment.

source/single_source_version.rst
tacaswell: As near as I can tell, the push to get rid of `__version__` came from people protesting to this pattern (as a waste of runtime on import).

webknjaz: Yeah, I'm starting to understand this better now, having read another thread: https://github.com/pypa/packaging.python.org/pull/1578#discussion_r1697056367

source/single_source_version.rst
webknjaz: @tacaswell perhaps, drop the “top-level” bit? Any level of nesting would work for as long as it's importable (which usually means not triggering imports of the deps…)

Also, it seems like “de facto” is spelled separately: https://en.wikipedia.org/wiki/De_facto.

source/single_source_version.rst
flying-sheep: It also comes from the fact that `__version__` is just a conventional place to duplicate the information in that already has a canonical place: not very pythonic

source/single_source_version.rst
tacaswell: Edited in-place to drop the top-level and fixed the typos.

tacaswell: I also updated the year to be the first commit in Matplotlib that used `__version__` not the date of pep 396.

source/single_source_version.rst
tacaswell: > It also comes from the fact that __version__ is just a conventional place to duplicate the information in that already has a canonical place: not very pythonic

I respectfully, but completely disagree with this characterization.  `__version__` is the (defacto) standard which predates the existence of pip by at least 7 years (in Matplotlib I have a commit from 2004 using `__version__`) let alone the current push to bootstrap another (Python -specific) packaging ecosystem.

From my perspective the "duplicate" here is `dist-info` and as I noted on [discourse](https://discuss.python.org/t/please-make-package-version-go-away/58501/38) I am not worried about this.  Package managers (apt/dpkg, yum, pacman, conda, pip, ....) all need their own copy of the version metadata (and dependencies) to do their jobs.  This is OK and expected as each of these systems, and runtime, have different needs and hence (may) need their own copy of the information. 

It comes across as very hostile to invent a new place to put the version information and then declare the _existing_ location an upythonic duplicate!

As a thought experiment, if I said "We can scrap all of `dist-info`, we can query pacman to get all of it when we need it" I don't think it would be taken seriously.  "just use the metadata for pip" reads exactly the same to me.

webknjaz: @flying-sheep I think that it's a bit more complicated than that. Duplicating the hardcoded values is bad. But if only one place is the actual source of the information and another just reflects the same information though a different API, that should be okay.

source/single_source_version.rst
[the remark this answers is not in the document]
webknjaz: Because there's no standard or a convention for disallowing this. I think, it's okay to highlight the problems and controversy. However, making imperative recommendations is reaching too far.

source/single_source_version.rst
webknjaz: @tacaswell I additionally edited your suggestion to remove the word “standard” as I believe it may be taken the wrong way.

source/single_source_version.rst
webknjaz: > "just use the metadata for pip" reads exactly the same to me.

@tacaswell TBF, from the Python packaging perspective, that is not “metadata for pip” by a standardized machine-readable information. That said, *if* a project is meant to be installed and the installed dist metadata contains a version, it's reasonable to threat it as the source for the most accurate information (even if it originated from some importables that could be accessed separately). From that perspective, it might be the source of truth in many contexts.

But it's rather tricky to explain/document this in an accessible manner. Hence, my suggestion to attempt visualizing these relationships with my diagram https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176 (or something close to it).

source/single_source_version.rst
tacaswell: Thanks @webknjaz .

source/single_source_version.rst
flying-sheep: > if only one place is the actual source of the information and another just reflects the same information though a different API, that should be okay.

there should be one – and preferably only one – way to do it

the preferable way is the standard way

> I respectfully, but completely disagree with this characterization.

` __version__` is older, but it’s not standardized in a PEP, so it’s not the standard.

also putting that info into an **import package** doesn’t make sense, as the versioned unit is the **distribution package** – i.e. the thing that has the `dist-info` directory in it that contains the standard metadata. *most* distribution packages might come with one main import package (e.g. `Pillow` contains the import package `PIL`), but that’s not a rule, so the only place that will reliably match 1:1 with versioned units is the distribution package, complete with its metadata.

ChrisBarker-NOAA:
OK all -- I *think* I've addressed everyone's comments.

Also -- please see my updated initial comment -- trying to keep the scope of this very narrow.

source/single_source_version.rst
```diff
@@ -5,113 +5,44 @@ Single-sourcing the Project Version
...
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the installed package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
 
-        def read(*names, **kwargs):
-            with io.open(
-                os.path.join(os.path.dirname(__file__), *names),
-                encoding=kwargs.get("encoding", "utf8")
-            ) as fp:
-                return fp.read()
+In the cases where a package does not set a top level ``__version__`` attribute, the version may still be accessible using ``importlib.metadata.version("distribution_name")``.
```
flying-sheep, line 17: This is not true

*Every* distribution package will *always* have standard metadata, so no “may” here

How about you delete that here …

```suggestion
```

source/single_source_version.rst
```diff
@@ -5,113 +5,44 @@ Single-sourcing the Project Version
...
 :Page Status: Complete
-:Last Reviewed: 2015-09-08
+:Last Reviewed: 2024-??
 
+One of the challenges in building packages is that the version string can be required in multiple places.
 
-There are a few techniques to store the version in your project code without duplicating the value stored in
-``setup.py``:
+* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
+   - That will assure that it is properly assigned in the distribution file name, and in the installed package metadata.
```
flying-sheep, line 14: … then you add it back in the place where you mention standard metadata first:

```suggestion
   - the version is accessible at runtime via ``importlib.metadata.version("distribution_name")``.

```

source/single_source_version.rst
```diff
@@ -5,113 +5,44 @@ Single-sourcing the Project Version
...
 
-#.  Read the file in ``setup.py`` and parse the version with a regex. Example (
-    from `pip setup.py <https://github.com/pypa/pip/blob/1.5.6/setup.py#L33>`_)::
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the installed package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
```
flying-sheep, line 15: Finally you need to improve the wording of “to provide runtime access”, because the line above mentioning `importlib.metadata.version` already describes how to get runtime access.

flying-sheep (changes requested):

source/single_source_version.rst
tacaswell: As I keep saying, not everything is a "distribution package" and there is no guarantee that the thing you imported (which is what matters) is the thing that the packaging system thinks is installed.

tacaswell, line 15:
```suggestion
* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <how_to_set_version_links>`.
```

change "installed" (which is a packaging time concept) to "imported" (which is a runtime concept).

source/single_source_version.rst
webknjaz: So `s/may still/is always/` then.

bwoodsend: I meant *may* to mean *can* here as opposed to *sometimes can* when I originally suggested that sentence. Just replacing *may* with either *can* or *is* sounds like the way to avoid that potential interpretation?

flying-sheep: neither, the whole phrasing implies that `__version__` is the primary way to do it, and `importlib.metadata.version()` is a kind of fallback, which it isn’t. So more like

```diff
-In the cases where a package does not set a top level ``__version__`` attribute,
 the version
-may still be accessible using
+is accessible at runtime via
 ``importlib.metadata.version("distribution_name")``.
```

source/single_source_version.rst
flying-sheep: Yes, everything this guide is about is contained in a distribution package. .deb/.rpm, … packages contain them. conda packages contain them.

There are some counter examples like fintech, but `__version__` doesn’t apply to these either.

source/single_source_version.rst
flying-sheep: @tacaswell why the :-1:? what’s wrong?

source/single_source_version.rst
tacaswell: If I delete the `dist-info` folder does anything break at runtime?

source/single_source_version.rst
tacaswell: I do not think either edit should be made.

source/single_source_version.rst
abravalheri: @tacaswell, if you manually delete the `dist-info` folder lots of stuff will break at runtime (e.g. entry-points). I believe that only package managers are meant to modify such files.

For every distribution installed using the default procedure a corresponding `dist-info` is supposed to be present (otherwise the system is in an inconsistent state).

source/single_source_version.rst
abravalheri: > As I keep saying, not everything is a "distribution package" and there is no guarantee that the thing you imported (which is what matters) is the thing that the packaging system thinks is installed.


I think that for the sake of simplicity developers should seek to align versions of all import packages with the version of the corresponding distribution package and differ only if there is a good reason for doing so. Having `pip install distname==2.0` resulting in `__import__("distname").__version__ == "1.0"` would be too chaotic otherwise. Also in the case the same distribution contains multiple import packages (which is already considered an edge case), it would be good practice to coincide them unless there is a good reason not to.

ChrisBarker-NOAA:
> @ChrisBarker-NOAA could you try working my diagram from [#1276 (comment)](https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176) into this?

Hmm -- I'm trying to keep this page as simple as possible and not get into too many tech details.

But if the Doc Editors (who are they?) think it should be included, then sure.

ChrisBarker-NOAA:
By the way, I have not idea what to do with this:

The required .readthedocs.yaml configuration file was not found at repository's root. Learn how to use this file in our [configuration file tutorial](https://docs.readthedocs.io/en/stable/config-file/index.html).

But I suppose that explains the lack of a preview.

webknjaz:
> By the way, I have not idea what to do with this:
> 
> The required .readthedocs.yaml configuration file was not found at repository's root. Learn how to use this file in our [configuration file tutorial](https://docs.readthedocs.io/en/stable/config-file/index.html).
> 
> But I suppose that explains the lack of a preview.

I believe RTD would build against the merge commit of your PR branch into main. Since there's merge conflicts, there's no merge commit, hence no place to check the config from.

Action item: address the merge conflicts (I personally like rebasing) and it should work then.

ChrisBarker-NOAA:
Turns out I had not synced my fork before doing this, and it was a serious mess, due to the restructuring of the docs. So I wan't able to fix it in place.

But the discussion had gotten pretty convoluted anyway, so a new PR has its benifits:

#1580

PR #1580, opened 2024-08-01
Simplify single source

ChrisBarker-NOAA:
I'm no longer calling this a WIP -- I hope it's close to as good as it's going to get.

This is a follow-up to #1578 -- I made a mess of it when I tried to rebase.

The goal here is to have a simple, and not opinionated, page that guides folks to their build tool of choice's way to handle version strings.

Not sure if it's a guide or a discussion -- I put in discussion, but maybe it's more of a guide.

Now that I think about it -- if it's in discussions, then maybe it *could* be a lot longer, and go into why one might want to (or not want to) provide a version attribute in the importable package.

But I don't know that I want to moderate THAT discussion…

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1580.org.readthedocs.build/en/1580/discussions/single-source-version/


source/discussions/single_source_version.rst
```diff
@@ -0,0 +1,47 @@
...
+:Last Reviewed: 2024-??
+
+One of the challenges in building packages is that the version string can be required in multiple places.
+
+* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
+   - That will assure that it is properly assigned in the distribution file name, and in the installed package metadata.
+
+* A package may set a top level ``__version__`` attribute to provide runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.
+
+In any case, The installed distribution version should be accessible using ``importlib.metadata.version("distribution_name")``.
```
flying-sheep, line 17: The text below mentions three locations where the single source of truth of the version can be, but only two of these locations appear in this list.

- I added the third
- I established the connection between `importlib.metadata` and the package metadata for clarity

```suggestion
* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
   This will make it available installed package’s metadata, from where it will be accessible at runtime using ``importlib.metadata.version("distribution_name")``.

* A package may set a top level ``__version__`` attribute to provide an alternative means of runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.

* If the code is in in a version control system (VCS), e.g. Git, the version may appear in a *tag* such as ``v1.2.3``.
```

flying-sheep (changes requested):
I like the approach you went with with “goals” and “non-goals” in the predecessor PR. So let’s name a clear **goal** here:

- We’re talking about [distribution packages](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package) here:

	This repository is only about distribution packages, as stated in the readme, so let’s be clear that single-sourcing versions while packaging a distribution package is the goal for this PR.
	
	> http://packaging.python.org/
	> 
	> The "Python Packaging User Guide" (PyPUG) aims to be the authoritative resource on how to package and install _distributions_ in Python using current tools.

	Therefore [import packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package) that aren’t part of a distribution packages are a **non-goal**.

source/discussions/single_source_version.rst
ChrisBarker-NOAA: merging, but a thought:

> "This will make it available in the installed package’s metadata,.."

" ... installed distribution’s..."  ?

or

"... installed distribution package’s ..." ? 

the overloading of the term "package" is a challenge ...

ChrisBarker-NOAA: " but only two of these locations appear in this list."

The three potential sources of "truth" are for when the distribution is being built, not at run time -- there is no VCS for an installed package.

Not sure how to make this more clear, but the first point is:

When building the distribution:

 - The version must be specified in pyproject.toml (perhaps dynamically, but it needs to be there)
 - A ``__version__`` attribute can optionally be put (or pre-exist) in the importable code.

The second list is the how to make sure that that's all in sync.

Perhaps we could put your more specific language about tags in the first bullet in the next list.

I do like most of the rest of your language - thanks!

flying-sheep: > the overloading of the term "package" is a challenge ...

It sure is, the core devs should really just have used “packages”/“modules” to distinguish between distribution packages and import packages. Instead they wasted the distinction on single-file “modules” vs folder-with-`__init__.py` “packages”.

The problem with saying “distributions” is that people in the linux world use that term for Linux distributions. So to be clear, we should probably link “import packages” and “distribution packages” to the glossary and consistently use these terms. I think you do that via `` :term:`Import Package` ``

>

webknjaz:
> * This repository is only about distribution packages, as stated in the readme, so let’s be clear that single-sourcing versions while packaging a distribution package is the goal for this PR.
>       > [http://packaging.python.org](http://packaging.python.org/)
>       > The "Python Packaging User Guide" (PyPUG) aims to be the authoritative resource on how to package and install _distributions_ in Python using current tools.
>       
>       
>       Therefore [import packages](https://packaging.python.org/en/latest/glossary/#term-Import-Package) that aren’t part of a distribution packages are a **non-goal**.

So that statement is outdated: regardless of what's stated there, we do have mentions of packaging outside of distribution packages on this site. Besides, imports coupled with distributions because they provide them.

webknjaz:
@ChrisBarker-NOAA have you looked into the possibility of embedding a diagram? (https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176)
I think, it's important to demonstrate unidirectional flow of the version sourcing.

source/discussions/single_source_version.rst
webknjaz, line 1: Is this replacing `source/guides/single-sourcing-package-version.rst`?

P.S. I see the existing convention to naming the documents is kebab-case. They are not Python modules, so they follow different principles than snake_case.

ChrisBarker-NOAA: I'm following what's currently there with regard to naming files.

ChrisBarker-NOAA:
OK -- I *think* this is in good shape, and has addressed the concerns of everyone.

Open Question: whether to include a diagram -- see: https://github.com/pypa/packaging.python.org/pull/1580#issuecomment-2286522496

Other than that -- good to go.

Editors -- anything else?

source/discussions/single-source-version.rst
```diff
@@ -0,0 +1,47 @@
...
+
+To ensure that version numbers do not get out of sync, it is recommended that there is a single source of truth for the version number.
+
+In general, the options are:
+
+1) If the code is in a version control system (VCS), e.g. Git, then the version can be extracted from the VCS.
+
+2) The version can be hard-coded into the :file:`pyproject.toml` file -- and the build system can copy it into other locations it may be required.
+
+3) The version string can be hard-coded into the source code -- either in a special purpose file, such as :file:`_version.txt`, or as a attribute in the :file:`__init__.py`, and the build system can extract it at build time.
```
abravalheri, line 27: [Some people do use other modules](https://github.com/search?q=%2Fattr+%3D+%22.*VERSION%22%2F+language%3ATOML&type=code&l=TOML) to store the version (e.g., `version.py`, or in the case of single-module distributions, whichever module name happens to be the main module).

```suggestion
3) The version string can be hard-coded into the source code -- either in a special purpose file, such as :file:`_version.txt`, or as a attribute in a module, such as :file:`__init__.py`, and the build system can extract it at build time.
```

source/discussions/single-source-version.rst
```diff
@@ -0,0 +1,47 @@
...
+
+The following are links to some build system's documentation for handling version strings.
+
+* `Flit <https://flit.pypa.io/en/stable/>`_
+
+* `Hatchling <https://hatch.pypa.io/1.9/version/>`_
+
+* `PDM <https://pdm-project.org/en/latest/reference/pep621/#__tabbed_1_2>`_
+
+* `Setuptools <https://setuptools.pypa.io/en/latest/userguide/distribution.html#specifying-your-project-s-version>`_
```
abravalheri, line 45: Probably the link that comes closer to the intent of the text would be https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

```suggestion
* `Setuptools <https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata>`_
```

source/discussions/single-source-version.rst
```diff
@@ -0,0 +1,47 @@
...
+
+:Page Status: Complete
+:Last Reviewed: 2024-??
+
+One of the challenges in building packages is that the version string can be required in multiple places.
+
+* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
+   This will make it available in the installed package’s metadata, from where it will be accessible at runtime using ``importlib.metadata.version("distribution_name")``.
+
+* A package may set a top level ``__version__`` attribute to provide an alternative means of runtime access to the version of the imported package. If this is done, the value of ``__version__`` attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.
```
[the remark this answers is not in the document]
bwoodsend: Maybe say *(usually called `__version__`)* and leave it at that? Some packages also set a `__version_info__` or have separate long and short versions or ... there are lots of them. Let's just endorse the one that's more or less (albeit unofficially) seen as *standard*.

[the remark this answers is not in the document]
abravalheri: I don't see why endorsing it, because I think this is a matter of API design.
I also don't think it is seen as a *standard*, more like a popular *convention* (`VERSION` is also a very well stablished convention, just not as popular)...

But I am completely OK with @bwoodsend suggestion. Something would also work for me is:

```suggestion
* A package may set a module attribute (e.g., ``__version__``) to provide an alternative means of runtime access to the version of the imported package. If this is done, the value of the attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.
```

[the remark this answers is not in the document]
ChrisBarker-NOAA: > [people do use VERSION](https://github.com/search?q=%2Fattr+%3D+%22.*VERSION%22%2F+language%3ATOML&type=code&l=TOML) and [other things](https://github.com/search?q=%2Fattr+%3D+%22.*%5B%5E_%22%5D%22%2F+language%3ATOML+path%3Apyproject.toml&type=code)... Probably this is best left as a design choice for the API of each individual package (if they want to expose it through their API).

Sure they do -- but while there is no consensus about whether an import module should have an attribute that returns the version string, I hope there is consensus that it's better for the community if everyone uses the same name for it, and `__version__` is the most common convention at this point. In fact, a `__version__` PEP is being revived to hopefully codify it -- but in any case, in all the discussions I've seen on this topic, `__version__` is the ONLY name I've seen advocated for (as a standard). 

In any case, I've accepted @bwoodsend 's text to soften it a bit.

source/discussions/single-source-version.rst
ChrisBarker-NOAA: You're quite right, a single file module doesn't have an `__init__.py` -- thanks!

source/discussions/single-source-version.rst
ChrisBarker-NOAA: well, yes, but that's out of scope for this particular doc ..... I may submit a PR there, if I can find the time -- if anyone else gets to it first -- great! 

Hmm -- this doc is not necessarily endorsing the dynamic method per se -- but sure, that's probably a better link. Accepted.

source/discussions/single-source-version.rst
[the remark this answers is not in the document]
webknjaz: @ChrisBarker-NOAA at first, I was misunderstanding why the opposition existed — I thought it was because of the name. But having read the explanations later in the process, I realized that it was about having import-time logic.

When I brought this up in a code review of another project that relies on `asyncio`, I heard an interesting point — having heavy version computation in `__init__.py` during import-time (as opposed to having `__version__` under `__getattr__()` for lazy access) might be preferable because of the predictability, when it's known to be always used. And having it computed lazily would be blocking to the event loop at a non-deterministic point in time.

Just wanted to mention this...

That said, when I pointed out that setuptools has an option to source the version from importables through dynamic attributes, that resolved the problem even better as with that, there's no sudden computational cost to it. And that version in a Python module could be the original source of truth as the version data access flow is unidirectional — `[importable] -> [installed metadata] -> [reading from metadata]` or `[importable] -> [project-internal code reading from importable]` (which fits well in my diagram: https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176). This might be useful for programs that access and display their own versions without concerning themselves of how they're injected into venv.

So I think that it might be useful to mention the cost of having `__version__` with a static value vs. `version = importlib.metadata.version(__package__)`. These cases would just have different sources of where the version comes from initially. And the external programs would be still recommended to do `importlib.metadata.version(__package__)` in general case.

Perhaps, we shouldn't present the idea of having `__version__` as an “approved/recommended public API:tm:” for external programs but reframe that as “*sometimes* this attribute exists, *sometimes* it's called `__version__`, *some* projects may use it internally and for packaging metadata”

source/discussions/single-source-version.rst
webknjaz: @ChrisBarker-NOAA is it possible to use intersphinx instead of a hardcoded link? I'm pretty sure that setuptools' docs are plugged into Sphinx already.

webknjaz:
> Open Question: whether to include a diagram -- see: [#1580 (comment)](https://github.com/pypa/packaging.python.org/pull/1580#issuecomment-2286522496)

Thanks for acknowledging that you saw my request. What's your opinion? I don't think I saw you sharing it anywhere — have I missed it?

bwoodsend:
What would the diagram contain? I'm guessing some flow that depicts `pyproject.toml -> build backend -> project/__init__.py:__version__ -> version in package metadata`? Not sure how I'd translate those into picture form short of just writing those words with rectangles round them.

source/discussions/single-source-version.rst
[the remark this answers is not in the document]
ChrisBarker-NOAA: Surprisingly, this is a VERY contentious topic -- I am trying desperately to capture the key point, which is:

have one source of truth for yur version string when building a package.

While avoiding recommendations on what should actually be in the final, installed package.

I think that discussion is good, but should be elsewhere.

NOTE: the current structure of these docs does, indeed, have a "discussions" section -- -so it could be that a nice full discussion of the issue would be accepted -- but let's do that as another PR :-)

ChrisBarker-NOAA:
>> Open Question: whether to include a diagram -- 
>  What's your opinion? 

I don't think it's necessary -- it's just not that complicated.

But I'm happy for the doc editor(s) to decide.

-CHB

source/discussions/single-source-version.rst
ChrisBarker-NOAA: probably -- if you can figure out how, I'll accept the change.

bwoodsend (approved):

source/discussions/single-source-version.rst
abravalheri: > well, yes, but that's out of scope for this particular doc ..... I may submit a PR there, if I can find the time -- if anyone else gets to it first -- great!

Don't worry, that is setuptools stuff. I just added the note to remind myself.

source/discussions/single-source-version.rst
```diff
@@ -0,0 +1,47 @@
+.. _`Single sourcing the version`:
+
+===================================
+Single-sourcing the Project Version
+===================================
+
+:Page Status: Complete
+:Last Reviewed: 2024-??
```
webknjaz, line 8: TODO: enter the actual date pre-merge. *Do not mark this discussion as resolved until right before clicking merge.*

source/discussions/single-source-version.rst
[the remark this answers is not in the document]
webknjaz: Alright, noted. I think I might've subconsciously extended the scope of your effort in my head :)

Marking this thread as resolved for now.

webknjaz:
@bwoodsend I made a diagram in this comment a while back: https://github.com/pypa/packaging.python.org/pull/1276#issuecomment-1709024176. I imagined it'd be reused, maybe with some modifications. It's made with Mermaid.

@ChrisBarker-NOAA okay, I hear you. In the interest of unblocking the effort, I withdraw that request for now.

source/discussions/single-source-version.rst
```diff
@@ -0,0 +1,47 @@
+.. _`Single sourcing the version`:
```
webknjaz, line 1: This label duplicates the one already declared in another document, which legitimately breaks the CI @ https://github.com/pypa/packaging.python.org/actions/runs/10514372639/job/29137925991?pr=1580#step:5:126 with

```console
/home/runner/work/packaging.python.org/packaging.python.org/source/guides/single-sourcing-package-version.rst:5: WARNING: duplicate label single sourcing the version, other instance in /home/runner/work/packaging.python.org/packaging.python.org/source/discussions/single-source-version.rst
```

So it has to be different. I think, we can go for
```suggestion
.. _`Single sourcing the version discussion`:
```

source/discussions/single-source-version.rst
webknjaz:
```suggestion
:Last Reviewed: 2024-08-24
```

webknjaz (approved):

flying-sheep:
Thanks for slogging through all our comments, I think the end result is really nice and clear now!

ChrisBarker-NOAA:
Glad it got through to the end this time! 

Thanks all -- it was a slog, but it's a much better document as a result of all the input.

bwoodsend:
Here's to never seeing `FileNotFoundError: .../site-packages/xyz/version.txt` ever again!

PR #1269, opened 2023-06-19
Conda info

PythonCHB:
Updated info about conda -- it was a bit out of date.

This is the same as #1204, but from a topic branch.

Preview: https://python-packaging-user-guide--1269.org.readthedocs.build/en/1269/key_projects/#conda

PR #1581, opened 2024-08-01
Updated notes about conda

ChrisBarker-NOAA:
This is more or less, an update to: #1269, but adapted to the new structure.

However, conda is now talked about in two places:

Installing scientific packages

and

Project Summaries

perhaps one of them should simply be alink to the. other -- probably Project Summaries can ref to Installing scientific packages.

Though I'm not totally on board with that -- while conda was developed for, and widely used by the Scientific computing community -- it's a more universal tool than that so maybe?

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1581.org.readthedocs.build/en/1581/


ncoghlan:
I think the way you've done it here is a good compromise: enough detail inline for the scientific packages section to make sense, but the larger amount of info in the project description. If a reference were to be added, I'd make it from the scientific section to the project summary.

The build failure picked up a genuine structural issue, though: the Anaconda link is duplicated because one of the new paragraphs should be moved up to replace the old intro paragraph outright (the fact the old intro paragraph references Anaconda 5.0 suggests the previous text is at least 7 years old)

Leaving my approval active (since I agree with the gist of the change), but putting some comments in line with a suggestion to fix the build error.

source/guides/installing-scientific-packages.rst
```diff
@@ -134,5 +134,10 @@ native virtual environments. Conda makes environments first-class citizens,
 making it easy to create independent environments even for C libraries. It is
 written in Python, but is Python-agnostic. Conda manages Python itself as a
 package, so that :command:`conda update python` is possible, in contrast to
-pip, which only manages Python packages. Conda is available in Anaconda and
-Miniconda (an easy-to-install download with just Python and conda).
+pip, which only manages Python packages.
+
+The conda package manager is available in `Anaconda <https://docs.anaconda.com/anaconda/>`_, `miniconda <https://docs.anaconda.com/miniconda/>`_, `miniforge <https://github.com/conda-forge/miniforge>`_, and `pixi <https://pixi.sh/>`_.
```
ncoghlan, line 139: The `Anaconda` link here conflicts with the one in the first paragraph in this section.

Suggested resolution:

1. Delete the old first paragraph entirely (switching this section to introduce `conda` itself first, then explain ways to get it)
2. Edit the `conda` intro paragraph to remove the text "included in Anaconda" (it's unnecessary, and doesn't make sense when `conda` is being introduced in its own right first)

source/key_projects.rst
```diff
@@ -404,26 +404,30 @@ conda
...
-<https://docs.anaconda.com/anaconda/>`__ Python installations.
-Anaconda Python is a distribution from `Anaconda, Inc
-<https://www.anaconda.com/products/individual>`__ specifically aimed at the scientific
-community, and in particular on Windows where the installation of binary
-extensions is often difficult.
+Conda is a package, dependency, and environment management system for any language — Python, R,
+Ruby, C/C++, Fortran, and more. It is written in Python and
+widely used in the Python scientific computing community, due to its support for non-Python
+compiled libraries and extensions. It is used as the basis of the `Anaconda
+<https://docs.anaconda.com/anaconda/>`__ Python distribution from `Anaconda, Inc.specifically
```
ncoghlan, line 411:
```suggestion
<https://docs.anaconda.com/anaconda/>`__ Python distribution from Anaconda, Inc. It is specifically
```

source/key_projects.rst
```diff
@@ -404,26 +404,30 @@ conda
...
+aimed at the scientific community, community, but can also be used on its own, or with the
```
ncoghlan, line 412:
```suggestion
aimed at the scientific community, but can also be used on its own, or with the
```

source/key_projects.rst
```diff
@@ -404,26 +404,30 @@ conda
...
-<Distribution Package>` from PyPI. Also, `conda skeleton
-<https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/build-pkgs-skeleton.html>`__
-is a tool to make Python packages installable by conda by first
-fetching them from PyPI and modifying their metadata.
+many of their combined features, such as package management, virtual environment
+management and deployment of binary extensions and other binary code.
+
+Conda does not install packages from PyPI -- it can only manage packages built specifically
+for conda, which can be made available on a "conda channel", such as those hosted on
+`anaconda.org <https://anaconda.org/>`__ or a local (e.g. intranet) package server.
```
ncoghlan, line 422:
```suggestion
`anaconda.org <https://anaconda.org/>`__, or a local (e.g. intranet) package server.
```

source/key_projects.rst
```diff
@@ -408,7 +408,7 @@ Conda is a package, dependency, and environment management system for any langua
 Ruby, C/C++, Fortran, and more. It is written in Python and
 widely used in the Python scientific computing community, due to its support for non-Python
 compiled libraries and extensions. It is used as the basis of the `Anaconda
-<https://docs.anaconda.com/anaconda/>`__ Python distribution from `Anaconda, Inc.specifically
+<https://docs.anaconda.com/anaconda/>`__ Python distribution from Anaconda, Inc. It waas originally
```
sinoroc, line 411:
```suggestion
<https://docs.anaconda.com/anaconda/>`__ Python distribution from Anaconda, Inc. It was originally
```

source/guides/installing-scientific-packages.rst
ChrisBarker-NOAA: thanks -- I've cleaned that up, and applied your other copy-editing suggestions.

Good to go pending the build ...

ChrisBarker-NOAA:
@sinoroc, @ncoghlan: thanks for the copy-editing!

ChrisBarker-NOAA:
fixed the netlib link .. <cross fingers>

webknjaz:
@ChrisBarker-NOAA looks like the anaconda website link breaks linkcheck. Could you take a look at addressing that?

ChrisBarker-NOAA:
I'm confused -- this works for me right now:

https://anaconda.org/

?!?!

webknjaz:
@ChrisBarker-NOAA I restarted it before tagging you and the error was the same, I believe. This likely means that their website rejects various automated non-browser requests. This might be based on HTTP headers, the user agent string or the fact that it originated from the GitHub Actions platform.
I'll trigger it again but I wouldn't hold your breath. We'll likely need to add it to `linkcheck_ignore` or find a better link.

webknjaz:
yep... restarting did nothing: https://github.com/pypa/packaging.python.org/actions/runs/11584803433/job/32268247267?pr=1581#step:5:254

ChrisBarker-NOAA:
OK -- added it to linkcheck_ignore and it's passing -- thanks!

PR #1582, opened 2024-08-01
If the "build" package does not exist, you will get an error

alanmehio:
if the build package does not exist, there will be an error such "Not recognized .."

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1582.org.readthedocs.build/en/1582/


chrysle:
The lines just before your change guide the reader to install `build`.

alanmehio:
@chrysle  please fix the script . The code view does not render see attached image.
Thanks 
image

PR #1584, opened 2024-08-06
Update referenced RFC specification

AdamJel:
Update the referenced RFC specification:
  - from [obsoleted](https://datatracker.ietf.org/doc/html/rfc822.html) version `822`,
  - through newer [obsoleted](https://datatracker.ietf.org/doc/html/rfc2822) version `2822`,
  - to [current](https://datatracker.ietf.org/doc/html/rfc5322) version `5322`.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1584.org.readthedocs.build/en/1584/


chrysle (dismissed):

abravalheri:
The old RFC might be obsolete, but do we have confirmation if the existing PyPA standards and implementations are completely aligned with the new RFC?

PR #1586, opened 2024-08-12
clarify ``Provides-Dist`` example to match warehouse metadata validation

rjdbcm:
So I ran into the issue over of the weekend of getting 502s from PyPI on upload of one of my repos. After exhausting everything else I found out that using the format for ``Provides-Dist`` in the spec causes uploads to fail because warehouse uses ``packaging.requirements.Requirement`` to validate wheel METADATA. See https://github.com/pypi/warehouse/issues/16452 It should also be noted that auditwheel and twine did not flag any problems with the wheel format and the only point of failure was a HTTP 502 from warehouse. It seemed in my judgement that editing the spec to keep people from running into this is a good idea. I defer to the PyPA team, though, it may be that warehouse needs changed instead as I am not privy to the original intent of the spec writers.

Thanks,
Eden Ross Duff

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1586.org.readthedocs.build/en/1586/

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -725,7 +725,7 @@ This field may be followed by an environment marker after a semicolon.
 Examples::
 
     Provides-Dist: OtherProject
-    Provides-Dist: AnotherProject (3.4)
+    Provides-Dist: AnotherProject==3.4
     Provides-Dist: virtual_package; python_version >= "3.4"
 
 .. _core-metadata-obsoletes-dist:


di (approved):
This looks right to me. [PEP 566](https://peps.python.org/pep-0566/) says:

> Version numbering requirements and the semantics for specifying comparisons between versions are defined in [PEP 440](https://peps.python.org/pep-0440/).

The string `3.4` by itself is not a valid version specifier per PEP 440. Additionally:

> Following [PEP 508](https://peps.python.org/pep-0508/), version specifiers no longer need to be surrounded by parentheses in the fields Requires-Dist, Provides-Dist, Obsoletes-Dist or Requires-External, so e.g. requests >= 2.8.1 is now a valid value. The recommended format is without parentheses, but tools parsing metadata should also be able to handle version specifiers in parentheses.

Which means either of the following should be fine:

* `Provides-Dist: AnotherProject (==3.4)`
* `Provides-Dist: AnotherProject==3.4`

But the latter is preferred.

PR #1596, opened 2024-09-04
Update download-artifact plugin in publish-to-test-pypi.yml to fix vulnerability

markusgrotz:
Versions of actions/download-artifact before 4.1.7 are vulnerable to arbitrary file write when downloading and extracting a specifically crafted artifact that contains path traversal filenames.

Fore more details see: https://github.com/advisories/GHSA-6q32-hq47-5qq3

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1596.org.readthedocs.build/en/1596/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Build a binary wheel and a source tarball
       run: python3 -m build
     - name: Store the distribution packages
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: python-package-distributions
         path: dist/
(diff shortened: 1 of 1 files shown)


sinoroc (approved):
I wonder if it is in the long term manageable to keep those up to date? Are Dependabot and tools like this able to help with that?

sinoroc:
> Yes Dependabot supports bumping these versions https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

Even for the files that are in non-regular locations (i.e. not in `.github/workflows`)?

webknjaz:
> Even for the files that are in non-regular locations (i.e. not in `.github/workflows`)?

Not sure. However, I wouldn't want hashes or patch-level version bumps in the guide, which is why I kept the major-version refs in it.

They also need to be manually validated, since for example the upload+download actions of v4 have breaking changes. They probably don't affect the example workflow but would affect people with more complicated scenarios, having matrices of C-extensions.

webknjaz (approved):
Thanks!

PR #1607, opened 2024-10-06
Add notes on runtime version access

ncoghlan:
Also redirects the obsolete single-source version guide to the updated single-source version discussion.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1607.org.readthedocs.build/en/1607/


ncoghlan:
Seeing if we can make HTTP redirects work without an extension like https://pypi.org/project/sphinx-reredirects/

ncoghlan:
Found a surprising CPython docs issue here: https://github.com/python/cpython/issues/125018

webknjaz:
@ncoghlan thanks for taking this on! I was hoping to take a stub at actually writing some content, but it doesn't look like I'm going to get to that anytime soon, so I guess this is the way.

Seeing that you're repeating some of #1276, I think it would be nice to add a `Co-Authored-By: wim glenn <hey@wimglenn.com>` trailer to Git commits… That's possible via `git commit --amend --trailer='wim glenn <hey@wimglenn.com>'`. WDYT?

webknjaz (approved):

source/discussions/versioning.rst
```diff
@@ -172,6 +171,49 @@ since the latest release, setuptools-scm generates a version like
...
+   >>> importlib.metadata.version("pip")
+   '23.3.2'
+
+Many libraries also choose to version their top level
+:term:`import packages <Import Package>` by providing a package level
+``__version__`` attribute::
+
+   >>> import pip
+   >>> pip.__version__
+   '23.3.2'
```
ncoghlan, line 192: In the thread that prompted this updated PR, one of the key points of confusion was the assumption that because PEP 396 was rejected, packages shouldn't be setting `__version__` at all. The reason that PEP was rejected as a general requirement is the fact that the distribution package version and the import package API version *aren't* required to be the same (the names aren't even required to be the same).

Even though these versions are *often* the same, they're technically versioning different things (the distribution package and the import package's API), so requiring a specific relationship between them isn't valid.

The updated text in this PR aims to make that distinction clearer.

source/discussions/versioning.rst
```diff
@@ -172,6 +171,49 @@ since the latest release, setuptools-scm generates a version like
...
+
+Import packages are *not* required to be versioned independently of their
+distribution package version information (see the rejected proposal in
```
ChrisBarker-NOAA, line 195: I really wish Barry had Withdrawn, rather than Rejected, that one -- which I think is closer to the truth, but oh well. In any case, this text isn't opinionated -- thanks!

source/discussions/single-source-version.rst
```diff
@@ -1,31 +1,39 @@
...
+(as described in :ref:`runtime-version-access`).
 
-* It needs to be specified when building the package (e.g. in :file:`pyproject.toml`)
-   This will make it available in the installed package’s metadata, from where it will be accessible at runtime using ``importlib.metadata.version("distribution_name")``.
+It is also frequently desired that this version information be derived from a version
+control system *tag* (such as ``v1.2.3``) rather than being manually updated in the
+source code.
 
-* A package may set a module attribute (e.g., ``__version__``) to provide an alternative means of runtime access to the version of the imported package. If this is done, the value of the attribute and that used by the build system to set the distribution's version should be kept in sync in :ref:`the build systems's recommended way <Build system version handling>`.
+To ensure that version numbers do not get out of sync, it may be sufficient to add
```
ChrisBarker-NOAA, line 20: This feels out of place to me -- perhaps it can be moved to after the discussion of how to single source.

Having the build system do it for you is not an alternative to a test -- a test is a way to ensure that it was done correctly, regardless of the approach. Now that i think about it, relying on a test for a non-DRY method is an alternative to the DRY method, not the other way around.

[But I'm word-smithing here, not a blocker.]

ChrisBarker-NOAA (changes requested):
Thanks for finishing this up!

source/discussions/versioning.rst
ncoghlan: Heh, I had exactly the same thought when I was re-reading the PEP to check exactly why it was rejected :)

ncoghlan:
I posted a PR for the lack of semantic reference targets in the CPython docs, so assuming that review goes smoothly I may be able to fix this PR to use proper references before it gets merged: https://github.com/python/cpython/pull/125027

source/discussions/single-source-version.rst
ncoghlan: Yeah, I was genuinely unsure where best to include that note.

The other option I meant to add, but forgot to mention in the initial draft, is that https://peps.python.org/pep-0562/ now makes it possible to *lazily* populate `__version__` from `importlib.metadata.version(__spec__.name)`, so I'll definitely be making another pass over that bit (since it eliminates the main downside of doing that implicitly: the upfront import time cost, even if nobody ever accesses `__version__` on a given run).

source/discussions/versioning.rst
```diff
@@ -172,6 +171,49 @@ since the latest release, setuptools-scm generates a version like
...
+:pep:`PEP 396 <396>`), so this approach to retrieving runtime version
+information should only be used with libraries that are known to provide it.
+
+Library publishers wishing to ensure their reported distribution package and
```
eli-schwartz, line 199: This focuses on libraries specifically, but I'm not sure why. Applications, both CLI and GUI, often have the same concern -- more strongly for CLI than for GUI, as the latter are usually not installable via python packaging at all due to deficiencies in the wheel format.

Applications usually *must* report their version number at runtime somehow, and at least one of those locations is typically `progname --version` which is desired to complete as fast as possible in as few imports as possible. Applications may also support being packaged in a zipapp or PyInstaller or similar, in which case there is no "distribution metadata" anyway (but for the sake of people `pip install`ing the CLI application it's desirable to be able to specify the version to install, regardless of whether "distribution metadata" is also created).

All in all, the references to "libraries" comes across as weird to me.

source/discussions/single-source-version.rst
eli-schwartz: Note that lazy population makes the assumption that the rationale for avoiding importlib.metadata is the cost "when not accessed anyway" which I feel is a fairly big assumption even for libraries (why write even more complicated code just to make a certain part of your API even slower when you can just do it right to begin with for less code, less maintenance, and better performance?).

 But it's completely invalid for the application use case as the primary concern there is to make the happy path of running the application just to report the version be as fast as possible -- and of course that's the one case where you're absolutely guaranteed to (slowly) access this attribute. The rest of the time, you may not care if the attribute is eagerly loaded and incurs a universal slowdown since the application is probably doing loads of work anyway and importing the importlib.metadata indirector (sic) could end up drowned out in the noise.

source/discussions/versioning.rst
```diff
@@ -172,6 +171,49 @@ since the latest release, setuptools-scm generates a version like
...
+using the standard library's ``importlib.metadata.version`` function::
+
+   >>> importlib.metadata.version("pip")
+   '23.3.2'
+
+Many libraries also choose to version their top level
+:term:`import packages <Import Package>` by providing a package level
+``__version__`` attribute::
+
+   >>> import pip
```
pradyunsg, line 190: I would prefer if we used something other than pip as an example of this. 😅

We actively discourage using pip as an importable package.

source/guides/single-sourcing-package-version.rst
```diff
@@ -1,173 +1,8 @@
...
-        Although this technique is common, beware that it will fail if
-        ``sample/__init__.py`` imports packages from ``install_requires``
-        dependencies, which will very likely not be installed yet when
-        :file:`setup.py` is run.
-
-
-#.  Keep the version number in the tags of a version control system (Git, Mercurial, etc)
-    instead of in the code, and automatically extract it from there using
-    `setuptools_scm <https://pypi.org/project/setuptools-scm/>`_.
+If the page doesn't automatically refresh, click :ref:`here <single-source-version>`.
```
merwok, line 8: «here» is not a good practice for link text 🙂 

```suggestion
If the page doesn't automatically refresh, see :ref:`single-source-version`.
```

source/discussions/single-source-version.rst
```diff
@@ -1,31 +1,39 @@
...
 1) If the code is in a version control system (VCS), e.g. Git, then the version can be extracted from the VCS.
 
-2) The version can be hard-coded into the :file:`pyproject.toml` file -- and the build system can copy it into other locations it may be required.
-
-3) The version string can be hard-coded into the source code -- either in a special purpose file, such as :file:`_version.txt`, or as a attribute in a module, such as :file:`__init__.py`, and the build system can extract it at build time.
+2) The version can be hard-coded into the :file:`pyproject.toml` file -- and the build system can copy it
+   into other locations it may be required.
 
+3) The version string can be hard-coded into the source code -- either in a special purpose file,
+   such as :file:`_version.txt`, or as an attribute in a module, such as :file:`__init__.py`, and the build
```
bwoodsend, line 35: I don't want to start a big debate here so if there's any conflict then just say no and resolve this comment.

Could we not mention version files? Seeing defunct packages on PyPI produce `FileNotFoundError: version.txt` on build because they forgot to add the file to their manifest is my primary motivation for being so opposed to `__version__` attributes in the first place.

webknjaz:
@ncoghlan so I made a mistake in the suggested attribution command earlier (I've updated that comment to correct it since) and missed a chunk of the trailer in it, which is why GitHub didn't attribute that commit to wim glenn. It was supposed to be `git commit --amend --trailer='Co-Authored-By: wim glenn <hey@wimglenn.com>'`.
I must note, however, that this command modifies the last commit, and you seem to have produced a separate empty commit (I'm guessing with `--allow-empty`).

Normally, I'd suggest just putting this trailer as a `Co-Authored-By: wim glenn <hey@wimglenn.com>` line into the merge commit and use the squash button. But this repository is set up to use natural merge when accepting PRs, so you won't have this option. All of the individual commits here will become a part of `main` and the empty commit will be what's credited. So my recommendation is to squash the commits locally, with an interactive rebase, adding that line to the resulting commit message. This can be done right before merging.

ncoghlan (commented):
Assorted review comments addressed. Build will currently fail, as I went back to using semantic cross-references to `importlib.metadata` APIs (in anticipation of getting the stdlib docs update PR and its backports merged)

source/discussions/single-source-version.rst
ncoghlan: I reworked this to move the testing recommendation to the end, and ultimately decided not to mention the lazy loading idea (since @eli-schwartz is right and it's basically the worst possible solution).

ncoghlan:
@webknjaz Huh, interesting, I hadn't realised that squash merges could be disabled outright (although I can see what enabled the merge queue would do so). I'll do the full squash on the PR branch just before merge (since it messes up the comment threads if I do it beforehand)

webknjaz:
@ncoghlan yep, you're right. Technically that's not disabled directly but when merge queues is required for a branch, it only lets one select a single method. Overriding it is possible for admins as they'd be able to bypass the merge queues when force-merging.

ncoghlan:
The semantic link targets are live on docs.python.org, so I'll remove the `!` prefixes from the PR (since https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.version is a valid target now).

source/discussions/single-source-version.rst
```diff
@@ -1,34 +1,49 @@
...
 
 :Page Status: Complete
-:Last Reviewed: 2024-08-24
+:Last Reviewed: 2024-10-07
 
-One of the challenges in building packages is that the version string can be required in multiple places.
+Many Python :term:`distribution packages <Distribution Package>` publish a single
+Python :term:`import package <Import Package>` where it is desired that the runtime
+``__version__`` attribute on the import package report the same version specifier
+as :func:`!importlib.metadata.version` reports for the distribution package
```
ncoghlan, line 13:
```suggestion
as :func:`importlib.metadata.version` reports for the distribution package
```

source/discussions/versioning.rst
```diff
@@ -172,6 +171,55 @@ since the latest release, setuptools-scm generates a version like
...
 "0.5.dev1+gd00980f.d20231217".
 
+.. _runtime-version-access:
+
+Accessing version information at runtime
+========================================
+
+Version information for all :term:`distribution packages <Distribution Package>`
+that are locally available in the current environment can be obtained at runtime
+using the standard library's :func:`!importlib.metadata.version` function::
```
ncoghlan, line 181:
```suggestion
using the standard library's :func:`importlib.metadata.version` function::
```

source/discussions/versioning.rst
```diff
@@ -184,6 +232,15 @@ since the latest release, setuptools-scm generates a version like
    Brett Cannon <semver-brett-cannon_>`_. For a humoristic take, read about
    ZeroVer_.
 
+.. [#fallback-to-dist-version] A full list mapping the top level names available
+   for import to the distribution packages that provide those import packages and
+   modules may be obtained through the standard library's
+   :func:`!importlib.metadata.packages_distributions` function. This means that
```
ncoghlan, line 238:
```suggestion
   :func:`importlib.metadata.packages_distributions` function. This means that
```

tabedzki (commented):
Lines 101-114 seem to overlap with lines 154-167 since they both cover the versioning dynamically though only one points to the updated "single source version" discussion page while the other doesn't. 

Should we (1) update the example for static vs dynamic to not be version but some other attribute, (2) update lines 113-114 to point to the discussion as well, or (3) remove one of these two sections?

I will note that the [#1276](https://github.com/pypa/packaging.python.org/pull/1276/commits/cde56ce42eb10c77ad646486983bd923cddc29fa) only had this included in the static vs dynamic section and did not include it under version section itself.

source/guides/writing-pyproject-toml.rst
```diff
@@ -163,8 +163,8 @@ This field is required, although it is often marked as dynamic using
    dynamic = ["version"]
 
 This allows use cases such as filling the version from a ``__version__``
-attribute or a Git tag. Consult :ref:`Single sourcing the version` for more
-details.
+attribute or a Git tag. Consult the :ref:`single-source-version`
+discussion for more details.
```
tabedzki, line 167: Lines [101-114](https://github.com/ncoghlan/packaging.python.org/blob/bbf6d6618ebebaf0b609c637a2e107f0a1556f4e/source/guides/writing-pyproject-toml.rst?plain=1#L101-L114) seem to overlap with lines [154-167](https://github.com/ncoghlan/packaging.python.org/blob/bbf6d6618ebebaf0b609c637a2e107f0a1556f4e/source/guides/writing-pyproject-toml.rst?plain=1#L154-L167) since they both cover the versioning dynamically though only one points to the updated "single source version" discussion page while the other doesn't.

Should we (1) update the example for static vs dynamic to not be version but some other attribute, (2) update lines [113-114](https://github.com/ncoghlan/packaging.python.org/blob/bbf6d6618ebebaf0b609c637a2e107f0a1556f4e/source/guides/writing-pyproject-toml.rst?plain=1#L113-L114) to point to the discussion as well, or (3) remove one of these two sections?

I will note that the [#1276](https://github.com/pypa/packaging.python.org/pull/1276/commits/cde56ce42eb10c77ad646486983bd923cddc29fa) only had this included in the static vs dynamic section and did not include it under version section itself.

ChrisBarker-NOAA: >  (1) update the example for static vs dynamic to not be version but some other attribute

Please keep version as an example, it's one of the most common use cases.

I think it's OK for 113-114 to be like it is -- refer to the build system docs. while version is the example, that section is not specifically about version.

101--114 could be shortened some, so as not to be too repetitive.

ncoghlan: I'd prefer to keep the scope of this PR's effect on the pyproject.toml guide to just simplifying the link reference. Happy to take a look at a separate PR tightening up the wording, though.

ncoghlan:
Given the +1's above and the lack of objections on the discuss.python.org thread, I'm going to go ahead and merge this one to give us a solid foundation for any future updates.

ncoghlan:
@webknjaz Rather than force-pushing to this branch, I decided it would be cleaner to supersede this PR with a new one (so the comment history doesn't get messed up), but that does mean the new PR needs approval from a second reviewer: #1612

PR #1612, opened 2024-10-12
Add notes on runtime version access

ncoghlan:
Also redirects the obsolete single-source version guide to the updated single-source version discussion.

Squash merge of #1607 to give the desired co-author attribution on the final commit.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1612.org.readthedocs.build/en/1612/


webknjaz (approved):
Thanks!

ncoghlan:
Slightly delayed due to the Internet Archive DDoS (some of our links point to the wayback machine, so the link checks failed while that was down)

PR #1613, opened 2024-10-13
Add dependabot.yml to keep dependencies up to date

shenxianpeng:
Add dependabot.yml to update GitHub action and Python package dependencies to date automatically. 
(maybe it also can help with pull requests like #1614 )

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1613.org.readthedocs.build/en/1613/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
       actions:
         patterns:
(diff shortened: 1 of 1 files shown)


webknjaz:
This will probably not work for workflows embedded in the guides. Plus, it's best to verify those manually every time since there's no automation doing that.

webknjaz:
I don't like the frequent intervals in the config. Dependabot can be very noisy. Quarterly would be fine, I suppose, but daily is crazy.

webknjaz:
Monthly still feel like a lot. It might seem that it's not, but that's only when you get notifications from a single repository. The more repos, the more notifications. So for me personally, it'd just add to a lot of noise that's already present… Not sure if anybody else here would want this in.

webknjaz:
On that note, it'd be nice to make `pre-commit.ci`'s PRs less frequent too.

PR #1614, opened 2024-10-13
upgrade gh-action-sigstore-python to v3.0.0

### LINKED ISSUE #1610: The user guide's `publish.yml` fails on `ubuntu-latest` because of PEP 668
The publishing workflow at https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ uses `ubuntu-latest` as its platform, and also references `sigstore/gh-action-sigstore-python@v2.1.1`, a version that is broken on Ubuntu 24.04, because PEP 668 prevents it from working outside of a virtualenv.

Ubuntu 24.04 is the current referent of `ubuntu-latest`.

gh-action-sigstore-python version 3.0.0 incorporated this fix: https://github.com/sigstore/gh-action-sigstore-python/pull/145

I'm aware that the bottom of the page recommends updating all actions to their latest versions, but does not specify how to find out what the latest version numbers are.

The error you get from following the instructions makes it unclear what needs to be fixed, given that the user of the GitHub action does not have interactive access to the filesystem and is not even specifying the commands to run:

```
× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.
```

**Suggested fix**: The `@v2.1.1` in the example should be changed to `@v3.0.0`.

shenxianpeng:
closes #1610 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1614.org.readthedocs.build/en/1614/

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -68,7 +68,7 @@ jobs:
         name: python-package-distributions
         path: dist/
     - name: Sign the dists with Sigstore
-      uses: sigstore/gh-action-sigstore-python@v2.1.1
+      uses: sigstore/gh-action-sigstore-python@v3.0.0
       with:
         inputs: >-
           ./dist/*.tar.gz


webknjaz (approved):
Thanks!

PR #1620, opened 2024-10-23
PEP 396 was reclassifed as Withdrawn

ncoghlan:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1620.org.readthedocs.build/en/1620/


ncoghlan:
Since I was posted a PR anyway, I also fixed another bit of technically incorrect wording in my previous PR. Compliance with PEP 440 (or the lack thereof) isn't the most significant reason for avoiding `__version__` when reporting external API versions, it's the fact the external API version isn't the *module* version.

tacaswell (approved):
Was on my way here to make the same PR 😄

webknjaz (approved):
Thanks!

PR #1623, opened 2024-10-28
Add an initial specification doc for Dependency Groups

sirosen:
This is a first draft doc, based off of the original spec + reading the surrounding specs for examples of what's desirable.

I have in some cases reworded things from the original PEP document to try to simplify and clarify, but we can hew closer to the original spec text if that's preferable.
For example, I simplified the section which defines Dependency Group Includes: the spec defines `Dependency Object Specifiers`, the class of tables inside of dependency groups, and then defines one type of `Dependency Object Specifier`: a `Dependency Group Include`. I think that phrasing is good for the original PEP, which wants to lay out the idea that we could define other table types in the future, but not necessarily useful for someone who wants to just read and implement the spec as it currently stands.

In many cases, I very intentionally lifted-and-shifted original PEP text, since we worked hard on it and it's quite good.

Unlike the original PEP, there are not that many examples given here.
I also was uncertain if I should link out to the ``dependency-groups`` package instead of or in addition to the smaller and simpler reference implementation from the PEP.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1623.org.readthedocs.build/en/1623/

### DIFF
--- a/source/specifications/dependency-groups.rst
+++ b/source/specifications/dependency-groups.rst
@@ -36,47 +36,33 @@ and a similar table which defines ``test`` and ``coverage`` groups::
 The ``[dependency-groups]`` Table
 ---------------------------------
 
-This PEP defines a new section (table) in ``pyproject.toml`` files named
+Dependency Groups are defined as a table in ``pyproject.toml`` named
 ``dependency-groups``. The ``dependency-groups`` table contains an arbitrary
 number of user-defined keys, each of which has, as its value, a list of
-requirements (defined below). These keys must be
-`valid non-normalized names <https://packaging.python.org/en/latest/specifications/name-normalization/#valid-non-normalized-names>`__,
-and must be
-`normalized <https://packaging.python.org/en/latest/specifications/name-normalization/#normalization>`__
-before comparisons.
+requirements.
 
-Tools SHOULD prefer to present the original, non-normalized name to users by
-default. If duplicate names, after normalization, are encountered, tools SHOULD
-emit an error.
+``[dependency-groups]`` keys, sometimes also called "group names", must be
+:ref:`valid non-normalized names <name-format>`. Tools which handle Dependency
+Groups MUST :ref:`normalize <name-normalization>` these names before
+comparisons.
 
-Requirement lists under ``dependency-groups`` may contain strings, tables
-("dicts" in Python), or a mix of strings and tables.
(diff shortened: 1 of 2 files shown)


webknjaz:
May I suggest adding the PEP "as is" first. In a separate PR. This would allow seeing the diff between that and simplification. Alternatively, put it in here in a separate commit, leaving the changes out so they'd only show up in the second commit.
I think this would be good for transparency.

sirosen:
Sure, that's a good idea to make the diff more readable! I'll have to come back to this and rewrite the commit history to be
1. Add the PEP contents (probably with some omissions, e.g. appendices) 
2. Edit that down 

I'm not able to do it right now, but maybe tomorrow.

webknjaz:
@sirosen ack

sirosen:
Oops! I forgot about this PR! I'll make an update this week. Thanks for the ping!

sirosen:
It's slightly embarrassing to have forgotten again, but that's what happened. I just checked my open PRs and saw this. I got it done by grabbing PEP text, rewriting my commit, and then checking out my rewrite over the top of it.

Most of the edits are mild but I also see that some of the line break positions changed -- I can't recall the exact process by which I did the original work, but I may have been copying from a rendered copy of the PEP at times. As a result, there's some noise in the diff, but I hope it's still easier to review.

ncoghlan (approved):

ncoghlan:
@sirosen Added to the merge queue. Once it's live, you can submit the changes over in the PEPs repo to link to the canonical spec location.

webknjaz:
Urgh... It's https://github.com/pypa/packaging.python.org/issues/1744. It needs to be fixed before merging can succeed.

webknjaz:
@sirosen oh, and no, you don't need to rebase. Merge Queues make a temporary branch with your PR merged into `main` and then a GHA workflow runs against that. If it succeeds, then `main` is fast-forwarded to that merge commit. It didn't, so the PR was dropped as potentially breaking `main`.

When it doesn't fail because of some external circumstances (like in this case, with PyPI changing things), it is able to catch cases when two PRs have green CI but when they are merged together, it would fail (which isn't evident separately). Something like https://bors.tech/essay/2017/02/02/pitch/ but built-in (and more limited).

ncoghlan:
#1767 has been posted with the PyPI link check workaround

PR #1653, opened 2024-11-13
Add simple issue template

abravalheri:
This is a follow up on a discussion on PyPA discord.
The hope is that having some mandatory fields will reduce the amount of spam or at least confuse a little bit the bots.

I am using the text `I am aware that participants in this repository must follow the PSF Code of Conduct`, instead of `I agree to follow the PSF Code of Conduct`, just to workaround some criticism we received in the past (e.g. https://discuss.python.org/t/what-to-do-with-contributors-who-refuse-code-of-conduct/13287).

Pretty much everything can be changed, this is just a initial draft to get things started.

I haven't been able to double check this renders correctly (so it may have problems). Possibly a maintainer also have access to Github UI to do this.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,6 @@
+blank_issues_enabled: false
+contact_links:
+  - name: "Community Guidelines"
+    url: "https://github.com/pypa/.github/blob/main/CODE_OF_CONDUCT.md"
+    about: "Please make sure to follow the PSF Code of Conduct when participating in this repository."
+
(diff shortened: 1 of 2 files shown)


webknjaz:
@abravalheri https://github.com/abravalheri/packaging.python.org/blob/issue-templates/.github/ISSUE_TEMPLATE/general.yml is supposed to render it but there's syntax errors.

I'd start by inheriting https://github.com/pypi/support/blob/c38e8a9/.github/ISSUE_TEMPLATE/limit-request-file.yml?plain=1#L89-L98.

`id:` is not necessary to my knowledge, I'm not sure where that is coming from in some examples. `required` is a validator.

I think it might be beneficial to surface a paragraph of the CoC explanation at the beginning: https://github.com/aio-libs/propcache/blob/73fc000/.github/ISSUE_TEMPLATE/bug_report.yml#L7-L36.

abravalheri:
> @abravalheri https://github.com/abravalheri/packaging.python.org/blob/issue-templates/.github/ISSUE_TEMPLATE/general.yml is supposed to render it, but there's syntax errors.

Thank you very much @webknjaz, that is a good tip for checking the rendering, haven't thought about it.

I tried to follow more or less some of your suggestions in the newest changes.

ncoghlan (approved):
Thanks! I'm going to go ahead and merge this to see if it cuts down on the spam.

Yek666 (approved):
transferencia

PR #1654, opened 2024-11-14
Update packaging-projects.rst to detail saving to .pypirc

ekohilas:
Possible solution to https://github.com/pypa/packaging-problems/issues/841

(although more likely a bug with twine)


merwok:
.pypirc is the old way of saving credentials, with its own usability and security concerns.
It should not be recommended over an interactive prompt (the current doc).

PR #1661, opened 2024-11-18
note an alternative for running src packages locally.

ekohilas:
Upon converting a package to src structure, a user is left unable to run their code via the normal python command, `python src/awesome_project/__main__.py`

This PR adds an alternative approach that doesn't require an in file workaround.


sinoroc:
I don't think we should have this in the guide, the whole point of the `src` layout is to not do that, and by "that" I mean importing from the current working directory, because it is cause for shooting oneself in the foot.

If we want to have this, it should be a small foot note hidden somewhere no one looks. Well maybe not that bad, but it definitely should not be the first suggestion. The first suggestion should be to install (as editable).

ekohilas:
> If we want to have this, it should be a small foot note hidden somewhere no one looks. Well maybe not that bad, but it definitely should not be the first suggestion. The first suggestion should be to install (as editable).

Ah thank you, I see your point.

I have updated to address that.

chrysle (approved):

source/discussions/src-layout-vs-flat-layout.rst
```diff
@@ -86,12 +86,22 @@ Running a command-line interface from source with src-layout
...
-:doc:`Development Mode <setuptools:userguide/development_mode>`
-for testing purposes. Since this can be unpractical in some situations,
-a workaround could be to prepend the package folder to  Python's
-:py:data:`sys.path` when called via its :file:`__main__.py` file:
+interface can not be run directly from the :term:`source tree <Project Source Tree>`.
+
+To run a command-line interface directly, you can install the package in
+:doc:`Development Mode <setuptools:userguide/development_mode>` for testing purposes.
+
+Since this can be unpractical in some situations, you can also either:
```
webknjaz, line 94:
```suggestion
Since this can be impractical in some situations, you can also either:
```

source/discussions/src-layout-vs-flat-layout.rst
```diff
@@ -86,12 +86,22 @@ Running a command-line interface from source with src-layout
...
+
+Run :file:`__main__.py` via:
+
+.. code-block:: bash
+
+    cd src/
+    python -m awesome_package
+
+Or prepend the package folder to Python's :py:data:`sys.path` when called via its
```
webknjaz, line 103: If these are options, they should probably be bullet points. It's weird to have two consequent paragraphs ending with a colon.

webknjaz: https://python-packaging-user-guide--1661.org.readthedocs.build/en/1661/discussions/src-layout-vs-flat-layout/#running-a-command-line-interface-from-source-with-src-layout

webknjaz:
Honestly, I share @sinoroc's sentiment on including dark sorcery into this document. Runpy scripts usually need dependencies, that are declared in the packaging metadata. So it doesn't really make sense to showcase an execution method that is broken half the time, forcing the reader to track down what they have to install. OTOH, `pip install -e .` will always put both dependencies and the importables on `$PYTHONPATH`. Always correctly.
Should we really indulge a `PYTHONPATH=src/ python -m awesome_package` in this case? Wouldn't it create more confusion to an entire class of newcomers?

ncoghlan (changes requested):
My "requested changes" flag here is less about making specific changes, and more about recording the guide maintainer uncertainty about the addition (which I share). I used to do this kind of thing myself, and my recommendation these days is "don't do it, use `pdm run` and similar tool-supported idioms".

PR #1668, opened 2024-11-25
tool-recommendations: update Trusted Publisher providers

woodruffw:
The docs currently say that only GitHub Actions is supported, but this is no longer true.

I've updated the section to include the currently supported providers, as well as include a rough date so that future readers can infer how old (and therefore stale) the list may be.

(I also added a `.. warning` callout to a statement that looked like it needed one. But I can undo that if desired!)

CC @webknjaz 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1668.org.readthedocs.build/en/1668/

### DIFF
--- a/source/guides/tool-recommendations.rst
+++ b/source/guides/tool-recommendations.rst
@@ -139,15 +139,25 @@ to build distributable wheels.
 Uploading to PyPI
 =================
 
-For projects hosted on GitHub, it is recommended to use the :ref:`trusted publishing
-<trusted-publishing>`, which allows the package to be securely uploaded to PyPI
-from a GitHub Actions job. (This is not yet supported on software forges other
-than GitHub.)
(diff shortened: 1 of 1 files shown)


source/guides/tool-recommendations.rst
```diff
@@ -139,15 +139,25 @@ to build distributable wheels.
...
+* GitHub Actions (on ``https://github.com``)
+* GitLab CI/CD (on ``https://gitlab.com``)
+* ActiveState
+* Google Cloud
 
 The other available method is to upload the package manually using :ref:`twine`.
 
-**Never** use ``python setup.py upload`` for this task. In addition to being
-:ref:`deprecated <setup-py-deprecated>`, it is insecure.
+.. warning::
```
webknjaz, line 157: Perhaps, even
```suggestion
.. danger::
```
?

PR #1673, opened 2024-12-06
Clarify that the scripts directory is for files only

konstin:
Following https://github.com/astral-sh/uv/issues/9656 and the discussion in the pypa discord (https://discord.com/channels/803025117553754132/837243676814999553/1314391181656064040), add a clarifying sentence that the scripts directory in wheels if for files only.

A future specification may want to extend this to "regular files and symlinks", for not it's limited to regular for clear rewrite semantics.

For ecosystem impact, uv errors when the scripts directory contains non-files and #9656 is the first time this error was reported.

CC @charliermarsh @pradyunsg @uranusjr 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1673.org.readthedocs.build/en/1673/


ncoghlan (approved):

ncoghlan:
Thanks! I also added a History note about the clarification.

Issue #1744, opened 2024-12-11
`linkcheck` in CI is broken on PyPI URLs with anchors


webknjaz:
> The linkcheck reports on `https://pypi.org/project/pip/23.3.1/#files` (Non-existing anchor), but that's surprising: yesterday's cron job reported is as success and the link works (opens PyPI on the files tab). Also, it's not related to any changes made for this PR.

You're right. Over the past few days I noticed a quick Fastly loading screen showing up on PyPI, which then redirects to where I was going originally.
So I probed it with cURL just now and verified that this is what's happening, and HTML DOM no longer contains that in the HTTP first response (this is probably cookie-based):
```console
$ curl -v 'https://pypi.org/project/pip/23.3.1/#files'
* Host pypi.org:443 was resolved.
* IPv6: 2a04:4e42:600::223, 2a04:4e42:200::223, 2a04:4e42:400::223, 2a04:4e42::223
* IPv4: 151.101.0.223, 151.101.192.223, 151.101.64.223, 151.101.128.223
*   Trying [2a04:4e42:600::223]:443...
* Immediate connect fail for 2a04:4e42:600::223: Network is unreachable
*   Trying [2a04:4e42:200::223]:443...
* Immediate connect fail for 2a04:4e42:200::223: Network is unreachable
*   Trying [2a04:4e42:400::223]:443...
* Immediate connect fail for 2a04:4e42:400::223: Network is unreachable
*   Trying [2a04:4e42::223]:443...
* Immediate connect fail for 2a04:4e42::223: Network is unreachable
*   Trying 151.101.0.223:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / x25519 / RSASSA-PSS
[...]

PR #1762, opened 2024-12-19
Update conf.py automatically keep copyright date the current year

clintonsteiner:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1762.org.readthedocs.build/en/1762/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -2,7 +2,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 import os
-import time
+from time import strftime
 
 # Some options are only enabled for the main packaging.python.org deployment builds
 RTD_BUILD = bool(os.getenv("READTHEDOCS"))
@@ -14,7 +14,7 @@ RTD_CANONICAL_BUILD = (
 
 project = "Python Packaging User Guide"
 author = "Python Packaging Authority"
-cur_year = time.strfime("%Y")
+cur_year = strftime("%Y")
 copyright = f"{cur_year}, {author}"
 
 # -- General configuration -------------------------------------------------------------
--- a/source/conf.py
+++ b/source/conf.py
@@ -2,6 +2,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 import os
+from datetime import date
 
 # Some options are only enabled for the main packaging.python.org deployment builds
 RTD_BUILD = bool(os.getenv("READTHEDOCS"))
@@ -13,7 +14,7 @@ RTD_CANONICAL_BUILD = (
 
 project = "Python Packaging User Guide"
 
-copyright = "2013–2020, PyPA"
+copyright = f"2013–{date.today().year}, PyPA"
 author = "Python Packaging Authority"
 
 # -- General configuration -------------------------------------------------------------
--- a/source/conf.py
+++ b/source/conf.py
@@ -2,7 +2,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 import os
-from datetime import date
+import time
 
 # Some options are only enabled for the main packaging.python.org deployment builds
 RTD_BUILD = bool(os.getenv("READTHEDOCS"))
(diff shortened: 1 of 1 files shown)


webknjaz:
The correct fix for linkcheck is described in #1744. Submit it separately.
Don't change the documents, that's not necessary.
As for the copyright, I'm not sure. It might be a good idea to just keep the first year. FWIW, copyright is supposed to signal the content creation. Updating it to a random year just because somebody restarted Sphinx is plain wrong.

clintonsteiner:
Removed - then began reading about copyrights
I'll stand by the change as if we are using it incorrectly we should keep it up to date - stating that pypa still has copyright on the material

Agree with what you said that it doesn't serve a purpose, however it may as well indicate that the website is still maintained

ncoghlan:
The question of static years in copyright notices came up recently(ish) for the main CPython documentation:

* Issue for the change (with advice from PSF legal counsel): https://github.com/python/cpython/issues/126133
* Resulting ongoing guidance (just list the *first* year, not subsequent years): https://devguide.python.org/getting-started/pull-request-lifecycle/#copyrights

The CPython reasoning is applicable here as well, so just saying "2013" would make sense.

(there are other issues with the notice due to the PyPA not really existing in any legal sense, but we're not going to try to sue anyone for copyright infringement over the packaging guide, so it doesn't particularly bother me)

ncoghlan:
For reference, #1767 has the linkcheck workaround to avoid having pypi.org anchor checks fail due to the active botscraping load reduction defences.

webknjaz (changes requested):
> keep it up to date [...] indicate that the website is still maintained

Well, this would be just wrong because it has nothing to do with the maintenance status. It's just the indication that somebody wanted to update this string, and that's about it. Here's another explanation why this is pointless: https://hynek.me/til/copyright-years/.

So I don't think that creating a false feeling of maintenance for arbitrary individuals on the internet is enough justification for contributing to the confusion further.

An acceptable change would be deleting the end-year or both years. The LF and many other respected organizations and individuals all agree on this generic guidance.

Another possible improvement here would be extending “PyPA” to “PyPA with PyPUG maintainers and contributors”. Technically, the contributors do not surrender their copyright to us fully, as I understand (since we don't demand that people sign a CLA).

clintonsteiner:
@webknjaz I actually disagree with you here
The examples listed such as LF list current year for copyright
PyPi uses copyright current year here
Python.org uses Copyright beg year to cur year
Linux foundation uses copyright current year

image
[Linux Foundation](https://training.linuxfoundation.org/?campaignid=21372675077&adgroupid=171974510868&creative=704304945812&matchtype=e&network=g&device=c&keyword=linux%20foundation&hsa_acc=8666746580&hsa_cam=21372675077&hsa_grp=171974510868&hsa_ad=704304945812&hsa_src=g&hsa_tgt=kwd-319242482088&hsa_kw=linux%20foundation&hsa_mt=e&hsa_net=adwords&hsa_ver=3&gad_source=1&gbraid=0AAAAAD2R-lqgn4NVlj9HU6JIopz--dkKG&gclid=Cj0KCQiAvvO7BhC-ARIsAGFyToVlfil6jawEHqr9jNH3BkR4fqvMJ6V5GEJi7OUkibIMvHB3r1J_4X8aApI_EALw_wcB&utm_source=google&utm_medium=paid-search&utm_campaign=24q2-evergreen-lf_training&utm_term=tnc-na-search-lf-brand&utm_content=eg_rsa&utm_term=linux%20foundation&utm_campaign=tnc-na-search-lf)

Python software foundation
image
[Pypi](https://pypi.org/manage/account/#account-emails)

Python.org
image
[Python.org](https://www.python.org/psf-landing/)

clintonsteiner:
This is being hotly contested in the public forum of reddit https://www.reddit.com/r/webdev/comments/1hwbsou/do_you_not_code_the_copyright_year_to_be_dynamic/
Seems the community is split 50-50, although the logic shoudln't matter, companies keep the year up to date so it doesn't look abandoned
I'd argue to do the same here, but splitting hairs on a unimportant thing

PR #1763, opened 2024-12-19
Standardize GitHub actions setup python

clintonsteiner:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1763.org.readthedocs.build/en/1763/

### DIFF
--- a/source/contribute.rst
+++ b/source/contribute.rst
@@ -134,9 +134,9 @@ need:
 
       python -m pip install --user nox
 
-2. Python 3.11. Our build scripts are usually tested with Python 3.11 only.
+2. Python 3.13. Our build scripts are usually tested with Python 3.11 only.
    See the :doc:`Hitchhiker's Guide to Python installation instructions <python-guide:starting/installation>`
-   to install Python 3.11 on your operating system.
+   to install Python 3.13 on your operating system.
 
 To build the guide, run the following shell command in the project's root folder:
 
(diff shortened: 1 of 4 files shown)


clintonsteiner:
https://github.com/pypa/packaging.python.org/pull/1762/commits/e3d9b36f4dd8f89be82072302a1dfb32f281c57c

This commit will fix the error here as well and are unrelated to the scope of the pr

webknjaz:
Looks like the version isn't updated everywhere. Also, I wouldn't rush to bump it. Updating the actions is fine, though. But the PR title/description should say so.

clintonsteiner:
Updating it just in the contributing section which is discussing specifically what's being used in ci to build the docs

Fixed the missing version update, thanks @webknjaz

source/contribute.rst
```diff
@@ -134,9 +134,9 @@ need:
 
       python -m pip install --user nox
 
-2. Python 3.11. Our build scripts are usually tested with Python 3.11 only.
+2. Python 3.13. Our build scripts are usually tested with Python 3.11 only.
```
webknjaz, line 137: This says we use 3.11 in tests.

webknjaz:
I think we were usually trying to have older Python versions to accommodate for contributors unable to upgrade to the latest… I'm not entirely sure if we should jump right into 3.13. Perhaps other people would have extra opinions, though.

PR #1765, opened 2024-12-22
Use environment variables instead of contexts

tomkins:
Closes #1748

Updates the "Signing the distribution packages" section to use environment variables rather than GitHub Actions context variables. No additional setting of environment variables are needed - these are all set/available already.

I've got a repo where we're using trusted publishing, although it's a slight variation on this, but hopefully it's an example that the environment variables work just fine:

https://github.com/developersociety/django-findreplace/blob/7fcf87397590b984dacaca58719f2d8b737d4f77/.github/workflows/publish.yml#L81
https://github.com/developersociety/django-findreplace/actions/runs/12457101550

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1765.org.readthedocs.build/en/1765/

PR #1768, opened 2025-01-02
Add and configure zizmor

di:
As a follow-on to #1765, this adds a workflow which runs zizmor on the example workflow to ensure it doesn't regress.


woodruffw (approved):
LGTM!

webknjaz:
> ```console
> (guides/installing-scientific-packages: line   91) broken    https://scipy.org/install/#distributions - Anchor 'distributions' not found
> ```

webknjaz (approved):

PR #1775, opened 2025-01-07
Update ci workflows

clintonsteiner:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1775.org.readthedocs.build/en/1775/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,6 @@ jobs:
         run: |
           python -m pip install --upgrade nox virtualenv
 
-      - name: Nox ${{ matrix.noxenv }}
+      - name: Nox ${{ inputs.nox-env }}
         run: |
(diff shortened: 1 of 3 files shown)


webknjaz (commented):
The idea is initially interesting, but I have a better example.
Mimic these if you want to continue working on this PR: https://github.com/ansible/receptor/blob/794e4c402c7045091e111a3bfc97722a2af214ba/.github/workflows/reusable-nox.yml#L58-L64 / https://github.com/ansible/receptor/blob/794e4c402c7045091e111a3bfc97722a2af214ba/.github/workflows/pull_request.yml#L87-L120. Without the go or project-specific stuff, of course. Follow the naming conventions too.

The "module" should be named `reusable-*`. The conditionals should live outside that reusable module. The call to nox should be split into provisioning and execution.

Besides that, I'd rather have version bumps in a separate PR.

PR #1776, opened 2025-01-07
switch to uv and pyproject.toml

clintonsteiner:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1776.org.readthedocs.build/en/1776/


pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+]
+dependencies = [
+    "furo==2024.8.6",
+    "sphinx-autobuild==2021.3.14",
+    "sphinx==7.2.6",
+    "sphinx-inline-tabs==2023.4.21",
+    "sphinx-copybutton==0.5.2",
+    "sphinx-toolbox==3.5.0",
+]
+
```
webknjaz, line 27: This project is not a distribution package, it must not have any packaging metadata as it's not supposed to be packaged as such.
```suggestion
```

Using the dependency groups may be fine, provided that they list docs build environment deps.

pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+[dependency-groups]
+dev = [
```
webknjaz, line 29: Why is this called dev? What's the difference from non-dev? Why is sphinx-lint installed in the same env as nox? Why is virtualenv needed?

README.md
webknjaz, line 1: Why is this empty file added?

.readthedocs.yml
```diff
@@ -0,0 +1,17 @@
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
```
webknjaz, line 1: This is a deprecated filename that is warned about right at the top of the page, linked in this first line of the file. It must not be used.

.readthedocs.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  # adapted from uv recipe at https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv
+  # and comment at https://github.com/readthedocs/readthedocs.org/issues/11289#issuecomment-2103832834
+  commands:
+    - asdf plugin add uv
+    - asdf install uv latest
+    - asdf global uv latest
+    - asdf global uv latest
+    - uv sync --extra docs --frozen
+    - uv run -m sphinx -T -b html -d source/_build/doctrees source $READTHEDOCS_OUTPUT/html
```
webknjaz, line 17: Sphinx must remain integrated via nox.

webknjaz:
Looking into this, I don't feel like this is a desired change overall. Especially with no justification and zero prior discussion. There's a bunch of integrations and assumption being broken (like nox being the central workflow tool) among other things.

I don't think I'll be reviewing this anymore until there's at least some consensus among the maintainers that it's even slightly acceptable to look into replacing such a big chunk of automation with unclear gains.

chrysle:
> I don't think I'll be reviewing this anymore until there's at least some consensus among the maintainers that it's even slightly acceptable to look into replacing such a big chunk of automation with unclear gains.

I'm -1 on this too; the current setup suits PyPUG's needs perfectly well.

chrysle:
Closing for now.

PR #1783, opened 2025-01-08
Document that Requires-Dist can use explicit URLs

godlygeek:
The core-metadata documentation currently implies that `Requires-Dist` can only use name-based specifiers, but then references PEP 508 which allows for URL based specifiers. Update the documentation to reflect this alternative specifier format.

### DIFF
--- a/source/specifications/core-metadata.rst
+++ b/source/specifications/core-metadata.rst
@@ -539,19 +539,21 @@ Requires-Dist (multiple use)
    The field format specification was relaxed to accept the syntax used by
    popular publishing tools.
 
-Each entry contains a string naming some other distutils
-project required by this distribution.
+Each entry contains a string naming some other distribution required by this
+distribution.
 
-The format of a requirement string contains from one to four parts:
+A requirement string has one mandatory and three optional parts.
 
 * A project name, in the same format as the ``Name:`` field.
   The only mandatory part.
-* A comma-separated list of 'extra' names. These are defined by
-  the required project, referring to specific features which may
-  need extra dependencies. The names MUST conform to the restrictions
-  specified by the ``Provides-Extra:`` field.
-* A version specifier. Tools parsing the format should accept optional
-  parentheses around this, but tools generating it should not use
+* A comma-separated list of 'extra' names in square brackets, like
+  ``[extra1,extra2]``. These names are defined by the required
+  project, referring to specific features which may need extra
+  dependencies. The names MUST conform to the restrictions specified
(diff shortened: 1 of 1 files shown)


ncoghlan:
This section should reference https://packaging.python.org/en/latest/specifications/dependency-specifiers/ rather PEP 508 directly.

Any mention of direct URLs should point to their definition in https://peps.python.org/pep-0440/#direct-references and note that they are not permitted on PyPI (but are fine for use in packages that are only uploaded to org specific index servers).

The description of valid names should link to https://packaging.python.org/en/latest/specifications/name-normalization/#name-format

godlygeek:
I believe I've addressed all of your feedback @ncoghlan, please take another look

Kwpumpa (approved):

Kwpumpa (approved):

source/specifications/core-metadata.rst
```diff
@@ -539,31 +539,32 @@ Requires-Dist (multiple use)
...
+  :ref:`direct reference <direct-references>`. Tools parsing the format
+  should accept optional parentheses around a version specifier, but
+  tools generating it should not use parentheses. If a distribution
+  requires a direct reference it cannot be uploaded to PyPI, but private
+  repositories may allow such distributions.
 * An environment marker after a semicolon. This means that the
   requirement is only needed in the specified conditions.
 
-See :pep:`508` for full details of the allowed format.
+See :ref:`dependency-specifiers` for full details of the allowed format.
```
webknjaz, line 563: It's rendered as capitalized by default:
```suggestion
See :ref:`dependency specifiers <dependency-specifiers>` for full details of the allowed format.
```

source/specifications/core-metadata.rst
```diff
@@ -539,31 +539,32 @@ Requires-Dist (multiple use)
...
+
+A requirement string has one mandatory and three optional parts.
+
+* A project name, in :ref:`the same format <name-format>`
+  as the ``Name:`` field. The only mandatory part.
+* A comma-separated list of 'extra' names in square brackets, like
+  ``[extra1,extra2]``. These names are defined by the required
+  project, referring to specific features which may need extra
+  dependencies. The names MUST conform to the restrictions specified
+  by the ``Provides-Extra:`` field.
```
webknjaz, line 553: Opportunity to link https://python-packaging-user-guide--1783.org.readthedocs.build/en/1783/specifications/core-metadata/#provides-extra-multiple-use ?

Issue #24, opened 2014-02-01
Add an Example Deployment Strategies section


brettcannon:
The section would give people a launching-off point for solving their own deployment needs. I envision covering 3 types of deployments, but always using a venv and unpacking everything:
1. Always pull from PyPI (e.g. push a requirements file and then running pip on it)
2. Pulling from a private wheel repo (same as above but not hitting PyPI to save on speed/bandwidth)
3. cp a single wheel/zipfile that is either unzipped in-place or pip installs (e.g. what Twitter does to deploy code to not even have to reach across the network to a single repo server)

That covers the gamut of needs from a small deployment to internet-scale. All solutions should be cross-platform (e.g. avoid symlinks) so as to keep them simple and avoid the caveats.

Issue #22, opened 2014-01-27
No search hit for "requirements.txt"


guettli:
There is no search hit for "requirements.txt". The word is in the docs, but some how I does not appear if you search for it.

See: http://python-packaging-user-guide.readthedocs.org/en/latest/search.html?q=requirements.txt&check_keywords=yes&area=default#

PR #1791, opened 2025-01-27
A discussion on helping downstream packaging

mgorny:
A discussion aimed at explaining how downstream packaging works, what challenges are the packagers facing and how they can be helped.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1791.org.readthedocs.build/en/1791/

### DIFF
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -115,12 +115,16 @@ source archive for downstream use, for example by attaching it to a GitHub
 release. Alternatively, large files, such as test data, can be split into
 separate archives.
 
-A good idea is to **use your source distribution in the release workflow**.
-That is, build it first, then unpack it and perform all the remaining steps
-using the unpacked distribution rather than the Git repository — run tests,
-build documentation, build wheels. This ensures that it is well-tested,
-and reduces the risk that some users would hit build failures or install
-an incomplete package.
+A good idea is to use your source distribution in the release workflow.
+For example, the :ref:`build` tool does exactly that — it first builds a source
+distribution, and then uses it to build a wheel. This ensures that the source
+distribution actually works, and that it won't accidentally install fewer files
+than the official wheels.
+
+Ideally, use the source distribution also run tests, build documentation,
+and so on, or add specific tests to make sure that all necessary files were
+actually included. Understandably, this requires more effort, so it's fine
+not do that — downstream packagers will report any missing files promptly.
 
 
 .. _no-internet-access-in-builds:
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -21,6 +21,8 @@ and what additional challenges downstream packagers typically face. It aims
 to provide some optional guidelines that project maintainers may choose to
 follow which help make downstream packaging *significantly* easier
 (without imposing any major maintenance hassles on the upstream project).
+Note that this is not an all-or-nothing proposal — anything that upstream
+maintainers can do is useful, even if it's only a small part.
 
 Establishing a good relationship between software maintainers and downstream
 packagers can bring mutual benefits. Downstreams are often willing to share
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -367,6 +367,71 @@ Some specific suggestions are:
   test discovery.
 
 
+.. _aim-for-stable-releases:
+
+Aim for stable releases
+-----------------------
+
+Why?
+~~~~
+
+Many downstreams provide stable release channels in addition to the main
+package streams. The goal of these channels is to provide more conservative
+upgrades to users with higher stability needs. These users often prefer
+to trade having the newest features available for lower risk of issues.
+
+While the exact policies differ, an important criterion for including a new
+package version in a stable release channel is for it to be available in testing
+for some time already, and have no known major regressions. For example,
+in Gentoo Linux a package is usually marked stable after being available
+in testing for a month, and being tested against the versions of its
+dependencies that are marked stable at the time.
+
+However, there are circumstances which demand more prompt action. For example,
+if a security vulnerability or a major bug is found in the version that is
+currently available in the stable channel, the downstream is facing a need
+to resolve it. In this case, they need to consider various options, such as:
+
+- putting a new version in the stable channel early,
+
+- adding patches to the version currently published,
+
+- or even downgrading the stable channel to an earlier release.
+
+Each of these options involves certain risks and a certain amount of work,
+and packagers needs to weigh them to determine the course of action.
+
+How?
+~~~~
+
+There are some things that upstreams can do to tailor their workflow to stable
+release channels. These actions often are beneficial to the package's users
+as well. Some specific suggestions are:
+
+- Adjust the release frequency to the rate of code changes. Packages that
+  are released rarely often bring significant changes with every release,
+  and a higher risk of accidental regressions.
+
+- Avoid mixing bug fixes and new features, if possible. In particular, if there
+  are known bug fixes merged already, consider making a new release before
+  merging feature branches.
+
+- Consider making prereleases after major changes, to provide more testing
+  opportunities for users and downstreams willing to opt-in.
+
+- If your project is subject to very intense development, consider splitting
+  one or more branches that include a more conservative subset of commits,
+  and are released separately. For example, Django_ currently maintains three
+  release branches in addition to main.
+
+- Even if you don't wish to maintain additional branches permanently, consider
+  making additional patch releases with minimal changes to the previous
+  version, especially when a security vulnerability is discovered.
+
+- Split your changes into focused commits that address one problem at a time,
+  to make it easier to cherry-pick changes to earlier releases when necessary.
+
+
 .. _responses: https://pypi.org/project/responses/
 .. _vcrpy: https://pypi.org/project/vcrpy/
 .. _pytest-socket: https://pypi.org/project/pytest-socket/
@@ -374,3 +439,4 @@ Some specific suggestions are:
 .. _pytest: https://pytest.org/
 .. _pytest-rerunfailures: https://pypi.org/project/pytest-rerunfailures/
 .. _pytest-timeout: https://pypi.org/project/pytest-timeout/
+.. _Django: https://www.djangoproject.com/
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -1,8 +1,8 @@
 .. _downstream-packaging:
 
-=================================
-Simplifying downstream packaging
-=================================
+===============================
+Supporting downstream packaging
+===============================
 
 :Page Status: Draft
 :Last Reviewed: 2025-?
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -55,19 +55,19 @@ rather than use the upstream-provided binary packages. This is also true
 of pure Python packages that provide universal wheels. The reasons for using
 source distributions may include:
 
-- being able to audit the source code of all packages
+- Being able to audit the source code of all packages.
 
-- being able to run the test suite and build documentation
+- Being able to run the test suite and build documentation.
 
-- being able to easily apply patches, including backporting commits
-  from the project's repository and sending patches back to the project
+- Being able to easily apply patches, including backporting commits
+  from the project's repository and sending patches back to the project.
 
-- being able to build on a specific platform that is not covered
-  by upstream builds
+- Being able to build on a specific platform that is not covered
+  by upstream builds.
 
-- being able to build against specific versions of system libraries
+- Being able to build against specific versions of system libraries.
 
-- having a consistent build process across all Python packages
+- Having a consistent build process across all Python packages.
 
 While it is usually possible to build packages from a Git repository, there are
 a few important reasons to provide a static archive file instead:
@@ -205,11 +205,7 @@ However, none of these issues apply to downstream packaging, and downstreams
 have good reasons to prefer dynamically linking to system dependencies.
 In particular:
 
-- in many cases, reliably sharing dynamic dependencies between components is a large part
-  of the *purpose* of a downstream packaging ecosystem. Helping to support that makes it
-  easier for users of those systems to access upstream projects in their preferred format.
-
-- in many cases, reliably sharing dynamic dependencies between components is a large part
+- In many cases, reliably sharing dynamic dependencies between components is a large part
   of the *purpose* of a downstream packaging ecosystem. Helping to support that makes it
   easier for users of those systems to access upstream projects in their preferred format.
 
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -233,41 +233,67 @@ There is a number of things that you can do to help us test your package
 better. Some of them were already mentioned in this discussion. Some examples
 are:
 
-- Include the test files and fixtures in the source distribution, or make it
+- **Include the test files and fixtures in the source distribution**, or make it
   possible to easily download them separately.
 
-- Do not write to the package during testing. Downstream test setups sometimes
-  run tests on top of the installed package, and test-time modifications can
-  end up being part of the production package!
+- **Do not write to the package directories during testing.** Downstream test
+  setups sometimes run tests on top of the installed package, and modifications
+  performed during testing and temporary test files may end up being part
+  of the installed package!
 
-- Make the test suite work offline. Mock network interactions, using packages
-  such as responses_ or vcrpy_. If that is not possible, make it possible
-  to easily disable the tests using Internet access, e.g. via a pytest marker.
-  Use pytest-socket_ to verify that your tests work offline.
+- **Make the test suite work offline.** Mock network interactions, using
+  packages such as responses_ or vcrpy_. If that is not possible, make it
+  possible to easily disable the tests using Internet access, e.g. via a pytest
+  marker.  Use pytest-socket_ to verify that your tests work offline. This
+  often makes your own test workflows faster and more reliable as well.
 
-- Make your tests work without a specialized setup, or perform the necessary
+- **Make your tests work without a specialized setup**, or perform the necessary
   setup as part of test fixtures. Do not ever assume that you can connect
   to system services such as databases — in an extreme case, you could crash
   a production service!
 
-- Do not assume that the test suite will be run with ``-Werror``. Downstreams
+- **If your package has optional dependencies, make their tests optional as
+  well.** Either skip them if the needed packages are not installed, or add
+  markers to make deselecting easy.
+
+- More generally, **add markers to tests with special requirements**. These can
+  include e.g. significant space usage, significant memory usage, long runtime,
+  incompatibility with parallel testing.
+
+- **Do not assume that the test suite will be run with -Werror.** Downstreams
   often need to disable that, as it causes false positives, e.g. due to newer
   dependency versions. Assert for warnings using ``pytest.warns()`` rather
   than ``pytest.raises()``!
 
-- Aim to make your test suite reliable. Avoid flaky tests. Avoid depending
-  on specific platform details, don't rely on exact results of floating-point
-  computation, or timing of operations, and so on. Fuzzing has its advantages,
-  but you want to have static test cases for completeness as well.
+- **Aim to make your test suite reliable and reproducible.** Avoid flaky tests.
+  Avoid depending on specific platform details, don't rely on exact results
+  of floating-point computation, or timing of operations, and so on. Fuzzing
+  has its advantages, but you want to have static test cases for completeness
+  as well.
 
-- Split tests by their purpose, and make it easy to skip categories that are
-  irrelevant or problematic. Since the primary purpose of downstream testing is
-  to ensure that the package itself works, we generally are not interested
+- **Split tests by their purpose, and make it easy to skip categories that are
+  irrelevant or problematic.** Since the primary purpose of downstream testing
+  is to ensure that the package itself works, we generally are not interested
   in e.g. checking code coverage, code formatting, typing or running
   benchmarks. These tests can fail as dependencies are upgraded or the system
   is under load, without actually affecting the package itself.
+
+- If your test suite takes significant time to run, **support testing
+  in parallel.** Downstreams often maintain a large number of packages,
+  and testing them all takes a lot of time. Using pytest-xdist_ can help them
+  avoid bottlenecks.
+
+- Ideally, **support running your test suite via PyTest**. PyTest_ has many
+  command-line arguments that are truly helpful to downstreams, such as
+  the ability to conveniently deselect tests, rerun flaky tests
+  (via pytest-rerunfailures_), add a timeout to prevent tests from hanging
+  (via pytest-timeout_) or run tests in parallel (via pytest-xdist_).
 
 
 .. _responses: https://pypi.org/project/responses/
 .. _vcrpy: https://pypi.org/project/vcrpy/
 .. _pytest-socket: https://pypi.org/project/pytest-socket/
+.. _pytest-xdist: https://pypi.org/project/pytest-xdist/
+.. _pytest: https://pytest.org/
+.. _pytest-rerunfailures: https://pypi.org/project/pytest-rerunfailures/
+.. _pytest-timeout: https://pypi.org/project/pytest-timeout/
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -13,12 +13,22 @@ of other packaging ecosystems. These repackaging efforts are collectively called
 *downstream* packaging (your own efforts are called *upstream* packaging),
 and include such projects as Linux distributions, Conda, Homebrew and MacPorts.
 They often aim to provide good support for use cases that cannot be handled
-via Python packaging tools alone, such as good integration with non-Python
-software.
+via Python packaging tools alone, such as **good integration with non-Python
+software**.
 
 This discussion attempts to explain how downstream packaging is usually done,
 and what challenges are downstream packagers facing. It ultimately aims to give
-you some hints on how you can make downstream packaging easier.
+you some hints on how you can **make downstream packaging easier**.
+
+Establishing a good relationship between software maintainers and downstream
+packagers can bring mutual benefits. Downstreams are often willing to **share
+their experience, time and hardware** to improve your package. They are
+sometimes in a better position to see **the bigger picture**, and to provide
+you with **information about other packages** that would otherwise require you
+to put significant effort to obtain. Packagers often can **find bugs** before
+your users hit them on production, **provide bug reports of good quality**
+and **supply patches** whenever they can. For example, they are often
+in the vanguard when **a new Python version** comes out.
 
 Please note that downstream builds include not only binary redistribution,
 but also source builds done on user systems, in source-first distributions
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -136,9 +136,12 @@ Why?
 ~~~~
 
 Downstream builds are frequently done in sandboxed environments that cannot
-access the Internet. Even if this is not the case, and assuming that you took
-sufficient care to properly authenticate downloads, using the Internet
-is discouraged for a number of reasons:
+access the Internet. The package sources are unpacked into this environment,
+and all the necessary dependencies are installed.
+
+Even if this is not the case, and assuming that you took sufficient care to
+properly authenticate downloads, using the Internet is discouraged for a number
+of reasons:
 
 - The Internet connection may be unstable (e.g. due to poor reception)
   or suffer from temporary problems that could cause the process to fail
@@ -159,21 +162,30 @@ is discouraged for a number of reasons:
 How?
 ~~~~
 
-Your source distribution should either include all the files needed
-for the package to build, or allow provisioning them externally. Ideally,
-it should not even attempt to access the Internet at all, unless explicitly
-requested to. If that is not possible to achieve, the next best thing
-is to provide an opt-out switch to disable all Internet access.
+If the package is implementing any custom build *backend* actions that use
+the Internet, for example by automatically downloading vendored dependencies
+or fetching Git submodules, its source distribution should either include all
+of these files or allow provisioning them externally, and the Internet must not
+be used if the files are already present.
 
-When such a switch is used, the build process should fail if some
-of the required files are missing, rather than try to fetch them automatically.
-This could be done e.g. by checking whether a ``NO_NETWORK`` environment
-variable is set to a non-empty value. Please also remember that if you are
-fetching remote resources, you must *verify their authenticity* (usually against
-a hash), to protect against the file being substituted by a malicious party.
+Note that this point does not apply to Python dependencies that are specified
+in the package metadata, and are fetched during the build and installation
+process by *frontends* (such as :ref:`build` or :ref:`pip`). Downstreams use
+frontends that use local provisioning for Python dependencies.
+
+Ideally, custom build scripts should not even attempt to access the Internet
+at all, unless explicitly requested to. If any resources are missing and need
+to be fetched, they should ask the user for permission first. If that is not
+feasible, the next best thing is to provide an opt-out switch to disable
+all Internet access. This could be done e.g. by checking whether
+a ``NO_NETWORK`` environment variable is set to a non-empty value.
 
 Since downstreams frequently also run tests and build documentation, the above
 should ideally extend to these processes as well.
+
+Please also remember that if you are fetching remote resources, you absolutely
+must *verify their authenticity* (usually against a hash), to protect against
+the file being substituted by a malicious party.
 
 
 .. _support-system-dependencies-in-builds:
--- a/source/discussions/downstream-packaging.rst
+++ b/source/discussions/downstream-packaging.rst
@@ -1,84 +1,96 @@
 .. _downstream-packaging:
 
-========================================
-How to make downstream packaging easier?
-========================================
+=================================
+Simplifying downstream packaging
+=================================
 
 :Page Status: Draft
 :Last Reviewed: 2025-?
 
 While PyPI and the Python packaging tools such as :ref:`pip` are the primary
-means of distributing your packages, they are often also made available as part
+means of distributing Python packages, they are also often made available as part
 of other packaging ecosystems. These repackaging efforts are collectively called
 *downstream* packaging (your own efforts are called *upstream* packaging),
 and include such projects as Linux distributions, Conda, Homebrew and MacPorts.
-They often aim to provide good support for use cases that cannot be handled
-via Python packaging tools alone, such as **good integration with non-Python
-software**.
+They generally aim to provide improved support for use cases that cannot be handled
+via Python packaging tools alone, such as native integration with a specific operating
+system, or assured compatibility with specific versions of non-Python software.
 
 This discussion attempts to explain how downstream packaging is usually done,
-and what challenges are downstream packagers facing. It ultimately aims to give
-you some hints on how you can **make downstream packaging easier**.
+and what additional challenges downstream packagers typically face. It aims
+to provide some optional guidelines that project maintainers may choose to
+follow which help make downstream packaging *significantly* easier
+(without imposing any major maintenance hassles on the upstream project).
 
 Establishing a good relationship between software maintainers and downstream
-packagers can bring mutual benefits. Downstreams are often willing to **share
-their experience, time and hardware** to improve your package. They are
-sometimes in a better position to see **the bigger picture**, and to provide
-you with **information about other packages** that would otherwise require you
-to put significant effort to obtain. Packagers often can **find bugs** before
-your users hit them on production, **provide bug reports of good quality**
-and **supply patches** whenever they can. For example, they are often
-in the vanguard when **a new Python version** comes out.
+packagers can bring mutual benefits. Downstreams are often willing to share
+their experience, time and hardware to improve your package. They are
+sometimes in a better position to see how your package is used in practice,
+and to provide information about its relationships with other packages that
+would otherwise require significant effort to obtain.
+Packagers can often find bugs before your users hit them in production,
+provide bug reports of good quality, and supply patches whenever they can.
+For example, they are regularly active in ensuring the packages they redistribute
+are updated for any compatibility issues that arise when a new Python version
+is released.
 
 Please note that downstream builds include not only binary redistribution,
-but also source builds done on user systems, in source-first distributions
-such as Gentoo Linux.
+but also source builds done on user systems (in source-first distributions
+such as Gentoo Linux, for example).
 
 
-.. _Provide complete source distributions:
+.. _provide-complete-source-distributions:
 
 Provide complete source distributions
 -------------------------------------
+
 Why?
 ~~~~
+
 The vast majority of downstream packagers prefer to build packages from source,
 rather than use the upstream-provided binary packages. This is also true
 of pure Python packages that provide universal wheels. The reasons for using
 source distributions may include:
 
-- being able to **audit the source code** of all packages
+- being able to audit the source code of all packages
 
-- being able to **run the test suite and build documentation**
+- being able to run the test suite and build documentation
 
-- being able to **easily apply patches**, including backporting commits
-  from your repository and sending patches back to you
+- being able to easily apply patches, including backporting commits
+  from the project's repository and sending patches back to the project
 
-- being able to **build on a specific platform** that is not covered
+- being able to build on a specific platform that is not covered
   by upstream builds
 
-- being able to **build against specific versions of system libraries**
+- being able to build against specific versions of system libraries
 
 - having a consistent build process across all Python packages
 
 While it is usually possible to build packages from a git repository, there are
 a few important reasons to provide a static archive file instead:
 
-- Fetching a single file is often **more efficient, more reliable and better
-  supported** than e.g. using a git clone. This can help users with a shoddy
-  Internet connection.
+- Fetching a single file is often more efficient, more reliable and better
+  supported than e.g. using a git clone. This can help users with poor
+  Internet connectivity.
 
 - Downstreams often **use checksums to verify the authenticity** of source files
   on subsequent builds, which require that they remain bitwise identical over
   time. For example, automatically generated git archives do not guarantee
   that.
 
-- Archive files can be **mirrored**, reducing both upstream and downstream
+- Archive files can be mirrored, reducing both upstream and downstream
   bandwidth use. The actual builds can afterwards be performed in firewalled
   or offline environments, that can only access source files provided
   by the local mirror or redistributed earlier.
 
+- Explicitly publishing archive files can ensure that any dependencies on version control
+  system metadata are resolved when creating the source archive. For example, automatically
+  generated git archives omit all of the commit tag information, potentially resulting in
+  incorrect version details in the resulting builds.
+
 How?
 ~~~~
+
 Ideally, **a source distribution archive should include all the files**
 necessary to build the package itself, run its test suite, build and install
 its documentation, and any other files that may be useful to end users, such as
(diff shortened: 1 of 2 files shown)


source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+However, none of these issues apply to downstream packaging, and downstreams
+have good reasons to prefer **dynamically linking to system dependencies**.
+In particular:
+
+- Static linking and vendoring obscures the use of external dependencies,
+  **making source auditing harder**.
+
+- Dynamic linking makes it possible to easily and **quickly replace the used
+  libraries**, which can be particularly important when they turn out to
+  be vulnerable or buggy.
```
ncoghlan, line 193:
```suggestion
- Dynamic linking makes it possible to quickly and systematically replace the used
  libraries across an entire downstream packaging ecosystem, which can be particularly
  important when they turn out to contain a security vulnerability or critical bug.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Using system dependencies makes the package benefit from **downstream
+  customization** that can improve the user experience on a particular platform,
+  without the downstream maintainers having to consistently patch
+  the dependencies vendored in different packages. This can include
+  **compatibility improvements and security hardening**.
```
ncoghlan, line 199:
```suggestion
- Using system dependencies makes the package benefit from downstream
  customization that can improve the user experience on a particular platform,
  without the downstream maintainers having to consistently patch
  the dependencies vendored in different packages. This can include
  compatibility improvements and security hardening.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Static linking and vendoring could result in **multiple different versions
+  of the same library being loaded in the same process** (e.g. when you use two
+  Python packages that link to different versions of the same library).
+  This can cause no problems, but it could also lead to anything from subtle
+  bugs to catastrophic failures.
```
ncoghlan, line 205:
```suggestion
- Static linking and vendoring can result in multiple different versions of the
  same library being loaded in the same process (for example, attempting to
  import two Python packages that link to different versions of the same library).
  This sometimes works without incident, but it can also lead to anything from library
  loading errors, to subtle runtime bugs, to catastrophic system failures.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Last but not least, static linking and vendoring results in duplication,
+  and may increase the **use of both the disk space and memory**.
```
ncoghlan, line 208:
```suggestion
- Last but not least, static linking and vendoring results in duplication,
  and may increase the use of both disk space and memory.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+How?
+~~~~
```
ncoghlan, line 211:
```suggestion
~~~~

```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+A good compromise between the needs of both parties is to **provide a switch
+between using vendored and system dependencies**. Ideally, if the package has
```
ncoghlan, line 213:
```suggestion
A good compromise between the needs of both parties is to provide a switch
between using vendored and system dependencies. Ideally, if the package has
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+multiple vendored dependencies, it should provide both individual switches
+for each dependency, and a general switch to control the default for them,
+e.g. via a ``USE_SYSTEM_DEPS`` environment variable.
+
+If the user requests using system dependencies, and **a particular dependency
+is either missing or incompatible, the build should fail** with an explanatory
```
ncoghlan, line 219:
```suggestion
If the user requests using system dependencies, and a particular dependency
is either missing or incompatible, the build should fail with an explanatory
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+message rather than fall back to a vendored version. This gives the packager
+the opportunity to notice their mistake and a chance to consciously decide
+how to solve it.
+
```
ncoghlan, line 223:
```suggestion

Note that it is reasonable for upstream projects to leave *testing* of building with
system dependencies to their downstream repackagers. The goal of these guidelines
is to facilitate more effective collaboration between upstream projects and downstream
repackagers, not to suggest upstream projects take on tasks that downstream repackagers
are better equipped to handle.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+.. _Support downstream testing:
```
ncoghlan, line 225:
```suggestion
.. _support-downstream-testing:
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+Support downstream testing
+--------------------------
```
ncoghlan, line 228:
```suggestion
--------------------------

```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+Why?
+~~~~
```
ncoghlan, line 230:
```suggestion
~~~~

```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+A variety of downstream projects run some degree of testing on the packaged
+Python projects. Depending on the particular case, this can range from minimal
+smoke testing to comprehensive runs of the complete test suite. There can
+be various reasons for doing this, for example:
+
+- Verifying that the downstream **packaging did not introduce any bugs**.
```
ncoghlan, line 236:
```suggestion
- Verifying that the downstream packaging did not introduce any bugs.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Testing on **additional platforms** that are not covered by upstream testing.
```
ncoghlan, line 238:
```suggestion
- Testing on additional platforms that are not covered by upstream testing.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Finding subtle bugs that can only be reproduced on a **particular hardware,
+  system package versions**, and so on.
```
ncoghlan, line 241:
```suggestion
- Finding subtle bugs that can only be reproduced with particular hardware,
  system package versions, and so on.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Testing the released package against **newer dependency versions** than
+  the ones present during upstream release testing.
```
ncoghlan, line 244:
```suggestion
- Testing the released package against newer (or older) dependency versions than
  the ones present during upstream release testing.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- Testing the package in an environment closely resembling **the production
+  setup**. This can detect issues caused by nontrivial interactions between
```
ncoghlan, line 247:
```suggestion
- Testing the package in an environment closely resembling the production
  setup. This can detect issues caused by nontrivial interactions between
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  different installed packages, including packages that are not dependencies
+  of your package, but nevertheless can cause issues.
+
+- Testing the released package against **newer Python versions** (including
```
ncoghlan, line 251:
```suggestion
- Testing the released package against newer Python versions (including
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  newer point releases), or less tested Python implementations such as PyPy.
+
+Admittedly, sometimes downstream testing may yield false positives or
+inconvenience you about scenarios that you are not interested in supporting.
```
ncoghlan, line 255:
```suggestion
Admittedly, sometimes downstream testing may yield false positives or bug
reports about scenarios the upstream project is not interested in supporting.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+However, perhaps even more often it does provide early notice of problems,
+or find nontrivial bugs that would otherwise cause issues for your users
+in production. And believe me, the majority of **downstream packagers are doing
+their best to double-check their results, and help you triage and fix the bugs
+that they report**.
```
ncoghlan, line 260:
```suggestion
However, perhaps even more often it does provide early notice of problems,
or find non-trivial bugs that would otherwise cause issues for the upstream
project's users. While mistakes do happen, the majority of downstream packagers
are doing their best to double-check their results, and help their upstream projects
triage and fix the bugs that they report.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+How?
+~~~~
```
ncoghlan, line 263:
```suggestion
~~~~

```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+There is a number of things that you can do to help us test your package
+better. Some of them were already mentioned in this discussion. Some examples
+are:
+
+- **Include the test files and fixtures in the source distribution**, or make it
+  possible to easily download them separately.
```
ncoghlan, line 269:
```suggestion
- Include the test files and fixtures in the source distribution, or make it
  possible to easily download them separately.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+or find nontrivial bugs that would otherwise cause issues for your users
+in production. And believe me, the majority of **downstream packagers are doing
+their best to double-check their results, and help you triage and fix the bugs
+that they report**.
+
+How?
+~~~~
+There is a number of things that you can do to help us test your package
+better. Some of them were already mentioned in this discussion. Some examples
+are:
```
ncoghlan, line 266:
```suggestion
There are a number of things that upstream projects can do to help downstream
repackagers test their packages efficiently and effectively, including some of the suggestions
already mentioned above. These are typically improvements that make the test suite more
reliable and easier to use for everyone, not just downstream packagers.
Some specific suggestions are:
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- **Include the test files and fixtures in the source distribution**, or make it
+  possible to easily download them separately.
+
+- **Do not write to the package directories during testing.** Downstream test
```
ncoghlan, line 271:
```suggestion
- Do not write to the package directories during testing. Downstream test
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  setups sometimes run tests on top of the installed package, and modifications
+  performed during testing and temporary test files may end up being part
+  of the installed package!
+
+- **Make the test suite work offline.** Mock network interactions, using
```
ncoghlan, line 276:
```suggestion
- Make the test suite work offline. Mock network interactions, using
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  packages such as responses_ or vcrpy_. If that is not possible, make it
+  possible to easily disable the tests using Internet access, e.g. via a pytest
```
ncoghlan, line 278:
```suggestion
  possible to easily disable the tests using Internet access, e.g. via a pytest_
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  marker.  Use pytest-socket_ to verify that your tests work offline. This
+  often makes your own test workflows faster and more reliable as well.
+
+- **Make your tests work without a specialized setup**, or perform the necessary
```
ncoghlan, line 282:
```suggestion
- Make your tests work without a specialized setup, or perform the necessary
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  setup as part of test fixtures. Do not ever assume that you can connect
+  to system services such as databases — in an extreme case, you could crash
+  a production service!
+
+- **If your package has optional dependencies, make their tests optional as
+  well.** Either skip them if the needed packages are not installed, or add
```
ncoghlan, line 288:
```suggestion
- If your package has optional dependencies, make their tests optional as
  well. Either skip them if the needed packages are not installed, or add
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  markers to make deselecting easy.
+
+- More generally, **add markers to tests with special requirements**. These can
```
ncoghlan, line 291:
```suggestion
- More generally, add markers to tests with special requirements. These can
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  include e.g. significant space usage, significant memory usage, long runtime,
+  incompatibility with parallel testing.
+
+- **Do not assume that the test suite will be run with -Werror.** Downstreams
```
ncoghlan, line 295:
```suggestion
- Do not assume that the test suite will be run with ``-Werror``. Downstreams
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  often need to disable that, as it causes false positives, e.g. due to newer
+  dependency versions. Assert for warnings using ``pytest.warns()`` rather
+  than ``pytest.raises()``!
+
+- **Aim to make your test suite reliable and reproducible.** Avoid flaky tests.
```
ncoghlan, line 300:
```suggestion
- Aim to make your test suite reliable and reproducible. Avoid flaky tests.
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  Avoid depending on specific platform details, don't rely on exact results
+  of floating-point computation, or timing of operations, and so on. Fuzzing
+  has its advantages, but you want to have static test cases for completeness
+  as well.
+
+- **Split tests by their purpose, and make it easy to skip categories that are
+  irrelevant or problematic.** Since the primary purpose of downstream testing
```
ncoghlan, line 307:
```suggestion
- Split tests by their purpose, and make it easy to skip categories that are
  irrelevant or problematic. Since the primary purpose of downstream testing
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  is to ensure that the package itself works, we generally are not interested
+  in e.g. checking code coverage, code formatting, typing or running
+  benchmarks. These tests can fail as dependencies are upgraded or the system
+  is under load, without actually affecting the package itself.
```
ncoghlan, line 311:
```suggestion
  is to ensure that the package itself works, downstreams are not generally interested
  in tasks such as checking code coverage, code formatting, typechecking or running
  benchmarks. These tests can fail as dependencies are upgraded or the system
  is under load, without actually affecting the package itself.
```

ncoghlan: I left the examples of skipped checks as is in my edit, but skipping typechecking when upgrading dependencies actually seems a bit dubious - typecheckers will complain if an upgrade removes or changes interfaces in a dependency that a project accesses from lines that the test suite doesn't cover.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+- If your test suite takes significant time to run, **support testing
+  in parallel.** Downstreams often maintain a large number of packages,
```
ncoghlan, line 314:
```suggestion
- If your test suite takes significant time to run, support testing
  in parallel. Downstreams often maintain a large number of packages,
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  and testing them all takes a lot of time. Using pytest-xdist_ can help them
+  avoid bottlenecks.
+
+- Ideally, **support running your test suite via pytest**. pytest_ has many
```
ncoghlan, line 318:
```suggestion
- Ideally, support running your test suite via ``pytest``. pytest_ has many
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  command-line arguments that are truly helpful to downstreams, such as
+  the ability to conveniently deselect tests, rerun flaky tests
+  (via pytest-rerunfailures_), add a timeout to prevent tests from hanging
+  (via pytest-timeout_) or run tests in parallel (via pytest-xdist_).
```
ncoghlan, line 322:
```suggestion
  (via pytest-timeout_) or run tests in parallel (via pytest-xdist_).
  Note that test suites don't need to be *written* with ``pytest`` to be *executed* with
  ``pytest``: ``pytest`` is able to find and execute any test cases that are compatible with
  the standard library's ``unittest`` test discovery.
```

ncoghlan (commented):
Thanks for this!

*Lots* of comments inline, but many of them are just mechanical and stylistic things:

* suggested adding blank lines after the headings
* suggested changing the heading anchors to use-shorter-phrases-with-hyphens rather than repeating the titles
* suggested dropping most of the bold emphasis from the text (mostly dropped entirely, occasionally scaled back to just italics rather than bold)
* suggested rewording for anything containing the pronouns "you" or "your" (it's really hard to make second person pronouns sound non-confrontational in English, so it's simpler to avoid them entirely)
* suggested rewording anything containing the pronouns "we" or "us" or "I" (as a shared community documentation project, first person pronouns sometimes make sense as a way of referring to Python developers or project contributors in general, but we don't want to use them to refer to specific community subsets)
* took a close look at anything containing other pronouns (third party pronouns are generally fine, but the text needs to be clear on the intended reference)

There are a few more substantial comments mixed in:

* suggesting a different discussion title
* one case where I think a comment in the text isn't accurate (for projects that support typechecking, downstreams probably *should* be running a static typecheck to check for potential problems with dependency upgrades)
* a couple of cases where I suggested added a new bullet point to the rationale for a section 
* a few cases of more substantial suggested changes to the wording of various points, or adding entirely new paragraphs to the text

mgorny:
Thanks for your feedback.

> * suggested dropping most of the bold emphasis from the text (mostly dropped entirely, occasionally scaled back to just italics rather than bold)

Not sure about this. It's been pointed out to me that there are people who would like to skim the text rather than reading it in detail. Bold makes it easy to find the key points, especially if you've read it once and now just want to quickly consult.

> * one case where I think a comment in the text isn't accurate (for projects that support typechecking, downstreams probably should be running a static typecheck to check for potential problems with dependency upgrades)

To be honest, type checking has never caused us anything but trouble. I suppose type checking may make sense if there is a poor test coverage. However, if the test coverage is reasonably good, any real breakage affecting users should be detected through regular tests. Type checking, on the other hand, will generally report things that are "technically wrong", but do not really affect affect end users (i.e. there is no hurry to backport fixes to stable versions of a package). On top of that, MyPy is being constantly improved, so packages that passed typing tests in the past, start failing with newer MyPy versions.

Hmm, perhaps I should also mentioned the existence of "stable" channels somewhere.

mgorny:
> * suggested rewording for anything containing the pronouns "you" or "your" (it's really hard to make second person pronouns sound non-confrontational in English, so it's simpler to avoid them entirely)

Perhaps https://packaging.python.org/en/latest/contribute/#conventions-and-mechanics needs to be updated for that, because using "you" is not my usual style, and I've done that specifically to follow the guidelines.

source/discussions/downstream-packaging.rst
mgorny: Not sure if "system failures" won't be misleading here. My point was more like risk of data loss.

source/discussions/downstream-packaging.rst
mgorny: I'm okay with changing this, but my idea was to avoid "why aren't you using up-to-date packages?" argument.

source/discussions/downstream-packaging.rst
mgorny: I think "their" is mixing subjects here at the end, or at least "downstream packagers' upstream projects" sounds weird to me.

source/discussions/downstream-packaging.rst
mgorny: That's very nice, thanks!

source/discussions/downstream-packaging.rst
mgorny: Oh, I see that you've clarified here it's about uncovered tests. I've already replied to the top-level comment.

source/discussions/downstream-packaging.rst
mgorny: I think this might be technically incorrect — I think pytest has some limitations / deprecations, but right now I'm not 100% sure if there were related only to nosetests leftovers, or if they included some "technically valid" unittest stuff as well.

mgorny (commented):
Thanks for your review. I've applied most of the suggestions, left a few with comments.

source/discussions/downstream-packaging.rst
mgorny: https://docs.pytest.org/en/7.1.x/how-to/unittest.html says "almost all", so I've just added "almost" to your text.

mgorny (commented):
Okay, I've attempted to address the remaining suggestions (modulo the one about Rust), taking my comments into consideration. Hope that addresses your concerns. I'm going to add one more section now, regarding release stability.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,442 @@
+.. _downstream-packaging:
+
+=================================
+Simplifying downstream packaging
```
mgorny, line 4: Hmm, thinking about it, I'm not sure if "simplifying" is really the correct term. It suggests the discussion would be about how to change downstream packaging to make it simpler. Perhaps "helping" would be a better term?

pfmoore: "Supporting"?

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,444 @@
...
+between using vendored and system dependencies. Ideally, if the package has
+multiple vendored dependencies, it should provide both individual switches
+for each dependency, and a general switch to control the default for them,
+e.g. via a ``USE_SYSTEM_DEPS`` environment variable.
+
+If the user requests using system dependencies, and a particular dependency
+is either missing or incompatible, the build should fail with an explanatory
+message rather than fall back to a vendored version. This gives the packager
+the opportunity to notice their mistake and a chance to consciously decide
+how to solve it.
```
abravalheri, line 251: This seems like a big ask, might not be viable maintenance from the perspective of the devs.

I think that in order to do this we need more substantial changes in terms of developing the infrastructure to allow it (e.g. feature from the importlib/import machinery directly).

mgorny: I'm sorry but I don't understand how importlib would help when linking to system libraries.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,445 @@
...
+  and may increase the use of both disk space and memory.
+
+How?
+~~~~
+
+A good compromise between the needs of both parties is to provide a switch
+between using vendored and system dependencies. Ideally, if the package has
+multiple vendored dependencies, it should provide both individual switches
+for each dependency, and a general switch to control the default for them,
+e.g. via a ``USE_SYSTEM_DEPS`` environment variable.
```
konstin, line 245: I would prefer something that is part of the PEP 517 interface instead of using environment variables.

There should also be some discussion about how PEP 517 build systems forward this to their native build systems, e.g. how do pyo3 users setup their `Cargo.toml` so that maturin can use this switch?

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+
+
+.. _Support building against system dependencies:
+
+Support building against system dependencies
+--------------------------------------------
+Why?
+~~~~
+Some Python projects have non-Python dependencies, such as libraries written
+in C or C++. Trying to use the system versions of these dependencies
```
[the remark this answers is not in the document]
konstin: cryptography and pydantic are the https://github.com/pypa/packaging.python.org/issues/22 and https://github.com/pypa/packaging.python.org/issues/24 most downloaded packages on pypi ([hugovk.github.io/top-pypi-packages](https://hugovk.github.io/top-pypi-packages/)) and both use a shared library written in rust. In my experience, maturin, uv and py-spy being rust binaries are the exception. Usually, they can dynamically or statically link system dependencies written in C or C++, but the vendoring concerns apply to them too.

source/discussions/downstream-packaging.rst
abravalheri: The impression that I had when reading the text is that it emphasizes on linking but it does mention dependencies in general. For vendored dependencies that compile a Python extension, I believe the importlib/import machinery is relevant.

If the concern is exclusive to linking, then I agree with you. Maybe the text could clarify the intention?

mgorny: I focused specifically on external C/C++ dependencies, as mention on the top. I didn't want to get into details of Python dependencies — especially that vendoring Python dependencies is rather rare.

abravalheri: Does it make sense to have as title `Support linking against binary system libraries`?

source/discussions/downstream-packaging.rst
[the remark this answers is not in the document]
mgorny: > cryptography and pydantic are the https://github.com/pypa/packaging.python.org/issues/22 and https://github.com/pypa/packaging.python.org/issues/24 most downloaded packages on pypi ([hugovk.github.io/top-pypi-packages](https://hugovk.github.io/top-pypi-packages/)) and both use a shared library written in rust.

I'm confused now. Are we talking of a system library written in Rust, or a library built internally?

I feel like we're splitting hairs here. I'm not saying there could be no relevant Rust dependency. What I'm saying is that adding "Rust" here is likely to add more confusion than gain, and these are just a few examples, not an exhaustive list of programming languages used to produce shared libraries.

source/discussions/downstream-packaging.rst
mgorny: Perhaps. What I really wanted to convey here that there could be dependencies that aren't technically binary libraries, such as Eigen (a header-only library), or tools like CMake.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,445 @@
...
+Some projects are concerned about increasing the size of source distribution,
+or do not wish Python packaging tools to fall back to source distributions
+automatically.  In these cases, a good compromise may be to publish a separate
+source archive for downstream use, for example by attaching it to a GitHub
+release. Alternatively, large files, such as test data, can be split into
+separate archives.
+
+A good idea is to **use your source distribution in the release workflow**.
+That is, build it first, then unpack it and perform all the remaining steps
+using the unpacked distribution rather than the git repostiry — run tests,
```
pawamoy, line 111:
```suggestion
using the unpacked distribution rather than the git repository — run tests,
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,445 @@
...
+
+How?
+~~~~
+
+Ideally, **a source distribution archive should include all the files**
+necessary to build the package itself, run its test suite, build and install
+its documentation, and any other files that may be useful to end users, such as
+shell completions, editor support files, and so on.
+
+Some projects are concerned about increasing the size of source distribution,
```
pawamoy, line 102:
```suggestion
Some projects are concerned about increasing the size of source distributions,
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,445 @@
...
+have good reasons to prefer dynamically linking to system dependencies.
+In particular:
+
+- in many cases, reliably sharing dynamic dependencies between components is a large part
+  of the *purpose* of a downstream packaging ecosystem. Helping to support that makes it
+  easier for users of those systems to access upstream projects in their preferred format.
+
+- in many cases, reliably sharing dynamic dependencies between components is a large part
+  of the *purpose* of a downstream packaging ecosystem. Helping to support that makes it
+  easier for users of those systems to access upstream projects in their preferred format.
```
pawamoy, line 214: Duplicated item?

pawamoy (commented):
Other general suggestions:

- write "Git" instead of "git" everywhere 
- use consistent casing in bullet points: sentence starts with capital letter, or not

Very nice document! Super interesting and helpful :)

source/discussions/downstream-packaging.rst
mgorny: Good catch, tanks. Must've accidentally got duplicated when merging changes.

mgorny:
Hopefully this version is clearer. In particular, I've tried to emphasize that we're talking about explicitly fetching stuff such as vendored dependencies, and not installing dependencies as done by a build frontend.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,466 @@
...
+release. Alternatively, large files, such as test data, can be split into
+separate archives.
+
+A good idea is to use your source distribution in the release workflow.
+For example, the :ref:`build` tool does exactly that — it first builds a source
+distribution, and then uses it to build a wheel. This ensures that the source
+distribution actually works, and that it won't accidentally install fewer files
+than the official wheels.
+
+Ideally, use the source distribution also run tests, build documentation,
```
JelleZijlstra, line 124:
```suggestion
Ideally, also use the source distribution to run tests, build documentation,
```

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,466 @@
...
+This point applies only to the files belonging to the package itself.
+The downstream packaging process, much like Python package managers, will
+provision the necessary Python dependencies, system tools and external
+libraries that are needed by your package and its build scripts. However,
+the files listing these dependencies (for example, ``requirements*.txt`` files)
+should also be included, to help downstreams determine the needed dependencies,
+and check for changes in them.
+
+Some projects are concerned about increasing the size of source distributions,
+or do not wish Python packaging tools to fall back to source distributions
```
JelleZijlstra, line 112: I sort of feel this assumes too much. I know enough about Python packaging (I think!) to know that this is alluding to sdist files on PyPI (or other indexes), but I think that could made more explicit. It's probably worth explicitly saying something like "put your sdist on PyPI" (or other indexes), and linking terms like "sdist" to definitions.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,466 @@
...
+
+Even if this is not the case, and assuming that you took sufficient care to
+properly authenticate downloads, using the Internet is discouraged for a number
+of reasons:
+
+- The Internet connection may be unstable (e.g. due to poor reception)
+  or suffer from temporary problems that could cause the process to fail
+  or hang.
+
+- The remote resources may become temporarily or even permanently
```
JelleZijlstra, line 150: Also worth adding that the external resource may change, making your build not reproducible?

source/discussions/downstream-packaging.rst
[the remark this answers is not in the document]
JelleZijlstra: Agree C and C++ make more sense to mention here. The libraries could also be written in Fortran or Rust but C/C++ are by far the most common in this context.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,466 @@
...
+message rather than fall back to a vendored version. This gives the packager
+the opportunity to notice their mistake and a chance to consciously decide
+how to solve it.
+
+Note that it is reasonable for upstream projects to leave *testing* of building with
+system dependencies to their downstream repackagers. The goal of these guidelines
+is to facilitate more effective collaboration between upstream projects and downstream
+repackagers, not to suggest upstream projects take on tasks that downstream repackagers
+are better equipped to handle.
+Note that it is reasonable for upstream projects to leave *testing* of building with
```
JelleZijlstra, line 279: Paragraph is duplicated.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,466 @@
...
+for each dependency, and a general switch to control the default for them,
+e.g. via a ``USE_SYSTEM_DEPS`` environment variable.
+
+If the user requests using system dependencies, and a particular dependency
+is either missing or incompatible, the build should fail with an explanatory
+message rather than fall back to a vendored version. This gives the packager
+the opportunity to notice their mistake and a chance to consciously decide
+how to solve it.
+
+Note that it is reasonable for upstream projects to leave *testing* of building with
```
JelleZijlstra, line 274:
```suggestion
It is reasonable for upstream projects to leave *testing* of building with
```
Style nit: "Note that" generally adds no meaning, prefer fewer words to say the same thing.

JelleZijlstra (commented):
Thanks, this is very helpful!

source/discussions/downstream-packaging.rst
mgorny: Thanks. GitHub is clearly messing merging suggestions up sometimes :-(.

source/discussions/downstream-packaging.rst
mgorny: Will add, thanks.

source/discussions/downstream-packaging.rst
mgorny: Ah, good idea. I've gotten mixed up between assuming "source distribution" as PyPI specific term, and simultaneously using it to mean source archives in general.

mgorny:
@JelleZijlstra, hope this addresses your concern. I've rewritten the paragraphs to make it clearer when I'm taking of PyPI, and when of some other hosting. I've also mentioned the possibility of installing tests.

source/discussions/downstream-packaging.rst
```diff
@@ -108,20 +108,30 @@ the files listing these dependencies (for example, ``requirements*.txt`` files)
...
+Some projects have concerns related to Python package managers using source
+distributions from PyPI. They do not wish to increase their size with files
+that are not used by these tools, or they do not wish to publish source
+distributions at all, as they enable a problematic or outright nonfunctional
+fallback to building the particular project from source. In these cases, a good
+compromise may be to publish a separate source archive for downstream use
+elsewhere, for example by attaching it to a GitHub release. Alternatively,
+large files, such as test data, can be split into separate archives.
+
+On the other hand, some projects (NumPy_, for instance) decide to install tests
```
JelleZijlstra, line 120:
```suggestion
On the other hand, some projects (NumPy_, for instance) decide to include tests
```

mgorny: Not sure about this. It could be ambiguous over what "Python package" means here, in particular whether it doesn't just mean including them in the source distribution of the package.

mgorny: Perhaps "include tests in their installed packages"?

JelleZijlstra: I think that would be clearer, yes.

source/discussions/downstream-packaging.rst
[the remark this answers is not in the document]
ncoghlan: The Rust libraries also typically won't *have* system versions available to potentially link against, so agreed that it makes sense to omit Rust from the list of examples here.

source/discussions/downstream-packaging.rst
ncoghlan: Improving the suggested techniques here feels like something that is best  left to a future revision, since we don't *have* a good standardised mechanism (other than process environment variables) for passing information through to build systems.

Even `pip` itself relies on environment variables to consistently apply constraints files to isolated builds.

source/discussions/downstream-packaging.rst
ncoghlan: A potential wording improvement here might be to emphasise that this is less about folks writing these mechanisms themselves, and more about being open to PRs from downstream repackagers that add them (so the downstream redistributors aren't all having to carry a common set of patches to enable the use of system dependencies).

When submitting such PRs, the repackagers can link here to explain the purpose of the change rather than having to explain the concept from scratch themselves.

source/discussions/downstream-packaging.rst
ncoghlan: Sounds good to me!

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+a few important reasons to provide a static archive file instead:
+
+- Fetching a single file is often **more efficient, more reliable and better
+  supported** than e.g. using a git clone. This can help users with a shoddy
+  Internet connection.
+
+- Downstreams often **use checksums to verify the authenticity** of source files
+  on subsequent builds, which require that they remain bitwise identical over
+  time. For example, automatically generated git archives do not guarantee
+  that.
```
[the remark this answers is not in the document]
ncoghlan: For this use case, there's an important property that hashes offer that checksums don't: resistance against the *intentional* generation of collisions. If we're only going to mention one of the two categories, it should be hashes.

Checksums offer the property that small bit differences in the input will change the checksum, but don't offer *any* assurance that significantly different inputs will give different checksum (e.g. some checksum algorithms won't detect full byte swaps, since they're not dependent on the order of operations). Most significantly, given a current checksum and a desired checksum, it's often possible to determine what input data to insert or append to get the desired result. Still good for checking transmission integrity (since they're usually cheaper to calculate than hashes), but not trustworthy when it comes to preventing malicious substitution of artifacts.

Hash algorithms offer much stronger collision resistance guarantees, to the point that they're considered broken if there are known techniques for producing distinct inputs that emit the same hash more efficiently than would be expected given the number of bits in the hash output. This can detect both transmission errors *and* intentional substitution.

(And after writing all that, I noted the "Outdated" tag and realised the wording had already been changed accordingly, so there's no further changes needed here)

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,331 @@
...
+  or suffer from temporary problems that could cause the process to fail
+  or hang.
+
+- The remote resources may **become temporarily or even permanently
+  unavailable**, making the build no longer possible. This is especially
+  problematic when someone needs to build an old package version.
+
+- Accessing remote servers poses a **privacy** issue and a potential
+  **security** issue, as it exposes information about the system building
+  the package.
```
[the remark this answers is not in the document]
ncoghlan: Makes sense, nothing further to do here.

source/discussions/downstream-packaging.rst
ncoghlan: Fair, something like `... catastrophic failures (like losing data, or rendering the system unable to boot)` could be clearer.

source/discussions/downstream-packaging.rst
ncoghlan: Reasonable concern, but I think prioritising the "newer" case in the text will cover it (maybe that's overly optimistic, but we can always drop the parenthetical bit later if we need to)

source/discussions/downstream-packaging.rst
ncoghlan: Yeah, the wording you used instead is better than what I proposed.

source/discussions/downstream-packaging.rst
ncoghlan: Good point. Since "to typecheck, or not to typecheck" is a decision downstream packagers will want to make on a case by case basis, it's a valid entry in this particular list of examples.

ncoghlan:
There are still a few inline conversations taking place (and the overall PR is still marked as a draft), so I'm not going to merge this yet, but I wanted to note that the proposed text *is* at a point now where I would be prepared to accept and publish it (we can always make further clarifications after its initial publication, after all).

source/discussions/downstream-packaging.rst
mgorny: Well, I think we're already making it clear that we're willing to prepare patches for everything in the top section, so I don't think we should keep repeating it. However, the point about patches is very good, so I've added it to that section — also noting that duplication may result in inconsistent solutions.

mgorny:
Thanks. I think I've addressed all the feedback so far to the best of my ability, either by applying changes or by commenting. I see GitHub hasn't marked some of the suggestions resolved, even though they were applied — unless I've missed one, they should be marked resolved now. So yeah, unless others have comments, I think this is ready to land.

source/discussions/downstream-packaging.rst
pfmoore: Reiterating my point about this being a co-operative process, I think it's worth *also* talking about what downstream repackagers can do to ease the maintenance burden of taking such a PR. Because it's often not the initial PR that's the problem, but the change in workflow, or the extra knowledge needed from the maintainer, which is the real cost.

mgorny: Do you have something specific in mind?

mgorny:
@ncoghlan, gentle ping. Can it be merged now?

mgorny:
Gentle ping.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,480 @@
...
+
+.. _provide-complete-source-distributions:
+
+Provide complete source distributions
+-------------------------------------
+
+Why?
+~~~~
+
+The vast majority of downstream packagers prefer to build packages from source,
```
mcepl, line 56: “Prefer” is not enough strong word here. For most distributions (all Linux ones for sure) not having source code available is the red line, which we don’t cross over.

source/discussions/downstream-packaging.rst
```diff
@@ -0,0 +1,480 @@
...
+must *verify their authenticity* (usually against a hash), to protect against
+the file being substituted by a malicious party.
+
+
+.. _support-system-dependencies-in-builds:
+
+Support building against system dependencies
+--------------------------------------------
+
+Why?
```
mcepl, line 211: I think the order of paragraphs in the section is in the wrong order: I would first explain why dynamic linking and using system libraries is important and only then add that there might be some problems with it and static linking/vendoring may be the only solution.

mcepl (commented):
Please, consider my comments for the further improvement of this text. Generally, I like it very much! Thank you!

source/discussions/downstream-packaging.rst
mgorny: Well, you have a point here, though I need to think how to word it better. Yes, there is a fair number of Linux distributions that distribute free packages only, but definitely **not all of them**. And then, "having source available" isn't necessarily equivalent to actually using it (I mean, there are packagers who use upstream binary packages for some reason even though they could use sources). And we are talking of Python here, and for pure Python packages "wheels" supply the "source code" just as well as source distributions.

But yes, I think it's a fair point to make that for some downstreams building from source is a must. I just don't think we can use "vast majority" anymore then.

source/discussions/downstream-packaging.rst
mgorny: I think this order makes comprehension easier. It says "you may want to do it because A, however none of A apply to downstream redistribution, and there are important reasons B to provide an option not to do it".

I think the other order of "you shouldn't do it because B, but you will do it anyway because of A, however A doesn't really apply to us, so give us an option" is more back-and-forth.

ncoghlan: Given the target audience, I agree with @mgorny on this one - the current order starts by saying things the target audience (upstream developers) will typically agree with, which is then likely to put them in a more receptive mood when we explain the differences between downstream and upstream.

If we start the other way around (by covering the downstream perspective first), upstream developers will likely have already tuned out by the time we attempt to explain that we understand that "build for cross-platform portability" vs "build for platform internal consistency" drives genuinely different priorities and desirable outcomes in this space.

ncoghlan (approved):
Thanks for the nudge, as checking back on the discussion here had indeed dropped off my radar.

Let's get this published, and then potentially iterate on specific subsections from there.

mgorny:
Thank you!

PR #1794, opened 2025-01-28
add pip and conda comparison

sujaudd1n:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1794.org.readthedocs.build/en/1794/


source/guides/tool-recommendations.rst
```diff
@@ -42,7 +42,13 @@ making use of the same Python interpreter (especially on Linux).
 
 For scientific software specifically, consider :ref:`Conda` or :ref:`Spack`.
 
-.. todo:: Write a "pip vs. Conda" comparison, here or in a new discussion.
+.. note::
+
+  Pip manages Python packages, while Conda manages packages for multiple
+  languages and environments. Conda excels at dependency management and creating
+  isolated environments, but has limited Python packages. Pip is widely used,
+  but struggles with dependencies and virtual environments. Using both together
```
webknjaz, line 50: > struggles with dependencies and virtual environments

This doesn't sound true as phrased right now.

notatallshaw:
Is this PR being seriously considered by the repo maintainers?

I just read through it and it has **many** factual inaccuracies that need fixing, and duplication of points, I assume it was researched via LLM. 

I would like to know if I should spend time suggesting corrections or not?

zanieb:
I think a comparison like this should probably be written by active maintainers in the packaging ecosystem.

webknjaz:
I've looked through the content more closely and it doesn't seem fixable. I'm closing this and if somebody wants a better document, I recommend they start a discussion first being ready to be responsible for representing a diverse variety of tooling and neighboring ecosystems, not just Conda.

PR #1804, opened 2025-02-05
Document iOS, Android and macOS platform tag schemes.

### LINKED ISSUE #1801: Document PEP 738 in "Platform compatibility tags" specification
### Issue Description

PEP 738 is accepted: https://peps.python.org/pep-0738/

Per https://github.com/python/peps/pull/4250, https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/ should be updated to include PEP 738, as well as any updates/clarifications necessary to implement it.

### Code of Conduct

### LINKED ISSUE #1802: Document PEP 730 in "Platform compatibility tags" specification
### Issue Description

PEP 730 is accepted: https://peps.python.org/pep-0730/

https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/ should be updated to include PEP 730, as well as any updates/clarifications necessary to implement it.

### Code of Conduct

freakboy3742:
Adds documentation for the iOS PEP 730, Android PEP 738 platform tags. 
Also adds documentation for macOS tags, given they have their own quirks (especially regarding the ``universal`` fat architecture specification).

Fixes #1801 
Fixes #1802

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1804.org.readthedocs.build/en/1804/

### DIFF
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -227,7 +227,7 @@ Android's release documentation contains the `full list of Android versions and
 their corresponding API levels
 <https://developer.android.com/tools/releases/platforms>`__.
 
-There are 4 `supported ABIs <https://developer.android.com/ndk/guides/abis>`__. 
+There are 4 `supported ABIs <https://developer.android.com/ndk/guides/abis>`__.
 Normalized according to the rules above, they are:
 
 * ``armeabi_v7a``
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -207,7 +207,9 @@ For example, specifying an architecture of ``universal2`` indicates that
 binaries support *both* x86_64 *and* arm64.
 
 The minimum supported macOS version may also be constrained by architecture. For
-example, macOS 11 (Big Sur) was the first release to support arm64.
+example, macOS 11 (Big Sur) was the first release to support arm64. These
+additional constraints are enforced transparently by the macOS compilation
+toolchain when building a ``universal2`` binary.
 
 .. _android:
 
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -217,10 +217,10 @@ the following list that describes the set of supported architectures:
 ``arch``       Architectures supported
 ============== ========================================
 ``universal2`` ``arm64``, ``x86_64``
-``universal``  ``ppc64``, ``i386``, ``x86_64``
+``universal``  ``i386``, ``ppc``, ``ppc64``, ``x86_64``
 ``intel``      ``i386``, ``x86_64``
-``fat``        ``ppc``, ``ppc64``, ``i386``, ``x86_64``
-``fat32``      ``ppc``, ``i386``
+``fat``        ``i386``, ``ppc``
+``fat3``       ``i386``, ``ppc``, ``x86_64``
 ``fat64``      ``ppc64``, ``x86_64``
 ============== ========================================
 
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -227,6 +227,12 @@ Android's release documentation contains the `full list of Android versions and
 their corresponding SDK versions
 <https://developer.android.com/tools/releases/platforms>`__.
 
+There are 4 supported architectures:
+* ``armeabi_v7a``
+* ``arm64_v8a``
+* ``x86``
+* ``x86_64``
+
 By default, Python 3.13 is compiled using SDK 24 (i.e, Android 7); Python 3.14
 uses SDK 27 (i.e, Android 8.1).
 
@@ -245,6 +251,9 @@ The iOS platform has two SDKs: ``iphoneos`` for physical devices; and
 but are incompatible at the binary level, even if they are running on the same
 architecture. Code compiled for an arm64 simulator will not run on an arm64
 device.
+
+The ``iphonesimulator`` SDK supports 2 architectures: ``arm64`` and ``x86_64``.
+The ``iphoneos`` SDK only supports the ``arm64`` architecture.
 
 By default, Python is compiled with a minimum iOS compatibility version of 13.0.
 
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -109,8 +109,8 @@ subset of Linux platforms, and allows building wheels tagged with the
 ``manylinux`` platform tag which can be used across most common Linux
 distributions.
 
-The current standard is the future-proof ``manylinux_x_y`` standard. It defines
-tags of the form ``manylinux_x_y_arch``, where ``x`` and ``y`` are glibc major
+The current standard is the future-proof ``manylinux_<x>_<y>`` standard. It defines
+tags of the form ``manylinux_<x>_<y>_<arch>``, where ``x`` and ``y`` are glibc major
 and minor versions supported (e.g. ``manylinux_2_24_xxx`` should work on any
 distro using glibc 2.24+), and ``arch`` is the architecture, matching the value
 of :py:func:`sysconfig.get_platform()` on the system as in the "simple" form above.
@@ -151,7 +151,7 @@ auditwheel  ``>=1.0.0``     ``>=2.0.0``        ``>=3.0.0``        ``>=3.3.0`` [#
 
 The ``musllinux`` family of tags is similar to ``manylinux``, but for Linux
 platforms that use the musl_ libc rather than glibc (a prime example being Alpine
-Linux). The schema is ``musllinux_x_y_arch``, supporting musl ``x.y`` and higher
+Linux). The schema is ``musllinux_<x>_<y>_<arch>``, supporting musl ``x.y`` and higher
 on the architecture ``arch``.
 
 The musl version values can be obtained by executing the musl libc shared
@@ -189,6 +189,64 @@ There are currently two possible ways to find the musl library’s location that
 Python interpreter is running on, either with the system ldd_ command, or by
 parsing the ``PT_INTERP`` section’s value from the executable’s ELF_ header.
 
+.. _macos:
+
+macOS
+-----
+
+macOS uses the ``macosx`` family of tags (the ``x`` suffix is a historical
+artefact of Apple's official macOS naming scheme). The schema for compatibility
+tags is ``macosx_<x>_<y>_<arch>``, indicating that the wheel is compatible with
+macOS ``x.y`` or later on the architecture ``arch``. The version number always
+includes a major and minor version, even if Apple's official version numbering
+only refers to the major value. For example, a ``macosx_11_0_arm64`` indicates
+compatibility with macOS 11 or later, on arm64 (i.e., Apple Silicon) hardware.
+
+Recent macOS binaries distributed on Python.org are compiled with a minimum
+macOS compatibility version of 11.0, as macOS 11 (Big Sur) was the first release
+to support the ARM64 Apple Silicon architecture. Python binaries obtained from
+other sources may have a different compatibility version.
+
+macOS also supports the use of a combined, or "fat" architecture specification.
+For example, specifying an architecture of ``universal2`` indicates that
+binaries support *both* x86_64 *and* arm64.
+
+.. _android:
+
+android
+-------
+
+Android uses the schema ``android_<sdk>_<arch>``, indicating compatibility with
+Android SDK ``sdk`` or later, on the architecture ``arch``. Android makes no
+distinction between physical devices and emulated devices.
+
+Note that this tag schema uses the *SDK* version, not the Android OS version
+number. The Android release known publicly as Android 12 (code named "Snow
+Cone") uses SDK 31 or 32, depending on the specific Android version in use.
+Android's release documentation contains the `full list of Android versions and
+their corresponding SDK versions
+<https://developer.android.com/tools/releases/platforms>`__.
+
+By default, Python 3.13 is compiled using SDK 24 (i.e, Android 7); Python 3.14
+uses SDK 27 (i.e, Android 8.1).
+
+.. _ios:
+
+iOS
+---
+
+iOS uses the schema ``ios_<x>_<y>_<arch>_<sdk>``, indicating compatibility with
+iOS ``x.y`` or later, on the ``arch`` architecture, using the ``sdk`` SDK. The
+version number always includes a major and minor version, even if Apple's
+official version numbering only refers to the major value.
+
+The iOS platform has two SDKs: ``iphoneos`` for physical devices; and
+``iphonesimulator`` for simulated devices. These SDKs have the same API surface,
+but are incompatible at the binary level, even if they are running on the same
+architecture. Code compiled for an arm64 simulator will not run on an arm64
+device.
+
+By default, Python is compiled with a minimum iOS compatibility version of 13.0.
 
 Use
 ===
@@ -339,7 +397,8 @@ History
 - November 2019: The ``manylinux_x_y`` perennial tag was approved through
   :pep:`600`.
 - April 2021: The ``musllinux_x_y`` tag was approved through :pep:`656`.
-
+- December 2023: The tags for iOS were approved through :pep:`730`.
+- March 2024: The tags for Android were approved through :pep:`738`.
 
 
 .. _musl: https://musl.libc.org
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -234,9 +234,11 @@ toolchain when building binaries that support multiple architectures.
 Android
 -------
 
-Android uses the schema :file:`android_{api_level}_{abi}`, indicating compatibility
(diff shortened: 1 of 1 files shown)


freakboy3742:
FYI @mhsmith

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+tags is ``macosx_<x>_<y>_<arch>``, indicating that the wheel is compatible with
+macOS ``x.y`` or later on the architecture ``arch``. The version number always
+includes a major and minor version, even if Apple's official version numbering
+only refers to the major value. For example, a ``macosx_11_0_arm64`` indicates
+compatibility with macOS 11 or later, on arm64 (i.e., Apple Silicon) hardware.
+
+Recent macOS binaries distributed on Python.org are compiled with a minimum
+macOS compatibility version of 11.0, as macOS 11 (Big Sur) was the first release
+to support the ARM64 Apple Silicon architecture. Python binaries obtained from
+other sources may have a different compatibility version.
```
ned-deily, line 208: I'm not sure what this is referring to. The universal2 pythons provided by the python.org installers for macOS have always used deployment targets less than 11.0. Perhaps just omit that part.

freakboy3742: I was thinking of older Python versions (which used 10.6, 10.9, etc), or binaries from sources like homebrew; IIRC, homebrew Python builds tend to use the macOS version they've been bottled for. 

However, if you think that detail is more confusion than it's worth, I'm happy to remove (or modify) that line.

ned-deily: The reference to Python.org isn't accurate (we use at least 10.15 today). Perhaps delete the whole paragraph and add something to the following one with something like "The minimum supported version may be further constrained by architecture, for example, macOS supported arm64 beginning with 11.0."?

freakboy3742: Ah - I forgot about x86-64 support :-)

I've modified the text as you've suggested.

ned-deily: Thanks. Perhaps also add that something like "Such constraints may be managed transparently by the universal build tool chain." ?

freakboy3742: Done.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
 
+.. _macos:
+
+macOS
+-----
+
+macOS uses the ``macosx`` family of tags (the ``x`` suffix is a historical
+artefact of Apple's official macOS naming scheme). The schema for compatibility
+tags is ``macosx_<x>_<y>_<arch>``, indicating that the wheel is compatible with
+macOS ``x.y`` or later on the architecture ``arch``. The version number always
```
mhsmith, line 200: There should be a definition of `arch`. For example, the`manylinux` section defines it as "matching the value of sysconfig.get_platform()", though that isn't exactly correct, because `get_platform` also returns the word "linux".

This also applies to iOS below.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+The iOS platform has two SDKs: ``iphoneos`` for physical devices; and
+``iphonesimulator`` for simulated devices. These SDKs have the same API surface,
+but are incompatible at the binary level, even if they are running on the same
+architecture. Code compiled for an arm64 simulator will not run on an arm64
+device.
+
+The ``iphonesimulator`` SDK supports 2 architectures: ``arm64`` and ``x86_64``.
+The ``iphoneos`` SDK only supports the ``arm64`` architecture.
+
+By default, Python is compiled with a minimum iOS compatibility version of 13.0.
```
di, line 258: I think this should be explicit that the list of `arch` components is:

* `arm64_iphoneos`
* `arm64_iphonesimulator`
* `x86_64_iphonesimulator`

...so the reader doesn't have to assume how to combine these two values.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+distinction between physical devices and emulated devices.
+
+Note that this tag schema uses the *SDK* version, not the Android OS version
+number. The Android release known publicly as Android 12 (code named "Snow
+Cone") uses SDK 31 or 32, depending on the specific Android version in use.
+Android's release documentation contains the `full list of Android versions and
+their corresponding SDK versions
+<https://developer.android.com/tools/releases/platforms>`__.
+
+There are 4 supported architectures:
```
mhsmith, line 230: Android documentation more often uses the terms "API level" and "ABI".

```suggestion
Android uses the schema ``android_<api_level>_<abi>``, indicating compatibility
with the given Android API level or later, on the given ABI. Android makes no
distinction between physical devices and emulated devices.

Note that this tag schema uses the API level, not the Android OS version
number. The Android release known publicly as Android 12 (code named "Snow
Cone") uses API level 31 or 32, depending on the specific Android version in use.
Android's release documentation contains the `full list of Android versions and
their corresponding API levels
<https://developer.android.com/tools/releases/platforms>`__.

There are 4 `supported ABIs <https://developer.android.com/ndk/guides/abis>`__:
```

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+* ``armeabi_v7a``
+* ``arm64_v8a``
+* ``x86``
+* ``x86_64``
+
+By default, Python 3.13 is compiled using SDK 24 (i.e, Android 7); Python 3.14
+uses SDK 27 (i.e, Android 8.1).
```
mhsmith, line 237: I don't think this needs to be included on this page. It would add a maintenance burden to the page, it isn't relevant to the major users of the page, such as pip and warehouse, and it could be misunderstood as saying that these are the only acceptable combinations.

This also applies to iOS below.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+Cone") uses SDK 31 or 32, depending on the specific Android version in use.
+Android's release documentation contains the `full list of Android versions and
+their corresponding SDK versions
+<https://developer.android.com/tools/releases/platforms>`__.
+
+There are 4 supported architectures:
+* ``armeabi_v7a``
+* ``arm64_v8a``
+* ``x86``
+* ``x86_64``
```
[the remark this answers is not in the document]
mhsmith: I've added a suggested rephrasing and link in my review.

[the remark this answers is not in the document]
mhsmith: But it looks like the general rule is that once a packaging PEP is final, we should link to the live documentation in this repository rather than the PEP. So I've referred to the normalization rules on the same page.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -109,8 +109,8 @@ subset of Linux platforms, and allows building wheels tagged with the
 ``manylinux`` platform tag which can be used across most common Linux
 distributions.
 
-The current standard is the future-proof ``manylinux_x_y`` standard. It defines
-tags of the form ``manylinux_x_y_arch``, where ``x`` and ``y`` are glibc major
+The current standard is the future-proof ``manylinux_<x>_<y>`` standard. It defines
```
merwok, line 112: There is a role that includes variable text (rendering with the HTML `var` element), but I think it’s meant for filenames, so arguably appropriate for other cases.

source/specifications/platform-compatibility-tags.rst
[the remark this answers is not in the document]
di: Just fyi @mhsmith, I'm not seeing that suggestion in your review, did it get submitted?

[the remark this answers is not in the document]
di: Never mind, I missed it!

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,73 @@ There are currently two possible ways to find the musl library’s location that
...
+Android uses the schema ``android_<sdk>_<arch>``, indicating compatibility with
+Android SDK ``sdk`` or later, on the architecture ``arch``. Android makes no
+distinction between physical devices and emulated devices.
+
+Note that this tag schema uses the *SDK* version, not the Android OS version
+number. The Android release known publicly as Android 12 (code named "Snow
+Cone") uses SDK 31 or 32, depending on the specific Android version in use.
+Android's release documentation contains the `full list of Android versions and
+their corresponding SDK versions
+<https://developer.android.com/tools/releases/platforms>`__.
```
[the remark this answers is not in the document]
di: In that case, can we be explicit here that the specification only permits positive integers?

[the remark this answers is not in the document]
mhsmith: I've done that in my suggested change.

source/specifications/platform-compatibility-tags.rst
freakboy3742: TIL about `:file:`'s `{}` syntax handling. Although this isn't *literally* a file name, I agree that the markup-based approach is a little nicer; I'll make that change.

source/specifications/platform-compatibility-tags.rst
freakboy3742: Good point - it's detail that isn't needed in a formal specification. However, I think it would be worth including the detail from PEP 738 about the applicability of x86 and x86_64 as targets.

source/specifications/platform-compatibility-tags.rst
```diff
@@ -189,6 +189,106 @@ There are currently two possible ways to find the musl library’s location that
...
+============== ========================================
+``arch``       Architectures supported
+============== ========================================
+``universal2`` ``arm64``, ``x86_64``
+``universal``  ``ppc64``, ``i386``, ``x86_64``
+``intel``      ``i386``, ``x86_64``
+``fat``        ``ppc``, ``ppc64``, ``i386``, ``x86_64``
+``fat32``      ``ppc``, ``i386``
+``fat64``      ``ppc64``, ``x86_64``
+============== ========================================
```
freakboy3742, line 225: @ned-deily I've added this table for completeness; since this is a formal specification, it seemed appropriate to include all legal values, not just ``universal2``. 

The notable detail - these values are [what Warehouse will accept](https://github.com/pypi/warehouse/blob/main/warehouse/forklift/legacy.py#L135)... but they don't match the values that ``configure`` accepts. I'm not sure what to do here.

I've taken a stab at how they line up - `universal` and `fat` are the two that I'm most unclear about, as it's not clear which should be `3-way`, and which should be `all`. 

Or, is this a bug in Warehouse that has never been tested because in practice - probably because the x86 transition was essentially complete by the time the wheel format was finalised? (Snow Leopard EOL was 2011; PEP 427 was approved in 2013).

In practice, it looks like `intel` and `universal2` are the only two of these tags that are used. So we could cut this table down to just those two, and note "other tags exist for historical reasons", and ignore the discrepancy.

Any thoughts on how to proceed here?

source/specifications/platform-compatibility-tags.rst
freakboy3742: The wording on the `manylinux` usage is a little ambiguous; while it could be read as meaning `get_platform()` returns the value for `arch`, I think the intention is that it's referring to the entire tag value. 

Regardless, I've modified the definitions in the macOS and iOS specification to refer to `platform.machine()`, since that *only* returns the architecture name.

source/specifications/platform-compatibility-tags.rst
ned-deily: The _definitive_ set of macOS values is in [cpython/Lib/_osx_support.py](https://github.com/python/cpython/blob/main/Lib/_osx_support.py#L546-L562). That code was centralized to be used by `distutils` and other extension builders; I believe `setuptools` used to use it, at least, not sure if it still does or what other builders use it or copies of it.

So it appears that ``fat32`` in Warehouse should have been ``fat`` and there should be a ``fat3`` with ``ppc``, ``ppc64``, ``i386``.

And the in the table you've added, ``fat`` should be ``ppc``, ``i386`` and ``universal`` should be ``ppc``, ``ppc64``, ``i386``, ``x86_64``.

All that said, I've been giving this subject some thought and think it is (past) time to deprecate and/or remove most of the older universal options. My plan is to document that in an upcoming macOS support PEP and take action in an upcoming 3.14 alpha. So I think adding the current values without qualification is OK for now and we can update the values after the support PEP is in place. Or we can could just leave it at ``intel`` and ``universal2`` for now as you suggest.

(Thanks for taking this on.)

ned-deily (commented):
The macOS tag information LGTM, thanks.

freakboy3742:
> The macOS tag information LGTM, thanks.

Thanks for the pointers on what the actual ground truth is here.

FYI: I've opened pypi/warehouse#17566 to track the issue on warehouse, and python/cpython#129712 to track the missing documentation (and inconsistency with `--with-universal-archs` specification).

di (approved):
LGTM, thanks a ton @freakboy3742!

source/specifications/platform-compatibility-tags.rst
```diff
@@ -196,66 +196,99 @@ macOS
...
-tags is ``macosx_<x>_<y>_<arch>``, indicating that the wheel is compatible with
-macOS ``x.y`` or later on the architecture ``arch``. The version number always
-includes a major and minor version, even if Apple's official version numbering
-only refers to the major value. For example, a ``macosx_11_0_arm64`` indicates
-compatibility with macOS 11 or later, on arm64 (i.e., Apple Silicon) hardware.
-
-macOS also supports the use of a combined, or "fat" architecture specification.
-For example, specifying an architecture of ``universal2`` indicates that
-binaries support *both* x86_64 *and* arm64.
+tags is :file:`macosx_{x}_{y}_{arch}``, indicating that the wheel is compatible
```
mhsmith, line 199:
```suggestion
tags is :file:`macosx_{x}_{y}_{arch}`, indicating that the wheel is compatible
```

source/specifications/platform-compatibility-tags.rst
```diff
@@ -196,66 +196,99 @@ macOS
...
+with macOS ``x.y`` or later on the architecture ``arch``.
+
+The value of ``x`` and ``y`` correspond to the major and minor version number of
```
mhsmith, line 202:
```suggestion
The values of ``x`` and ``y`` correspond to the major and minor version number of
```

source/specifications/platform-compatibility-tags.rst
```diff
@@ -196,66 +196,99 @@ macOS
...
+the macOS release, respectively. They must both be positive integers, with the
+``x`` value being ``>= 10``. The version number always includes a major *and*
+minor version, even if Apple's official version numbering only refers to
+the major value. For example, a ``macosx_11_0_arm64`` indicates compatibility
```
mhsmith, line 206:
```suggestion
the major value. For example, ``macosx_11_0_arm64`` indicates compatibility
```

source/specifications/platform-compatibility-tags.rst
```diff
@@ -196,66 +196,99 @@ macOS
...
+with the given Android API level or later, on the given ABI. Android makes no
 distinction between physical devices and emulated devices.
 
-Note that this tag schema uses the *SDK* version, not the Android OS version
-number. The Android release known publicly as Android 12 (code named "Snow
-Cone") uses SDK 31 or 32, depending on the specific Android version in use.
-Android's release documentation contains the `full list of Android versions and
-their corresponding SDK versions
+The API level should be a positive integer. Note that this value is the API
+level, and *not* the user-facing Android version. The release known as Android
```
mhsmith, line 242:
```suggestion
The API level should be a positive integer. This is *not* the same thing as
the user-facing Android version. For example, the release known as Android
```

source/specifications/platform-compatibility-tags.rst
```diff
@@ -196,66 +196,99 @@ macOS
...
 
 .. _android:
 
-android
+Android
 -------
 
-Android uses the schema ``android_<sdk>_<arch>``, indicating compatibility with
-Android SDK ``sdk`` or later, on the architecture ``arch``. Android makes no
+Android uses the schema :file:`android_{api_level}_{abi}`, indicating compatibility
```
mhsmith, line 237: Unfortunately the `:file:` formatting makes the schema less clear, because there are underscores both between and within the placeholders. This could be mitigated by adding an example of a full tag.

freakboy3742: I think simplifying the "variable name" `api_level` to `apilevel` gets around this ambiguity, since the variable name doesn't have any significance beside internal references in this standard.

Regardless, I agree adding an example would be worthwhile.

mhsmith (approved):

mhsmith:
I don't have access to resolve my conversations above, but everything looks good to me now.

willingc (approved):
Changes look good. Thanks @freakboy3742. If CI checks are happy, I will try to merge.

PR #1807, opened 2025-02-05
specifications: add build-details.json (PEP 739)

FFY00:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1807.org.readthedocs.build/en/1807/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,6 @@
 source/guides/github-actions-ci-cd-sample/* @webknjaz
 source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst @webknjaz
+
+# build-details.json
+source/specifications/build-details/ @FFY00
+source/specifications/specs/build-details-*.json @FFY00
--- a/source/conf.py
+++ b/source/conf.py
@@ -28,6 +28,7 @@ extensions = [
     "sphinx_inline_tabs",
     "sphinx_copybutton",
     "sphinx_toolbox.collapse",
+    "sphinx-jsonschema",
 ]
 
 nitpicky = True
@@ -128,6 +129,7 @@ texinfo_documents = [
 
 linkcheck_ignore = [
     "http://localhost:\\d+",
+    "https://packaging.python.org/en/latest/specifications/schemas/*",
     "https://test.pypi.org/project/example-package-YOUR-USERNAME-HERE",
     "https://pypi.org/manage/*",
     "https://test.pypi.org/manage/*",
--- a/source/conf.py
+++ b/source/conf.py
@@ -129,7 +129,7 @@ texinfo_documents = [
 
 linkcheck_ignore = [
     "http://localhost:\\d+",
-    "https://packaging.python.org/en/latest/specifications/schemas/*",
+    "https://packaging.python.org/en/latest/specifications/schemas/.*",
     "https://test.pypi.org/project/example-package-YOUR-USERNAME-HERE",
     "https://pypi.org/manage/*",
     "https://test.pypi.org/manage/*",
--- a/source/conf.py
+++ b/source/conf.py
@@ -83,6 +83,10 @@ if RTD_CANONICAL_BUILD:
     # https://plausible.io/packaging.python.org
     html_js_files.extend(_metrics_js_files)
 
+html_extra_path = [
+    "specifications/schemas",
+]
+
 # -- Options for HTML help output ------------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-help-output
 
--- a/source/specifications/build-details/index.rst
+++ b/source/specifications/build-details/index.rst
@@ -16,7 +16,7 @@ extension ABI details, and other information that is specific to that particular
 build of Python.
 
 Starting from Python 3.14, a ``build-details.json`` file is installed in the
-platform-independent standard library directory (``stdlib``, eg.
+platform-independent standard library directory (``stdlib``, e.g.
 ``/usr/lib/python3.14/build-details.json``).
 
 Please refer to the :ref:`latest version <build-details-v1.0>` for its
--- a/source/specifications/build-details/index.rst
+++ b/source/specifications/build-details/index.rst
@@ -1,8 +1,8 @@
 .. _build-details:
 
-======================
-``build-details.json``
-======================
+==========================
+:file:`build-details.json`
+==========================
 
 .. toctree::
    :hidden:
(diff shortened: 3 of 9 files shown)


FFY00:
@CAM-Gerlach, could you review this? Thanks!

source/specifications/build-details/index.rst
```diff
@@ -0,0 +1,52 @@
...
+   v1.0 <v1.0>
+
+
+The ``build-details.json`` file is a standardized file format that provides
+build-specfic information of a Python installation, such as its version,
+extension ABI details, and other information that is specific to that particular
+build of Python.
+
+Starting from Python 3.14, a ``build-details.json`` file is installed in the
+platform-independent standard library directory (``stdlib``, eg.
```
zanieb, line 19:
```suggestion
platform-independent standard library directory (``stdlib``, e.g.
```

source/conf.py
```diff
@@ -128,6 +129,7 @@
 
 linkcheck_ignore = [
     "http://localhost:\\d+",
+    "https://packaging.python.org/en/latest/specifications/schemas/*",
```
webknjaz, line 132: Why?

webknjaz: Also, shouldn't this be a regex?

source/specifications/build-details/index.rst
```diff
@@ -0,0 +1,52 @@
+.. _build-details:
+
+======================
+``build-details.json``
+======================
```
webknjaz, line 5: Let's add the file role here and across the doc.
```suggestion
==========================
:file:`build-details.json`
==========================
```

source/specifications/build-details/index.rst
```diff
@@ -0,0 +1,52 @@
...
+v1.0
+~~~~
+
+.. list-table::
+
+    * - Specification
+      - :ref:`build-details-v1.0`
+
+    * - Schema
+      - https://packaging.python.org/en/latest/specifications/schemas/python-build-info-v1.0.schema.json
```
webknjaz, line 49: Can this be a relative link?

.github/CODEOWNERS
webknjaz, line 1: GH shows this warning:
> [!warning]
> This CODEOWNERS file contains errors

*(due to you not having commit bit in the repo)*

Yedfree (approved):

source/conf.py
FFY00: Because the URL doesn't exist yet, so the CI will always fail.

source/specifications/build-details/index.rst
FFY00: I'd very strongly prefer to have it show the full canonical URL that can be used as the schema identifier.

.github/CODEOWNERS
FFY00: Ah, right. Can we change that? With this addition, I would be maintaining the `build-details.json` specification.

pradyunsg: Happy to add you as an editor. Lemme click the magic buttons...

pradyunsg:
With great power, comes an annoying amount of responsibility. Yield it carefully @ffy00, whom I've added to the packaging.python.org editors list.

chrysle (approved):

FFY00:
@webknjaz can you have a look again, so that we can unblock the PR?

source/specifications/build-details/index.rst
webknjaz: I'm wondering if it'd be possible to migrate to using something like the `:download:` role.. Any ideas?

webknjaz: Should we be storing this under `_static/` instead?
Additionally, it should be possible to do something like

```suggestion
      - `https://packaging.python.org/en/latest/_static/schemas/python-build-info-v1.0.schema.json <_static/schemas/python-build-info-v1.0.schema.json>`_
```

This needs testing. But if it works, we'd be able to get rid of the linkignore thing.

FFY00: > Should we be storing this under `_static/` instead?

I don't think having `_static` in the canonical ID URL looks good. I added the schemas directory to `html_extra_path`.

> Additionally, it should be possible to do something like
> 
> This needs testing. But if it works, we'd be able to get rid of the linkignore thing.

I tried using both a normal link like in that snippet, and the download role, neither fixed the link check.

source/conf.py
```diff
@@ -128,6 +133,7 @@
 
 linkcheck_ignore = [
     "http://localhost:\\d+",
+    "https://packaging.python.org/en/latest/specifications/schemas/.*",
     "https://test.pypi.org/project/example-package-YOUR-USERNAME-HERE",
     "https://pypi.org/manage/*",
```
ncoghlan, line 138: I worked out why these trailing globs seem to work: `linkcheck_ignore` does a *prefix* match on the given regexes, so finishing with `/*` (zero-or-more trailing slashes) is essentially identical to omitting that part of the pattern.

To force a whole string match instead of a prefix match you have to specifically terminate the pattern with `$`

ncoghlan (approved):

ncoghlan:
Looks good to me too, so I went ahead and marked this one as ready for merging.

@webknjaz I think your link cleanup requests are reasonable, but also easier to resolve once we've published a version that includes the schema file.

source/specifications/build-details/index.rst
```diff
@@ -0,0 +1,52 @@
...
+v1.0
+~~~~
+
+.. list-table::
+
+    * - Specification
+      - :ref:`build-details-v1.0`
+
+    * - Schema
+      - :download:`https://packaging.python.org/en/latest/specifications/schemas/build-details-v1.0.schema.json <../schemas/build-details-v1.0.schema.json>`
```
webknjaz, line 49: Looks like the actual generated href isn't preserved, as `:download:` copies things into another location.

source/conf.py
```diff
@@ -82,6 +83,10 @@
     # https://plausible.io/packaging.python.org
     html_js_files.extend(_metrics_js_files)
 
+html_extra_path = [
```
webknjaz, line 86: Why is this needed? I don't think it works the way you assume...

webknjaz:
@ncoghlan we can unblock and deal with the consequences later, but this is broken, FWIW.

source/specifications/build-details/index.rst
FFY00: You are right, thanks for noticing, we should go back to a normal link then.

source/conf.py
FFY00: Yep, sorry!

PR #1803, opened 2025-02-04
simple-repository-api: bump, explain api-version

woodruffw:
See discussion in https://discuss.python.org/t/does-the-simple-repository-api-version-1-3-exist/79361 -- this updates the living index API spec to correctly reflect the latest API version (1.3), as well as add some callouts for the fields that make a response look like a 1.3 response (namely, provenance).

CC @di @brettcannon 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1803.org.readthedocs.build/en/1803/

PR #1808, opened 2025-02-06
simple-repository-api: clean up, add API history

woodruffw:
This follows #1803, and performs a few cleanup actions on the simple repository API docs:

- I've removed the `Additional Fields for the Simple API for Package Indexes` section entirely and have folded its prose into the relevant parts of the `JSON Serialization` section. This makes the doc as a whole easier to follow, since the user doesn't need to tab down to a different adjacent section to learn about the changes made with API version 1.1. It's also consistent with that was done with API version 1.3, i.e. PEP 740.
- I've added an `API Version History` subsection that lists only API versions and the PEPs that stabilized them. This should make it easier to cross-reference changes.
- I've moved the standard RFC 2119 language to the top of the document, since the entire doc uses it and not just the subsection it previously was in.
- I've added a few instances of proper HTML syntax highlighting.

CC @di @brettcannon @nlhkabu 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1808.org.readthedocs.build/en/1808/

### DIFF
--- a/source/specifications/simple-repository-api.rst
+++ b/source/specifications/simple-repository-api.rst
@@ -5,6 +5,11 @@
 Simple repository API
 =====================
 
+The keywords "**MUST**", "**MUST NOT**", "**REQUIRED**", "**SHALL**",
+"**SHALL NOT**", "**SHOULD**", "**SHOULD NOT**", "**RECOMMENDED**", "**MAY**",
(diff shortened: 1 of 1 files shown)


brettcannon (approved):

di (approved):

willingc (approved):
@woodruffw Thanks for these helpful changes. The prose and rendering look good to me.

PR #1810, opened 2025-02-10
simple-repository-api: clean up PEP 658/PEP 714 bits

woodruffw:
This follows #1803 and #1808, as part of my effort to make the living index specification a bit easier to follow. Key changes:

1. I've removed the `Rename dist-info-metadata in the Simple API` section entirely, which seems to have come mostly verbatim from PEP 714. Instead, I've folded its prescriptions (renaming fields in both the HTML and JSON representations) into their relevant sections.
2. I've updated both the HTML and JSON sections to mention `data-core-metadata`/`core-metadata`, and have added admonitions noting that the old (PEP 658) names are considered legacy.

While doing the above, I noticed a minor errata (maybe even too small to be an errata?): PEP 658 and PEP 714 both change the Index APIs and happened after version 1.0 but *before* 1.1. Consequently, there was no version bump defined for either, despite the fact that they change fields in a forward compatible way (which other PEPs mark by bumping the minor version). 

In practice the above doesn't matter (since everything is compatible), but it means that PEP 658 and PEP 714 don't fit cleanly into the "API Version History" section.

CC @di @pypa/packaging-user-guide-editors 

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1810.org.readthedocs.build/en/1810/

### DIFF
--- a/source/specifications/simple-repository-api.rst
+++ b/source/specifications/simple-repository-api.rst
@@ -88,6 +88,26 @@ In addition to the above, the following constraints are placed on the API:
   associated signature, the signature would be located at
   ``/packages/HolyGrail-1.0.tar.gz.asc``.
 
+* A repository **MAY** include a ``data-core-metadata`` attribute on a file
+  link.
+
+  The repository **SHOULD** provide the hash of the Core Metadata file as the
+  ``data-core-metadata`` attribute's value using the syntax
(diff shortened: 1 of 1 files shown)


di (approved):

willingc (approved):
Thanks @woodruffw

PR #1816, opened 2025-02-18
Add anchor links for license and license-files

cdce8p:
Add anchor links for `license` and `license-files` to make cross-linking from other projects easier.

In particular, I'd like to cross reference https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-files from the setuptools docs.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1816.org.readthedocs.build/en/1816/


cdce8p:
/CC @webknjaz Would you mind taking a quick look at this? It would help us like to the correct section from the setuptools docs.

webknjaz (approved):
Thanks!

PR #1826, opened 2025-03-11
Extract glob specification to a separated document and add reference implementation

abravalheri:
This is a follow up on the discussion in https://github.com/python/peps/issues/4299.

I basically extracted the existing text, but tried to make it slightly more generic so that it is not tied to `license-files` and can be reused somewhere else. Then I added a reference implementation based on @cdce8p's [contribution to Setuptools](https://github.com/pypa/setuptools/pull/4841).

It seems that other [PEPs in writing](https://discuss.python.org/t/pep-770-improving-measurability-of-python-packages-with-software-bill-of-materials/76308/68) are looking into using the same glob definition, so that would also be a good reason for the extraction.

Closes python/peps#4299

/cc @befeleme

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1826.org.readthedocs.build/en/1826/

### DIFF
--- a/source/specifications/glob-patterns.rst
+++ b/source/specifications/glob-patterns.rst
@@ -1,0 +1,115 @@
+=================
+``glob`` patterns
+=================
+
+Some PyPA specifications, e.g. :ref:`pyproject.toml's license-files
+<pyproject-toml-license-files>`, accept certain types of *glob patterns*
+to match a given string containing wildcards and character ranges against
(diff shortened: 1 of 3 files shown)


befeleme (approved):
I think it's a great addition to the specification. Thank you.

chrysle (approved):

chrysle:
Thank you!

PR #1829, opened 2025-03-19
Add PBR on Project Summaries

bangseongbeom:
[PBR](https://docs.openstack.org/pbr/latest/) is a Setuptools-related library developed by OpenStack. I hope this is also included in the list.


webknjaz:
Technically, it's a setuptools plugin. I don't think there was a precedent of documenting plugins for build backends. If this is added, other plugins would also have to be listed.
I don't know if this can be qualified as a key project.

PR #1853, opened 2025-04-24
Add PEP 770 to specifications

sethmlarson:
* Adds reservation of directory names under `.dist-info` and reserves a few names already in use.
* Adds description of `.dist-info/sboms` directory.

I wasn't sure whether to add a "guide" or not yet, as there aren't any especially "user-facing" ways to have an SBOM end up in this directory beyond your build backend/build tool supporting it. Should I leave that for a later task, then?

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1853.org.readthedocs.build/en/1853/

### DIFF
--- a/source/specifications/binary-distribution-format.rst
+++ b/source/specifications/binary-distribution-format.rst
@@ -256,8 +256,8 @@ The .dist-info directory
 Subdirectories in :file:`.dist-info/`
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


webknjaz (commented):
There's probably no need for a guide. Especially if build backends don't have a toggle. It might be useful to note in the existing ones which backend versions implement it.

Also, it might be a good idea to migrate the PEP itself into the specs section.

I've made a few suggestions to RST as I'm trying to keep the style consistent and it's easier to distinguish dirs vs regular files based on the trailing slash..

webknjaz (approved):
Thanks!

PR #1844, opened 2025-04-11
grammar you->your

wade-cheng:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1844.org.readthedocs.build/en/1844/

PR #1855, opened 2025-05-01
minor copy-editing fixes

nedbat:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1855.org.readthedocs.build/en/1855/

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -291,7 +291,7 @@ Glossary
         <https://bitbucket.org/pypa>`_, and discuss issues on the
         `distutils-sig mailing list
         <https://mail.python.org/mailman3/lists/distutils-sig.python.org/>`_
-	and `the Python Discourse forum <https://discuss.python.org/c/packaging>`__.
+	and `the Python Discourse forum <https://discuss.python.org/c/packaging>`_.
 
 
     Python Package Index (PyPI)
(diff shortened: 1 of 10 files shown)


nedbat:
BTW, I see #1844 also covered one of these.

source/glossary.rst
```diff
@@ -291,7 +291,7 @@ Glossary
         <https://bitbucket.org/pypa>`_, and discuss issues on the
         `distutils-sig mailing list
         <https://mail.python.org/mailman3/lists/distutils-sig.python.org/>`_
-	and `the Python Discourse forum <https://discuss.python.org/c/packaging>`__.
+	and `the Python Discourse forum <https://discuss.python.org/c/packaging>`_.
```
webknjaz, line 294: What's the motivation here? I had a feeling that an anonymous style was used on purpose.

webknjaz (approved):
Thanks for doing this! Sorry for not getting to it earlier.

Could you clarify this one thing below?

source/glossary.rst
nedbat: As far as I can tell, these two syntaxes do the same thing, but maybe I am missing something?

nedbat: We can leave that out of the PR, I changed it because the anonymous syntax with an explicit URL seemed really odd to me.

nedbat: I've reverted that line.

webknjaz: Thanks! Marking this thread as resolved.

firas9941 (approved):

PR #1859, opened 2025-05-13
guides: remove manual Sigstore steps from publishing guide

### LINKED ISSUE #1788: Remove `gh-action-sigstore-python` from the GitHub Actions publishing guide
### Issue Description

I'm filing this as a reminder issue for myself, although if someone wants to get to it before I do please feel free 🙂 

TL;DR: The current GitHub Actions publishing guide has sections like [this](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#signing-the-distribution-packages) that reference [gh-action-sigstore-python](https://github.com/sigstore/gh-action-sigstore-python) for signing. This step is strictly superfluous now that [PEP 740](https://peps.python.org/pep-0740/) has been implemented on PyPI, since the publishing step (or more generally, any step that chooses to make an attestation) does the signing implicitly.

As such, references to `gh-action-sigstore-python` should probably be fully removed and replaced with notes about how attestations currently work by default when using a supported Trusted Publishing provider (currently GitHub and GitLab, but others as well in the future). Keeping the references in place is liable to cause user confusion:

1. The signatures produces by `gh-action-sigstore-python` aren't format-compatible with what PEP 740 expects, meaning they really only end up in the GitHub release artifacts, where they have limited value;
2. Having two+ signing steps means the user ends up with multiple sets of signatures per artifact, which is confusing.

Neither of these results in breakage, but we can eliminate them as sources of confusion 🙂 

### Code of Conduct

woodruffw:
These steps are superfluous now that gh-action-pypi-publish generates and uploads PEP 740-compatible attestations by default on its own.

(They also served a slightly different purpose than PEP 740 attestations, since they were never uploaded to PyPI and used a different format.)

Closes #1788.

+CC @webknjaz @di for review 🙂

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-test-pypi.yml
@@ -51,49 +51,6 @@ jobs:
     - name: Publish distribution 📦 to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
 
-  github-release:
-    name: >-
-      Sign the Python 🐍 distribution 📦 with Sigstore
-      and upload them to GitHub Release
-    needs:
-    - publish-to-pypi
-    runs-on: ubuntu-latest
-
-    permissions:
-      contents: write  # IMPORTANT: mandatory for making GitHub Releases
-      id-token: write  # IMPORTANT: mandatory for sigstore
-
-    steps:
-    - name: Download all the dists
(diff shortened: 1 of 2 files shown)


di (approved):
LGTM (but I am not an 'editor' here)

webknjaz (approved):
Thanks!

PR #1860, opened 2025-05-19
Add 'docs' to dependency groups example

sirosen:
This is a small example update based on some observed usage of Dependency Groups.
We wouldn't want to bloat the example, but many projects are using a `docs` group and this example therefore better exemplifies how Dependency Groups can be used to declare isolated/separate groups.

---

I argue for this change as net-good based on the fact that it shows that `docs` and `test` can be separate groups.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1860.org.readthedocs.build/en/1860/

### DIFF
--- a/source/specifications/dependency-groups.rst
+++ b/source/specifications/dependency-groups.rst
@@ -22,14 +22,16 @@ Specification
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


webknjaz (approved):
Thanks!

PR #1867, opened 2025-05-30
Add guide about `sdist` drawbacks

abravalheri:
This PR is motivated by the discussion in https://discuss.python.org/t/how-can-build-backends-avoid-breaking-users-when-they-make-backwards-incompatible-changes/85847.

The core idea is to provide advice for people with mission critic systems/pipelines or environments with very little tolerance for unforeseen (and sometimes accidental) disruptions when building packages.

### DIFF
--- a/source/guides/index.rst
+++ b/source/guides/index.rst
@@ -13,3 +13,4 @@ introduction to packaging, see :doc:`/tutorials/index`.
    section-hosting
    tool-recommendations
    analyzing-pypi-package-downloads
+   sdist-drawbacks-and-predictability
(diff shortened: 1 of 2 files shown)


ncoghlan:
Rather than framing this as a guide on the disadvantages of sdists, I'd prefer to see this framed topic as a guide on the advantages of publishing wheel archives in *addition* to publishing sdists.

sdists and wheels serve different purposes, so "publish both" is the only option that covers the full range of potential use cases (source control links don't provide assurance that the tag hasn't been moved since the release was made).

The packaging tutorial does suggest to [publish both](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives), but it would be good to have something to link to that explained *why* that is the recommendation.

PR #1868, opened 2025-06-01
Fix: CI failure (Linkcheck)

Daksh2000:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1868.org.readthedocs.build/en/1868/

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -146,6 +146,7 @@ linkcheck_ignore = [
     "https://pyscaffold.org/*",
     "https://anaconda.org",
     "https://www.cisa.gov/sbom",
+    "https://developers.redhat.com/products/softwarecollections/overview",
 ]
 linkcheck_retries = 5
 # Ignore anchors for common targets when we know they likely won't be found


webknjaz:
FTR the empty commit didn't trigger the main CI run since it wasn't approved.

Daksh2000:
> This needs justification. Why ignore it? Why not scope to a specific URL? Etc.

HI @webknjaz , from what I could gather upon researching , that redhat blocks the request maybe because of missing headers , I couldn't dig down deeper as the articles on redhat requires some sort of subscription to access, 
I have (for now) changed the generic url to the specific one , I am open to any suggestion on this issue.

abravalheri:
Thanks @Daksh2000.

I also saw this problem in https://github.com/pypa/packaging.python.org/pull/1867#issuecomment-2925010622.

webknjaz (approved):
Hey, sorry it took so long to get back to you. Let's try merging.

Though, if there's alternative web pages that work, I'd normally prefer replacing the links instead of maintaining a list of ignores.

PR #1871, opened 2025-06-18
Mention the metadata version at the top, and re-order the history

nedbat:
This should make the spec's status and history clearer.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1871.org.readthedocs.build/en/1871/


brettcannon (approved):

ncoghlan (approved):
Thanks!

PR #1874, opened 2025-06-28
Fix a small markup error in the contribute guide

koyuki7w:
The "Do not translate ..." part shoule be a definition list, not a blockquote, like the rest of this page.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1874.org.readthedocs.build/en/1874/


webknjaz (approved):
Thanks!

PR #1876, opened 2025-07-02
Add RepoFlow to list of existing PyPI compatible repositories

100tomer:
Added RepoFlow to the comparison table in the "Hosting your own simple repository" guide.

RepoFlow is a modern package registry that supports PyPI uploads and proxying. It is available as both a cloud-hosted and self-hosted solution, and also supports other package types like npm, Docker, and Maven.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1876.org.readthedocs.build/en/1876/

### DIFF
--- a/source/guides/hosting-your-own-index.rst
+++ b/source/guides/hosting-your-own-index.rst
@@ -126,6 +126,12 @@ Existing projects
        <https://httpd.apache.org/docs/current/mod/mod_cache_disk.html>`_,
        you can cache requests to package indexes through an Apache server
 
+   * - `RepoFlow <https://www.repoflow.io>`_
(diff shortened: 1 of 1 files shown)


webknjaz (changes requested):
I think we've been rejecting commercial product offerings in this table in the past. All the projects in there are FOSS and don't look like ads. Need to search the tracker for precedents.

webknjaz:
Found the respective decisions: https://github.com/pypa/packaging.python.org/pull/1366#issuecomment-1913818089 / https://github.com/pypa/packaging.python.org/pull/1367#issuecomment-1806700551.

PR #1879, opened 2025-07-12
simple-repository-api: add PEP 792, clean up layout

woodruffw:
This PR serves two purposes:

1. It integrates the recently-approved [PEP 792](https://peps.python.org/pep-0792/) into the living PyPUG specifications, so that I can complete the procedural step of linking to PyPUG from the PEP itself
2. In the process of integrating this spec into the living simple repository API spec, I attempted to clean up, streamline, and fix the "flow" of the current spec. In particular:

    * I've moved the copy of the "yanking" PEP to its own page with proper linking, and made it consistent with how the other amending PEPs are documented (field changes are noted under the HTML and JSON sections with links to the separate page for full semantics)
    * I've moved some of the more "mechanical" parts of the living spec (about normalized names, API versioning, etc.) to the top of the page, since it was previously awkward to not have their concepts introduced until halfway through the page (at which point they're already being used)
    * I've tweaked the layout/headings slightly on the HTML sections, so that they more closely mirror the JSON sections in terms of clearly distinguishing between the "top" and "detail" indices. 
    * Finally I've gone throughout and bumped the examples to API version 1.4, since that's what PEP 792 was approved with.

All changes besides the ones approved with PEP 792 should be non-semantic, i.e. they should only make the spec easier to read, not change any pre-existing behaviors.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1879.org.readthedocs.build/en/1879/

### DIFF
--- a/source/specifications/simple-repository-api.rst
+++ b/source/specifications/simple-repository-api.rst
@@ -1026,3 +1026,4 @@ History
   from a package, in :pep:`714`
 * November 2024: provenance metadata in the HTML and JSON formats, in :pep:`740`
 * July 2025: project status markers in the HTML and JSON formats, in :pep:`792`
+* July 2025: layout changes (dedicated page for file yanking, introduce concepts before API details)
--- a/source/specifications/simple-repository-api.rst
+++ b/source/specifications/simple-repository-api.rst
@@ -20,7 +20,7 @@ retrieving packages from an index server comes in two forms:
 Base API
 ========
 
-A repository that implements the simple API is defined by its base URL, this is
+A repository that implements the simple API is defined by its base URL. This is
 the top level URL that all additional URLs are below. The API is named the
 "simple" repository due to the fact that PyPI's base URL is
 ``https://pypi.org/simple/``.
@@ -636,7 +636,8 @@ As an example:
     {
       "meta": {
         "api-version": "1.4",
-        "project-status": "active"
+        "project-status": "active",
+        "project-status-reason": "this project is not yet haunted"
       },
       "name": "holygrail",
       "files": [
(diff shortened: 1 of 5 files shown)


woodruffw:
NB: Linkcheck is failing elsewhere:

```
(guides/installing-scientific-packages: line   19) broken    https://math-atlas.sourceforge.net/ - 403 Client Error: Forbidden for url: https://math-atlas.sourceforge.net/
```

source/specifications/simple-repository-api.rst
```diff
@@ -28,11 +29,115 @@ the top level URL that all additional URLs are below. The API is named the
...
+
+   def normalize(name):
+       return re.sub(r"[-_.]+", "-", name).lower()
+
+.. _simple-repository-api-versioning:
+
+Versioning PyPI's Simple API
+----------------------------
+
+This spec proposes the inclusion of a meta tag on the responses of every
```
ncoghlan, line 52: Huh, not a new problem, but this highlights that some of the phrasing didn't get fixed when migrating from "proposed API specification" to "actual API specification".

source/specifications/simple-repository-api.rst
```diff
@@ -1006,3 +1025,4 @@ History
 * June 2023: renaming the field which provides package metadata independently
   from a package, in :pep:`714`
 * November 2024: provenance metadata in the HTML and JSON formats, in :pep:`740`
+* July 2025: project status markers in the HTML and JSON formats, in :pep:`792`
```
ncoghlan, line 1028:
```suggestion
* July 2025: project status markers in the HTML and JSON formats, in :pep:`792`
* July 2025: layout changes (dedicated page for file yanking, introduce concepts before API details)
```

source/specifications/simple-repository-api.rst
woodruffw: Yep 😅 -- if you think it's OK, I can do a follow-up that tweaks some of these (including my own) to be less verbatim transfer and more normative ("This spec stipulates..." or similar).

woodruffw:
Looks like the link is fixed, but a different anchor is now broken. I'll fix that...

PR #1837, opened 2025-03-25
Latest setuptools and flit now support PEP 639

not-my-profile:
This updates the now outdated notes added in
faf4c7d882ef2be310d5dc46d54ca666aa18821c.

Closes #1831.

----
📚 Documentation preview 📚:
* https://python-packaging-user-guide--1837.org.readthedocs.build/en/1837/guides/writing-pyproject-toml/
* https://python-packaging-user-guide--1837.org.readthedocs.build/en/1837/guides/licensing-examples-and-user-scenarios/
* https://python-packaging-user-guide--1837.org.readthedocs.build/en/1837/tutorials/packaging-projects/

PR #1880, opened 2025-07-16
Add uv-build to table

cbrnr:
uv-build is stable since 0.7.19 and supports PEP 639.

### DIFF
--- a/source/guides/writing-pyproject-toml.rst
+++ b/source/guides/writing-pyproject-toml.rst
@@ -314,11 +314,13 @@ backend>` now support the new format as shown in the following table.
      - flit-core [#flit-core-pep639]_
      - pdm-backend
      - poetry-core
+     - uv-build
    * - 1.27.0
      - 77.0.3
      - 3.12
      - 2.4.0
      - `not yet <poetry-pep639-issue_>`_
+     - 0.7.19
 
 
 .. _license:


Mey1986 (approved):

cbrnr:
I'm happy to add uv-build in the other locations mentioned in the linked issue. Is it OK in this PR, or should I open a new one?

cbrnr:
I think it's easy enough to add this here.

cbrnr (commented):
I noticed that the table in `writing-pyproject-toml.rst` lists hatchling 1.27.0, whereas `build-backend-tabs.rst` contains 1.26.0. Should I fix this and if so, which one is the correct version?

webknjaz:
It's probably a good idea to separate such changes. Also, the broken link needs a separate PR too, to unblock this.

source/shared/build-backend-tabs.rst
```diff
@@ -32,3 +32,11 @@
         [build-system]
         requires = ["pdm-backend >= 2.4.0"]
         build-backend = "pdm.backend"
+
+.. tab:: uv-build
+
+    .. code-block:: toml
+
+        [build-system]
+        requires = ["uv_build >= 0.7.19, <0.9.0"]
```
webknjaz, line 41: Upper limits are normally discouraged. What's the justification here?

cbrnr: See the note in their official docs: https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend

konstin: We make (careful) breaking changes in the configuration format in 0.x releases, i.e. the configuration is only stable between >=0.x and <0.(x+1). Being a build backend, the usual concerns about upper bounds don't apply: The uv build backend is a binary with only a minimal Python shim, so older versions of the build backend will still work with newer Python versions. It has zero dependencies and is the only package in the build venv, so the aren't any conflicts the upper bound could cause.

johnthagen:
Another place it might be worth mentioning `uv_build`

- https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

cbrnr:
@johnthagen that's already part of this PR.

cbrnr:
@webknjaz the broken link works for me, can you re-run the failing test?

cbrnr:
You're probably right, but I think this problem should then be fixed in a different PR.

source/shared/build-backend-tabs.rst
webknjaz: > Being a build backend, the usual concerns about upper bounds don't apply

That may be true upstream, but may present a challenge for the downstream maintainers as they often have a single version of a thing across their entire repositories (per distro release).

cc @befeleme @hroncok @mgorny any downstream insights on this?

hroncok: Please don't encourage upper bounds here. If you do, we'll and up in a position where we cannot update uv-build until we lift them. We won't be even able to test the packages with the new uv-build version in Fedora because of the pin.

Thanks @webknjaz for the mention.

mgorny: It's roughly as you say. I get the point that some degree of pinning is necessary to account for API changes, but this sounds like the API may change a lot — which makes me think the backend is not yet ready to be recommended to people.

konstin:
Also note that we're only building pure Python `py3-none-any` wheels, so it is possible to build a package on any supported linux/windows/mac/... machine and reuse on any other system. Similarly, users of those packages never have to build from source, as `py3-none-any` wheels exist on the index.

Please see https://github.com/astral-sh/uv/issues/3957#issuecomment-2657825266, where we're addressing a lot of the common concerns there.

source/shared/build-backend-tabs.rst
musicinmybrain: I agree with @hroncok and @mgorny. However, I fear that our recommendations may have little effect no matter what we write here, since most projects will be likely arriving at the `uv_build` backend via something like `uv init --name foobar --build-backend uv_build`, or eventually just `uv init --name foobar` once the `uv_build` backend becomes the default, and `uv init` does write a SemVer-pinned dependency.

zanieb: It feels like the best place for a discussion on upper bounds for `uv-build` belongs in the uv issue tracker. It's just going to confuse users to recommend something here that differs from `uv init` and the uv documentation.

geofft: I agree we should move this to the uv issue tracker, but in that issue I'd be curious to understand why distros can't (if necessary) package the build backend multiple times at multiple versions. This seems pretty normal for build tooling in the general case, e.g., Fedora has [autoconf](https://packages.fedoraproject.org/pkgs/autoconf/autoconf), [autoconf213](https://packages.fedoraproject.org/pkgs/autoconf213/autoconf213/), and [autoconf2.7x](https://packages.fedoraproject.org/pkgs/autoconf2.7x/autoconf2.7x/), Debian doess imilar things, and I think this is also common for GCC, LLVM, etc. in many distros.

To be clear I am fully of the opinion that having to do this is suboptimal, but I would guess the API breaking changes are not intended to be onerous, so upstreams should be encouraged to move to the latest version anyway - which is (as I understand it) how stuff like autoconf works. If an upstream doesn't work with the latest version and it's hard to patch, then distros have the option to package an older version too, which is annoying but doable.

konstin: In uv, we are very deliberate about breaking changes. We only stabilize features we have high confidence in, usually after a preview period, and we avoid changing features as much as possible.

uv has a large userbase with very diverse setups which we need to take into consideration for any change. See for example our latest breaking release, 0.8.0: https://github.com/astral-sh/uv/releases/tag/0.8.0. The large majority of users won't be broken by this, they do not have to change anything in their workflow. But these changes could affect users, and at our scale we know that there is some user that will be affected. We're conservative and bundle even changes such as making PEP 514 on-by-default or changing the behavior on broken workspace structures to be less surprising into a breaking release.

I think this is a pretty similar situation to how a lot of packages work? Especially if you look at non-Python, many ecosystem allow multiple package versions in the same dependency tree even without the separation that we get from build environments.

mgorny: > I'd be curious to understand why distros can't (if necessary) package the build backend multiple times at multiple versions

The primary reason is that normally there can be only a single `uv_build` package installed at the same time. Most of the time, distributions package multiple versions only if they have to — and it usually involves a lot of effort. For autoconf, it generally implies renaming the binaries and then having a whole framework to select the right version. For `uv_build`, this would likely be similar — having to rename the installed package and manually patch every `pyproject.toml` to use the correct renamed version.

notatallshaw: Going back to the original comment, `flit_core` also advises using an upper bound. And downstream maintainers seem to handle that okay?

And in general I've come around to the idea this is the correct thing for a build backend, if the expectation is it will not it be broken by new Python releases (e.g. because its very simple or minimally interacts with Python).

The correct thing to do for `setuptools` _may_ be different, but we should treat `setuptools` as special, not the rule.

musicinmybrain: > I agree we should move this to the uv issue tracker, but in that issue I'd be curious to understand why distros can't (if necessary) package the build backend multiple times at multiple versions. This seems pretty normal for build tooling in the general case, e.g., Fedora has [autoconf](https://packages.fedoraproject.org/pkgs/autoconf/autoconf), [autoconf213](https://packages.fedoraproject.org/pkgs/autoconf213/autoconf213/), and [autoconf2.7x](https://packages.fedoraproject.org/pkgs/autoconf2.7x/autoconf2.7x/), Debian doess imilar things, and I think this is also common for GCC, LLVM, etc. in many distros.

It’s probably technically possible to do a compat package like `python-uv-build0.7` in Fedora, but there are disadvantages that likely outweigh the advantages.

> I think this is a pretty similar situation to how a lot of packages work? Especially if you look at non-Python, many ecosystem allow multiple package versions in the same dependency tree even without the separation that we get from build environments.

Indeed, we use compat packages much more freely in Fedora for other language ecosystems like Rust, although we still try to limit them to reduce our maintenance burden. A key difference is that there can only be one system-wide copy of a particular Python package at a time. A compat package would have to conflict with the “main” `python-uv-build` package because both would be installed in e.g. `/usr/lib64/python3.14/site-packages/uv_build` and would provide e.g. `python3dist(uv-build)` and `/usr/bin/uv-build`. (The binary name could be patched with some care, but the Python package name couldn’t.) We try to avoid conflicts like this [whenever possible](https://docs.fedoraproject.org/en-US/packaging-guidelines/Conflicts/#_compat_package_conflicts), although they are sometimes necessary. As with C and C++ `-devel` packages, the conflict would be much more tolerable because there should be few or no packages that depend on `uv-build` at runtime.

In any case, plenty of Python package upstreams are slow-moving at best, so we would still end up having to patch many of the dependent packages ourselves in order to avoid maintaining `python-uv-build` compat packages indefinitely. Given that, I anticipate that we would choose instead to do all the patching up front and avoid making a compat package in the first place. We would probably ask dependent packages to carry downstream patches (perhaps `sed`-patches) removing the upper bound on the version of `uv_build`, so that we don’t have to touch *every package* that uses `uv_build` on *every* SemVer-breaking `uv_build` update. This would still be a game of whack-a-mole to add and maintain patches, but it’s probably the least burdensome approach available to us. We would still impact-check SemVer-breaking updates rather than shipping them blindly, of course.

So far, we have exactly one package in Fedora built with `uv_build`, but I expect that it will not be long before we see many more.

I hope that adds some useful context without excessively prolonging the discussion in this issue.

musicinmybrain: > Going back to the original comment, `flit_core` also advises using an upper bound. And downstream maintainers seem to handle that okay?

It’s relevant that Flit / `flit_core` has been at 3.x for nearly five years. Similarly, Poetry / `poetry-core` 2.x was a significant one-time effort that will reach users in Fedora 43. SemVer-breaking updates in `uv` / `uv-build` are still happening every couple of months or so. This is understandable, given the project’s very active development and the entirely laudable level of care taken around potentially-breaking changes, but it does help explain why we are a bit more anxious about the potential downstream compatibility work.

notatallshaw: > It’s relevant that Flit / `flit_core` has been at 3.x for nearly five years.

That's likely to change soon: https://github.com/pypa/flit/issues/751#issuecomment-3064967232

hroncok: Also note that we could package several different uv-build versions in Fedora (each conflicting with the others), but then, when we start building with Python 3.15 in October and uv-build potentially fails a test with it, we need to backport the fix to all N uv-build packages we build and ship.

So, yes, it's possible, but if we can avoid that, it would be better.

zanieb: > when we start building with Python 3.15 in October and uv-build potentially fails a test with it, we need to backport the fix to all N uv-build packages we build and ship.

This seems very unlikely — `uv-build` only exposes the minimum required Python interface for spec-compliance, the implementation does not depend on Python.

hroncok: Perhpas it will fail with a new Rust compiler released in 10 years. The exact reason is not relevant, the reality is that maintaining a dozen different versions of a Python package in Fedora is not desirable. We don't just build it once and then forgot about it, it needs to keep being buildable and testable.

geofft: Ah, thanks, the point about Python being a single import namespace (and distros being geared towards a single-ish global namespace, instead of per-build isolated venvs) is a good one.

Because there's already an upper bound for another build backend (#1837) I think we're getting heavily off-topic for this PR. I opened astral-sh/uv#15000 about this, let's discuss further there.

webknjaz: It's commonly understood within PyPA, downstreams and surrounding communities that upper version bounds are harmful in many contexts (with rare exceptions).

It's probably a bit of a tribal knowledge, though, seeing that it's only mentioned in the context of python version, even in PyPUG (https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires / https://packaging.python.org/en/latest/guides/dropping-older-python-versions/#specify-the-version-ranges-for-supported-python-distributions).

I think it's a bad idea to show examples with upper bounds in generic docs (like PyPUG). Somebody would have to keep bumping the pins so the examples are up-to-day and nobody's up for committing to doing this. It's not a problem with lower bounds, obviously.

If the backends want to restrict this, they would do this in their own docs. But many people land in PyPUG from search engines and copy snippets into their new projects w/o any clue of the long-term consequences. I think it's our shared responsibility to avoid making lives of the downstream maintainers difficult. Showing such examples on PyPUG would work against them and will have negative consequences for years to come. I don't want that. And this is exactly why I invited downstreams into the review — most people around only represent upstreams exclusively.

If anything, we should drop the upper bound from the Flit example instead. But for consistency, we shouldn't show examples with upper bounds to the general public. Adding them should be a well-thought choice in each specific instance.

Perhaps, we could have a small admonition below examples, mentioning that some backends want upper bounds but that's harmful for downstreams. But the examples themselves would not show copy-pasteable snippets with such limits.

The build backends could make the situation better by contributing to better reproducibility standards ecosystem-wide, like this discussion: https://discuss.python.org/t/how-can-build-backends-avoid-breaking-users-when-they-make-backwards-incompatible-changes/85847.

geofft: I don't think encouraging users to drop upper bounds is helpful when a build backend is not committing to API stability. That's harmful for downstreams, too - the build-dependency resolution process will succeed, but the package will be incorrectly built, or fail to build.

I've run into this myself as a downstream, vendoring packages into a corporate monorepo - I believe it was with setuptools (I can't remember for sure, maybe it was something else), which would do discovery in older versions that it wouldn't do in newer versions. We upgraded setuptools treewide, builds passed, and a couple of packages "built" but with no files in the wheel, and we only noticed when users reported their code breaking. I would have _much_ preferred seeing the upper bound and knowing to pay attention to potential incompatibilities!

In other words - I think pushing back on upper bounds here is effectively encouraging build backends to be less proactive/careful about backwards compatibility (i.e. a backend that says "I'm pretty sure this won't break anyone, I'll just ship the change" is not receiving any pushback, but a backend that says "I'm worried this might break and so I'll tell users to use upper bounds" is told that it's doing something wrong). I don't think that's good for downstreams or anyone else, and I don't think that's the message you want to communicate here.

I understand that the goal is for people to not break backwards compatibility at all, and I think everyone agrees on that. :) If a backend does manage not to break backwards compatibility, then downstream packagers can carry very tiny patches to bump the bounds and make no other changes. As I mentioned in that uv issue, I think this is indeed the intention of `uv_build`.

geofft: Also to be precise I'm talking here about packages that are only used as build dependencies. For normal dependencies, I agree that upper bounds are an anti-pattern and you should aim never to break backwards compatibility, because that causes diamond dependency problems. If you must, release a new package under a new dist name and importable name to avoid the diamond dependency. I also agree that this isn't communicated enough (and arguably PEP 440's support for semver/compatible release makes people think that doing breaking changes in major versions is okay).

But for packages that are used only as build dependencies, this argument doesn't apply. The PyPA ecosystem's tools default to build isolation and strongly encourage it, and so diamond dependencies don't happen. They only come up if you're in an ecosystem that doesn't do build isolation—which is a position I've been in and very much sympathize with, but I don't think that PyPA docs should encourage designing around that constraint.

webknjaz: Perhaps, we should wait for the uv-side discussion to happen, then? Are you suggesting we don't add any admonition at all? Maybe encourage the readers to update their build backend regularly?

---

On a related note, a few Packaging Summit attendees at EuroPython were talking about ways to include the exact versions of build backend into the dists somewhere in the packaging metadata. This might be worth exploring in the context of giving rebuilders the context regarding "what actually works upstream".

zanieb: See also https://discuss.python.org/t/how-can-build-backends-avoid-breaking-users-when-they-make-backwards-incompatible-changes/85847

mgorny: I think @webknjaz is making a very important point, though it's not really limited to this particular case — adding upper bounds to snippets is likely going to result in people thoughtlessly copying these upper bounds to their own projects, without even considering what the newest version is right now, and effectively pinning new projects to severely outdated build backend versions. What's worse is that these snippets tend to be copied around and republished, so even if you commit to keeping them up-to-date, you'd eventually have a lot of copies all over the place.

I'm saying this as a person who sent dozens of pull requests to remove `wheel` from `build-system.requires` from lots of new projects — because people still manage to find some outdated snippet to copy it from.

sinoroc: Should there be any version constraint at all? For the most common cases isn't it fine to declare all build dependencies without version constraint? Isn't it typically much easier to handle cases where version constraints are too lax compared to cases where the version constraints are too strict?

Also maybe we should wait until `uv_build` becomes more mature and see if version constraints really do matter or not for this build backend. What's the rush?

zanieb: >  For the most common cases isn't it fine to declare all build dependencies without version constraint? 

I don't think it's actually fine. There are numerous discussions about the problems this causes because... then build backends can't make breaking changes or, when they do (as `setuptools` recently did), they break the whole ecosystem because locking build dependencies isn't available in the Python ecosystem. The lack of bounds is justified by concerns about Python version forward compatibility, which... just isn't a concern for `uv-build`. These topics are discussed in the linked discuss.python.org thread.

The actually think the most compelling argument against upper bounds for `uv-build` here is that nobody is going to bother updating it because the process is too contentious.

zanieb: (That's a bit snarky... but I genuinely do believe it'll go stale unless the uv team has a stake in maintaining it)

sinoroc: I do not think it is snarky at all. Generally, not just for build dependencies, I do not think having real version constraints in this guide makes sense, because indeed they can not be kept up to date. There should be a very obvious note leading to an explanation of why one should set version constraints with examples of how to write such version constraints and an advice to go check each project's own documentation regarding their versioning policies.

notatallshaw: > , I do not think having real version constraints in this guide makes sense

When the advise from those projects is to add constraints the packaging guide should not provide conflicting advise. Creating two conflicting streams of advise is going to create user confusion.

> because indeed they can not be kept up to date

I see regular PRs to update the setuptools lower bound, why would this be any different?

> There should be a very obvious note leading to an explanation of why one should set version constraints with examples of how to write such version constraints and an advice to go check each project's own documentation regarding their versioning policies.

Every other build back end has lower bound version constraints, flit has lower and upper: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

Why are you adding this requirement for uv_build when it's not been required to any other build back end?

sinoroc: I should refrain from posting when I do not have the time to post clear messages...

By "real version constraints", I mean version constraints with actual numbers. I wonder if instead of writing `uv_build >= 0.7.19, <0.9.0` we could write `uv_build >= M.m.b, < M+1.0.0` and then explain what that means, why it might be useful, and recommend users to refer to the project's documentation for more details to figure out what are the best current actual numbers. And of course, we would do that for all build back-ends (not just for Uv) and wherever else in the guide we might have such things.

But if people are willing to update those version strings in the guide all the time, it is fine by me.

ncoghlan: Getting back to the specific question of *this* PR review (rather than the more general ones of showing exact version numbers and upper bounds in build dependencies):

* mentioning `uv-build` is a reasonable change to make
* the exact build backend version number problem is an existing issue, not new in this PR
* the upper bound problem is an existing issue (due to `flit`), not new in this PR

That says to me that we should define a new PyPUG issue regarding the maintainability of those config snippets, and avoid having that concern block *this* PR.

I've created that issue here: #1886 

@webknjaz Given the creation of the dedicated issue, are you happy to consider the question resolved for the purpose of adding `uv-build` to the example backends?

ncoghlan (approved):
LGTM, but I'll give @webknjaz time to consider my suggestion on the version number/bounds maintainability question.

source/shared/build-backend-tabs.rst
webknjaz: @ncoghlan I'm worried that accepting this as is would make the existing problem worse. And doing so is too impactful to just disregard it. The downstreams have expressed legitimate concern. Perhaps, we should address that issue you filed first so this PR wouldn't be as harmful?

---

@notatallshaw

> > because indeed they can not be kept up to date
>
> I see regular PRs to update the setuptools lower bound, why would this be any different?

This is different because the example with setuptools is adding new features and when documenting them, the bound has to be updated for things to work. Inevitably.

But the same isn't happening when things break. Nobody comes in to update the guide documenting breakages and adjusting related bounds. And so there's no process for when this would be happening.

> > There should be a very obvious note leading to an explanation of why one should set version constraints with examples of how to write such version constraints and an advice to go check each project's own documentation regarding their versioning policies.
>
> Every other build back end has lower bound version constraints, flit has lower and upper: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/
>
> Why are you adding this requirement for uv_build when it's not been required to any other build back end?

Lower bounds are fine. Flit was mostly missed. It should also have an admonition explaining that upper bounds should be updated often. Maybe this admonition makes sense to have below all the examples, not selectively.

ncoghlan: I guess my view is that the PR itself isn't really making things worse, it's instead making this existing problem with the status quo more obvious.

Omitting the upper bounds creates an "old project releases may not build" scenario that affects potentially everyone, whereas including them has a more limited impact specifically on "unified build environment" scenarios (which could potentially be mitigated by build tooling changes, such as an option to ignore upper bounds in build dependency declarations and see if anything *actually* breaks with the newer build tool versions).

webknjaz:
Alright, let's give it a go.

source/shared/build-backend-tabs.rst
webknjaz: @mgorny @hroncok while merging this didn't exactly cater to your needs, I'd like to suggest discussing a systemic change to adding mentions of what's important for downstreams throughout PyPUG, if you're interested. Maybe, some admonitions linking to #1791 where relevant.

konstin: I've filed a PR to automatically update the `uv_build` version range, to address the concern that the bounds would go stale: https://github.com/pypa/packaging.python.org/pull/1899

PR #1881, opened 2025-07-22
Ignore linkcheck for ATLAS URL

cbrnr:
This should prevent 403 errors.


cbrnr:
OK, I think I've fixed all errors...

webknjaz:
Why does this have so many unrelated changes? Is it possible to make the diff small?

woodruffw:
It looks like @cbrnr went ahead and fixed some of the other URLs too (which were probably over-matching due to `.` being interpreted in regexes). I don't have a strong opinion on whether those changes should be included or not though.

cbrnr:
Exactly, the `linkcheck_ignore` items are all regexes, and previously dots were not escaped, so matched any character. This is usually not really a problem in practice, but I thought while I'm at it I might as well fix all entries.

webknjaz:
I tend to prefer atomic PRs. The title says it fixes one single URL but what I see in the diff is a bunch of other URLs with zero justification. Plus refactoring. Maybe something I missed in between these too.

PR #1882, opened 2025-07-27
Add "namespace package" to glossary

thejcannon:
I'm a masochist, so here I go mucking around in the world of namespace packages. Really, I just wanted to link to a definition in an upcoming blog post, and realized there wasn't one.

I'm not the best wordsmith so happy to take suggestions.

(I was thinking we _could_ include definitions for "explicit" or "implicit", but incremental improvement is still improvement)

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -187,6 +187,13 @@ Glossary
         types: :term:`Pure Module`, or :term:`Extension Module`.
 
 
+    Namespace Package
+
+        A mechanism for splitting a single :term:`Import Package <Import Package>`
+        across multiple directories on disk. Can either be "explicit" or "implicit",
+        depending on the mechanism. Defined in :pep:`420`.
+
+
     Package Index
 
         A repository of distributions with a web interface to automate


source/glossary.rst
```diff
@@ -187,6 +187,13 @@ Glossary
         types: :term:`Pure Module`, or :term:`Extension Module`.
 
 
+    Namespace Package
+
+        A mechanism for splitting a single :term:`Import Package <Import Package>`
+        across multiple directories on disk. Can either be "explicit" or "implicit",
```
sinoroc:
```suggestion
        across multiple directories on disk. Can either be "explicit" or "implicit",
```

I guess this does not cover the (nowadays most common?) case where the package is all in one place on disk (i.e. in `site-packages`) but was installed from different distribution packages (i.e. from different packages/libraries on PyPI for example).

thejcannon: I updated the PR description to include (because I forgot to mention) most of this is just lifted straight from the opening line of [PEP 420](https://peps.python.org/pep-0420/):

> Namespace packages are a mechanism for splitting a single Python package across multiple directories on disk.

That being said, what do you think about this change:

```suggestion
        across one or more directories on disk. Can either be "explicit" or "implicit",
```

thejcannon: Actually, I just found [this](https://docs.python.org/3/reference/import.html#namespace-packages) so "disk" is incorrect.

ncoghlan: Aye, it's import locations rather than specifically directories on disk. The language reference definition covers implicit namespace packages, while the [`pkgutil.extend_path`](https://docs.python.org/3/library/pkgutil.html#pkgutil.extend_path) docs cover explicit namespace packages.

Having a glossary entry here that links to both of those reference definitions (as well as summarising them) is a good suggestion.

thejcannon: I ended up just cleaning this up and omitting "implicit" and "explicit" (for now). Those can be added as alternative entries and linked to in a follow-up.

LMK if anyone wants to push back on that and I can try and finagle it here.

thejcannon:
Happy to close if that one merges!

PR #1883, opened 2025-07-31
Convert linkcheck_ignore entries to regexes

cbrnr:
As the title says, this PR converts `linkcheck_ignore` entries to regexes so that the URLs really only match exactly (otherwise, a dot is interpreted as "any character").

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1883.org.readthedocs.build/en/1883/


webknjaz (approved):
Thank you!

PR #1890, opened 2025-08-04
Fix broken setuptools-scm link

edgarrmondragon:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1890.org.readthedocs.build/en/1890/

### DIFF
--- a/source/discussions/setup-py-deprecated.rst
+++ b/source/discussions/setup-py-deprecated.rst
@@ -115,7 +115,7 @@ A possible replacement solution (among others) is to rely on setuptools-scm_:
 
 * ``python -m setuptools_scm``
 
-.. _setuptools-scm: https://setuptools-scm.readthedocs.io/en/latest/usage/#as-cli-tool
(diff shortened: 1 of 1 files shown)


source/discussions/setup-py-deprecated.rst
```diff
@@ -115,7 +115,7 @@ A possible replacement solution (among others) is to rely on setuptools-scm_:
 
 * ``python -m setuptools_scm``
 
-.. _setuptools-scm: https://setuptools-scm.readthedocs.io/en/latest/usage/#as-cli-tool
+.. _setuptools-scm: https://setuptools-scm.readthedocs.io/en/latest/usage.html#as-cli-tool
```
ncoghlan:
```suggestion
.. _setuptools-scm: https://setuptools-scm.readthedocs.io/en/latest/usage#as-cli-tool
```

If we omit both the `/` *and* the `.html`, that seems the most likely to work consistently even if the build layout temporarily changes.

ncoghlan (changes requested):

PR #1898, opened 2025-08-21
Correct regex for metadata 'name' format

di:
The current regex permits strings that end in the newline character, which is counter to what the description for the field states ("ASCII letters and numbers, period, underscore and hyphen").

This updates the regex to use `\Z` instead of `$` to match at the end of the string and exclude newline characters.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1898.org.readthedocs.build/en/1898/


source/specifications/name-normalization.rst
```diff
@@ -17,7 +17,7 @@ underscore and hyphen. It must start and end with a letter or number.
 This means that valid project names are limited to those which match the
 following regex (run with :py:data:`re.IGNORECASE`)::
 
-    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$
+    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])\Z
```
ncoghlan, line 20: The history block below should get an entry noting the bug fix in the regex. Something like:

```
- August 2025: The suggested name validation regex was fixed to match the field specification
  (it previously finished with ``$`` instead of ``\Z``, incorrectly permitting trailing newlines)
```

miketheman (approved):

brettcannon (approved):

PR #1899, opened 2025-08-25
Update uv_build version automatically

konstin:
In https://github.com/pypa/packaging.python.org/pull/1880, the concern was raised that the uv_build upper bound in the docs will go stale. This PR adds a GitHub Actions workflow that automatically updates the version daily with the latest uv(-build) version.

I tested this change on my fork, but I unfortunately can't test this in pypa/packaging.python.org itself. You can see an example pull request created by this action in https://github.com/konstin/packaging.python.org/pull/2.

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -5,10 +5,36 @@ name: Cron
 on:
   schedule:
   - cron: "0 6 * * *"  # daily at 6am
+  workflow_dispatch:
 
 jobs:
   test:
     if: github.repository_owner == 'pypa'  # suppress noise in forks
     uses: ./.github/workflows/test.yml
+
+  update-uv-build-version:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
+      - name: Update uv_build version
+        id: update_script
+        run: uv run scripts/update_uv_build_version.py
+      - # If there are no changes, no pull request will be created and the action exits silently.
+        name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update uv_build version to ${{ steps.update_script.outputs.version }}
+          title: Update uv_build version to ${{ steps.update_script.outputs.version }}
+          body: |
+            Automated update of uv_build version bounds for uv ${{ steps.update_script.outputs.version }}.
+            
+            This PR was created automatically by the cron workflow, ping `@konstin` for problems.
+          branch: bot/update-uv-build-version
+          delete-branch: true
 
 ...
--- a/scripts/update_uv_build_version.py
+++ b/scripts/update_uv_build_version.py
@@ -1,0 +1,56 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "httpx>=0.28.1,<0.29",
+#     "packaging>=25.0",
+# ]
+# ///
+import re
+from pathlib import Path
+
+import httpx
+from packaging.utils import parse_wheel_filename
+from packaging.version import Version
+
+
+def main():
+    response = httpx.get(
+        "https://pypi.org/simple/uv-build/",
+        headers={"Accept": "application/vnd.pypi.simple.v1+json"},
+    )
+    response.raise_for_status()
+    data = response.json()
+    current_release = None
+    for file in data["files"]:
+        if not file["filename"].endswith(".whl"):
+            continue
+        _name, version, _build, _tags = parse_wheel_filename(file["filename"])
+        if version.is_prerelease:
+            continue
+        if current_release is None or version > current_release:
+            current_release = version
+
+    [major, minor, _patch] = current_release.release
+    if major != 0:
+        raise NotImplementedError("The script needs to be updated for uv 1.x")
+    upper_bound = Version(f"{major}.{minor + 1}.{0}")
+
+    repository_root = Path(__file__).parent.parent
+    existing = repository_root.joinpath("source/shared/build-backend-tabs.rst").read_text()
+    replacement = f'requires = ["uv_build >= {current_release}, <{upper_bound}"]'
+    searcher = re.compile(re.escape('requires = ["uv_build') + ".*" + re.escape('"]'))
+    if not searcher.search(existing):
+        raise RuntimeError("Could not `uv-build` entry")
+    updated = searcher.sub(replacement, existing)
+
+    if existing != updated:
+        print("Updating source/shared/build-backend-tabs.rst")
+        Path("source/shared/build-backend-tabs.rst").write_text(updated)
+        print(f"::set-output name=version::{current_release}")
+        print(f"::set-output name=updated::true")
+    else:
+        print("Already up-to-date source/shared/build-backend-tabs.rst")
+        print(f"::set-output name=updated::false")
+
+if __name__ == '__main__':
+    main()


scripts/update_uv_build_version.py
```diff
@@ -0,0 +1,56 @@
...
+    replacement = f'requires = ["uv_build >= {current_release}, <{upper_bound}"]'
+    searcher = re.compile(re.escape('requires = ["uv_build') + ".*" + re.escape('"]'))
+    if not searcher.search(existing):
+        raise RuntimeError("Could not `uv-build` entry")
+    updated = searcher.sub(replacement, existing)
+
+    if existing != updated:
+        print("Updating source/shared/build-backend-tabs.rst")
+        Path("source/shared/build-backend-tabs.rst").write_text(updated)
+        print(f"::set-output name=version::{current_release}")
```
webknjaz: This syntax was deprecated due to security vulnerabilities in GitHub Runners years ago. Use the new syntax.

scripts/update_uv_build_version.py
```diff
@@ -0,0 +1,56 @@
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "httpx>=0.28.1,<0.29",
+#     "packaging>=25.0",
```
webknjaz, line 5: Since this script is app-like and not a library for reuse, I'd probably go for exact pins. Or perhaps use a pylock file to pin the dep tree — this would be even better!

scripts/update_uv_build_version.py
```diff
@@ -0,0 +1,56 @@
+# /// script
+# requires-python = ">=3.12"
```
webknjaz: Additionally, we have a convention of making specs look sparse in PyPUG for the sake of readability.
```suggestion
# requires-python = ">= 3.12"
```

.github/workflows/cron.yml
```diff
@@ -5,10 +5,36 @@ name: Cron
...
   schedule:
   - cron: "0 6 * * *"  # daily at 6am
+  workflow_dispatch:
 
 jobs:
   test:
     if: github.repository_owner == 'pypa'  # suppress noise in forks
     uses: ./.github/workflows/test.yml
 
+  update-uv-build-version:
```
webknjaz: In my experience, frequent runs cause a lot of noise (which is most noticeable for people with access to hundreds of repos). It's best to do this at least quarterly. So it needs to move into a separate workflow.

webknjaz: Additionally a similar fork check is needed.

.github/workflows/cron.yml
```diff
@@ -5,10 +5,36 @@ name: Cron
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v5
```
webknjaz: Missed this originally, but each step must have a name.

scripts/update_uv_build_version.py
konstin: Are there currently any implementations that support pylock files for running PEP 723 scripts?

I can offer a uv script lockfile, but naturally that's tool specific. It is use with the `uv run` in the GitHub workflow FWIW.

I wouldn't use exact pins here as they don't apply to transitive dependencies.

.github/workflows/cron.yml
konstin: We release frequently, and at least for major releases it shouldn't take three months to update; The period between latest two uv major releases was slightly less than 3 months.

If the packaging.python.org maintainers feel that e.g. weekly update with the uv releases are too much, we can make the script have only check for major releases and ignore lower bound changes for minor releases otherwise.

scripts/update_uv_build_version.py
konstin: I'm not sure what the recommended use for the new syntax from within a Python script is (the docs only show a `run: ` step), I've updated it with examples I found online.

konstin:
> OTOH, I'd probably prefer this data extraction to be an in-tree Sphinx extension. It could get fresh versions on build and make use of Sphinx's built-in caching mechanism.

I can contribute a GitHub Actions workflow, but I can't contribute a Sphinx extension. To me, the workflow has the advantage that it allow explicit review and doesn't need an external service on build.

.github/workflows/update-uv-build-version.yml
```diff
@@ -0,0 +1,42 @@
+---
+
+name: Cron
```
ncoghlan:
```suggestion
name: Update uv build version
```

scripts/update_uv_build_version.py
ncoghlan: We don't currently have a full transitive lock even for the docs build deps, so I wouldn't consider transitive locking a blocking requirement here.

scripts/update_uv_build_version.py
woodruffw: I _think_ you should be able to use `GITHUB_OUTPUT` here, i.e. something roughly like:

```python
with open(os.getenv("GITHUB_OUTPUT"), mode="w+") as io:
    print("...", file=io)
```

webknjaz: Something like that. I usually go for `mode="a"`: https://github.com/cherrypy/cheroot/blob/ac9b6e5543278439312404aa7a0d83f6b5f80981/.github/workflows/ci-cd.yml#L184-L189

.github/workflows/cron.yml
```diff
@@ -5,6 +5,7 @@ name: Cron
 on:
   schedule:
   - cron: "0 6 * * *"  # daily at 6am
+  workflow_dispatch:
```
webknjaz: This doesn't seem necessary.

scripts/update_uv_build_version.py
```diff
@@ -0,0 +1,64 @@
...
+    replacement = f'requires = ["uv_build >= {current_release}, <{upper_bound}"]'
+    searcher = re.compile(re.escape('requires = ["uv_build') + ".*" + re.escape('"]'))
+    if not searcher.search(existing):
+        raise RuntimeError("Could not `uv-build` entry")
+    updated = searcher.sub(replacement, existing)
+
+    if existing != updated:
+        print("Updating source/shared/build-backend-tabs.rst")
+        Path("source/shared/build-backend-tabs.rst").write_text(updated)
+        if github_output := os.environ.get("GITHUB_OUTPUT"):
```
webknjaz, line 52: I don't think a conditional is needed if it's always expected to run within GHA.

scripts/update_uv_build_version.py
webknjaz: Alright, marking this as resolved.

.github/workflows/update-uv-build-version.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+      - name: Update uv_build version
+        id: update_script
+        run: uv run scripts/update_uv_build_version.py
+      - # If there are no changes, no pull request will be created and the action exits silently.
+        name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
webknjaz, line 32: Using this special token will make it so the CI (or any other integrations) won't get triggered. This will result in a PR that is forever blocked (unless a human comes in and closes+reopens it). This isn't great UX. One workaround I picked up from GH is creating PRs as draft, adding a comment telling a human to undraft and configuring the main CI workflow to react to "ready for review".

.github/workflows/cron.yml
webknjaz: Personally, for me it's a lot of noise. It'll just pile up more stuff in my inbox that will contribute to ignoring this repo, postponing reviews for longer. We can't keep up with already open PRs, adding more, even just once in 3 months isn't going to make the maintainers more responsive but itstead contribute to the underlying problem.

woodruffw: I agree about the volume, but I think the solution here is to maybe add more reviewers on this repo: I'd personally be happy to review a bit of additional extra volume here, since I already do periodic automated PR triage on my own PyPA repos. 

To that end: is there a way we can get myself or another PyPA person as another committer on this repo (assuming that's OK by the existing committers)? We could additionally keep the noise down by adding a `CODEOWNERS` file that only selectively assigns people to e.g. automated PRs.

webknjaz: Yeah, I've been just thinking of adding you.. The notifications would still be noisy for me but it's better than no reviews at all.

woodruffw: I'm happy to help however I can!

(Related: I wish GitHub actually allowed site-level notifications to be filtered by CODEOWNER like that, I have similar problems on other orgs. Oh well.)

.github/workflows/update-uv-build-version.yml
webknjaz: @konstin here's what's necessary to implement that draft-based UX if you're up for it: https://github.com/ansible/awx-plugins/blob/48664b8254527fa8c65dfe32c7f4d410e62db106/.github/workflows/pip-tools.yml#L372-L422 + https://github.com/ansible/awx-plugins/blob/48664b8254527fa8c65dfe32c7f4d410e62db106/.github/workflows/ci-cd.yml#L18-L22.

.github/workflows/cron.yml
webknjaz: Yep, that's something to complain about in the @maintainers community.. It's growing exponentially the more high-profile projects you gain access too. I start my Mondays by removing dependabot PR emails from my inbox across many orgs.

.github/workflows/update-uv-build-version.yml
woodruffw: Huh, TIL that the "create as draft" trick works -- I've been manually cycling PRs through "closed" to get them to trigger the CI for years 😅

scripts/update_uv_build_version.py
woodruffw: +1, I'm okay with letting this fail with a `KeyError` or similar if `GITHUB_OUTPUT` isn't set.

(Another option that I sometimes do is have `GITHUB_OUTPUT` fall back to `sys.stderr` for debugging. But that seems low-priority here 🙂)

woodruffw:
> Maybe other maintainers would want to keep up with these PRs but I have a lot of other projects that need attention.

I'll go ahead and take responsibility for these PRs -- I agree with your rationale about a Sphinx extension in the long term, but I'm okay with taking on the review responsibility/load for this in the medium term as a measure of expedience 🙂

.github/workflows/cron.yml
konstin: I added it as a feature to manually trigger an update if we want one below the monthly cycle for whatever reason, it's also useful for testing.

scripts/update_uv_build_version.py
konstin: You can run this script locally too :) It makes developing on it easier, but if we want the assurance that this can't slip in CI I can change it to unconditional too.

.github/workflows/update-uv-build-version.yml
zanieb: Oh that's cool. We should take that :D

konstin: Oh that's a tough problem. I've toggled to create PRs a draft for now. https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs has a list of options, I can additionally do the PR comment as in https://github.com/ansible/awx-plugins/blob/48664b8254527fa8c65dfe32c7f4d410e62db106/.github/workflows/pip-tools.yml#L415-L422 or we can use PAT.

woodruffw: This looks good to me 🙂 -- I'm personally OK without an extra bot comment, since we could always have the PR body itself mention the CI trigger step.

(But I think it's totally fine as is, since there's a nice forcing function here where the user has to undraft the PR to merge anyways.)

woodruffw (approved):
Thanks @konstin! I'm going to do a last pass on this tonight, but otherwise this LGTM. I'll also see about the link check failure then.

(For posterity, I'll record here as well that I'll take ownership of triaging these automated PRs as they come in.)

.github/workflows/update-uv-build-version.yml
webknjaz: > Huh, TIL that the "create as draft" trick works -- I've been manually cycling PRs through "closed" to get them to trigger the CI for years 😅

I actually stole this idea from one of the CodeQL repos owned by GH itself. A year or two ago, I think. I've been using it in like 3-4 repos for a while now, and it does make the UX somewhat bearable. In one repo, I still have a despicable hack of using `${{ github.token }}` to create a PR and a repo-bound SSH deployment key to perform a hacky push that would propagate events 🤪 (https://github.com/cherrypy/cheroot/blob/ac9b6e5543278439312404aa7a0d83f6b5f80981/.github/workflows/pip-tools.yml#L313-L377).

One other possibility is to use a GH App with its secrets stored in the repo. I have a better solution in mind for when I get to playing with GH Apps again — it should be possible to have a service that would allocate tokens based on OIDC (scoped to a dedicated job, of course). I just never got to experimening with this idea.

> or we can use PAT.

I don't like using PAT for CI in general. Those tokens expire, they need to be rotated. Somebody with a privilige higher level would have to update these secrets in the repo settings. I don't even have this kind of access myself. It'll be a can of worms.

> This looks good to me 🙂 -- I'm personally OK without an extra bot comment, since we could always have the PR body itself mention the CI trigger step.

I've shown this to some Ansible community folks and they figured that having additional comments might be noisy and also suggested just having this trigger mention in the PR description.

Note that if a PR isn't getting merged, the automation would need to push commits to the same branch (likely force-push). This will have an effect of not triggering the CI and so for such updates, the PR must be turned back into a draft. This can accumulate a bunch of duplicate comments.

> (But I think it's totally fine as is, since there's a nice forcing function here where the user has to undraft the PR to merge anyways.)

Yep. The only downside is that you can't click "auto-merge" until you undraft it. One must be aware of it.

Additionally, this will cause double-runs if contributors send PRs in draft (or draft them in the process). I think this might be unexpected for some but definitely better: their `pull_request` `synchronize` event might be days apart from `pull_request` `ready_for_review` so undrafting might trigger testing against newer main branch.

.github/workflows/cron.yml
webknjaz: But this is a different workflow. Having it in the update one would be enough.

woodruffw: Agreed, I'll remove it from this one.

scripts/update_uv_build_version.py
webknjaz: @konstin I usually do `GITHUB_OUTPUT=/dev/stderr ./my-script.sh` locally when testing various GHA-related stuff, FTR. Would that be a good middle ground?

.github/workflows/update-uv-build-version.yml
```diff
@@ -0,0 +1,43 @@
...
+          commit-message: Update uv_build version to ${{ steps.update_script.outputs.version }}
+          title: Update uv_build version to ${{ steps.update_script.outputs.version }}
+          draft: true # Trigger CI by un-drafting the PR, otherwise `GITHUB_TOKEN` PRs don't trigger CI.
```
webknjaz, line 35: This is not enough. You must update the `pull_request` trigger *in the CI workflow* to include the undrafting sub-type, it's not there by default. See my earlier examples: https://github.com/pypa/packaging.python.org/pull/1899/files#r2349425964.

woodruffw: Oops, thanks for catching that.

scripts/update_uv_build_version.py
```diff
@@ -0,0 +1,64 @@
...
+        _name, version, _build, _tags = parse_wheel_filename(file["filename"])
+        if version.is_prerelease:
+            continue
+        if current_release is None or version > current_release:
+            current_release = version
+
+    [major, minor, _patch] = current_release.release
+    if major != 0:
+        raise NotImplementedError("The script needs to be updated for uv 1.x")
+    upper_bound = Version(f"{major}.{minor + 1}.{0}")
```
webknjaz: Why have curly braces around that trailing zero?
```suggestion
    upper_bound = Version(f"{major}.{minor + 1}.0")
```

webknjaz: I wasn't even sure this would work but checked that it does.

.github/workflows/update-uv-build-version.yml
woodruffw: PTAL; I used your examples as reference here 🙂

webknjaz: Marking this thread as resolved.

webknjaz (approved):
:shipit:

PR #1900, opened 2025-08-26
Clarify that the ``License-Expression`` field applies to the containing distribution file, not the project itself

brettcannon:


pfmoore (approved):

notatallshaw:
Sorry, I've not read through the glossary recently, but is the distribution file being referred to here the sdist and wheel of the METADATA file in the .dist directory? Or something else?

brettcannon:
> is the distribution file being referred to here the sdist and wheels? Or the METADATA file in the .dist directory? Or something else?

sdist and wheels.

source/specifications/core-metadata.rst
```diff
@@ -473,6 +473,9 @@ Text string that is a valid SPDX
 :term:`license expression <License Expression>`,
 as specified in :doc:`/specifications/license-expression`.
 
+Note that the expression in this field only applies to the **distribution** file
+containing the metadata, not the project itself or other distribution files.
```
befeleme: What "other distribution files" are meant here? Can we clarify in the text? It's confusing to me, even though I'm fairly familiar with the topic.

pfmoore: It means the expression only applies to the file containing it, not other files. IMO it would be clearer if "distribution" wasn't emphasised on its first occurrence, but if you think it needs more than that can you say what you'd like to see?

The key point is that you can only assume the license expression applies to the single file it's read from. And in addition, you can't assume that the "project license" is the same as the license of any one of its individual distribution files.

brettcannon: I tweaked the sentence slightly to try and make it clearer. Otherwise I will need a concrete suggestion of how to change it.

notatallshaw: I was also confused about the term "other distribution files", until I realized it meant other sdists or wheels. As opposed to, say, the files built, installed, or extracted from the distribution.

pfmoore: "distribution file" is a technical term (it's in a glossary somewhere) but it's clumsy and hence not frequently used...

befeleme: My take:

```
Note that this license expression applies only to the specific :term:`Distribution Archive` that contains the metadata
(e.g., this particular :term:`source distribution <Source Distribution (or "sdist")>` or :term:`wheel <Wheel>`).
It cannot be assumed to be the license for the project as a whole, nor for other distribution packages of the same project.
```

merwok: Or: _other distributions_

Or even: _other distributions (other wheels or sdists for the same release)_

brettcannon:
Tried to clarify again by taking inspiration from @befeleme on phrasing by inlining essentially examples of what distribution archives are.

pfmoore (approved):

PR #1901, opened 2025-08-27
Clarify that the Dynamic metadata field only applies when building from sdist

pfmoore:
This change was previously discussed, and approved as a clarification, in https://discuss.python.org/t/split-from-pep-639-expressing-project-vs-distribution-licenses-post-pep-639-mod-titled/90314/189 (and the previous comment https://discuss.python.org/t/split-from-pep-639-expressing-project-vs-distribution-licenses-post-pep-639-mod-titled/90314/179).

I'm creating this as draft as I want to wait a few days for any feedback on Discourse before merging.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1901.org.readthedocs.build/en/1901/

### DIFF
--- a/source/overview.rst
+++ b/source/overview.rst
@@ -339,7 +339,7 @@ originated and where the technologies below work best:
 Bringing your own kernel
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-Most operating systems support some form of classical virtualization,
+Most desktop operating systems support some form of classical virtualization,
 running applications packaged as images containing a full operating
 system of their own. Running these virtual machines, or VMs, is a
 mature approach, widespread in data center environments.
@@ -348,9 +348,13 @@ These techniques are mostly reserved for larger scale deployments in
 data centers, though certain complex applications can benefit from
 this packaging. The technologies are Python agnostic, and include:
 
-* `Vagrant <https://www.vagrantup.com/>`_
-* `VHD <https://en.wikipedia.org/wiki/VHD_(file_format)>`_, `AMI <https://en.wikipedia.org/wiki/Amazon_Machine_Image>`_, and :doc:`other formats <openstack:user/formats>`
-* `OpenStack <https://www.redhat.com/en/topics/openstack>`_ - A cloud management system in Python, with extensive VM support
+* KVM on Linux
+* Hyper-V on Windows
+* `VHD <https://en.wikipedia.org/wiki/VHD_(file_format)>`_,
+  `AMI <https://en.wikipedia.org/wiki/Amazon_Machine_Image>`_,
+  and :doc:`other formats <openstack:user/formats>`
+* `OpenStack <https://www.redhat.com/en/topics/openstack>`_ -
+  A cloud management system written in Python, with extensive VM support
 
 Bringing your own hardware
 ^^^^^^^^^^^^^^^^^^^^^^^^^^


source/specifications/core-metadata.rst
```diff
@@ -133,6 +133,18 @@ only, and indicates that the field value was calculated at wheel build time,
 and may not be the same as the value in the sdist or in other wheels for the
 project.
 
+Note in particular that if you have a wheel, you cannot assume that a field
+which is not marked as ``Dynamic`` will have the same value in other wheels, as
+some wheels are not built directly from the sdist, but are modified from
+existing wheels (the ``cibuildwheel`` tool does this, for example). Such
```
oscarbenjamin: I think it would be better to mention auditwheel here rather than cibuildwheel.

The input to cibuildwheel is an sdist/vcs-checkout and the outputs are wheels but it builds each wheel in two steps:

- run pypa/build for PEP 517 build to make non-portable wheels.
- run auditwheel (or delocate etc for other OS) to repair the wheels by bundling shared libraries.

So it is auditwheel that takes wheels as input and outputs wheels (even if cibuildwheel runs auditwheel by default). There are also many projects that don't use cibuildwheel but still use auditwheel.

I don't know if it is preferred not to link to PEPs here but auditwheel was introduced in [PEP 513](https://peps.python.org/pep-0513/#auditwheel) which explains what it is doing and why. There can of course be other reasons for post-processing wheels but in my experience running auditwheel is the most common case by far so PEP 513 is a good further reading for anyone wanting to understand this.

Another point that is perhaps out of place here but would be helpful if someone is actually trying to understand this would be to mention that it is very common practice to use something like auditwheel for PyPI wheels. Libraries with extension modules that depend on any non-Python (e.g. C) external libraries are likely doing this. Note specifically *PyPI* wheels rather than wheels built/used in other situations. Again PEP 513 explains specifically why this is needed for wheels uploaded to PyPI (at least in the Linux case but the same applies to other OS).

pfmoore: Thanks, I forgot which tool it was. You're right, I've changed this to auditwheel. Let me know if the updated text looks OK.

pfmoore: I've not linked to PEP 513 - I don't think we should be linking to this level of explanatory detail, but I'm willing to be persuaded otherwise if more people think it would be helpful.

oscarbenjamin: That looks good.

Perhaps "if you have a wheel" needs clarification. The text above here only really mentions the idea of wheels being built from an sdist.

The relevant situation is that you obtained a prebuilt wheel directly from somewhere (likely PyPI) rather than building a wheel from the sdist yourself.

pfmoore: Updated.

oscarbenjamin: Looks good to me.

source/specifications/core-metadata.rst
```diff
@@ -133,6 +133,19 @@ only, and indicates that the field value was calculated at wheel build time,
...
+in other wheels, as some wheels are not built directly from the sdist, but are
+modified from existing wheels (the ``auditwheel`` tool does this, for example,
+and it's commonly used when building wheels for PyPI). Such modifications
+*could* include changing metadata (even non-dynamic metadata).  Similarly, if
+you have a sdist and a wheel which you didn't build from that sdist, you cannot
+assume that the wheel's metadata matches that of the sdist, even if the field
+is not marked as ``Dynamic``.
+
+It is advisable, but not required, that tools which modify wheel metadata add
+the modified fields to the generated wheel's ``Dynamic`` field.
```
zooba: This feels entirely unnecessary given the above. We might as well say "tools which modify wheel metadata do not need to modify ``Dynamic``", just to reinforce to anyone reading that they can't trust `Dynamic` at all in a given wheel (which is what the above says).

pfmoore: So are you saying it should just be omitted? I'm fine with that. I only added it because of your comment on Discourse "Should you also clarify how non-build backends should handle this?"

zooba: What you were proposing on Discourse wasn't quite as explicit about how users should interpret `Dynamic` in a wheel. Since you are being that explicit here, it's quite clear enough for tools like auditwheel that they aren't obliged to touch other parts of the metadata.

brettcannon (approved):

rgommers (approved):
LGTM and is a very helpful clarification, thanks @pfmoore

PR #1912, opened 2025-09-16
Vagrant link is failing CI, drop it entirely

ncoghlan:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--1912.org.readthedocs.build/en/1912/


ncoghlan:
@webknjaz Rather than ignoring the failing Vagrant link, I decided to just drop it entirely (it was just from the overview section, when talking about VM tech)

PR #1915, opened 2025-09-16
Updates table to contain poetry-core version supporting PEP 639

Hanse00:
I happened to notice that the latest version of poetry does indeed support PEP639 (See: https://python-poetry.org/history/#220---2025-09-14) so I figured this table should be updated 🙂 

FWIW I did also validate myself that the same dummy project build with poetry 2.1.4 and 2.2.0 result in slightly different PKG-INFO outputs as expected.

Old:
```
Metadata-Version: 2.3
License: MIT
```

New:
```
Metadata-Version: 2.4
License-Expression: MIT
```

### DIFF
--- a/source/guides/writing-pyproject-toml.rst
+++ b/source/guides/writing-pyproject-toml.rst
@@ -319,7 +319,7 @@ backend>` now support the new format as shown in the following table.
      - 77.0.3
      - 3.12
      - 2.4.0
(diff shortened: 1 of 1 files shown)


Hanse00:
See also, the closed poetry-core PR listing this change: https://github.com/python-poetry/poetry-core/pull/885

ncoghlan (approved):
Thanks!

PR #1847, opened 2025-04-14
dependency-groups specification cross link Added

Daksh2000:
Fixes: https://github.com/pypa/packaging.python.org/issues/1840

PR #1918, opened 2025-09-27
Convert 'Dependency Groups' to lowercase

sirosen:
Outside of titles and other contexts, convert this term to lowercase.
This usage better matches other terms defined in the packaging specifications, such as "script metadata" and "dependency specifiers".

For reference, this change was inspired by:
https://github.com/pypa/packaging.python.org/pull/1847#discussion_r2266571023

---

I like having some kind of styling which indicates a term.
However, I would like to keep docs consistent. If there's some reasonable choice of styling (e.g., boldface) which can indicate a specialized term, I would be happy to work on applying it consistently across the site.


webknjaz:
```rst
:term:`dependency groups`
```
on first use in a document, followed by italics in the following uses?

I recently discovered that CPython has a great style guide: https://devguide.python.org/documentation/style-guide/#links

alialobidm (approved):

sirosen:
I'll rebase happily when I have some downtime. I'm currently traveling, very limited availability. 

I wasn't clear, btw, if your reference to the Python docs style guide was a request for changes. I'm happy to start using a new style. But I think that would need buy in from other maintainers -- which in my mind means @ncoghlan, unless there are others I don't know about? If the two of you agree on some desired style for technical terms, based on the Python one or otherwise, I'm happy to not only apply it here but also try to make all of the pages uniform in using it.

webknjaz:
@sirosen oh, I forgot you're afk for a while.. I'll accept it as is for now.

I did want to follow the CPython guide but you're right, it can be handled separately.

PR #1923, opened 2025-10-04
Update glossary.rst for accessibility (just added a hyperlink)

mahmoudadelaziz:
I only added a hyperlink to the "source tree" definition at its first occurrence in the Glossary, under the definition of a "Build Backend", for accessibility.

I needed to read this authoritative definition of a "Build Backend" and I found the term "source tree" just a bit confusing, I had to look it up myself in the glossary, which was not a big deal, but it might be for someone else. So, this is merely an attempt to make it slightly more accessible.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1923.org.readthedocs.build/en/1923/

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -160,7 +160,7 @@ Glossary
 
         A string with valid SPDX license expression syntax,
         including one or more SPDX :term:`License Identifier`\(s),
-        which describes a :term:`Project`'s license(s)
+        which describes a :term:`Distribution Archive`'s license(s)
         and how they inter-relate.
         Examples:
         ``GPL-3.0-or-later``,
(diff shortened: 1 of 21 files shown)


ncoghlan (approved):
Thanks!

webknjaz:
@mahmoudadelaziz looks like the term is no longer there or is called differently.

source/glossary.rst
```diff
@@ -14,7 +14,7 @@ Glossary
 
     Build Backend
 
-        A library that takes a source tree
+        A library that takes a :term:`source tree <Project-Source-Tree>`
```
merwok:
```suggestion
        A library that takes a :term:`source tree <Project Source Tree>`
```

merwok: I would also add a link to the next entry, Build Frontend

PR #1936, opened 2025-10-21
Index hosted attestations: Add optional timestamps

jku:
Sigstore public good instance is transitioning to Rekor v2 transparency log: v2 no longer includes integrated time in the log entries so external timestamps are needed.

This extends the attestation spec so that timestamps can be included.

The companion PR to this is https://github.com/pypi/pypi-attestations/pull/143. There is an overall tracking issue in https://github.com/pypi/pypi-attestations/issues/147

### DIFF
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -63,7 +63,7 @@ Versions may be specified according to the rules of the
 :ref:`Version specifier specification <version-specifiers>`. (Note:
 URI is defined in :rfc:`std-66 <3986>`)::
 
-    version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
+    version_cmp   = wsp* '<=' | '<' | '!=' | '===' | '==' | '>=' | '>' | '~='
     version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+
     version_one   = version_cmp version wsp*
     version_many  = version_one (',' version_one)* (',' wsp*)?
@@ -339,7 +339,7 @@ Complete Grammar
 The complete parsley grammar::
 
     wsp           = ' ' | '\t'
-    version_cmp   = wsp* <'<=' | '<' | '!=' | '==' | '>=' | '>' | '~=' | '==='>
+    version_cmp   = wsp* <'<=' | '<' | '!=' | '===' | '==' | '>=' | '>' | '~='>
     version       = wsp* <( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+>
     version_one   = version_cmp:op version:v wsp* -> (op, v)
     version_many  = version_one:v1 (',' version_one)*:v2 (',' wsp*)? -> [v1] + v2
@@ -529,6 +529,8 @@ History
 - August 2025: The suggested name validation regex was fixed to match the field
   specification (it previously finished with ``$`` instead of ``\Z``,
   incorrectly permitting trailing newlines)
+- December 2025: Ensure ``===`` before ``==`` in grammar, to allow arbitrary
+  equality comparisons to be parsed.
 
 
 References


source/specifications/index-hosted-attestations.rst
```diff
@@ -85,6 +85,13 @@ object is provided as pseudocode below.
       and certificate.
       """
 
+      timestamps: list[bytes] | None
```
woodruffw: Thinking out loud: can we make this `list[bytes]` only, with an empty list signaling the equivalent of the `None` state? I think it's nice to have as few "null" states as possible 🙂 

(This is the PEP 740 format, so we're not constrained like the Sigstore bundle is in terms of valid-but-representable states.)

jku: I can do that ...  what I was thinking of here was a sort of  backwards compatibility nod towards current attestations that don't have `timestamps` field at all but I suppose I could just mention this in a comment

source/specifications/index-hosted-attestations.rst
```diff
@@ -85,6 +85,13 @@ object is provided as pseudocode below.
...
+      """
+      Optional list of base64 encoded RFC3161 timestamp responses. Timestamps
+      are optional if `transparency_entries` only contains entries with an
+      integrated_time.
+      """
+
```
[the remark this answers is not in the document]
jku: I've added a bit of detail in the verification section below

source/specifications/index-hosted-attestations.rst
woodruffw: Oh, that's a good point -- I hadn't thought about the existing attestations. Let's keep this as is, then!

woodruffw:
> * Are the only potential incompatibilities within `pypi-attestations` project or should I look elsewhere too?

That's the main place, yeah 🙂 

> * The attestation has a version (1), should that be bumped? the pypi-attestations code does not need it but at least I could then mention in a comment that version 1 did not have the timestamps field...

Yeah, I think we probably _should_ bump it -- it's technically a semantic change to verification.

source/specifications/index-hosted-attestations.rst
```diff
@@ -43,7 +43,10 @@ object is provided as pseudocode below.
   class Attestation:
       version: Literal[1]
```
woodruffw, line 46: Should this be:

```suggestion
      version: Literal[1, 2]
```

based on the comment below?

We could also model this more precisely, i.e. have `type Attestation = AttestationV1 | AttestationV2`.

source/specifications/index-hosted-attestations.rst
```diff
@@ -85,15 +88,25 @@ object is provided as pseudocode below.
...
+      """
+
 A full data model for each object in ``transparency_entries`` is provided in
 :ref:`appendix`. Attestation objects **SHOULD** include one or more
 transparency log entries, and **MAY** include additional keys for other
 sources of signed time (such as an :rfc:`3161` Time Stamping Authority or a
 `Roughtime <https://blog.cloudflare.com/roughtime>`__ server).
 
-Attestation objects are versioned; this PEP specifies version 1. Each version
+Attestation objects are versioned; this PEP specifies version 2. Each version
```
woodruffw: This was my error when importing this into PyPUG, but we should fix it now to be clear that this is a change to the living standard 🙂 

```suggestion
Attestation objects are versioned; this document specifies version 2. Each version
```

source/specifications/index-hosted-attestations.rst
```diff
@@ -85,15 +88,25 @@ object is provided as pseudocode below.
...
       """
 
+      timestamps: list[bytes]
+      """
+      List of base64 encoded RFC3161 timestamp responses.
+
+      Added in Attestation version 2.
+
+      Note that list may be empty if `transparency_entries` only contains entries
+      with an integrated_time (in other words entries of kind "dsse 0.0.1").
```
woodruffw: I might be missing something, but where do we enforce that this is true for "DSSE 0.0.1"? The "version 1" above doesn't refer to the DSSE version, it's an arbitrary version picked for PEP 740.

jku: What I meant is that an empty list is fine -- but only attestations with rekor v1  entries (so dsse 0.0.1) can be considered valid without RFC3161 timestamps.

source/specifications/index-hosted-attestations.rst
jku: sure

source/specifications/index-hosted-attestations.rst
jku: I will rewrite the "verification algorithm" text  so it's based on the Attestation version.

source/specifications/index-hosted-attestations.rst
jku: I tried this `Attestation = AttestationV1 | AttestationV2` suggestion -- it's not exactly pretty but I suppose it does make it clearer what exactly the differences are. Let me know which you prefer

PR #1944, opened 2025-11-09
Allow errors for missing `[build-system]` table with no metadata

pradyunsg:
Permit installers to present an error if the working directory does not seem like a location that the user intended to install from.

x-ref https://discuss.python.org/t/do-we-want-to-keep-the-build-system-default-for-pyproject-toml/104759/15

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1944.org.readthedocs.build/en/1944/


notatallshaw:
One issue with this is the language in PEP 517 doesn't reflect these changes and that PEP does not point to the living spec.

PR #1889, opened 2025-08-04
Host PyPA spec schemas in packaging.python.org/

edgarrmondragon:
Closes https://github.com/pypa/packaging.python.org/issues/1875.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1889.org.readthedocs.build/en/1889/

PR #1950, opened 2025-11-12
Restore linkcheck for specs

FFY00:
This PR restores the `linkcheck` functionality for `https://packaging.python.org/en/latest/...` links.

It does so by adding a Sphinx plugin that hooks into the `linkcheck-process-uri` event and replaces the original URLs with the corresponding local path.

We've had a couple issues with schema links, which need to be specified via the full remote URL (eg. https://github.com/pypa/packaging.python.org/pull/1889#issuecomment-3287539359). We currently need to ignore them in `linkcheck` as it fails to take into account local changes. With this PR, the ability to catch these issues during lint is restored.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1950.org.readthedocs.build/en/1950/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,9 @@
 source/guides/github-actions-ci-cd-sample/* @webknjaz
 source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst @webknjaz
 
+# Sphinx extension
+pug_sphinx_extensions/ @FFY00
+
 # build-details.json
 source/specifications/build-details/ @FFY00
 source/specifications/specs/build-details-*.json @FFY00


FFY00:
CI is passing now :blush:

pug_sphinx_extensions/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    .. code-block:: python
+
+        >>> resolve_local_html_link('https://packaging.python.org/en/latest/flow/')
+        '{srcdir}/flow.rst'
+        >>> resolve_local_html_link('https://packaging.python.org/en/latest/flow.html')
+        '{srcdir}/flow.rst'
+        >>> resolve_local_html_link('https://packaging.python.org/en/latest/specifications/schemas/')
+        '{srcdir}/specifications/schemas/index.rst'
+        >>> resolve_local_html_link('https://packaging.python.org/en/latest/specifications/schemas/build-details-v1.0.schema.json')
+        '{html_extra_path0}/specifications/schemas/build-details-v1.0.schema.json'
```
edgarrmondragon, line 32: typo?

```suggestion
        '{html_extra_path}/specifications/schemas/build-details-v1.0.schema.json'
```

FFY00: It was on purpose, to subtly hint that `html_extra_path` is not a string, rather a list :sweat_smile:

edgarrmondragon: Ah, gotcha! 👍

webknjaz (approved):

Issue #1933, opened 2025-10-17
Clarify allowed minor version of macos tags


anuraaga:
### Issue Description

I am working on publishing a macos wheel and learned platform tags with minor versions such as 14_5 aren't allowed past 11_0

https://github.com/astral-sh/uv/issues/16337

This makes some sense because before 11, the minor version was the major version and since 11 it is more traditional.

Does it make sense to update the docs with this requirement? I'm happy to send a PR after confirming my understanding.

Also should pypi reject wheels with the disallowed tags? I was quite confused to be able to publish a wheel but not install it with uv.

### Code of Conduct

- [x] I am aware that participants in this repository must follow the PSF Code of Conduct.

anuraaga:
Thanks @notatallshaw - FWIW now that I know the issue, I won't have any problems and my goal was to try for others to not run into the same problem. The docs site has a "Contribute to this guide" link that pointed to this repo, and not DPO, so I filed this issue.

I like using familiar OSS tools and don't want to create another forum account so won't be pursuing this on DPO - I like contributing but also have boundaries. If it seems fine to maintain the status quo where the tools don't match since it's a corner case, let's close this issue. Otherwise I hope for a maintainer to take up the cause.

PR #1958, opened 2025-11-21
Document `macosx_{x}_0_{arch}` pattern for macOS 11+

konstin:
Following https://discuss.python.org/t/document-that-macos-platform-tags-use-minor-version-0-for-macos-11/104616:

 * Document that for macOS 11 and later, the platform tag is `macosx_{x}_0_{arch}`
 * Document that platform tags may change over time

Fixes https://github.com/pypa/packaging.python.org/issues/1933

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1958.org.readthedocs.build/en/1958/

### DIFF
--- a/source/specifications/platform-compatibility-tags.rst
+++ b/source/specifications/platform-compatibility-tags.rst
@@ -82,6 +82,11 @@ decide how to best use the ABI tag.
 Platform Tag
 ============
 
+.. important::
+    Platform tags are dependent on the versioning of the operating system or
+    platform they represent and may change over time as the underlying platform
+    changes its versioning.
+
 Basic platform tags
 -------------------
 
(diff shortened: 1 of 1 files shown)


source/specifications/platform-compatibility-tags.rst
```diff
@@ -82,6 +82,11 @@ decide how to best use the ABI tag.
 Platform Tag
 ============
 
+.. important::
+    Platform tags are dependent on the versioning of the operating system or
+    platform they represent and may change over time as the underlying platform
+    changes its versioning.
```
konstin, line 88: Technically, some parts are bound by PEPs while others were defined by `packaging`, while again others are defined here at all (e.g., BSDs), but platforms authors are not bound by PEPs and may make changes counter to PEP assumptions. The new macOS versioning is an example of where Apple made a change counter to packaging spec assumptions, and packaging had to adapt.

PR #1959, opened 2025-11-22
Specify arbitrary equality case insensitivity.

notatallshaw:
Following https://discuss.python.org/t/clarify-that-arbitrary-equality-is-supposed-to-be-case-insensitive/104885 specify that for ASCII letters arbitrary equality MUST be case insensitive.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1959.org.readthedocs.build/en/1959/

### DIFF
--- a/source/specifications/version-specifiers.rst
+++ b/source/specifications/version-specifiers.rst
@@ -1272,4 +1272,4 @@ History
 - August 2014: This specification was approved through :pep:`440`.
 - May 2025: Clarify that development releases are a form of pre-release when
   they are handled.
-- Nov 2025: Specify arbitrary equality case insensitivity.
+- Nov 2025: Make arbitrary equality case insensitivity explicit.


woodruffw (approved):

source/specifications/version-specifiers.rst
```diff
@@ -1271,3 +1272,4 @@ History
 - August 2014: This specification was approved through :pep:`440`.
 - May 2025: Clarify that development releases are a form of pre-release when
   they are handled.
+- Nov 2025: Specify arbitrary equality case insensitivity.
```
ncoghlan:
```suggestion
- Nov 2025: Make arbitrary equality case insensitivity explicit.
```

It was previously implied by only semantic details of the version text being excluded.

notatallshaw: I'm fine with this change and no one else has objected in 2 days, so committed to PR.

webknjaz (approved):

PR #1979, opened 2025-12-24
Clarify definition of native and implicit namespace packages

stevepiercy:
This pull request clarifies the definition of native and implicit namespace packages as synonyms, and adds a glossary term for them.

From a discussion in Discord at https://discord.com/channels/935215565872693329/935215566334079058/1451178701956976853, by an initially very confused reader who's relatively new to the terminology.

Feedback welcome. Thank you for your consideration.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1979.org.readthedocs.build/en/1979/

### DIFF
--- a/source/guides/packaging-namespace-packages.rst
+++ b/source/guides/packaging-namespace-packages.rst
@@ -167,8 +167,8 @@ Legacy namespace packages
 
 These two methods, that were used to create namespace packages prior to :pep:`420`,
 are now considered to be obsolete and should not be used unless you need compatibility
-with packages already using this method. Also, :doc:`pkg_resources <setuptools:pkg_resources>`
-has been deprecated.
+with packages already using this method. Also,
+:doc:`pkg_resources <setuptools:deprecated/pkg_resources>` has been deprecated.
 
 To migrate an existing package, all packages sharing the namespace must be migrated simultaneously.
 


source/glossary.rst
```diff
@@ -119,6 +119,20 @@ Glossary
...
+        "Native namespace packages" and "implicit namespace packages" are two names for
+        the same type of namespace package. It's defined in :pep:`420`, and is available
+        in Python 3.3 and later.
+
+        The term "native" derives from the built-in feature of Python to create packages
+        without installing an extra package, such as setuptools, from outside the standard
+        library. The term "implicit" derives from the fact that you don't have to explicitly
+        declare the namespace package in a :file:`__init__.py` file.
+
+		This guide uses these names interchangeably.
```
AA-Turner:
```suggestion
        This guide uses these names interchangeably.
```

AA-Turner (approved):

source/guides/packaging-namespace-packages.rst
```diff
@@ -71,6 +71,12 @@ from which the latter is discouraged:
 #. Use `legacy namespace packages`_. This comprises `pkgutil-style namespace packages`_
    and `pkg_resources-style namespace packages`_.
 
+.. note::
+
+    :term:`Native namespace <Native Namespace Package>` and
+    :term:`implicit namespace <Implicit Namespace Package>` packages are two
```
merwok: I think this should say either

> *native* and *implicit* namespace packages

or 

> *native namespace packages* and *implicit namespace packages*

but not use two words out of the three

stevepiercy: Did the latter in https://github.com/pypa/packaging.python.org/pull/1979/commits/60f94b7504efc1fe4ea8529dcebddc8ca57380ec

ncoghlan:
Something that occurred to me while reviewing the complementary PR at #1882 is that it would make sense to mutually cross link the new glossary entry with the more detailed description at 
https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

PR #1980, opened 2025-12-25
pypirc: stipulate UTF-8 encoding

woodruffw:
This is a kind of interesting/nuanced thing, and I don't know if this PR is actually the right approach 😅 -- I'm opening it for discussion!

Context: `.pypirc` was never PEP'd; it was seemingly grandfathered in as a distutils implementation detail that gets respected by twine and flit (and perhaps others?). As a result, there's no stipulation as to what charset a `.pypirc` should be encoded with.

In practice this is mostly a non-issue: most users will default to UTF-8, or in the worst case they'll use whatever system character encoding Python picks and _hopefully_ their consuming tool will too. However this does unfortunately break down sometimes, like with https://github.com/pypa/twine/pull/1268 recently.

I think the best fix here is just to rip the bandaid off and stipulate that `.pypirc` files should be UTF-8, especially since upcoming Python versions will use UTF-8 in a blanket fashion by default. I've gone with "should" instead of "must" since twine _does_ make an effort to parse using the system charset, but this is a fallback we'd probably like to eventually remove. I'm not sure what flit does, but "should" also felt appropriate given that it's entirely open-ended at the moment 🙂 

Open questions:

* Can we just do this, without any additional standardization procedure? I think the answer is yes, since `.pypirc` is (1) kind of a legacy component anyways, and (2) never went through a packaging PEP (AFAICT). However, I'd like to hear if others feel otherwise.
* Are there other publishing tools that use this file that this needs to be coordinated with? I know twine and flit for certain, and that uv doesn't (yet?) support it. Ref: https://github.com/astral-sh/uv/issues/7676

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1980.org.readthedocs.build/en/1980/

### DIFF
--- a/source/guides/multi-version-installs.rst
+++ b/source/guides/multi-version-installs.rst
@@ -37,7 +37,3 @@ time, but that approach does mean that standard command line invocations of
 the affected tools can't be used - it's necessary to write a custom
 wrapper script or use ``python3 -c '<command>'`` to invoke the application's
 main entry point directly.
-
-Refer to the `pkg_resources documentation
-<https://setuptools.readthedocs.io/en/latest/pkg_resources.html#workingset-objects>`__
-for more details.
--- a/source/guides/packaging-namespace-packages.rst
+++ b/source/guides/packaging-namespace-packages.rst
@@ -159,8 +159,7 @@ Legacy namespace packages
 
 These two methods, that were used to create namespace packages prior to :pep:`420`,
 are now considered to be obsolete and should not be used unless you need compatibility
-with packages already using this method. Also, :doc:`pkg_resources <setuptools:pkg_resources>`
-has been deprecated.
+with packages already using one of these methods.
 
 To migrate an existing package, all packages sharing the namespace must be migrated simultaneously.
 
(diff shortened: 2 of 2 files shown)


woodruffw:
@pypa/packaging-user-guide-editors does anybody have any objections to this?

I raised it on DPO here as well: https://discuss.python.org/t/psa-proposed-change-to-the-pypirc-spec/105581

(Beyond that, I'm not sure if I need to do anything else procedurally before this can be merged, given that this "spec" is in a legacy/grandfathered category.)

source/specifications/pypirc.rst
```diff
@@ -5,6 +5,13 @@
 The :file:`.pypirc` file
 ========================
 
+.. important::
+
+    The :file:`.pypirc` file **SHOULD** be UTF-8 encoded.
+
+    Tools that read or write :file:`.pypirc` files may not function correctly
+    if another character encoding is used.
```
merwok: Wording is ok for me, but 1) I would not have the note before the paragraph below (that explains what even this file is) 2) I don’t think a callout («important» markup) is needed.

woodruffw: Thanks, I'll adjust!

PR #1987, opened 2026-01-09
Clarify that TOML v1.0 is used for specs

sirosen:
- State "version 1.0" for pylock.toml and pyproject.toml
- State that "TOML 1.0" is used for script metadata

For motivation, see:

  https://discuss.python.org/t/adopting-toml-1-1/105624

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1987.org.readthedocs.build/en/1987/

### DIFF
--- a/source/specifications/inline-script-metadata.rst
+++ b/source/specifications/inline-script-metadata.rst
@@ -74,7 +74,7 @@ script type
 -----------
 
 The first type of metadata block is named ``script``, which contains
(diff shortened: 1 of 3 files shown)


notatallshaw:
Just a note, and I'm sure it would already be the case, can we make sure not to merge this until the DPO thread has resolved.

The discussion has gone in a direction that I am now -1 on this update, and I want to take  but some time to review and consider the points before responding.

PR #1988, opened 2026-01-10
Attempt to clarify environment marker evaluation

ncoghlan:
Preparation for the release of packaging 25.1 revealed multiple deficiencies in the specification of environment marker evaluation. Review of the proposed amendments to resolve those deficiencies highlighted multiple other problems, including some dating from the original PEP 508 specification:

* other pages still referencing PEP 508 instead of the living spec
* direct reference to PEP 685 instead of the core metadata spec
* the "extra" special case not being properly defined
* lacking guidance to tool developers regarding what should be considered errors to disallow entirely vs issues to work around

Inspired by the initial PR at #1971

discuss.python.org thread: https://discuss.python.org/t/spec-change-bugfix-dependency-specifiers-simplification-pep-508

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1988.org.readthedocs.build/en/1988/

### DIFF
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -6,21 +6,25 @@
 Dependency specifiers
 =====================
 
-This document describes the dependency specifiers format as originally specified
-in :pep:`508`.
+This document defines the format used to specify dependencies on other projects.
+The language defined is a compact line based format which was adapted from the
+format originally used in ``pip`` requirements files.
 
 The job of a dependency is to enable tools like pip [#pip]_ to find the right
 package to install. Sometimes this is very loose - just specifying a name, and
 sometimes very specific - referring to a specific file to install. Sometimes
-dependencies are only relevant in one platform, or only some versions are
+dependencies are only relevant on one platform, or only some versions are
 acceptable, so the language permits describing all these cases.
 
-The language defined is a compact line based format which is already in
-widespread use in pip requirements files, though we do not specify the command
-line option handling that those files permit. There is one caveat - the
-URL reference form, specified in :ref:`Versioning specifier specification <version-specifiers>`
-is not actually implemented in pip, but we use that format rather
-than pip's current native format.
+Whether tools should be strict or permissive in their processing of dependency
+specifiers is largely dependent on the role of the tool in the wider ecosystem:
+
+* publishing tools and index servers SHOULD be strict in their processing for
+  new releases, encouraging the consistency of published specifiers to improve
+  over time
+* locking and installation tools MAY be permissive in their processing, allowing
+  consumption of older packages which may contain dependency specifiers that are
+  arguably nonsensical
 
 Specification
 =============
@@ -30,7 +34,7 @@ Examples
 
 All features of the language shown with a name based lookup::
 
-    requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "2.7"
+    requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "3.7"
 
 A minimal URL based lookup::
 
@@ -108,8 +112,6 @@ field::
     extras_list   = identifier (wsp* ',' wsp* identifier)*
     extras        = '[' wsp* extras_list? wsp* ']'
 
-Restrictions on names for extras is defined in :pep:`685`.
-
 Giving us a rule for name based requirements::
 
     name_req      = name wsp* extras? wsp* versionspec? wsp* quoted_marker?
@@ -126,23 +128,22 @@ Whitespace
 ----------
 
 Non line-breaking whitespace is mostly optional with no semantic meaning. The
-sole exception is detecting the end of a URL requirement.
+sole exceptions are detecting the end of a URL requirement and inside user
+supplied constants in environment markers.
 
 .. _dependency-specifiers-names:
 
 Names
 -----
 
-Python distribution names are currently defined in :pep:`345`. Names
-act as the primary identifier for distributions. They are present in all
+Distribution names are defined in the :ref:`Core metadata <core-metadata-name>`.
+Names act as the primary identifier for distributions. They are present in all
 dependency specifications, and are sufficient to be a specification on their
-own. However, PyPI places strict restrictions on names - they must match a
-case insensitive regex or they won't be accepted. Accordingly, in this
-document we limit the acceptable values for identifiers to that regex. A full
-redefinition of name may take place in a future metadata PEP. The regex (run
-with re.IGNORECASE) is::
+own.
 
-    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])\Z
+Valid distribution names are defined in the :ref:`name format specification
+<name-format>`.
+
 
 .. _dependency-specifiers-extras:
 
@@ -163,6 +164,12 @@ are listed in the "security" extra of requests.
 
 If multiple extras are listed, all the dependencies are unioned together.
 
+Restrictions on names for extras are defined in the
+:ref:`Core metadata specification <core-metadata-provides-extra>`. Publication
+tools SHOULD enforce these restrictions in dependency specifiers, while locking
+and installation tools MAY normalize invalid extra names in order to accept
+published metadata using core metadata versions prior to 2.3.
+
 .. _dependency-specifiers-versions:
 
 Versions
@@ -172,7 +179,7 @@ See the :ref:`Version specifier specification <version-specifiers>` for
 more detail on both version numbers and version comparisons. Version
 specifications limit the versions of a distribution that can be
 used. They only apply to distributions looked up by name, rather than
-via a URL. Version comparison are also used in the markers feature. The
+via a URL. Version comparisons are also used in environment markers. The
 optional brackets around a version are present for compatibility with
 :pep:`345` but should not be generated, only accepted.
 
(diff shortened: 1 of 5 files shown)


source/specifications/dependency-specifiers.rst
```diff
@@ -183,63 +190,140 @@ Environment Markers
...
+
+For ``Version`` fields, the comparison operations are defined by the
+:ref:`Version specifier specification <version-specifiers>` when either both
+the marker field value and the user supplied constant can be parsed as valid
+version specifiers or the ``===`` arbitrary equivalence comparison operator
+is used. When an operator other than ``===`` is used, publishing tools SHOULD
+emit an error if the user supplied constant cannot be parsed as a valid version
+specifier, while locking and installation tools MAY either emit an error or else
+fall back to ``String`` field comparison logic if either the marker field value
+or the user supplied constant cannot be parsed as a valid version specifier.
```
ncoghlan, line 281: Need to note that `in` and `not in` aren't valid for version fields.

source/specifications/dependency-specifiers.rst
```diff
@@ -183,63 +190,140 @@ Environment Markers
...
+
+Composing marker expressions
+''''''''''''''''''''''''''''
+
+More complex marker expressions may be composed using the ``and`` and ``or``
+logical operators. Parentheses may be used as necessary to control operand
+precedence (with all comparison operations having a higher precedence).
+
+Python's comparison chaining (such as ``3.4 < python_version < 3.9``) is NOT
+supported in environment markers.
```
ncoghlan: Add a couple of examples here, such as checking for `SMP` in the platform description after constricting the platform to Linux, or checking for a minimum macOS version.

source/specifications/dependency-specifiers.rst
```diff
@@ -183,63 +190,154 @@ Environment Markers
...
+explicitly discouraged (as it makes no semantic sense in the packaging context),
+and publishing tools SHOULD emit an error, while locking and installation tools
+SHOULD implement the following behavior:
+
+* treat ``>=`` and ``<=`` as equivalent to ``==``
+* treat ``>`` and ``<`` as always being False
+
+For ``Set of String`` fields, as there is no marker syntax for set literals,
+the only valid operations are ``in`` and ``not in`` comparisons with a user
+supplied string literal as the left operand.
```
ncoghlan: Publishing tools should emit an error for other operations, consuming tools may just treat the operations as false

source/specifications/dependency-specifiers.rst
```diff
@@ -183,63 +190,154 @@ Environment Markers
...
+Note that ``in`` and ``not in`` containment checks are NOT valid for ``Version``
+fields and publishing tools SHOULD emit an error, while locking and installation
+tools MAY treat them as always being False.
+
+For ``Version | String`` fields, comparison operations are defined as they are
+for ``Version`` fields, while ``in`` and ``not in`` containment checks are
+defined as they are for ``String`` fields. However, there is no expectation
+that the parsing of the marker field value or the user supplied constant as a
+valid version will succeed, so tools MUST fall back to processing the field as
+a ``String`` field. Alternatively, tools MAY unconditionally treat such fields
```
ncoghlan: The established use of `platform_release` as a version on macOS means that this field really does need to be checked as a version first.

Due to [iOS and Android](https://docs.python.org/3/library/platform.html#platform.version), `platform_version` potentially needs the same handling.

source/specifications/dependency-specifiers.rst
```diff
@@ -183,63 +190,154 @@ Environment Markers
...
+them (usually based on the text encoding of the file or stream containing the
+dependency specifier). This may be revisited in the future if it becomes more
+common for the runtime variables typically referenced in environment markers to
+contain non-ASCII text that users wish to perform comparisons against.
+
+Unknown marker fields
+'''''''''''''''''''''
+
+References to unknown marker fields MUST raise an error rather than resulting
+in a comparison that evaluates to True or False.
```
ncoghlan: Explain that this so unknown markers result in a clear installation failure rather than an installation with potentially missing dependencies that then fails at runtime. Soften it to a SHOULD.

source/specifications/dependency-specifiers.rst
henryiii: `platform_release` is _only_ useful if it's a `Version`. The string fallback means you can't use it (unless you really are shipping a package for literally only one exact version of a specific platform - maybe also on Windows if this just always reports `"11"`). So we do want the "Version" behavior to be required, with a string fallback, just like `Version`.

To me, the difference between `Version` and `Version | String` is that we are telling the user that we expect it to always be a Version, vs. we expect it to fallback to something useless (but not an error). This generally means a user can't rely on `Version | String` unless it's gated by something that ensures it's a Version, like a check for the platform.

It would be nice if we had some sort of "always Version" field instead, but it's defined as just the raw underlying `platform.release` and it would be a much bigger effort to develop some sort of always-Version transform for every platform, so I think we should just try to make the best of what we have for now.

I think `platform_version` is always a String? It's not convertible to a version on macOS, only `platform.release()` is. Also Pydodide (emscripten) it's not a version. We don't consider it to be possible to convert it to a version here: https://github.com/pypa/packaging/blob/7c0307115f734c1a6b7b54bf676ed1b635fc7c16/src/packaging/markers.py#L37-L41

henryiii: Oh, I realize there's nothing actionable above. I'd like to require `Version | String` to do the `Version` processing if possible for `platform_release`; that way it is possible for a package to build a marker using platform-specific versioning on some platforms, like macOS. If we leave that up to tools, then that becomes tool specific, instead of just platform specific.

ncoghlan: Yeah, I agree `Version | String` should always try and fallback - the `MAY` section (as currently worded) is a bad idea. The difference from `Version` is that whether the parsing failing is a problem or not is platform dependent. I'll tweak the `MAY` section to instead say something about tools being permitted to emit a warning if they know the field *should* be a version on a given platform, but it fails parsing.

For `platform_version`, it's specifically the implementation of iOS and Android platform support in the standard library that uses that as an actual version field, so it's not something most packaging utilities are going to be exposed to. I think it makes sense to formally specify it that way, though.

source/specifications/dependency-specifiers.rst
ncoghlan: Adding a second sentence with SHOULD clauses for both publishers and consumers.

source/specifications/dependency-specifiers.rst
ncoghlan: Significantly reworded the `Version | String` section, and noted example platforms in `platform_release` and `platform_version` where they're known to contain valid versions.

I also added explanatory notes to several of the other environment marker fields (leaving this discussion open for now, rather than marking it as resolved).

source/specifications/dependency-specifiers.rst
```diff
@@ -183,133 +190,277 @@ Environment Markers
...
+Unknown marker fields
+'''''''''''''''''''''
+
+References to unknown marker fields SHOULD raise an error rather than resulting
+in a comparison that evaluates to True or False. This is so that attempted
+installations involving unknown marker fields result in a clear installation
+failure, rather than an apparently successful installation that then fails at
+runtime due to missing dependencies (if the unknown marker is treated as
+False) or a potentially cryptic installation failure of a dependency that is
+not valid for the current platform (if the unknown marker is treated as True)
```
konstin: With my wheel variants hat, I'd argue that tools should discard this file or version and backtrack, rather than hard fail.

source/specifications/dependency-specifiers.rst
```diff
@@ -183,133 +190,277 @@ Environment Markers
...
      - String
-     - ``x86_64``
+     - ``x86_64``, ``aarch64``, ``AMD64``, ``arm64``
+       (note that this value is provided by the operating system, so the same
+       CPU architecture may use different strings on different platforms)
    * - ``platform_python_implementation``
      - :py:func:`platform.python_implementation()`
      - String
-     - ``CPython``, ``Jython``
+     - ``CPython``, ``PyPy``, ``Jython``
```
konstin: nit: We can replace Jython with Graal Python while we're at it

source/specifications/dependency-specifiers.rst
```diff
@@ -183,133 +190,277 @@ Environment Markers
...
+lock file to selectively install optional parts of the locked dependency tree).
+Publishing tools SHOULD emit an error if projects attempt to use them in their
+published metadata, and index servers SHOULD NOT accept uploads referencing
+these fields. Outside lock file processing, marker evaluation environments
+DO NOT need to define these fields.
+
+The ``extra`` field is also special, as it expects set-like behaviour, but
+predates the addition of ``Set of strings`` as a defined marker field type.
+Accordingly, for this field only, ``extra == "name"`` is equivalent to
+``"name" in extras``, while ``extra != "name"`` is equivalent to
```
konstin: Not quite, as they are different sources of extras, and you have to evaluate them separately.

konstin (approved):
The extras thing we should fix to avoid confusing people.

source/specifications/dependency-specifiers.rst
ncoghlan: As in, `... == extra` is only valid for comparing against dependency declaration extras (added via `pkg[name]`, while `... in extras` is only valid when installing from a lock file (added via a tool specific option)?

That feels like a genuine ambiguity in PEP 751 to me, as my assumption was that there's only one set of requested extras for each package, just different ways of adding new entries to it, and two syntaxes for querying it.

konstin: The `extras` and `dependency_groups` refer to the extras and the groups of the entire lockfile, while `extra` refers to the extras of the current package. This allows e.g. `tool install --extra foo --group bar pylock.toml`.

ncoghlan: How can a lock file have freestanding extras? I can see it having freestanding groups (since those are always project level rather than part of the package metadata), but extras only make sense for projects that publish a package.

konstin: My understanding was that you create `pylock.toml` from a `pyproject.toml` with `project.optional-dependencies`, and map those to the global extras.

> Tools should explicitly set this key to an empty array to signal that the inputs used to generate the lock file had no extras (e.g. a pyproject.toml file had no [project.optional-dependencies] table), signalling that the lock file is, in effect, multi-use even if it only looks to be single-use.

ncoghlan: Oh, you're right, I misremembered exactly how this works in `pylock.toml`:

* `name in extras`: only valid in `packages.marker` fields, `extras` refers to the currently selected subset of names from the global `extras` field. Allows for deps-only lock files that don't include the project itself, as well as keeping extras optional even when the project _is_ included.
* `extra == name`: only valid in `packages.dependencies` entries (inherited from published package metadata), has its usual meaning of referring to the extras specifier when dependents declare their dependencies on the project

The comparison operation itself is the same in both cases, the subtlety is in how the `extras` set is defined.

Yeesh, wording that clearly is going to be hard...

source/specifications/dependency-specifiers.rst
ncoghlan: I'm not sure we can, as `platform.python_implementation` is currently an ill-defined bit of `sys.version` parsing rather than something that implementations can reliably override without explicit stdlib patching: https://github.com/python/cpython/issues/116505

So I'll leave this one out of the current PR.

(Edit: I realised I could at least drop `Jython` from the examples, even though I'm not sure we can safely add `Graal` as a replacement, since I'm not sure what `platform.python_implementation()` emits there)

source/specifications/dependency-specifiers.rst
ncoghlan: I reworded this to also allow for ignoring the offending versions rather than aborting the whole resolution process (push coming after I sort out the extras comparisons).

source/specifications/dependency-specifiers.rst
ncoghlan: OK, I have wording for consideration that I think is at least not wrong. Whether it rises to the level of actually being _clear_? I'm not convinced I've achieved that, but it's still inevitably clearer than the current state of "not specified at all".

woodruffw (approved):

PR #1994, opened 2026-01-26
Discourage use of version epochs

mgorny:
Add a note discouraging use of epochs in version specifiers, and update the versioning discussion accordingly.  While at it, suggest using two-digit years for CalVer, as mentioned in the discussion thread.

Discussion: https://discuss.python.org/t/discouraging-use-of-epoch-segments-in-versions/105811/1

----
📚 Documentation preview 📚: https://python-packaging-user-guide--1994.org.readthedocs.build/en/1994/

### DIFF
--- a/source/guides/multi-version-installs.rst
+++ b/source/guides/multi-version-installs.rst
@@ -37,7 +37,3 @@ time, but that approach does mean that standard command line invocations of
 the affected tools can't be used - it's necessary to write a custom
 wrapper script or use ``python3 -c '<command>'`` to invoke the application's
 main entry point directly.
-
-Refer to the `pkg_resources documentation
-<https://setuptools.readthedocs.io/en/latest/pkg_resources.html#workingset-objects>`__
-for more details.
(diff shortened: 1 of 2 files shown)


source/discussions/versioning.rst
```diff
@@ -132,7 +133,9 @@ A key advantage of date-based versioning, or `calendar versioning <calver_>`_
 a particular release is given just the version number.
 
 Calendar version numbers typically take the form *year.month* (for example,
-23.12 for December 2023).
+23.12 for December 2023). When using this scheme, it is recommended to reference
+the year using two digits rather than four, to make it easier to change the
+versioning scheme without having to use very large version numbers.
```
pfmoore: IMO, this advice is a separate matter, unrelated to the question of whether we discourage the use of the epoch field. I'd rather we removed it from this PR.

mgorny: Done. I suppose no harm in filing another PR for this, if you think it makes sense to do so.

pfmoore: I would actually prefer it if we just didn't try to advise people on how to structure their version numbers at all. That should be a project decision, and not something the standards should weigh in on.

We have too many well-meaning people who treat style guides as cast iron rules. I don't think we should fuel that tendency.

mgorny:
I've already removed that statement. Is there anything else left for me to do?

pfmoore (approved):

PR #2023, opened 2026-03-22
Add uv

bangseongbeom:
Add [uv](https://docs.astral.sh/uv/) to [Project Summaries](https://packaging.python.org/en/latest/key_projects/).

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2023.org.readthedocs.build/en/2023/

### DIFF
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -836,7 +836,7 @@ uv
 `GitHub <https://github.com/astral-sh/uv>`__ |
 `PyPI <https://pypi.org/project/uv/>`__
 
-An extremely fast Python package and project manager, written in Rust. It
+A Python package and project manager, written in Rust for high performance. It
 supports creating and managing virtual environments, installing packages,
 locking dependencies, and managing Python versions and projects.
 
--- a/source/key_projects.rst
+++ b/source/key_projects.rst
@@ -827,6 +827,19 @@ scientific applications on clusters and supercomputers.
 Spack is not in PyPI (yet), but it requires no installation and can be
 used immediately after cloning from GitHub.
 
+.. _uv:
+
+uv
+==
+
+`Docs <https://docs.astral.sh/uv/>`__ |
+`GitHub <https://github.com/astral-sh/uv>`__ |
+`PyPI <https://pypi.org/project/uv/>`__
+
+An extremely fast Python package and project manager, written in Rust. It
+supports creating and managing virtual environments, installing packages,
+locking dependencies, and managing Python versions and projects.
+
 .. _zestreleaser:
 
 zest.releaser


source/key_projects.rst
```diff
@@ -827,6 +827,19 @@ scientific applications on clusters and supercomputers.
...
+.. _uv:
+
+uv
+==
+
+`Docs <https://docs.astral.sh/uv/>`__ |
+`GitHub <https://github.com/astral-sh/uv>`__ |
+`PyPI <https://pypi.org/project/uv/>`__
+
+An extremely fast Python package and project manager, written in Rust. It
```
webknjaz: Can we make this not look like an ad?

bangseongbeom: It's a phrase that I brought from the uv website, and it definitely feels like an ad. As you mentioned, I revised it a little to make it sound less promotional!

woodruffw (approved):
It seems reasonable to me for this to be included, but in the interest of transparency/conflict avoidance I think another editor should also approve this 🙂

henryiii (approved):

aleyan:
@bangseongbeom this PR is approved. Is there anything blocking it from merging?

henryiii:
Maybe @webknjaz?

woodruffw:
Yes, it needs someone other than me with the commit bit to approve (to resolve any potential CoI).

bangseongbeom:
> @bangseongbeom this PR is approved. Is there anything blocking it from merging?

As he mentioned, I'm waiting for approval from the other reviewer.

source/key_projects.rst
```diff
@@ -827,6 +827,19 @@ scientific applications on clusters and supercomputers.
...
+.. _uv:
+
+uv
+==
+
+`Docs <https://docs.astral.sh/uv/>`__ |
+`GitHub <https://github.com/astral-sh/uv>`__ |
+`PyPI <https://pypi.org/project/uv/>`__
+
+A Python package and project manager, written in Rust for high performance. It
```
webknjaz, line 839: Sounds like it fits the definition of a workflow tool more (it's actually listed as such @ https://packaging.python.org/en/latest/guides/tool-recommendations/#workflow-tools). Perhaps, this should just say it's a workflow tool at the beginning and being a package/project manager/interpreter installer could be a part of the list of things in the last sentence?

cc @woodruffw @henryiii

zanieb: That's fine with me. We don't use that terminology in the uv documentation, but it seems well defined here.

henryiii: It's kind of everything except a task runner, which is what I think I'd define as a key part of a workflow tool. You can get away without setting up tasks (much of the time `uv run pytest` works out-of-the-box in most projects I work on).

Nox/tox/poe/pixi don't seem to be listed here at all, so there's not much to compare with, I think that's what would come to mind for "workflows". In the workflow list, I do see Pipenv, that doesn't have tasks. I don't think `flit` does either. Poetry does have poe, so maybe. uv is not in that list, by the way, maybe that should be added here too?.

Given that list is pretty general, I think calling it a workflow tool is also fine, but it's also a package and project manager, and I'd say that fits a little better than "workflow", especially without task support. +/- 0.

webknjaz: Yeah, that list is kinda loosely defined. Other entries in this document also seem to focus more on what the projects do rather than how they're implemented. And so I'm struggling to formalize why this seems to stand out writing style-wise.

pradyunsg: I'd say let's go with this language for now and we can tweak it in a follow up if needed.

webknjaz: Alright, I won't be blocking this. Just wanted to document my concerns..

webknjaz (approved):

webknjaz:
Looks like Fedora's docs might be down blocking this
```console
(guides/installing-using-linux-tools: line   52) broken    https://fedoraproject.org/wiki/EPEL - HTTPSConnectionPool(host='fedoraproject.org', port=443): Read timed out. (read timeout=30)
(discussions/package-formats: line  146)
[...]
(guides/installing-using-linux-tools: line   65) broken    https://fedoraproject.org/wiki/Infrastructure/Fedorahosted-retirement - HTTPSConnectionPool(host='fedoraproject.org', port=443): Read timed out. (read timeout=30)
```

PR #2025, opened 2026-03-24
Improve pylock requires-python exemple

sbidoul:
Since `requires-python` is a version specifier, `==3.12` actually means `==3.12.0` which is not how one would intuitively interpret it.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2025.org.readthedocs.build/en/2025/


sbidoul:
@brettcannon this comes from https://github.com/pypa/packaging/pull/1092#discussion_r2969606504

brettcannon (approved):

henryiii:
CC @woodruffw or @webknjaz, perhaps?

webknjaz (approved):

PR #2032, opened 2026-04-13
Fix linkcheck timeouts

woodruffw:
See #2024 for some context. This will hopefully make linkcheck time out less and, when it does time out, will ensure we don't hold onto the job for hours.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2032.org.readthedocs.build/en/2032/


webknjaz (approved):
Finally! Thanks ♥️

PR #2040, opened 2026-05-04
Warn about artifact action behavior for wheel matrices

### LINKED ISSUE #1454: Add a caution/warning box about `actions/upload-artifact@v4` @ GHA publishing guide
So I want to have some warning displayed right before the readers first see `actions/upload-artifact@v3` @ https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions.

The text should tell people to avoid upgrading the action to v4 (and the same for the download one), if they build many platform-specific wheels, and not just a pure-python wheel+sdist.
The main conversation of how to live with the incompatible/breaking changes is happening @ https://github.com/actions/upload-artifact/issues/472.

_More context and some explanation is linked @ https://github.com/orgs/aio-libs/discussions/31. I didn't want to copy it over as is._

Herrtian:
Adds a warning before the artifact upload/download workflow example for people adapting the guide to several platform-specific wheel jobs.

The sample now uses current artifact actions rather than the older v3 versions mentioned in the issue, so this keeps the caution framed around the still-relevant v4+ behavior: don't have multiple wheel jobs upload to the same artifact name; give each job its own artifact and adjust downloads accordingly.

Checks:
- git diff --check
- uv run --with pre-commit pre-commit run --files source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
- uv run --with nox nox -s build

Fixes #1454

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2040.org.readthedocs.build/en/2040/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions

### DIFF
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -107,6 +107,15 @@ build the distribution packages.
 First, we'll define the job for building the dist packages of
 your project and storing them for later use:
 
+.. warning::
+
+   This artifact configuration is intended for a single build job that uploads
+   one source distribution and one wheel. If you adapt it for several
+   platform-specific wheel jobs, use separate artifact names for each job and
+   adjust the download step accordingly. The v4+ artifact actions do not support
+   multiple jobs uploading to the same artifact; see
+   `actions/upload-artifact#472`_.
+
 .. literalinclude:: github-actions-ci-cd-sample/publish-to-pypi.yml
    :language: yaml
    :start-at: jobs:
@@ -228,6 +237,8 @@ sure that your release pipeline remains healthy!
    https://github.com/actions/download-artifact
 .. _`upload-artifact`:
    https://github.com/actions/upload-artifact
+.. _`actions/upload-artifact#472`:
+   https://github.com/actions/upload-artifact/issues/472
 .. _Secrets:
    https://docs.github.com/en/actions/reference/encrypted-secrets
 .. _Trusted Publishing: https://docs.pypi.org/trusted-publishers/
--- a/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
+++ b/source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
@@ -111,7 +111,8 @@ your project and storing them for later use:
 
    If you adapt this workflow to build multiple platform-specific wheels, use
    uniquely named artifacts for each build job and adjust the download step
-   accordingly.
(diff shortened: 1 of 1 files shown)


webknjaz:
I think this should be a tip, not a warning. That issue was filed years ago and nobody uses v3. So I don't see why we need to highlight it. Today, I'd rather have a shorter note along the lines of "when building multiple wheels, you'll end up needing using uniquely named artifacts per build job".

Herrtian:
Thanks, agreed. I pushed c2b1179 to make this a short tip instead, without the old v3/v4 framing.

Checked locally with `git diff --check`, file-level pre-commit, and `nox -s build`.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
```diff
@@ -107,6 +107,12 @@ build the distribution packages.
 First, we'll define the job for building the dist packages of
 your project and storing them for later use:
 
+.. tip::
+
+   If you adapt this workflow to build multiple platform-specific wheels, use
+   uniquely named artifacts for each build job and adjust the download step
+   accordingly.
```
webknjaz: I just remembered that `cibuildwheel` has examples. So we should also link https://cibuildwheel.pypa.io/en/latest/ci-services/#github-actions. ~But since it's a Sphinx website, use an RST role. You'll need to add an entry into `source/conf.py` for this to work.~ Never mind, it's an mkdocs site so use a normal link. I've filed https://github.com/pypa/cibuildwheel/issues/2840 for them to add intersphinx support on their side.

source/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.rst
webknjaz, line 1: Live change preview: https://python-packaging-user-guide--2040.org.readthedocs.build/en/2040/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions

webknjaz (approved):
Thanks!

webknjaz:
Given https://www.dayswithoutgithubincident.com, it may take a few attempts to merge.

webknjaz:
This one might actually be legit:
> ```console
> /home/runner/work/packaging.python.org/packaging.python.org/source/overview.rst:353: WARNING: unknown document: 'openstack:user/formats'
> ```

(though unrelated to the PR)

webknjaz:
A rerun passed. So it was flaky.

PR #2042, opened 2026-05-08
Document separate build and publish jobs

### LINKED ISSUE #1627: Document that building the dists in the same job as publishing is unsupported @ GHA publishing guide
Initially reported here: https://github.com/pypa/gh-action-pypi-publish/issues/237

Herrtian:
Closes #1627.

Adds a short note that publishing jobs should upload already-built distributions instead of building them.

Checked with `nox -s build`.


webknjaz (approved):
Thanks 🙌

PR #2045, opened 2026-05-12
Normalize case and make small changes in the reference implementation for dependency groups resolution

jonathandung:
```diff
-raise TypeError("Dependency Groups table is not a dict")
+raise TypeError("Dependency groups table is not a dict")
```

```diff
 dependency_groups_raw = pyproject["dependency-groups"]
 dependency_groups = _normalize_group_names(dependency_groups_raw)
-print("\n".join(resolve(pyproject["dependency-groups"], sys.argv[1])))
+print("\n".join(resolve(dependency_groups, sys.argv[1])))
```

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2045.org.readthedocs.build/en/2045/

### DIFF
--- a/source/shared/build-backend-tabs.rst
+++ b/source/shared/build-backend-tabs.rst
@@ -38,5 +38,5 @@
     .. code-block:: toml
 
         [build-system]
-        requires = ["uv_build >= 0.11.32, <0.12.0"]
+        requires = ["uv_build >= 0.12.0, <0.13.0"]
         build-backend = "uv_build"


webknjaz:
@jonathandung motivation? Justification?

cc @sirosen

jonathandung:
For the first one, I notice that the last commit to this file [`1700f85`](https://github.com/pypa/packaging.python.org/commit/1700f858e4066beda60c1fab2650013b9a789af9) changed dependency groups to lowercase where suitable, and I think this qualifies. For the second, it is simply strange to me that `pyproject["dependency-groups"]` was used rather than `dependency_groups_raw` which had just been defined as exactly that.

sirosen:
I have no strong feeling about the string in the `TypeError`. It's fine either way. The PEP was consistent about using Title Case, and the PyPUG docs are generally consistent about using lowercase. I would not bother changing it but I don't think it's a big enough deal to block the change over. (@jonathandung, I recommend removing that change, only because it may prevent this from getting merged, and my opinion is that the casing there is not as important as your other change.)

The other is a bugfix. Non-normalized names are being used by the example code, but resolution must use normalized names. I'm on my phone at the moment so I can't verify, but I'm pretty sure that is you try to resolve `foo` and your TOML data uses the name `FOO`, it will (incorrectly) fail.

jonathandung:
`pyproject` is an ordinary dictionary which is not modified between its lookups, so I don't really understand your mention about the resolution possibly failing. Can you please clarify? If there is a bug in my new version, the same bug must have existed in the old version. I was not aiming to fix a bug, but if that is indeed a bug, I am open to try and tackle it.

sirosen:
> I was not aiming to fix a bug, but if that is indeed a bug, I am open to try and tackle it.

I'm saying you already _did_ fix a bug! :grinning: :tada: 

The current version fails on the following input:
```toml
[dependency-groups]
FOO = ["requests"]
```

If you try to resolve `foo` against this, it raises a `LookupError` because the code is accidentally using `dependency_groups_raw`, not the normalized version.

The code in the original reference implementation in the PEP text also has this bug, in very slightly different form. Out of scope for a change here, but worth noting.

juanruiz34391-wq (approved):

webknjaz:
@juanruiz34391-wq plz don't spam.

webknjaz (approved):
Sounds like @sirosen is in favor of merging, which is enough for me to accept the patch.

webknjaz:
@jonathandung pro tip: creating PRs from `main` in your fork is known to cause problems in various settings. It's recommended that you always create a branch before opening a PR: https://hynek.me/articles/pull-requests-branch/.

anbazhakan-singlifeph (approved):

jonathandung:
Linkcheck failure is not related to my changes.

sirosen (approved):
I think we should merge this.

woodruffw (approved):

woodruffw:
image


what are you smoking, github

woodruffw:
For some reason I can't force a merge in the web UI or locally. @jonathandung could you merge or rebase your PR against the latest main?

(In the future, please consider using a feature branch instead of your fork's `main` as well.)

merwok:
Probably because the PR originates from the fork `main` branch. Don’t use your main branch!

jonathandung:
@woodruffw I'll do so after my vacation.

jonathandung:
#2100

PR #2060, opened 2026-05-22
Clarify script entry point targets

### LINKED ISSUE #1993: Clarify entry point's object reference specification
### Issue Description

In the entry point specification, section [data model](https://packaging.python.org/en/latest/specifications/entry-points#data-model),

> The object reference points to a Python object. It is either in the form `importable.module`, or `importable.module:object.attr`. Each of the parts delimited by dots and the colon is a valid Python identifier. It is intended to be looked up like this:
>
> ```python
> import importlib
> modname, qualname_separator, qualname = object_ref.partition(':')
> obj = importlib.import_module(modname)
> if qualname_separator:
>     for attr in qualname.split('.'):
>         obj = getattr(obj, attr)
> ```

First, I propose to disallow the former form `importable.module` as its invocation is never mentioned.

Second and more importantly, I want to explicitly enforces `object` on the right of the colon (:) to be an attribute of `importable.module` on the left.  To add the confusion , later in the section [use for scripts](https://packaging.python.org/en/latest/specifications/entry-points#use-for-scripts):

> the entry point `mycmd = mymod:main` would create a command `mycmd` launching a script like this:
>
> ```python
> import sys
> from mymod import main
> sys.exit(main())
> ```

Because of [the way the import-from statement works](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement):

> The `from` form uses a slightly more complex process:
> 1. find the module specified in the from clause, _loading_ and initializing it if necessary;
> 2. for each of the identifiers specified in the import clauses:
>    1. check if the imported module has an attribute by that name
>    2. if not, attempt to import a submodule with that name and then check the imported module again for that attribute
>    [...]

package authors may wrongly assume the first identifier `object` on the right of the colon to possibly be a module as well ([exhibit A](https://github.com/mps-youtube/yewtube/pull/1334)).  While both `pkg_resources`'s and the builtin `importlib.metadata`'s `EntryPoint.load` would fail on such incorrect entry point, they aren't run on common build processes and the most popular installer pip generates working scripts in a format the same as in the `mycmd = mymod:main` example.

It would be nice if the specification is more explicit over what is and is not allowed instead of a reference implementation (_[an object reference] is intended to be looked up like this_).  In Guix we got [entry point script generation (hopefully) right](https://codeberg.org/guix/guix/issues/4509) on [the third try](https://codeberg.org/guix/guix/pulls/5708) and it shouldn't be like this for other projects.

### Code of Conduct

Herrtian:
Fixes #1993.

Summary:
- clarify that script entry points need a module:object target

Checks:
- uv run --with pre-commit pre-commit run --files source/guides/writing-pyproject-toml.rst
- uv run --with nox nox -s build
- git diff --check

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2060.org.readthedocs.build/en/2060/

### DIFF
--- a/source/guides/writing-pyproject-toml.rst
+++ b/source/guides/writing-pyproject-toml.rst
@@ -208,9 +208,10 @@ In this example, after installing your project, a ``spam-cli`` command
 will be available. Executing this command will do the equivalent of
 ``import sys; from spam import main_cli; sys.exit(main_cli())``.
 
-Script entry points must use the ``module:object`` form. The object
-reference should point to a function that can be called with no arguments,
-not to a module by itself.
+Script entry points are written as :ref:`object references <entry-points>`,
+such as ``importable.module`` or ``importable.module:object.attr``. For
+script commands, the object reference should usually point to a callable object
+that can be called with no arguments, rather than to a module by itself.
 
 On Windows, scripts packaged this way need a terminal, so if you launch
 them from within a graphical application, they will make a terminal pop
--- a/source/guides/writing-pyproject-toml.rst
+++ b/source/guides/writing-pyproject-toml.rst
@@ -208,6 +208,10 @@ In this example, after installing your project, a ``spam-cli`` command
 will be available. Executing this command will do the equivalent of
 ``import sys; from spam import main_cli; sys.exit(main_cli())``.
 
+Script entry points must use the ``module:object`` form. The object
+reference should point to a function that can be called with no arguments,
+not to a module by itself.
+
 On Windows, scripts packaged this way need a terminal, so if you launch
 them from within a graphical application, they will make a terminal pop
 up. To prevent this from happening, use the ``[project.gui-scripts]``


source/guides/writing-pyproject-toml.rst
```diff
@@ -208,6 +208,10 @@ In this example, after installing your project, a ``spam-cli`` command
 will be available. Executing this command will do the equivalent of
 ``import sys; from spam import main_cli; sys.exit(main_cli())``.
 
+Script entry points must use the ``module:object`` form. The object
```
woodruffw: Per EPS, this is called an "object reference." I think we should probably use that language consistently and link to it as well:


```suggestion
Script entry points are written as object references like ``importable.module``, or importable.module:object.attr``. The object
```

Ref: https://packaging.python.org/en/latest/specifications/entry-points/#data-model

source/guides/writing-pyproject-toml.rst
```diff
@@ -208,6 +208,10 @@ In this example, after installing your project, a ``spam-cli`` command
...
+reference should point to a function that can be called with no arguments,
+not to a module by itself.
```
woodruffw: This is de facto true (I think), but in principle I think script entry points _can_ be modules if the user does something cursed like forcefully instantiating a callable module instance.

Not sure if that's worth qualifying though, the "should" is probably good enough 🙂

Herrtian:
Updated this to use the object reference terminology, link to the entry points spec, and avoid requiring the target to be strictly a function.

PR #2070, opened 2026-06-13
Document publishing job boundary

Himanshuagrawal4:
## Summary

- add an explicit note that build and publish steps should remain in separate jobs
- clarify that the publishing job intentionally avoids checkout/build commands
- explain why combining the jobs is unsupported for this trusted publishing workflow

## Why

The guide currently shows separate jobs, but the security boundary can be missed by readers adapting the workflow. This makes the intended structure clearer near the workflow example.

## Validation

- `sphinx-build --color --keep-going -j auto -b html -n -W source build`
- `git diff --check`

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2070.org.readthedocs.build/en/2070/


woodruffw (changes requested):
This is self-evident in context.

PR #2071, opened 2026-06-13
Update README metadata guidance

Himanshuagrawal4:
## Summary

- recommend the `pyproject.toml` `[project] readme` field first
- keep the older `setup.py` long_description guidance for projects that still need it
- note that the referenced README should be present in the source distribution

## Why

The README guide still led with older `setup.py` metadata. This updates the main path for modern projects while preserving the legacy example.

## Validation

- `sphinx-build --color --keep-going -j auto -b html -n -W source build`
- `git diff --check`

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2071.org.readthedocs.build/en/2071/

### DIFF
--- a/source/guides/making-a-pypi-friendly-readme.rst
+++ b/source/guides/making-a-pypi-friendly-readme.rst
@@ -18,23 +18,47 @@ Formats supported by `PyPI's README renderer <https://github.com/pypa/readme_ren
 * Markdown (`GitHub Flavored Markdown <https://github.github.com/gfm/>`_ by default,
   or `CommonMark <https://commonmark.org/>`_)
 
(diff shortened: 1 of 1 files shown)


webknjaz:
Based on the account activity in my other projects, other maintainers' reactions in other projects and overall unauthentic interactions, I believe we don't need to invest effort in reviewing or encouraging similar-quality submissions from this account.

I'll leave these external links here for now but we should probably adopt s more formal policy in the future:
* https://palletsprojects.com/contributing/llm-ai
* https://dontquotetheai.com
* https://github.com/jazzband/pip-tools/discussions/2278 / https://github.com/jazzband/pip-tools/pull/2318 / https://pip-tools.rtfd.io/en/stable/contributing/#llm-generated-contributions

PR #2041, opened 2026-05-04
Add PyPI extlink alias

Herrtian:
Adds the requested PyPI extlinks alias, so docs can use the :pypi: role for links to project pages.

I kept this to the alias itself and did not do any broad link rewrite in the same change.

Checks:
- git diff --check
- uv run --with pre-commit pre-commit run --files source/conf.py
- uv run --with nox nox -s build

Fixes #1370

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2041.org.readthedocs.build/en/2041/

PR #2076, opened 2026-06-17
Integrate sphinx-issues; convert inline PyPI URLs to :pypi:

### LINKED ISSUE #1370: Add extlinks alias for PyPI
```suggestion
    "user": (f"{github_sponsors_url}/%s", "@%s"),
    "pypi": ("https://pypi.org/project/%s", "%s"),
```

I think this could be real helpful.

_Originally posted by @chrysle in https://github.com/pypa/packaging.python.org/pull/1362#discussion_r1387128645_

sadiqkhzn:
Closes #1370.

Integrates sphinx-issues so the `:pypi:` role is available, per @webknjaz's feedback. The existing `issue`/`pr`/`commit`/`user` extlinks are moved over to sphinx-issues as well; `gh` stays in extlinks since sphinx-issues doesn't provide it.

Also converts ten inline `pypi.org/project` URLs in three files to use `:pypi:`, only in places where the visible link text already matched the project name.

One side effect to flag: `:pr:` usages in `news.rst` now render as `(#N)` instead of `(PR #N)`. sphinx-issues 5.0.1 restricts `issues_pr_prefix` to `#`, `@`, or `!`, so the previous prefix can't be configured. Same end state as [pypa/pip#12616](https://github.com/pypa/pip/pull/12616). Happy to convert the 109 `:pr:` lines to inline links if you'd rather preserve the prose.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -35,6 +35,7 @@ extensions = [
     "sphinx_copybutton",
     "sphinx_toolbox.collapse",
     "sphinx-jsonschema",
+    "sphinx_issues",
 ]
 
 nitpicky = True
@@ -194,12 +195,16 @@ github_repo_issues_url = f"{github_url}/{github_repo_slug}/issues"
 github_sponsors_url = f"{github_url}/sponsors"
 
 extlinks = {
-    "issue": (f"{github_repo_issues_url}/%s", "#%s"),
-    "pr": (f"{github_repo_url}/pull/%s", "PR #%s"),
-    "commit": (f"{github_repo_url}/commit/%s", "%s"),
     "gh": (f"{github_url}/%s", "GitHub: %s"),
-    "user": (f"{github_sponsors_url}/%s", "@%s"),
 }
+
+# -- Options for sphinx_issues -------------------------------------------------------
+# https://github.com/sloria/sphinx-issues#installation-and-configuration
+# Provides the :issue:, :pr:, :commit:, :user:, and :pypi: roles.
+
+issues_github_path = github_repo_slug
+# Preserve the previous `user` role behaviour, which pointed at GitHub Sponsors.
+issues_user_uri = f"{github_sponsors_url}/{{user}}"
 
 # -- Options for intersphinx ----------------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#configuration
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ sphinx-inline-tabs==2023.4.21
 sphinx-copybutton==0.5.2
 sphinx-toolbox==3.5.0
 sphinx-jsonschema==1.19.1
+sphinx-issues==5.0.1
--- a/source/discussions/deploying-python-applications.rst
+++ b/source/discussions/deploying-python-applications.rst
@@ -57,7 +57,7 @@ Windows
 Pynsist
 ^^^^^^^
 
-`Pynsist <https://pypi.org/project/pynsist>`__ is a tool that bundles Python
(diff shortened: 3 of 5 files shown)


sadiqkhzn:
Heads-up for reviewers: I noticed #2041 by @Herrtian also closes #1370 and adds the same one-line extlink alias. I missed it when picking this issue up — apologies. This PR additionally does the inline-link conversions (10 sites across 3 files), with rationale in the description for what was deliberately not converted.

I'm happy whichever path the maintainers prefer:
1. Merge #2041 first and I'll rebase this to just the conversion sweep on top, or
2. Merge this and close #2041 as superseded, or
3. Close this in favor of #2041 and I'll open a separate PR for the conversions later.

Let me know which works best.

webknjaz (changes requested):
> `sphinx-issues` does not register a `pypi` role, so it cannot fully cover this case. The pip project went the same route in [pypa/pip#12616](https://github.com/pypa/pip/pull/12616), keeping `pypi` as a plain extlink alongside `sphinx-issues`. This PR follows the same pattern.

That's not true. The plugin defines that role. Integrate it instead, as discussed in the issue.

Additionally, to keep this reviewable, could you delete most of the llm noise in the PR description?

sadiqkhzn:
You're right, sorry — I was relying on outdated info from an earlier issue comment. `sphinx-issues` has had the `:pypi:` role since 4.0.0.

Pushed an update that integrates it: `sphinx_issues` is now in extensions, the `pypi`/`issue`/`pr`/`commit`/`user` extlinks are removed (only `gh` stays since sphinx-issues doesn't have it), and `issues_user_uri` is set to preserve the Sponsors URL. Trimmed the description too.

One thing to flag: sphinx-issues 5.0.1 (latest compatible with the pinned Sphinx 7.2.6) hardcodes `issues_pr_prefix` to one of `#`/`@`/`!`, so the 109 `:pr:` lines in `news.rst` now render as `(#N)` instead of `(PR #N)`. Same outcome pip ended up with. If you'd rather keep the `PR #N` prose, I can convert those 109 lines to inline links in a follow-up commit — let me know.

source/conf.py
```diff
@@ -194,13 +195,17 @@
 github_sponsors_url = f"{github_url}/sponsors"
 
 extlinks = {
-    "issue": (f"{github_repo_issues_url}/%s", "#%s"),
-    "pr": (f"{github_repo_url}/pull/%s", "PR #%s"),
-    "commit": (f"{github_repo_url}/commit/%s", "%s"),
     "gh": (f"{github_url}/%s", "GitHub: %s"),
```
webknjaz, line 198: Do we use any use of this role anywhere?

source/conf.py
```diff
@@ -194,13 +195,17 @@
...
-    "user": (f"{github_sponsors_url}/%s", "@%s"),
 }
 
+# -- Options for sphinx_issues -------------------------------------------------------
+# https://github.com/sloria/sphinx-issues#installation-and-configuration
+# Provides the :issue:, :pr:, :commit:, :user:, and :pypi: roles.
+
+issues_github_path = github_repo_slug
+# Preserve the previous `user` role behaviour, which pointed at GitHub Sponsors.
```
webknjaz, line 206: This is the default, so no need

source/conf.py
```diff
@@ -194,13 +195,17 @@
 github_sponsors_url = f"{github_url}/sponsors"
```
webknjaz, line 195: Drop this

webknjaz:
Looks like #2041 is more complete than this, FTR.

sadiqkhzn:
Confirmed. Closing in favour of #2041. Thanks for the review.

PR #2083, opened 2026-06-29
chore: use non-legacy hook name for ruff-check

henryiii:
Noticed this warning in the pre-commit output. Switching to the non-legacy hook name.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2083.org.readthedocs.build/en/2083/


webknjaz (approved):

henryiii:
Link fix in #2084.

PR #2084, opened 2026-07-02
docs: fix link to activestate python

henryiii:
As far as I can tell, this is roughly the equivalent link now.

Fixes the broken linkcheck (the old page is a 404).

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2084.org.readthedocs.build/en/2084/

PR #2085, opened 2026-07-04
 Ban negative extras (`extra !=` markers)

konstin:
Negative extras (`extra !=`) are currently technically allowed in the grammar, but they behave strangely: `extra !=` means that package could transitively remove a package from selection through activating an extra. This turns incremental resolution into a global problem: We can't exit a resolution with a conflict when we've exhausted all versions in range including backtracking, we also have to exhaustively search for any package version that may activate the extra which deactivates the `extra !=` requirement which is part of the conflict. This would be fatal to resolver complexity and performance, preventing most of the short-circuiting that makes resolving tractable.

Much weirder, with negative extras, adding a new dependency can make a previously impossible resolution solvable:

```
[project]
name = "foo"
version = "0.1.0"
dependencies = [
  "numpy >2",
  "numpy <2; extra != 'old-science'",
]
```

Only by adding a package with `Requires-Dist: foo[old-science]`, this resolves.

Using PEP 621, users only get positive extras: `project.optional-dependencies.foo` gets translated into a conjunction of `extra == "foo"` with each requirement's marker.

While this may look like a breaking change, it only clarifies the current state of support in tools. From [pip not supporting it](https://github.com/pypa/pip/issues/14139) and [uv not supporting it](https://github.com/astral-sh/uv/issues/20107), we know it is not currently used and safe to remove. This solves the [confusion for users who want to use negative extras for conflicting packages](https://github.com/astral-sh/uv/issues/20107).

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2085.org.readthedocs.build/en/2085/

### DIFF
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -708,6 +708,8 @@ History
   work. [#marker_comparison_logic]_
 - January 2026: fix outdated references to other documents that were
   inadvertently retained from :pep:`508`
+- July 2026: Explain that ``extra != "name"`` is not (properly) supported
+  (https://discuss.python.org/t/ban-negative-extras-for-extras-marker/108022/24)
 
 
 References
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -458,13 +458,15 @@ DO NOT need to define these fields.
 The ``extra`` field is also special, as it expects set-like behaviour, but
 predates the addition of ``Set of strings`` as a defined marker field type.
 It only supports one operation, ``extra == "name"``, which is similar to
-``"name" in extras``. For dependency marker evaluations, the set of extra
-names used for these comparisons is the full set of requested extras for *that
-particular package*, whether requested directly in a top level dependency
-declaration, or indirectly in a transitive dependency declaration. Other
-comparison operations on ``extra`` are not defined and publishing tools SHOULD
-emit an error, index servers MAY disallow uploads containing such environment
-markers, while locking and installation tools SHOULD evaluate them as False.
+``"name" in extras``. ``extra != "name"`` SHOULD NOT be used anymore and MAY
+be rejected or evaluated to false. For dependency marker evaluations, the set
+of extra names used for these comparisons is the full set of requested extras
+for *that particular package*, whether requested directly in a top level
+dependency declaration, or indirectly in a transitive dependency declaration.
+Other comparison operations on ``extra`` are not defined and publishing tools
+SHOULD emit an error, index servers MAY disallow uploads containing such
+environment markers, while locking and installation tools SHOULD evaluate them
+as False.
 
 Unlike the newer ``extras`` field, environment markers using this field SHOULD
 be accepted by both publishing tools and index servers. Marker evaluation
--- a/source/specifications/dependency-specifiers.rst
+++ b/source/specifications/dependency-specifiers.rst
@@ -457,9 +457,8 @@ DO NOT need to define these fields.
 
 The ``extra`` field is also special, as it expects set-like behaviour, but
 predates the addition of ``Set of strings`` as a defined marker field type.
-Accordingly, ``extra == "name"`` in a dependency declaration is similar to
-``"name" in extras``, while ``extra != "name"`` is similar to
-``"name" not in extras``. For dependency marker evaluations, the set of extra
+It only supports one operation, ``extra == "name"``, which is similar to
+``"name" in extras``. For dependency marker evaluations, the set of extra
 names used for these comparisons is the full set of requested extras for *that
 particular package*, whether requested directly in a top level dependency
 declaration, or indirectly in a transitive dependency declaration. Other


konstin:
DPO thread: https://discuss.python.org/t/ban-negative-extras/108022

source/specifications/dependency-specifiers.rst
ncoghlan, line 265: Do we also need to amend the "Set of Strings" definition? If `extra != "name"` doesn't make sense, do `"name" not in extras` and `"name" not in dependency_groups` actually make sense?

ncoghlan, line 713: History entry is needed for the change

ncoghlan: Perhaps explicitly call out "... other comparison operations on `extra` (including `!=`) are not`" so it's clear the omission from the earlier sentence isn't accidental?

ncoghlan (dismissed):
The required change before approval is the history entry.

The proposed wording tweak is genuinely just a suggestion, and the "Set of Strings" question could be resolved in a follow-up PR if we're not sure the exact same concern applies.

dimbleby:
@radoering

radoering:
@dimbleby Thanks for the notification.

> While this may look like a breaking change, it only clarifies the current state of support in tools.

If "tools" is only pip and uv, you are probably right. However, I think it is kind of breaking for Poetry, see e.g. https://python-poetry.org/docs/dependency-specification/#exclusive-extras

I mean, Poetry can still document that this is not allowed anymore by the standard and should only be used in Poetry-only projects.

notatallshaw:
> If "tools" is only pip and uv, you are probably right. However, I think it is kind of breaking for Poetry, see e.g. https://python-poetry.org/docs/dependency-specification/#exclusive-extras

Hi @radoering, I checked Poetry's dependency selection behavior and found it is almost identical to uv's: it evaluates negative extras (`extra != ...`) as always false in dependency metadata. There is already an open issue for this: python-poetry/poetry#10120.

So I would consider that Poetry also doesn't support negative extras in dependency metadata.

Where Poetry does honor a negative extra is when it installs the root project itself (`poetry install` keeps a `; extra != 'cuda'` dependency, `poetry install --extras cuda` drops it). It does not carry them any further than that: `poetry-core` drops negative-extra dependencies at build time, so a Poetry-built wheel never publishes them. I'm not sure the root-project behavior is a concern for interoperability standards, since we don't have any standards governing how a tool resolves its own project, but perhaps we need to be clear on the language here?

So banning `extra !=` in dependency metadata wouldn't break anything Poetry publishes.

Independently, I'd strongly recommend Poetry document the limitation that the pattern only works for the top-level project: `poetry-core` drops the dependency from built wheels, and Poetry evaluates the marker as false when it appears in a dependency's metadata.

FYI here's an MRE for my findings: https://gist.github.com/notatallshaw/b0ec3db17c7647e3e9e8db7d9b6288a3

> I mean, Poetry can still document that this is not allowed anymore by the standard and should only be used in Poetry-only projects.

My findings indicate that this is effectively the case already, just undocumented.

radoering:
@notatallshaw Thanks for the analysis.

I checked why poetry-core ignores dependencies with negative extras. Actually, that does not seem to be intentionally but is just a side effect of an unrelated bugfix. We are probably lucky now that python-poetry/poetry#10120 has not been fixed yet. Same for ignoring negative extras from wheels.

> > I mean, Poetry can still document that this is not allowed anymore by the standard and should only be used in Poetry-only projects.
> 
> My findings indicate that this is effectively the case already, just undocumented.

You are right. Due to the bugs, the impact of banning negative extras is less severe than I thought. It is probably sufficient to update Poetry's documentation and pay attention not to fix the bugs in a way that ignores the ban.

notatallshaw:
> Actually, that does not seem to be intentionally but is just a side effect of an unrelated bugfix. We are probably lucky now that [python-poetry/poetry#10120](https://github.com/python-poetry/poetry/issues/10120) has not been fixed yet. Same

Ah I see, but even if it was fixed and the marker evaluation honored `extra != ...` I suspect it would still lead to resolver inconsistencies that I mention here: https://discuss.python.org/t/ban-negative-extras-for-extras-marker/108022/4 (I have [my own](https://github.com/notatallshaw/nab/tree/main/nab-resolver) experimental pubgrub style resolver that I am working on so I am trying to pin this down).

ncoghlan:
From https://discuss.python.org/t/ban-negative-extras-for-extras-marker/108022/18, it sounds like the _intended_ use case for `!=` support in poetry is for the tool-specific "conflicting extras" case. As described in the post, this is the same reason that `not in` support actually makes sense in lock files for both extras and dependency groups.

The guidance we give for attempts to use the "Set of Strings" fields in public dependency declarations is:

> Publishing tools SHOULD emit an error if projects attempt to reference the extras or dependency_groups fields in their published dependency declaration metadata, and index servers SHOULD NOT accept uploads referencing these fields.

The actual comparison operations themselves are left defined as normal rather than giving tools permission to treat them as `False` without evaluation.

This isn't the effect we get from undefining `extra != "name"` entirely. Undefining the comparison instead drops us into this clause (which is weaker than we want for index servers, and stronger than we want for locking and installation tools):

> Other comparison operations on `extra` are not defined and publishing tools SHOULD emit an error, index servers MAY disallow uploads containing such environment markers, while locking and installation tools SHOULD evaluate them as False.

Accordingly, I think the end result will be clearer if we add an explicit clause for negative extra markers along the lines of:

> Excluding dependencies with `extra != "name"` is only considered valid in already resolved dependency trees, so publishing tools SHOULD emit an error for such comparisons, index servers SHOULD disallow uploads containing them, and locking and installation tools MAY evaluate them as `False`.

(Having that explicit clause there should also help remind us that even if we someday allow `"name" in extras` on PyPI, we should still disallow `"name" not in extras`)

henryiii:
Link check fix in https://github.com/pypa/packaging.python.org/pull/2084.

konstin:
I added commentary on backwards compatibility with `extra != "name"`, this should be ready now.

source/specifications/dependency-specifiers.rst
konstin: I'm impartial (not a cause of the problem I'm trying to solve here), though poetry seems to like it.

henryiii:
I scanned the PyPI dataset for operations with `extra`:

- `==` → 1,462,072 instances
- `!=` → 31 instances (18 distinct packages)
- anything else → 0

Data is at https://github.com/sethmlarson/pypi-data/releases/tag/2026.07.21 (thanks @sethmlarson!).

**Packages using `extra != …`:**

| Package                                           | Rows   | Pattern                                                            |
| ------------------------------------------------- | -----: | ------------------------------------------------------------------ |
| `ipfabric-netbox`                                 | 9      | long `extra!='ipfabric-7-x' and …` chains — a version-matrix guard |
| `choccy`                                          | 2      | `extra!='cpu-only' and extra!='no-numba'`                          |
| `gato-hep`                                        | 2      | `extra!='gpu'`                                                     |
| `pymare`                                          | 2      | `python_version!='3.9' or extra!='stan'`                           |
| `turnkeyml`                                       | 2      | `platform_system=='…' and extra!='llm-oga-cuda'`                   |
| `wyzely-detect`                                   | 2      | `extra!='cuda'`                                                    |
| `callooktools`, `callsignlookuptools`, `qrztools` | 1 each | `extra!='async'`                                                   |
| `deepvelo`, `hufr`, `opendetect`                  | 1 each | `extra!='gpu'`                                                     |
| `django-citext`                                   | 1      | `extra!='psycopg2'`                                                |
| `feast-custom`                                    | 1      | `extra!='flink'`                                                   |
| `hevy-unofficial`                                 | 1      | `extra!='core'`                                                    |
| `reframe-hpc`                                     | 1      | `extra!='no-analytics'`                                            |
| `spinepose`                                       | 1      | `platform_system!='darwin' and extra!='gpu'`                       |
| `tart`                                            | 1      | `extra!='minimal'`                                                 |

konstin:
@ncoghlan Is there anything missing to merge this?

henryiii:
Could you rebase to get rid of the red X? ;)

konstin:
I rebased (the check is also non-mandatory now to avoid blocking specs changes on external failures).

woodruffw (approved):

source/specifications/dependency-specifiers.rst
woodruffw: Marking as resolved per https://github.com/pypa/packaging.python.org/pull/2085#pullrequestreview-4631239937, but this seems worth tracking in a follow-up issue/PR 🙂

PR #2089, opened 2026-07-12
Clarify handling of script metadata

pfmoore:
The specification states that only top-level comments define script metadata. But recognising precisely top-level comments is difficult, and requires a full Python parse, which (for example) the canonical regex does not do.

Clarify that tools MAY use a simple textual scan, like the regex does, but that users must not rely on comments within other Python constructs being ignored.

See https://discuss.python.org/t/clarify-that-pep-723-inline-script-metadata-recognition-is-textual/108089 for a discussion of this issue.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2089.org.readthedocs.build/en/2089/

### DIFF
--- a/source/specifications/inline-script-metadata.rst
+++ b/source/specifications/inline-script-metadata.rst
@@ -75,11 +75,11 @@ recognised as containing metadata. However, parsing Python code is non-trivial,
 and therefore:
 
 * Tools MAY choose to do a simple textual scan, rather than a full Python parse.
-  For example, the canonical regular expression provided above does a textual
-  scan.
 * As a result of the previous point, the behaviour of scripts that contain data
   that looks like metadata within another Python construct such as a multi-line
   string is tool-dependent and should not be relied on.
+* The canonical regular expression provided above is an example of an
+  implementation that does a simple textual scan.
 
 script type
 -----------
--- a/source/specifications/inline-script-metadata.rst
+++ b/source/specifications/inline-script-metadata.rst
@@ -70,6 +70,17 @@ and the regular expression, the text specification takes precedence.
 Tools MUST NOT read from metadata blocks with types that have not been
 standardized by this specification.
 
+Note that the specification only requires that *top-level* comment blocks are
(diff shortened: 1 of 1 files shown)


pfmoore:
The linkcheck failure appears to only be for files I haven't changed...

source/specifications/inline-script-metadata.rst
```diff
@@ -70,6 +70,17 @@ and the regular expression, the text specification takes precedence.
 Tools MUST NOT read from metadata blocks with types that have not been
 standardized by this specification.
 
+Note that the specification only requires that *top-level* comment blocks are
+recognised as containing metadata. However, parsing Python code is non-trivial,
+and therefore:
+
+* Tools MAY choose to do a simple textual scan, rather than a full Python parse.
+  For example, the canonical regular expression provided above does a textual
+  scan.
```
zsol: Maybe worth pointing this out explicitly
```suggestion
  For example, the canonical regular expression provided above does a textual
  scan, and does not ignore metadata blocks e.g. in a multi-line string literal.
```

zsol (commented):
Thanks for doing this!

source/specifications/inline-script-metadata.rst
pfmoore: Is that not just repeating what the next bullet point says?

Maybe the following would work:

> * Tools MAY choose to do a simple textual scan, rather than a full Python parse.
> * As a result of the previous point, the behaviour of scripts that contain data that looks like metadata within another Python construct such as a multi-line string is tool-dependent and should not be relied on.
> * The canonical regular expression provided above is an example of an implementation that does a simple textual scan.

zsol: Yeah that's even better

brettcannon (approved):

henryiii:
The link check fix in #2084.

pfmoore:
I'm not sure how to get the attention of "someone with write access" - @pypa/packaging-user-guide-editors is this the right group?

ewdurbin (approved):

PR #2090, opened 2026-07-12
Add GitHub Actions workflow for Python package with Conda

smarthomespropertieske-pixel:
This workflow sets up a Python package build using Conda on Ubuntu, installs dependencies, lints the code with flake8, and runs tests with pytest.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2090.org.readthedocs.build/en/2090/

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -1,0 +1,34 @@
+name: Python Package using Conda
+
+on: [push]
+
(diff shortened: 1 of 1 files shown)


merwok:
Can you explain how this is useful for this project?

woodruffw (changes requested):
Yeah, I don't see how this is relevant.

PR #2094, opened 2026-07-22
docs: fix typo e. g. to e.g. in dropping older python versions guide

LaRoyBot:
----
📚 Documentation preview 📚: https://python-packaging-user-guide--2094.org.readthedocs.build/en/2094/


merwok:
Why not say _for example_ instead of latin abbreviations?

woodruffw (approved):

woodruffw:
(I'm fine either way, approving this because it's a typo fix. If we want to switch to a non-Latin abbreviation we can merge that too 🙂)

PR #2099, opened 2026-07-30
Update description of wheels to "built distributions"

nlhkabu:
Context:

I am working on a new design for the PyPI release files page.

At the top of the page, we will direct users to https://packaging.python.org/en/latest/discussions/package-formats/#package-formats for more information about source distributions and built distributions:

Screenshot 2026-07-30 at 22-11-05 Versio · Warehouse

This PR updates the package formats documentation to refer to wheels as "built distributions" instead of "binary distributions"

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2099.org.readthedocs.build/en/2099/

### DIFF
--- a/source/discussions/package-formats.rst
+++ b/source/discussions/package-formats.rst
@@ -8,7 +8,7 @@ This page discusses the file formats that are used to distribute Python packages
 and the differences between them.
 
 You will find files in two formats on package indices such as PyPI_: **source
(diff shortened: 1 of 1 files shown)


merwok:
I think built and binary were quite equivalent in distutils docs, and the PEP and live docs for wheel do use binary: https://packaging.python.org/en/latest/specifications/binary-distribution-format/

PR #2080, opened 2026-06-29
docs: fix minor grammar issues in contributing guide

dhruvatr:
## Summary

This PR fixes a couple of minor grammar issues in the contributing guide.

### Changes
- Add a missing comma for readability.
- Change "in web browser" to "in a web browser".

These changes improve readability without changing the meaning of the documentation.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2080.org.readthedocs.build/en/2080/

PR #2101, opened 2026-08-02
Fix missing article in browser documentation

dhruvatr:
Add the missing article "a" before "web browser" in the documentation for improved grammar and clarity.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2101.org.readthedocs.build/en/2101/


henryiii:
Adding "a" before "web browser" would be a bit clearer, but the PR actually adds it in the middle, "web a browser"!

henryiii:
Also this fix is already in #2080.

PR #2106, opened 2026-08-14
Upgrade GitHub Actions

cclauss:
Do not encourage visitors to rely on out-of-date GitHub Actions.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2106.org.readthedocs.build/en/2106/

PR #2107, opened 2026-08-15
zizmor: Full security lint of publish-to-pypi.yml

cclauss:
A **superset** of #2106 that passes zizmor security linting checks.
* #2106 

@webknjaz Your review, please.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2107.org.readthedocs.build/en/2107/

### DIFF
--- a/source/guides/github-actions-ci-cd-sample/publish-to-pypi.yml
+++ b/source/guides/github-actions-ci-cd-sample/publish-to-pypi.yml
@@ -1,6 +1,7 @@
 name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
 
 on: push
+permissions: {}
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


source/guides/github-actions-ci-cd-sample/publish-to-pypi.yml
```diff
@@ -1,18 +1,19 @@
 name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
 
 on: push
+permissions: {}
```
webknjaz, line 4: This one might be okay provided the build job also lists the exact privileges required and the PR is scoped to only adding `permissions:` to examples.
Also, add one empty-line separator above.

webknjaz (changes requested):
The rest is not acceptable, though, no, same as #2106.

cclauss:
Closing in favor of #2106 because we want to keep the `uses: pypa/gh-action-pypi-publish@release/v1` syntax.

webknjaz:
@cclauss it's probably still a good idea to tell users to use Zizmor in some tip admonition in the guide, though, if it still doesn't do that. One important thing that people miss in my experience is separating the publishing and the building jobs, though. Showing different explicit permissions might be a good hint for them. Unfortunately, it seems like they tend to copy their examples from some weird projects on the internet that already don't follow this recommendation.

cclauss:
I find [`excessive-permissions`](https://docs.zizmor.sh/audits/#excessive-permissions) to be the Zizmor rule I am least confident about proposing solutions for.  It is helpful that the `excessive-permissions` example covers these GitHub Actions.

I agree that bad examples abound, but I believe that these samples should be the source of truth for visitors.

If this repo were to adopt Dependabot for GitHub Actions with a 7-day cooldown plus full Zizmor, then these samples could be best-practice examples that use current actions and pass Zizmor.

I know this is more PRs for maintainers to review and merge, but at least visitors would not be misled.

PR #2108, opened 2026-08-18
sdist format: qualify pax compatibility

woodruffw:
I'm opening this as a direct living spec PR, since it's _arguably_ just a clarification of the existing requirement, not a new one 🙂. However, if there's disagreement about that, I'm happy to take this to DPO for discussion first.

My rationale: the sdist spec already says that the tarball must be a POSIX.1-2001 pax archive. This implies a bunch of things that people may not realize, but are already fully covered by that requirement. For example, GNU-style tar files are not pax-conforming (they use a different magic and have special extended value encoding rules), and the `S` typeflag is GNU-specific.

There's a passable argument to be made that `GNU.sparse.*` is pax-conforming, since they're pax-style records. However:

1. Some versions of `GNU.sparse.*` are illegal to represent in pax, since they use duplicated record names;
2. `GNU.sparse.*` in general changes the framing logic, making it hard to correctly model the parse of a sparse member with a standard pax-conforming parser.

Finally, I think the most controversial part of this is adding a `NOT RECOMMENDED` for other pax-style (but not standard) vendor records. These are technically allowed in pax, just not defined by the standard themselves. However, they're also subject to a bunch of interoperability problems. For example, `SCHILY.xattr` is sometimes encoded as raw binary, which is not pax-conforming and causes problems for standard parsers. 

(And more generally, Python source distributions have no business encoding xattrs or anything like that, from what I can tell 🙂)

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2108.org.readthedocs.build/en/2108/

### DIFF
--- a/source/specifications/source-distribution-format.rst
+++ b/source/specifications/source-distribution-format.rst
@@ -74,10 +74,19 @@ at their respective paths relative to the root directory of the sdist
 No other content of a sdist is required or defined. Build systems can store
 whatever information they need in the sdist to build the project.
 
-The tarball must use the modern POSIX.1-2001 pax tar format, which specifies
+The tarball MUST use the modern POSIX.1-2001 pax tar format, which specifies
 UTF-8 based file names. In particular, source distribution files must be readable
 using the standard library tarfile module with the open flag 'r:gz'.
 
+Source distributions MUST NOT use non-pax features, even if those features are
+structurally compatible with the pax standard. Installers and other consumers MAY
+reject source distributions that are not strictly pax-conforming.
+
+Examples of non-pax features include GNU-style or legacy ("V7")-style tar members
+and sparse member encodings (e.g. ``GNU.sparse`` extensions or ``S`` typeflag members).
+
+The use of pax-style vendor extensions (such as ``SCHILY.xattr`` or ``LIBARCHIVE.xattr``)
+is NOT RECOMMENDED for interoperability reasons.
 
 .. _sdist-archive-features:
 
--- a/source/specifications/source-distribution-format.rst
+++ b/source/specifications/source-distribution-format.rst
@@ -75,11 +75,11 @@ No other content of a sdist is required or defined. Build systems can store
 whatever information they need in the sdist to build the project.
 
 The tarball MUST use the modern POSIX.1-2001 pax tar format, which specifies
-UTF-8 based file names. In particular, source distribution files must be readable
+UTF-8 based file names. In particular, source distribution files MUST be readable
 using the standard library tarfile module with the open flag 'r:gz'.
 
 Source distributions MUST NOT use non-pax features, even if those features are
-structurally compatible with the pax standard. Installers and other consumers MAY
+structurally compatible with the pax standard. Installers and other consumers SHOULD
 reject source distributions that are not strictly pax-conforming.
 
 Examples of non-pax features include GNU-style or legacy ("V7")-style tar members


woodruffw:
CCing some other folks who I think might be interested: @konstin @miketheman

source/specifications/source-distribution-format.rst
```diff
@@ -74,10 +74,19 @@ at their respective paths relative to the root directory of the sdist
 No other content of a sdist is required or defined. Build systems can store
 whatever information they need in the sdist to build the project.
 
-The tarball must use the modern POSIX.1-2001 pax tar format, which specifies
+The tarball MUST use the modern POSIX.1-2001 pax tar format, which specifies
```
woodruffw, line 77: I suspect the lowercasing of this "must" wasn't meant to de-emphasize its normativity, so I'm uppercasing it.

source/specifications/source-distribution-format.rst
```diff
@@ -74,10 +74,19 @@ at their respective paths relative to the root directory of the sdist
...
 
+Source distributions MUST NOT use non-pax features, even if those features are
+structurally compatible with the pax standard. Installers and other consumers MAY
+reject source distributions that are not strictly pax-conforming.
+
+Examples of non-pax features include GNU-style or legacy ("V7")-style tar members
+and sparse member encodings (e.g. ``GNU.sparse`` extensions or ``S`` typeflag members).
+
+The use of pax-style vendor extensions (such as ``SCHILY.xattr`` or ``LIBARCHIVE.xattr``)
+is NOT RECOMMENDED for interoperability reasons.
```
miketheman, line 89: I hear you on the trying to soften the blow, but if one goal of the Python packaging ecosystem is to remain compatible and interoperable, why not draw the line stronger, and save some future selves from having to deal with this?

source/specifications/source-distribution-format.rst
```diff
@@ -74,10 +74,19 @@ at their respective paths relative to the root directory of the sdist
...
 whatever information they need in the sdist to build the project.
 
-The tarball must use the modern POSIX.1-2001 pax tar format, which specifies
+The tarball MUST use the modern POSIX.1-2001 pax tar format, which specifies
 UTF-8 based file names. In particular, source distribution files must be readable
 using the standard library tarfile module with the open flag 'r:gz'.
 
+Source distributions MUST NOT use non-pax features, even if those features are
+structurally compatible with the pax standard. Installers and other consumers MAY
+reject source distributions that are not strictly pax-conforming.
```
miketheman, line 83: Maybe come right out with it? 
"sdists must not include any pax [Vendor Extensions](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_02:~:text=The%20standard%20developers%20have%20reserved%20keyword%20name%20space%20for%20vendor%20extensions)"

woodruffw: My only reason was that these are *technically* pax conforming in a more meaningful sense. So I didn't want to jump to changing the plain meaning of the spec itself, but if nobody objects...

woodruffw: Oh, I read this in the wrong section. Yeah, I'll strengthen this.

source/specifications/source-distribution-format.rst
woodruffw: I didn't strengthen this one because I think technically it's a standards change versus just a clarification -- pax vendor extensions are perfectly valid, they just do all kinds of random things. So I wanted to make this PR easy to accept without any standards work, but I agree this should be eliminated entirely 🙂

source/specifications/source-distribution-format.rst
```diff
@@ -74,10 +74,19 @@ at their respective paths relative to the root directory of the sdist
...
 
+Source distributions MUST NOT use non-pax features, even if those features are
+structurally compatible with the pax standard. Installers and other consumers SHOULD
+reject source distributions that are not strictly pax-conforming.
+
+Examples of non-pax features include GNU-style or legacy ("V7")-style tar members
+and sparse member encodings (e.g. ``GNU.sparse`` extensions or ``S`` typeflag members).
+
+The use of pax-style vendor extensions (such as ``SCHILY.xattr`` or ``LIBARCHIVE.xattr``)
+is NOT RECOMMENDED for interoperability reasons.
```
sethmlarson, line 89: TIL "NOT RECOMMENDED" is RFC 2119 compliant.

miketheman (approved):

sethmlarson (approved):
LGTM, thanks @woodruffw!

PR #2048, opened 2026-05-15
Document pip index configuration migration

Herrtian:
Closes #463.

Adds a short note to the PyPI.org migration guide about checking pip configuration when downloads still use the old host.

Checks: pre-commit on the changed file; nox -s build.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2048.org.readthedocs.build/en/2048/

PR #2119, opened 2026-08-30
docs: clarify pip index configuration during PyPI migration

### LINKED ISSUE #463: Add pip.conf heads-up to PyPI.org migration guide
Followup to https://github.com/pypa/packaging-problems/issues/135 :

A user had a `.pip` directory containing a `pip.conf` file that pointed to the old PyPI URL:

~~~
[global]
index-url=https://pypi.python.org/pypi/
~~~

I think https://packaging.python.org/guides/migrating-to-pypi-org/#downloading-packages should include a heads-up about that. Once you figure out the language to add, and add it, we'll revise the https://pyfound.blogspot.com/2018/03/warehouse-all-new-pypi-is-now-in-beta.html#Migrating section to point to it.

harshitethic:
Summary

* document that pip index-url configuration can override the default PyPI index
* show how to locate active pip configuration files with pip config list -v
* document updating a legacy [pypi.python.org](http://pypi.python.org/) index URL to [pypi.org/simple](http://pypi.org/simple)

Fixes #463

Validation

* Focused documentation-only change in source/guides/migrating-to-pypi-org.rst
* Verified the commit diff contains only the intended documentation addition

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2119.org.readthedocs.build/en/2119/

### DIFF
--- a/source/guides/migrating-to-pypi-org.rst
+++ b/source/guides/migrating-to-pypi-org.rst
@@ -135,6 +135,21 @@ Downloading packages
 
 ``pypi.org`` is the default host for downloading packages.
 
+If you have configured pip to use a custom ``index-url``, that setting can
+override the default PyPI index. Check your pip configuration files if pip
+continues to use the old PyPI URL after migration. The command ``pip config
+list -v`` can be used to show the configuration files pip is loading.
+
+For example, a configuration containing::
+
+    [global]
+    index-url = https://pypi.python.org/pypi
+
+should be updated to use the current PyPI URL::
+
+    [global]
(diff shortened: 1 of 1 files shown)


webknjaz:
It's unclear if PyPUG should contain individual tools' docs instead of pointing to them. Have you checked previous PR discussions? Said tools evolve independently and get own documentation updates sooner. We should probably give a more generic advice as some of the PRs attempted but didn't get to complete the effort. #2048 seems to have been solid but didn't get early reviews.

PR #2122, opened 2026-08-31
Check only new links in pull requests

### LINKED ISSUE #2037: Linkcheck for newly add links in PRs
### Issue Description

As follow-up to https://github.com/pypa/packaging.python.org/pull/2035#pullrequestreview-4203119819, we should change the linkcheck setup so that it check for new links in PRs, so that PR reviewers know that all newly added links are valid.

### Code of Conduct

Soumoditya:
Fixes #2037.\n\nRestore the latest successful default-branch linkcheck report for pull requests and add its URLs to Sphinx's ignore list. Linkcheck failures then block a pull request only when a baseline report was restored, so only newly added links can cause the failure. Scheduled and default-branch runs continue to check every link and refresh the baseline only after a successful run.\n\nValidation: python -m compileall -q source/conf.py; cached-report parsing verified with valid and malformed JSONL input; git diff --check.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2122.org.readthedocs.build/en/2122/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,18 +25,11 @@ jobs:
     if: ${{ github.repository_owner == 'pypa' || github.event_name != 'schedule' }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
-    continue-on-error: >-
-      ${{ fromJSON(matrix.continue-on-error) }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


woodruffw:
This is not a good issue for first time contributors, especially not with LLMs.

PR #2126, opened 2026-09-06
Document hashes for manually hosted package indexes

### LINKED ISSUE #1951: Expand on the "Manual" repository example
### Issue Description

https://packaging.python.org/en/latest/guides/hosting-your-own-index/ mentions hosting release artifacts as static files, but it doesn't mention that relying on the default index pages generated by something like `http.server` may result in lock files that don't contain expected artifact hashes (e.g. `uv` won't download the artifacts if hashes are not provided by the index server, it just omits the hashes from the affected lockfile)

Adding the hashes to a *pre-generated* project detail page is intentionally straightforward, so I don't think it would be unreasonable to actually cover how to do that:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Links for local_project</title>
  </head>
  <body>
    <h1>Links for local_project</h1>
<a href="./local_project-1.0-cp311-cp311-win_arm64.whl#sha256=eb8a8cd0d4667994ac218a1ed5faca4a634670f1500bbc9401cd212654efeaae" >local_project-1.0-cp311-cp311-win_arm64.whl</a><br />
</body>
</html>
```

Extracting the wheel metadata files and including them in the index entries is a little bit more involved, but also potentially worth documenting.

Finally, the mention of `dumb_pypi` should probably be moved up to this section rather than being left in the main self-hosting table (since it's actually a tool for translating a list of artifacts into a static simple repository API structure rather than being a server in its own right)

### Code of Conduct

roli-lpci:
The manual repository example currently recommends serving an automatically
generated directory listing, but those listings usually do not include artifact
hashes. This can leave package-manager lock files without hashes for artifacts
downloaded from the repository.

This adds a hash-bearing project detail page example, links to the Simple
Repository API contract, and points readers to `dumb-pypi` when they want a
static site generator that supplies hashes and other repository metadata.

Closes #1951.

Validation:

- `git diff --check`
- `nox -s build` (strict Sphinx HTML build with missing-reference warnings treated as errors)

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2126.org.readthedocs.build/en/2126/

This contribution was produced by agents through [Hermes Labs](https://hermes-labs.ai)’ engineering infrastructure. [Rolando Bosch](https://github.com/roli-lpci) is the responsible human contributor and authorized publication from his personal GitHub account.

### DIFF
--- a/source/guides/hosting-your-own-index.rst
+++ b/source/guides/hosting-your-own-index.rst
@@ -52,6 +52,26 @@ directory with autoindex enabled. For an example using the built in Web server
 in `Twisted`_, you would simply run ``twistd -n web --path .`` and then
 instruct users to add the URL to their installer's configuration.
 
+An automatically generated directory listing usually doesn't include artifact
+hashes. Without hashes in the repository links, tools may be unable to record
+them in lock files. To include a SHA-256 hash, generate an ``index.html`` file
+in each project directory and append the hash to each artifact URL as described
+by the :ref:`Simple Repository API <simple-repository-html-project-detail>`:
+
+.. code-block:: html
+
+   <!DOCTYPE html>
+   <html lang="en">
+     <head><title>Links for foo</title></head>
+     <body>
+       <h1>Links for foo</h1>
+       <a href="Foo-1.0.tar.gz#sha256=3571b...">Foo-1.0.tar.gz</a>
+     </body>
+   </html>
+
+Replace ``3571b...`` with the full SHA-256 digest of the artifact. You can use
+:ref:`dumb-pypi` to generate a static repository with hashes and other Simple
+Repository API metadata from a directory of package artifacts.
 
 Existing projects
 =================
@@ -104,11 +124,6 @@ Existing projects
      -
      -
      - also mirroring; manual synchronisation
-
-   * - :ref:`dumb-pypi`
-     -
-     -
-     - not a server, but a static file site generator
 
    * - :ref:`httpserver`
      -
--- a/source/guides/hosting-your-own-index.rst
+++ b/source/guides/hosting-your-own-index.rst
@@ -52,11 +52,12 @@ directory with autoindex enabled. For an example using the built in Web server
 in `Twisted`_, you would simply run ``twistd -n web --path .`` and then
 instruct users to add the URL to their installer's configuration.
 
-An automatically generated directory listing usually doesn't include artifact
-hashes. Without hashes in the repository links, tools may be unable to record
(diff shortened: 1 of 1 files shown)


source/guides/hosting-your-own-index.rst
```diff
@@ -105,11 +125,6 @@ Existing projects
      -
      - also mirroring; manual synchronisation
 
-   * - :ref:`dumb-pypi`
-     -
-     -
-     - not a server, but a static file site generator
-
    * - :ref:`httpserver`
```
merwok, line 131: Why is this section removed?

merwok (commented):
> Validation:
> - git diff --check

What do you mean by that?

source/guides/hosting-your-own-index.rst
```diff
@@ -52,6 +52,26 @@ directory with autoindex enabled. For an example using the built in Web server
 in `Twisted`_, you would simply run ``twistd -n web --path .`` and then
 instruct users to add the URL to their installer's configuration.
 
+An automatically generated directory listing usually doesn't include artifact
+hashes. Without hashes in the repository links, tools may be unable to record
+them in lock files. To include a SHA-256 hash, generate an ``index.html`` file
+in each project directory and append the hash to each artifact URL as described
+by the :ref:`Simple Repository API <simple-repository-html-project-detail>`:
```
woodruffw: I'm not 100% sure what this means: most tools/commercial products that I'm aware of _do_ include artifact hashes. Is there a specific tool/flow you ran into where this isn't the case?

(More generally, I think this particular guide is really stale and no longer reflects received wisdom in the community about how to host your own index. For example, as of PEP 833, we consider the HTML index frozen and encourage people to use the JSON one when possible.)

source/guides/hosting-your-own-index.rst
roli-lpci: `dumb-pypi` was moved rather than dropped: it is now linked in the manual-repository section immediately above, where it is described as the static site generator for this flow. I removed the table row because it is not a server, matching issue #1951’s request.

source/guides/hosting-your-own-index.rst
roli-lpci: Thanks—the specific flow is the bare autoindex described in this section, not package-index products. I reproduced it with `python -m http.server` and current uv 0.11.32: the generated link had no hash fragment, and `uv.lock` recorded the wheel URL without a hash. Commit fb577c1 narrows “usually” to “may” for generic autoindexes and adds the JSON preference. The HTML example remains only for this minimal static-hosting case; PEP 833 leaves HTML supported.

PR #2128, opened 2026-09-08
Don't block merges on linkcheck

woodruffw:
I believe this was an oversight in https://github.com/pypa/packaging.python.org/pull/2035 -- it doesn't make sense to allow linkchecks to pass PR gates, only to bounce them at the merge group.

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2128.org.readthedocs.build/en/2128/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,10 @@ jobs:
         - false
         include:
         - noxenv: linkcheck
-          continue-on-error: >-  # Don't block PRs on linkcheck unrelated failures
-            ${{ toJSON(github.event_name == 'pull_request' || github.event_name == 'merge_group') }}
+          # Don't block PRs, merge groups, or non-main pushes on unrelated linkcheck failures.
+          continue-on-error: >-
+            ${{ toJSON(github.event_name == 'pull_request' || github.event_name == 'merge_group' ||
+              (github.event_name == 'push' && github.ref != 'refs/heads/main')) }}
 
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
         - noxenv: linkcheck
           # Don't block PRs, merge groups, or non-main pushes on unrelated linkcheck failures.
           continue-on-error: >-
-            ${{ toJSON(github.event_name == 'pull_request' || github.event_name == 'merge_group' ||
+            ${{ toJSON(contains(['merge_group', 'pull_request'], github.event_name) ||
               (github.event_name == 'push' && github.ref_name != github.event.repository.default_branch)) }}
 
     steps:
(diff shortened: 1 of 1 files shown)


woodruffw:
I had to widen this to non-main pushes as well, since first-party PRs will _also_ trigger this since they receive the `push` event.

.github/workflows/test.yml
```diff
@@ -35,8 +35,10 @@ jobs:
         - false
         include:
         - noxenv: linkcheck
-          continue-on-error: >-  # Don't block PRs on linkcheck unrelated failures
-            ${{ toJSON(github.event_name == 'pull_request') }}
+          # Don't block PRs, merge groups, or non-main pushes on unrelated linkcheck failures.
+          continue-on-error: >-
+            ${{ toJSON(github.event_name == 'pull_request' || github.event_name == 'merge_group' ||
+              (github.event_name == 'push' && github.ref != 'refs/heads/main')) }}
```
webknjaz:
```suggestion
              (github.event_name == 'push' && github.ref_name != github.event.repository.default_branch)) }}
```

.github/workflows/test.yml
```diff
@@ -35,8 +35,10 @@ jobs:
         - false
         include:
         - noxenv: linkcheck
-          continue-on-error: >-  # Don't block PRs on linkcheck unrelated failures
-            ${{ toJSON(github.event_name == 'pull_request') }}
+          # Don't block PRs, merge groups, or non-main pushes on unrelated linkcheck failures.
+          continue-on-error: >-
+            ${{ toJSON(github.event_name == 'pull_request' || github.event_name == 'merge_group' ||
```
webknjaz: Something like this seems nicer
```suggestion
            ${{ toJSON(contains(['merge_group', 'pull_request'], github.event_name) ||
```

webknjaz:
> since first-party PRs will _also_ trigger this since they receive the `push` event.

Not sure we really need to care about pushes, though. What fails is merge queue events.

woodruffw:
> > since first-party PRs will _also_ trigger this since they receive the `push` event.
> 
> Not sure we really need to care about pushes, though. What fails is merge queue events.

FWICT the merge queue only becomes available after the push-triggered checks succeed, since the linkcheck is marked as required on that trigger still.

webknjaz:
@woodruffw no, all GH platform events are independent, they all are happening in isolated contexts and don't know of each other. But if a PR is created from within upstream, sure, both events will happen upstream and cause runs in the same repo. This means they'll create Checks API statuses that will show up in the PR. But this only highlights a repo configuration problem — we should never create topic branches upstream in the first place 🤷‍♂️. If I had enough privileges in the repo, I'd set up branch rulesets to reflect this but I don't.. Maybe one day I'll finally get to catch @pradyunsg and we'll set this up.

I don't think we should be permissive with pushes.

Meanwhile, #2129 will fix the broken links, unblocking other merges.

woodruffw:
> no, all GH platform events are independent, they all are happening in isolated contexts and don't know of each other.

Yeah, I'm not saying they're dependent. I'm saying that the "required" state is set for `Test / check`, and on `push` that was causing a failure despite essentially being a duplicate run of `pull_request`.

> we should never create topic branches upstream in the first place

Sorry, I don't follow -- you're saying we should only ever do PRs from forks for this repo?

(As before, I'm generally agnostic on _how_ things get accomplished. But the linkcheck continues to be a significant drag on this repo's velocity.)

PR #2136, opened 2026-09-22
Add Rye

bangseongbeom:
Add [Rye](https://github.com/astral-sh/rye) to [Project Summaries](https://packaging.python.org/en/latest/key_projects/).

Rye was a well-known Python project with more than 14,000 GitHub stars, and it was the predecessor to [uv](https://github.com/astral-sh/uv). Although the project is already archived, I would like to suggest adding it for historical context.

The description is sourced directly from the official [Rye website](https://rye.astral.sh/).

----
📚 Documentation preview 📚: https://python-packaging-user-guide--2136.org.readthedocs.build/en/2136/


woodruffw (changes requested):
This doesn't seem like a good idea to me. packaging.python.org already has a problem with conveying stale (and sometimes misleading) information to end users who are trying to make modern/idiomatic choices, and I think linking to an explicitly archived project will deepen that confusion (even with a warning admonition).

(Maybe if we had some kind of "history" section? But in general I think we should be aggressively deleting things from this page rather than adding them 🙂)

bangseongbeom:
Thanks for the clear explanation. I understand and agree with the concern about keeping the page current and idiomatic. I'll close this PR.

